PR #60, opened 2019-02-12
Corrected the --training-stats parameters to --training_statistics

us:
On the docs page `--training-stats` parameters changed to `--training_statistics`.


w4nderlust:
I actually already corrected this myself before seeing your pull request. I'm closing it, but thank you really much for sending it!

PR #62, opened 2019-02-12
Changed numpy minimum version to 1.15

us:
When I'm building from the source code I got a error like this and I changed to numpy version to 1.15 and problem is solved. 

```
Processing dependencies for ludwig==0.1.0
error: numpy 1.14.2 is installed but numpy>=1.15.0 is required by {'spacy'}
```

Note: I wrongly create a new branch from the not master so #61 issue include a 2 commit so I deleted it and corrected to this.


Barathwaja:
Yep spacy requires numpy >= 1.15.0 which I faced please update requirements.txt .

PR #67, opened 2019-02-13
Add Dockerfile for running Ludwig on a CPU

dettmering:
Adding a Dockerfile for running Ludwig on a CPU using the Tensorflow base image.

As soon as I get home, I will try to create a Dockerfile for using a GPU.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM tensorflow/tensorflow:latest-py3
+
+RUN apt-get install -y --no-install-recommends git
+RUN git clone https://github.com/uber/ludwig.git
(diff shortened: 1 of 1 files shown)


rajputakhil:
@dettmering by @w4nderlust help, I just managed to compile and successfully run Dockerfile for running Ludwig. I think this Dockerfile is to run Ludwig on CPU. Am I correct? If I want to run Ludwig on GPU, would changing "tensorflow" to "tensorflow-gpu" in requirements.txt file suffice? Thanks a lot!

w4nderlust:
@rajputakhil use the `--gpus` parameter, you can find info in the documentation. And yes the docker image will use requirements.txt, you should replace `tensorflow` with `tensorflow-gpu` if you want to use the gpus in the docker.

PR #80, opened 2019-02-14
swap slots/intent in NLU table header

ivanhe:
order_food seems to be intent; B-* seem to be slots


w4nderlust:
Thanks for noticing this @ivanhe ! Please sign the CLA and I'll merge this.

w4nderlust:
I ended up changing the order of the column in that example. Thank you anyway!

PR #106, opened 2019-02-16
Improve Dockerfile

philippgille:
- Added newlines for readability
- Added depth parameter to git clone so not the full git history is cloned
- Combined git clone and Ludwig installation into one layer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,12 @@
 FROM tensorflow/tensorflow:latest-py3
 
-RUN apt-get install -y --no-install-recommends git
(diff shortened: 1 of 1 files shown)


philippgille:
Memo: I first also added a line with `&& rm -rf /var/lib/apt/lists/*` but then realized that 1) it's not necessary because no `apt update` is done, which again isn't necessary because the TensorFlow image doesn't remove its apt lists and 2) deleting the inherited files wouldn't reduce the size of the layer (and thus the whole image) but instead would just force people who inherit from this image to `apt update` again.

w4nderlust:
Thank you for your contribution!

PR #153, opened 2019-03-01
Remove outdated note that TensorFlow does not work with Python 3.7

floscha:
Since [version 1.13.1](
https://github.com/tensorflow/tensorflow/releases/tag/v1.13.1), TensorFlow now supports Python 3.7 on all operating systems. Consequently, the note to use Python 3.6 as the recommended version can be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ sudo apt install python3  # on ubuntu
 brew install python3      # on mac
 ```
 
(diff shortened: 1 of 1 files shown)


w4nderlust:
Thanks! I added a note in the minimum version of tf needed to run on python 3.7.

PR #156, opened 2019-03-01
Fix KeyError

rolisz:
`dropout` doesn't exist in the default dictionary, it's `dropout_rate`


w4nderlust:
Thanks for spotting this!

PR #158, opened 2019-03-01
Add Keras fuel economy example to docs

c-m-hunt:
Add Keras example to docs based on https://www.tensorflow.org/tutorials/keras/basic_regression as suggested by @w4nderlust at https://twitter.com/w4nderlus7/status/1101333499817078784

Other example needs a little refinement as I'd hacked around with the data and it doesn't work out of the box. Will add over weekend.


w4nderlust:
Thank you so much for submitting this, I will reformat a little and then merge it.

Issue #159, opened 2019-03-02
Inconsistent arguments for 'get_experiment_description'


graytowne:
It seems when calling 'get_experiment_description' in ```train.py```([link](https://github.com/uber/ludwig/blob/master/ludwig/train.py#L200)), the argument order is inconsistent with the one from function definition ([link](https://github.com/uber/ludwig/blob/master/ludwig/utils/misc.py#L29)).

w4nderlust:
Your PR solves the issue, thanks!

PR #160, opened 2019-03-02
Fix inconsistent issue when calling get_experiment_description.

graytowne:
Quick bug fix for issue #159.


w4nderlust:
Thanks for sending this PR

PR #165, opened 2019-03-05
fixed preprocessing_defaults error in bag_feature.py

C0deFxxker:
'preprocessing_defaults' in bag_feature.py has error.

'most_common' should be **10000** and 'lowercase' should be **False**


w4nderlust:
Thank you!

PR #173, opened 2019-03-05
Update README.md

Shivam-Beeyani:
Corrected the installation instructions, changed pip to pip3.


w4nderlust:
This depends on the version of python and pip you have installed in your system, if you are running from a virtual environment (which is recommended), so i prefer to leave it as it is. Thank you for submitting anyway

PR #179, opened 2019-03-06
Fix in timeseires placeholder and PassthroughEncoder

w4nderlust:


BenMacKenzie:
hidden_size = input.sequence.shape[-1] rather than
hidden_size = input.sequence.shape()[-1] on line 72,

PR #200, opened 2019-03-11
Fix KeyError when running via initialize_model() for image features.

laserbeam3:
Training a model using `model.initialize_model(); model.train_online();` fails for image inputs.

As far as I understand, that field (`model_definition['data_hdf5_fp']`) is only created in `preprocess_for_training`, which only is called from `model.train()`. This change fixes the resulting KeyError, however it feels symptomatic. It would require a deeper knowledge of the code base than what I have at this time to ensure preprocessing is done the same in `model.train()` and `model.initialize_model()`

### DIFF
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -235,7 +235,7 @@ class ImageInputFeature(ImageBaseFeature, InputFeature):
         for dim in ['height', 'width', 'num_channels']:
             input_feature[dim] = feature_metadata['preprocessing'][dim]
         input_feature['data_hdf5_fp'] = (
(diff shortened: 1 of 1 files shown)


w4nderlust:
Thanks for this PR. @ydudin3 can you please take a look at the deeper problem here? (the assumption that there's always a hdf5 file available) and also can you confirm if this solves the problem for the moment?

w4nderlust:
@laserbeam3 we did some changes in the image feature in the https://github.com/uber/ludwig/pull/219 commit. Can you please check if it solves the issue that originated this PR?

w4nderlust:
Closing this because other PRs should have solved this issue already. Feel free to reopen if the problem emerges again.

PR #204, opened 2019-03-12
Transaction example

BenMacKenzie:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

Note that the documentation HTML files are a in `docs/` while the Markdown sources are in `mkdocs/docs`.

If you are proposing a modification to the documentation you should change only the Markdown files, then recreate the documentation as examplained in the `mkdocs/README.md` file with `mkdocs build`, which will create the HTML files automatically, and only after this create a commit.

`API.md` is automatically generated from the docstrings in the code, so if you want to change something in that file, first modify `ludwig/api.py` docstring, then run `mkdocs/code_docs_autogen.py`, which will create `mkdocs/docs/api.md` and then finally run `mkdocs build` which will generate the HTML in `docs/`.


w4nderlust:
Thanks for the PR @BenMacKenzie . I still have to figure out how and if i want to have notebooks with related data in the project (probably I do, but I don't want to deal with data issues, hence the thought that needs to go into it).
In the mean time, would you consider to add your transaction examples to `mkdocs/docs/examples.md` with another PR? It would be greatly appreciated.

w4nderlust:
I'm closing this as i added the transaction already to the examples. Thank you for contributing this anyway, I may consider adding notebooks, in that case I will reopen it

PR #247, opened 2019-03-26
Fixes two bugs introduced while migrating to tensorflow 1.13

msaisumanth:
- Fixes two bugs introduced while migrating to tensorflow 1.13

1. Typo in loss_modules
2. [tf.SparseTensor](https://www.tensorflow.org/api_docs/python/tf/sparse/SparseTensor) expects the shape to be of type tf.int64 (we were using tf.int32)

Screen Shot 2019-03-26 at 2 41 24 PM

Tests:
All tests pass

### DIFF
--- a/ludwig/models/modules/loss_modules.py
+++ b/ludwig/models/modules/loss_modules.py
@@ -109,8 +109,9 @@ def cross_entropy_sequence_loss(logits, targets, sequence_length):
             logits=logits, labels=targets)
         # Mask out the losses we don't care about
         loss_mask = tf.sequence_mask(
-            tf.cast(sequence_length, tf.int32), tf.cast(tf.shape(targets)[1],
-            tf.int32))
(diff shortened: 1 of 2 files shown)


ludwig/models/modules/loss_modules.py
```diff
@@ -109,8 +109,8 @@ def cross_entropy_sequence_loss(logits, targets, sequence_length):
             logits=logits, labels=targets)
         # Mask out the losses we don't care about
         loss_mask = tf.sequence_mask(
-            tf.cast(sequence_length, tf.int32), tf.cast(tf.shape(targets)[1]),
-            tf.int32)
+            tf.cast(sequence_length, tf.int32), tf.cast(tf.shape(targets)[1],
```
ydudin3: Break this into one argument per line, e.g.

fun(
  x1, 
  x2
)

Issue #198, opened 2019-03-11
Feature request: GAN support


bitcoinmeetups:
Hello,

Does Ludwig support GAN's?

How about a feature where the user inputs a list of images and receives as output a bunch of cool GAN generated images?

BM

w4nderlust:
Well at prediction time you don't need to provide anything, you just sample from noise and you produce images from there :)
Anyway, yes it's on out list, there needs to be a little piece before it, the image output feature so that first Ludwig can output images, then we can add a loss that is a GAN loss which contains also a discriminator.
More thought needs to be put into it, but it's definitely something that we eventually want to support.

sk1210:
Hi @w4nderlust , I would like to work on adding this feature, Can you guide me in that, like how to start or any reference.

bitcoinmeetups:
Awesome @w4nderlust!

Following this!

w4nderlust:
@sk1210 the first step would be to implement an image output feature, with a way to produce an image (deconvolution for instance), a way to calculate loss and a way to compute measures.
I would use the `cetegory_feature.py` module as a template to copy from.

EdwardFKit:
Has there been progress on this - this would be of great use!

w4nderlust:
@EdwardFKit this is currently not being worked on, there after the release of v0.3 (a month or so), we will reasses all users requests to rigure out priorities for v0.4 But we did not forget about this, it's just that with the resources at our disposal, it has not been prioritized yet.

sk1210:
@w4nderlust, @EdwardFKit  I have now started working on this, I will keep you updated.

EdwardFKit:
Great – please do – FYI what we are working on is not image / gradient based – it is more related to NLP / Keyword / Phrase / special purpose text generation – basically to generate automated Keyword software test cases for software systems -

w4nderlust:
> @w4nderlust, @EdwardFKit I have now started working on this, I will keep you updated.

@sk1210 that sound great, thank you for being willing to help, that is hreatly appreciated.
Assing GANs to Ludwig is non trivial at the moment because of its architectures. Some of the changes we are making for v0.3 will definitely make it easier.
Because it's non trivial I suggest you to fisrt study in detail the codebase, the best way being training a model by going through it with a debugger, line by line.
[...]

PR #256, opened 2019-03-28
Handles the alpha channel issue

msaisumanth:
# Code Pull Requests

- Handles the case where the dataset may have images with different number of channels
- Cleans up some repetitive code in image_feature. 
- Throws an exception If there are no images in the dataset
- If the user doesn't specify number of channels in the definition, we use the first image in the data as the source of truth. Now if some image has a different number of channels, we throw an exception.
- If the user specifies the number of channels and an image has a different number of channels, we append/drop the channels accordingly and warn the user.

- Issue #198
- Integration test added

- All tests pass


ludwig/features/image_feature.py
```diff
@@ -50,6 +51,62 @@ def get_feature_meta(column, preprocessing_parameters):
             'preprocessing': preprocessing_parameters
         }
 
+    @staticmethod
+    def _read_image_and_resize(filepath, im_width, im_height,
```
ydudin3: keep one argument per line

ludwig/features/image_feature.py
```diff
@@ -50,6 +51,62 @@ def get_feature_meta(column, preprocessing_parameters):
...
+                               should_resize, num_channels,
+                               resize_method, user_specified_num_channels):
+        """
+        :param filepath: path to the image
+        :param im_width: expected width of the image
+        :param im_height: expected height of the image
+        :param should_resize: Should the image be resized?
+        :param resize_method: type of resizing method
+        :param num_channels: expected number of channels in the first image
+        :param user_specified_num_channels: did the user specify num channels?
```
ydudin3: can we get rid of this variable and have num_channels with None initializer and then check 
if num_channels is not None instead?

ludwig/features/image_feature.py
```diff
@@ -50,6 +51,62 @@ def get_feature_meta(column, preprocessing_parameters):
...
+        im_num_channels = num_channels_in_image(img)
+        if im_num_channels == 1:
+            img = img.reshape((img.shape[0], img.shape[1], 1))
+
+        if user_specified_num_channels is False:
+            # If the image isn't like the first image, raise exception
+            if im_num_channels != num_channels:
+                raise ValueError(
+                    'Image {0} has {1} channels, unlike the first image, which'
+                    ' has {2} channels'.format(filepath, im_num_channels,
```
ydudin3: same, one argument per line

ludwig/features/image_feature.py
```diff
@@ -133,23 +195,16 @@ def add_feature_data(
...
-                        img,
-                        (height, width),
-                        preprocessing_parameters['resize_method']
-                    )
-                # todo: temporary workaround for images with alpha channel, replace
-                if img.ndim == 3 and img.shape[2] != num_channels:
-                    img = img[:, :, :num_channels]
+
+                img = ImageBaseFeature._read_image_and_resize(
+                    filepath, width, height, should_resize,
```
ydudin3: one argument per line

ludwig/features/image_feature.py
```diff
@@ -163,20 +218,16 @@ def add_feature_data(
...
-                        img = img.reshape((img.shape[0], img.shape[1], 1))
-                    if should_resize:
-                        img = resize_image(
-                            img,
-                            (height, width),
-                            preprocessing_parameters['resize_method'],
-                        )
+
+                    img = ImageBaseFeature._read_image_and_resize(
+                        filepath, width, height, should_resize,
```
ydudin3: one argument per line

tests/integration_tests/test_experiment.py
```diff
@@ -477,6 +478,87 @@ def test_visual_question_answering(csv_filename):
     os.rmdir(image_dest_folder)
 
 
+def test_image_resizing_num_channel_handling(csv_filename):
```
ydudin3: instead of adding a whole new integration test (which are pretty slow) can we add a couple new features to one of the existing integration tests involving images?

ludwig/features/image_feature.py
msaisumanth: Yeah, I tried that actually. It failed because we're updating the metadata which is in turn updating the preprocessing parameters. We will lose track of the fact that the user didn't actually mention a value.

tests/integration_tests/test_experiment.py
msaisumanth: Yes the tests are slow. But the tricky part here is that I have to generate two datasets with different number of channels. We can eliminate some of the image related tests because of this. Maybe I can do that as part of another task?

ludwig/features/image_feature.py
msaisumanth: will do

ludwig/features/image_feature.py
```diff
@@ -50,6 +51,67 @@ def get_feature_meta(column, preprocessing_parameters):
...
+                raise ValueError(
+                    'Image {0} has {1} channels, unlike the first image, which'
+                    ' has {2} channels'.format(filepath,
+                                               im_num_channels,
+                                               num_channels))
+
+        else:
+            # Number of channels is specified by the user
+            img_padded = np.zeros((im_height, im_width, num_channels))
+            for i in range(min(num_channels, im_num_channels)):
```
w4nderlust: This for loop is slow, we should compute the `min_num_channels` beforehand and then assign with `img_padded[:, :, :min_num_channels] = img[:, :, :min_num_channels]`

ludwig/features/image_feature.py
```diff
@@ -50,6 +51,67 @@ def get_feature_meta(column, preprocessing_parameters):
...
+        in the dataset as the source of truth. If any image in the dataset
+        doesn't have the same number of channels as the first image,
+        raise an exception.
+
+        If the user specifies a number of channels, we try to convert all the
+        images to the specifications by dropping channels/padding 0 channels
+        """
+
+        img = imread(filepath)
+        im_num_channels = num_channels_in_image(img)
```
w4nderlust: as we call the image img, here i would expect img_num_channels

ludwig/features/image_feature.py
```diff
@@ -50,6 +51,67 @@ def get_feature_meta(column, preprocessing_parameters):
...
+        if im_num_channels == 1:
+            img = img.reshape((img.shape[0], img.shape[1], 1))
+
+        if user_specified_num_channels is False:
```
w4nderlust: I would rather check if it is true and invert the order of the code blocks

ludwig/features/image_feature.py
```diff
@@ -50,6 +51,67 @@ def get_feature_meta(column, preprocessing_parameters):
             'preprocessing': preprocessing_parameters
         }
 
+    @staticmethod
+    def _read_image_and_resize(filepath,
+                               im_width,
```
w4nderlust: as for before, why im and not img or image?

ludwig/features/image_feature.py
msaisumanth: good point. Will update.

ludwig/features/image_feature.py
msaisumanth: makes sense. will update.

tests/integration_tests/test_experiment.py
ydudin3: sg

PR #260, opened 2019-03-30
Update docs for running tests

ydudin3:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests


w4nderlust:
@ydudin3 can you also rebuild the docs and commit again please?

w4nderlust:
In the checklist before the command, it is specified to have ludwig installed through setup, that shoul be removed as it is not needed anymore

ydudin3:
Ah, missed that one, will update.

PR #276, opened 2019-04-08
ContribClass instance in registry and generalize commands

dsblank:
This PR:

* creates an instance of the ContribClass from ContribClass.import_call and saves it in registry
* calls methods of instance, like ContribInstance.train(*args, **kwargs)
* For Comet, it uses a local config file to pass experiment_key down the pipeline

TODO: need to call contrib_command("METHOD_NAME") for places to add hooks.

### DIFF
--- a/ludwig/cli.py
+++ b/ludwig/cli.py
@@ -23,6 +23,7 @@ from ludwig import predict
 from ludwig import train
 from ludwig import visualize
 
+import ludwig.contrib
 
 class CLI(object):
     """CLI describes a command line interface for interacting with Ludwig, there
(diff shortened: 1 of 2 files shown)


dsblank:
I was thinking that this system might be a good way to add support for Jupyter notebooks (mentioned in the issues) and other environments too. Anything outside the main focus of the ludwig pipeline could be added through the contrib classes and methods.

w4nderlust:
This looks really nice. We can figure out together what additional functions need to be added and where.
Now we have a "import" time function and a "cli command" time function, we can figure out a bunch of others, like "visualization show" or "before training", "end of epoch", "end of training", "before prediction" and "after prediction".

One really minor consideration, which at the moment doesn't make any difference: at the moment we have one single registry for classes and instances, so each time we go over the registry and check if an instance has been created ( https://github.com/uber/ludwig/blob/89157e7e8cb9b5cdd3ac4fc7c618d0f0c1e810b7/ludwig/contrib.py#L117-L118 ). This is fine for now that there's only one contrib class. But I can imagine a moment when there are 10 contrib classes, of which only one or 2 are initialized at a time. Maybe keeping a separate list / set of initialized instances to iterate over would be a better option then.

Finally, do you want me to merge this already or should we figure out the visualization part first? (consider the viz part of the code is the one that received less love in terms of polish so far in the codebase)

dsblank:
Thanks! I think this will work nicely for many uses (I hope!)

Yes, you are right... that should be a faster method of finding the activated contribs. (contribs should probably be a folder too, as each one of these classes can get big).

I'd say go ahead and merge it, and we can continue to refine, if that is fine with you. And we can get more eyes on it.

dsblank:
I made a followup PR #277 that addresses three items.

PR #278, opened 2019-04-08
Adds a unittest for resnet encoder, a couple of simple tests for image utils

msaisumanth:
# Code Pull Requests

Adds unittests

1. For resnet image encoder 
   a. checks the expected type and shape of the hidden representation of resnet

2. sanity checks for image utility methods
  a. not important but wanted to provide examples


ydudin3:
1. I think the convention is to name test files starting with test_*.
2. Another convention is to have the "same" filenames for test files. For example if you're testing code in models/combiners.py the corresponding test file should be models/test_combiners.py

Are you following any particular patterns for setting up this unit test framework?

msaisumanth:
@ydudin3 pytest recognizes, test_.xpy or x_test.py. I felt that x_test.py is easier to read, but we can go with test_x.py also (we just need to be consistent). 

As for the pattern, all the encoders have a similar framework/flow, so it makes sense to test the basic functionality for all of them in the same manner. 

Take a look at: https://github.com/keras-team/keras/blob/master/tests/keras/layers/convolutional_test.py

tests/ludwig/models/modules/encoder_test.py
```diff
@@ -0,0 +1,72 @@
...
+from ludwig.models.modules.loss_modules import regularizer_registry
+
+
+L1_REGULARIZER = regularizer_registry['l1'](0.1)
+L2_REGULARIZER = regularizer_registry['l2'](0.1)
+NO_REGULARIZER = None
+DROPOUT_RATE = 0.5
+
+
+def encoder_test(encoder_type,
```
ydudin3: Let's stick to this format - most of the code base is like this:

def func(
    arg_1,
    arg_2
):

tests/ludwig/models/modules/encoder_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Instantiate the encoder
+    encoder = encoder_type(**encoder_args)
+
+    # Run the encoder
+    input_data = tf.convert_to_tensor(input_data)
+    dropout_rate = tf.convert_to_tensor(dropout_rate)
+    is_training = tf.convert_to_tensor(False)
+
+    hidden, _ = encoder(input_data, regularizer, dropout_rate,
```
ydudin3: If all arguments don't fit into one line => split into one argument per line.

tests/ludwig/models/modules/encoder_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+def test_image_encoders_resnet():
+    # Test the resnet encoder for images
+    encoder_args = {'resnet_size': 8, 'num_filters': 8, 'fc_size': 28}
+    image_size = (1, 10, 10, 3)
+
+    output_shape = [1, 28]
+    input_image = np.random.randint(0, 1, image_size).astype(np.float32)
+
+    encoder = encoder_test(ResNetEncoder, input_image, L1_REGULARIZER,
```
ydudin3: Same here

tests/ludwig/models/modules/encoder_test.py
```diff
@@ -0,0 +1,72 @@
...
+                           DROPOUT_RATE, np.float, output_shape, None,
+                           encoder_args)
+
+    assert encoder is not None
+    assert encoder.resnet.kernel_size == 3
```
ydudin3: Can we check all the settable attributes of the resnet encoder?

ludwig/utils/image_utils.py
```diff
@@ -77,6 +76,9 @@ def get_abs_path(data_csv_path, image_path):
 
 
 def num_channels_in_image(img):
+    if img is None or img.ndim < 2:
+        raise ValueError('Image does not have (at least 2) channels')
```
w4nderlust: Those 2 are height and width, not channels, plase change the error message.

PR #282, opened 2019-04-10
Better docs for contribs; new contrib access method: visualize_figure

dsblank:
This PR:

* provides more documentation for defining a contrib class
* adds a new contrib access method "visualize_figure"
* uses it from comet contrib


w4nderlust:
This is great, than you @dsblank !

PR #285, opened 2019-04-12
 New contrib methods: train_model, train_save, experiment_save

dsblank:
This PR injects three new contrib command methods:

1. train_model
2. train_save
3. experiment_save

In addition, the comet contrib is now in a complete state, and uses the released version of comet_ml. If you run experiment/train followed by many visualize commands, the figures will automatically go to the right experiment.

See results here: 
https://www.comet.ml/dsblank/ludwig/9f9b48c099734ff4ae91d9f2a02bfbea

Comet automatically logs Ludwig items:

* Chart tab and Metrics tab: accuracy and loss 
* Code tab: command line
* Hyperparameter tab: tensorflow hyperparameters
* Graph tab: textual description of model (useful for diffs)
* Output tab: output from training
* System metrics tab: os version, etc.
* Installed packages tab: all installed Python modules
* Graphics tab: all ludwig-visualized figures
* HTML tab: record of ludwig command lines for this experiment
* Asset tab: all results and figures


ludwig/contribs/comet.py
```diff
@@ -58,13 +68,46 @@ def train(self, *args, **kwargs):
...
+            model_definition = args[1]
+            model_definition_path = args[2]
+            if model:
+                self.experiment.set_model_graph(str(model.graph.as_graph_def()))
+            if model_definition:
+                if model_definition_path:
+                    base_name = os.path.basename(model_definition_path)
+                else:
+                    base_name = "model_definition.yaml"
+                if "." in base_name:
```
w4nderlust: Not 100% sure, why do you need to do this.

w4nderlust:
This is really cool @dsblank !
I added one little comment for one thing that I didn't get 100% regarding swapping yaml and json extensions.
Other two little considerations:
- I think it would be great if there was a vay to visualize the model definition in one of the comet tabs, not sure which one, ideally hyperparameters, but there you have key-values, while in Ludwig model definitions you have nested structures that don't fit in that pattern. Maybe in notes as a yaml text? Not sure but would be cool.
- now I feel like the only missing piece is keeping track of the measures at the end of the epoch. This part of the training loop https://github.com/uber/ludwig/blob/master/ludwig/models/model.py#L542-L587 computes the measures on training validation and test. They are added to `progress_tracker`. You could add an additional contrib function there, maybe `train_end_epoch` or something like this, so that you can send the contents of the progress tracker to comet to keep track of those per-epoch measures too alongside the per-batch measures that you already track through tensorflow. Let me know if this makes sense to you.

ludwig/contribs/comet.py
dsblank: Not sure either, but it appears that at this point in the process, model_definition is a JSON data structure and so we save it as json.

dsblank:
@w4nderlust Thanks! It does a lot already, and I think we can continue to tweak this as we move forward.

We have plans to visualize model graphs for a large number of types of ML frameworks. Stay tuned!

I will take a look at injecting another contrib method there. 

Thanks for the feedback and ideas!

ludwig/contribs/comet.py
w4nderlust: I checked, in that place `model_definition` has already been parsed and merged with the defaults, so it is a nested dictionary. You are passing the full dictionary to `self.experiment.log_asset_data` so I guess what that function does behing the scenes is to serialize the dictionary into a json file, so you need to provide a json file name, but that's something your function does, not Ludwig I believe. So depending of what you want to do, you may change the `self.experiment.log_asset_data` to provide also a strategy parameter on how to serialize (maybe default is json, but the user may specify yaml if they want), or maybe you want to make explicit in the docs of that function that, if a dictionary is provided, it is going to be serialized as json. Whatever works better for you.

Issue #286, opened 2019-04-14
Web service wrapper around ludwig


bertyhell:
I'm trying to make ludwig more accessable from other languages by wrapping it in a webservice (cherrypy)

When i try to make a call to the webservice to train a model i get the error:
```
 signal only works in main thread
```

I'm looking for some context around how these signals are used inside ludwig. I assume this could be resolved by starting ludwig in a separate process and then setup a socket connection between the ludwig process and the cherrypy web service.

This is the relevant code i have so far:

**service in cherrypy:**
```python
class AiWebService(object):

    @cherrypy.expose
    @cherrypy.tools.json_in()
    @cherrypy.tools.json_out()
    def train(self):
        data = cherrypy.request.json
        if 'training_samples' not in data:
            return RecursiveError('"training_samples" property is required on body to begin training').to_json()
        if 'definitions' not in data:
            return RecursiveError('"definitions" property is required on body to begin training').to_json()

        training_samples = pd.DataFrame(data.get('training_samples'))
        definitions = data.get('definitions')
        output = p.train(training_samples, definitions)
        return output

if __name__ == '__main__':
    cherrypy.config.update({'server.socket_host': '0.0.0.0'})
    cherrypy.quickstart(AiWebService())
```
https://gitlab.com/bertyhell/woody-puzzle-solver/blob/master/src/server/ludwigproxy.py

**wrapper around the ludwig api:**
```python
    def train(self, training_samples, model_definitions):
        model = LudwigModel(model_definitions)
        train_stats = model.train(training_samples)
        return {'message': 'ok'}
```
https://gitlab.com/bertyhell/woody-puzzle-solver/blob/master/src/server/service.py

**full error:**
```
ERROR:cherrypy.error.2203522445776:[14/Apr/2019:11:34:47] HTTP 
Traceback (most recent call last):
  File "C:\Users\verhe\AppData\Local\Programs\Python\Python36\lib\site-packages\cherrypy\_cprequest.py", line 628, in respond
    self._do_respond(path_info)
  File "C:\Users\verhe\AppData\Local\Programs\Python\Python36\lib\site-packages\cherrypy\_cprequest.py", line 687, in _do_respond
    response.body = self.handler()
  File "C:\Users\verhe\AppData\Local\Programs\Python\Python36\lib\site-packages\cherrypy\lib\encoding.py", line 219, in __call__
    self.body = self.oldhandler(*args, **kwargs)
[...]

PR #287, opened 2019-04-15
fix: Avoid using signal when not on main thread

bertyhell:
# Code Pull Requests

Please provide the following:

* a clear explanation of what your code does
    * This change makes sure no signals are fired if the ludwig code isn't called from the main thread. This should avoid errors when using ludwig inside a route inside a webservice like CherryPy
* if applicable, a reference to an issue
    * https://github.com/uber/ludwig/issues/286
    * clone this repo: https://gitlab.com/bertyhell/woody-puzzle-solver
    * checkout this commit: https://gitlab.com/bertyhell/woody-puzzle-solver/blob/50f84fde9db8f4aa80c86fb3d7fb8ba026172163/src
    * Run this file in python: https://gitlab.com/bertyhell/woody-puzzle-solver/blob/50f84fde9db8f4aa80c86fb3d7fb8ba026172163/src/server/service.py
    * Make post request: 
    ```
    curl -X POST \
  http://localhost:8080/train \
  -H 'Content-Type: application/json' \
  -H 'Postman-Token: 8a9abce5-f728-4f6e-99f7-99492677e419' \
  -H 'cache-control: no-cache' \
  -d '{"definitions" : {"input_features": [],	"output_features": []}, "training_samples":[]}'
    ```

### DIFF
--- a/ludwig/models/model.py
+++ b/ludwig/models/model.py
@@ -28,6 +28,7 @@ import re
 import signal
 import sys
 import time
+import threading
 from collections import OrderedDict
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


w4nderlust:
Thanks for this PR! Would you consider also contributing the cherrypy server itself? It owuld be really useful!

PR #288, opened 2019-04-15
Speedup integration tests

msaisumanth:
Speeds up integration tests
1.  Number of datapoints is reduced to 25 (further reductions didn't show much value) 
2. Some unncessary tests are reduced - tests which were using features/encoders that are already tested
3. Image sizes are reduced to 8x8 from 10x10

Testing time drops by ~20% overall on my laptop. All tests pass (repeated trials)


msaisumanth:
Testing time (end-to-end) is actually dropping almost in half (time python -m pytest).

PR #290, opened 2019-04-16
Adds data_dict to api.py: train()

msaisumanth:
Adds data_dict to the train method in api.py

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -271,6 +271,7 @@ class LudwigModel:
             data_train_hdf5=None,
             data_validation_hdf5=None,
             data_test_hdf5=None,
+            data_dict=None,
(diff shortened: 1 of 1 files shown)


w4nderlust:
In `train` we are doing:
```
if data_df is None:
            data_df = self._read_data(data_csv, data_dict)
            data_df.csv = data_csv
```
Why can't we do the same here too?

msaisumanth:
I'll redo this as we have updated api.py

msaisumanth:
Oh the reason for not doing it the same way as train_online was this:
If we do, data_df = self._read_data(data_csv, data_dict). Both data_csv and data_df may be not None. This will lead to some redundant data processing in full_train

PR #291, opened 2019-04-17
Refactor the preprocess module. Removes redundant code for data frames

msaisumanth:
1. Removes buggy code for preprocessing data frames
   > Currently there's logic to save the DF to hdf5 format after preprocessing. However, we don't have a 
      folder or filename to save the data to. If someone tries the api with a dataframe and
      skip_save_processed_input=False, things would fail.
2. Refactors the preprocess_for_training method. separates the logic for csvs, hdf5 and dataframes. 
3. Adds a util method to get the json/hdf5 file names corresponding to csvs. 

All tests pass

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -323,7 +323,8 @@ def preprocess_for_training(
     if all(data_sources_none):
         raise ValueError('No training data is provided!')
 
-    # Check if hdf5 and json already exist
+    # Check if hdf5 and json already exist. If they do, use the hdf5 data,
+    # instead of the csvs
     data_hdf5_fp = None
 
     if data_csv is not None:
@@ -393,6 +394,7 @@ def preprocess_for_training(
     ) = get_dataset_fun(dataset_type)
 
     if data_df is not None or data_train_df is not None:
+        # Preprocess data frames
         training_set, test_set, validation_set, train_set_metadata = \
             _preprocess_df_for_training(
                 features,
@@ -404,6 +406,7 @@ def preprocess_for_training(
                 random_seed
             )
     elif data_csv is not None or data_train_csv is not None:
+        # Preprocess csv data
         training_set, test_set, validation_set, train_set_metadata = \
             _preprocess_csv_for_training(
                 features,
@@ -510,6 +513,19 @@ def _preprocess_csv_for_training(
         preprocessing_params=default_preprocessing_parameters,
         random_seed=default_random_seed
 ):
+    """
+    Method to pre-process csv data
+    :param features: list of all features (input + output)
+    :param data_csv: path to the csv data
+    :param data_train_csv:  training csv data
+    :param data_validation_csv: validation csv data
+    :param data_test_csv: test csv data
+    :param skip_save_processed_input: if False, the pre-processed data is saved
+    as .hdf5 files in the same location as the csvs with the same names.
+    :param preprocessing_params: preprocessing parameters
+    :param random_seed: random seed
+    :return: training, test, validation datasets, training metadata
+    """
     if data_csv is not None:
         # Use data and ignore _train, _validation and _test.
         # Also ignore data and train set metadata needs preprocessing
@@ -610,6 +626,10 @@ def _preprocess_df_for_training(
         preprocessing_params=default_preprocessing_parameters,
         random_seed=default_random_seed
 ):
+    """ Method to pre-process dataframes. This doesn't have the optoin to save the
+    processed data as hdf5 as we don't expect users to do this as the data can
+    be processed in memory
+    """
 
     if data_df is not None:
         # needs preprocessing
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -317,9 +317,9 @@ def preprocess_for_training(
         preprocessing_params=default_preprocessing_parameters,
         random_seed=default_random_seed
 ):
-    data_sources = [data_df, data_train_df, data_csv, data_train_csv,
-                    data_hdf5, data_train_hdf5]
-    data_sources_none = [x is None for x in data_sources]
+    data_sources_provided = [data_df, data_train_df, data_csv, data_train_csv,
+                             data_hdf5, data_train_hdf5]
+    data_sources_none = [x is None for x in data_sources_provided]
     if all(data_sources_none):
         raise ValueError('No training data is provided!')
 
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -395,29 +395,37 @@ def preprocess_for_training(
 
     if data_df is not None or data_train_df is not None:
         # Preprocess data frames
-        training_set, test_set, validation_set, train_set_metadata = \
-            _preprocess_df_for_training(
-                features,
-                data_df,
-                data_train_df,
-                data_validation_df,
-                data_test_df,
-                preprocessing_params,
-                random_seed
-            )
+        (
+            training_set,
+            test_set,
+            validation_set,
+            train_set_metadata
+        ) = _preprocess_df_for_training(
+            features,
+            data_df,
+            data_train_df,
+            data_validation_df,
+            data_test_df,
+            preprocessing_params,
+            random_seed
+        )
     elif data_csv is not None or data_train_csv is not None:
         # Preprocess csv data
-        training_set, test_set, validation_set, train_set_metadata = \
-            _preprocess_csv_for_training(
-                features,
-                data_csv,
-                data_train_csv,
-                data_validation_csv,
-                data_test_csv,
-                skip_save_processed_input,
-                preprocessing_params,
-                random_seed
-            )
+        (
+            training_set,
+            test_set,
+            validation_set,
+            train_set_metadata
+        ) = _preprocess_csv_for_training(
+            features,
+            data_csv,
+            data_train_csv,
+            data_validation_csv,
+            data_test_csv,
+            skip_save_processed_input,
+            preprocessing_params,
+            random_seed
+        )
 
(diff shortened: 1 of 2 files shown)


ludwig/data/preprocessing.py
```diff
@@ -316,17 +317,22 @@ def preprocess_for_training(
         preprocessing_params=default_preprocessing_parameters,
         random_seed=default_random_seed
 ):
-    # Check if hdf5 and json already exist
+    data_sources = [data_df, data_train_df, data_csv, data_train_csv,
+                    data_hdf5, data_train_hdf5]
+    data_sources_none = [x is None for x in data_sources]
```
w4nderlust: would probably flip the logic here, check if `is not None` and call the variable `data_sources_provided`

ludwig/data/preprocessing.py
```diff
@@ -316,17 +317,22 @@ def preprocess_for_training(
...
+    if all(data_sources_none):
```
w4nderlust: This becomes `any`

ludwig/data/preprocessing.py
```diff
@@ -316,17 +317,22 @@ def preprocess_for_training(
...
-    data_validation_hdf5_fp = None
-    data_test_hdf5_fp = None
-    train_set_metadata_json_fp = 'metadata.json'
+
     if data_csv is not None:
-        data_hdf5_fp = os.path.splitext(data_csv)[0] + '.hdf5'
-        train_set_metadata_json_fp = os.path.splitext(data_csv)[0] + '.json'
-        if (os.path.isfile(data_hdf5_fp) and
-                os.path.isfile(train_set_metadata_json_fp)):
+        data_hdf5_fp = get_filepath_in_diff_format(data_csv, 'hdf5')
```
w4nderlust: would probably call this function `replace_extension` to make more generic

ludwig/data/preprocessing.py
```diff
@@ -350,22 +359,24 @@ def preprocess_for_training(
...
-            data_validation_csv)[0] + '.hdf5'
+        data_validation_hdf5_fp = get_filepath_in_diff_format(
+            data_validation_csv,
+            'hdf5'
+        )
         if os.path.isfile(data_validation_hdf5_fp):
             logging.info(
                 'Found hdf5 with the same filename of '
-                'the validation csv, using it instead'
+                'the validation csv, using them instead'
```
w4nderlust: why them?

w4nderlust:
This is a great starting point @msaisumanth .
Left a few comments.
Also, probably before merging we want to do the same refactoring also to the `preprocess_for_prediction` function, and it may probably be the case that bot functions are goiing to use the same building block function that now are called `_preprocess_csv_for_training ` and `_preprocess_df_for_training ` and so on, but could probably be generalized with a caouple of additional parameters into `_preprocess_csv`, `_preprocess_df` etc. that are used by both training and predict.
Also we can apply a strategy pattern with a registry here potentially, so that adding a new format would be straightforward and would easily allow for tsv, parquet and other formats.

ludwig/data/preprocessing.py
```diff
@@ -600,6 +503,172 @@ def preprocess_for_training(
...
+    elif data_train_df is not None:
+        # needs preprocessing
+        logging.info('Using training dataframe')
+        logging.info('Building dataset (it may take a while)')
+        concatenated_df = concatenate_df(
+            data_train_df,
+            data_validation_df,
+            data_test_df
+        )
+        data, train_set_metadata = build_dataset_df(
```
ydudin3: Is there a way to combine this call with the one in line 638?

For example first get a combined dataframe: df_combined = data_df (if data_df is not None) or df_combiner is a concatenated dataframe if data_train_df is not none, etc.

And then call build_dataset_df on that dataframe.

ludwig/data/preprocessing.py
```diff
@@ -641,13 +710,12 @@ def preprocess_for_prediction(
         model_definition['preprocessing']
     )
 
-    # Check if hdf5 and json already exist
+    # Check if hdf5 file already exists
     if data_csv is not None:
-        data_hdf5_fp = os.path.splitext(data_csv)[0] + '.hdf5'
+        data_hdf5_fp = get_filepath_in_diff_format(data_csv, 'hdf5')
```
ydudin3: maybe call it replace_file_extension to make it more clear what it does?
We can also declare string constants for `csv` and `hdf5`.

ludwig/utils/data_utils.py
```diff
@@ -281,6 +282,22 @@ def load_from_file(file_name, field=None, dtype=int):
...
+    a.csv, hdf5 -> a.hdf5
+    :param file_path: original file path
+    :param desired_format: desired file format
+    :return: file path with same name but different format
+    """
+    if '.' in desired_format:
+        # Handle the case if the user calls with '.hdf5' instead of 'hdf5'
+        desired_format = desired_format.replace('.', '').strip()
+
+    return os.path.splitext(file_path)[0] + '.' + desired_format
```
ydudin3: maybe make it more robust using regex - in case filenames will have additional `.`

ydudin3: Never mind, misread.

ydudin3: or actually, why do we need this helper method since it's only one line?

ludwig/data/preprocessing.py
msaisumanth: sure. makes sense.

ludwig/data/preprocessing.py
msaisumanth: oops. copy paste error. good catch!

ludwig/data/preprocessing.py
msaisumanth: yes! missed that. will do now

ludwig/data/preprocessing.py
msaisumanth: yeah, was thinking about the string constants. Will add that.

ludwig/utils/data_utils.py
msaisumanth: It's one line but we're using it several times. Makes the code a little more modular and testable.

ludwig/data/preprocessing.py
msaisumanth: sure. no problem

msaisumanth:
@w4nderlust about reusing the code for preprocess for predict, the flow is quite different. For training, we are handling the part of splitting the data into train, test and validation as necessary. But that step is not necessary for predict. Let me see if they can still be combined

ludwig/data/preprocessing.py
msaisumanth: I'll have to change it to 'if not any' right?

w4nderlust: yes

ludwig/data/preprocessing.py
msaisumanth: done

ludwig/data/preprocessing.py
msaisumanth: done

ludwig/data/preprocessing.py
msaisumanth: done

msaisumanth:
Using strategy pattern for preprocessing is a good idea, but requires quite a bit more refactoring. Will address that in another diff

w4nderlust:
lgtm, @ydudin3 can you do another review pass too please? Just to be more confident, as this is a relatively big change.

ludwig/data/preprocessing.py
```diff
@@ -600,6 +503,161 @@ def preprocess_for_training(
...
+        )
+        if not skip_save_processed_input:
+            logging.info('Writing dataset')
+            data_hdf5_fp = replace_file_extension(data_csv, 'hdf5')
+            data_utils.save_hdf5(data_hdf5_fp, data, train_set_metadata)
+            logging.info('Writing train set metadata with vocabulary')
+
+            train_set_metadata_json_fp = replace_file_extension(data_csv,
+                                                                'json')
+            data_utils.save_json(
```
ydudin3: Nit: stick everywhere to a consistent pattern. For example if you're opening a brace with an empty line, close it with an empty line as well. For example

x = foo(
    y
)

Not
x = foo(
    y)

ludwig/data/preprocessing.py
```diff
@@ -600,6 +503,161 @@ def preprocess_for_training(
...
+            preprocessing_params,
+            random_seed=random_seed
+        )
+        if not skip_save_processed_input:
+            logging.info('Writing dataset')
+            data_hdf5_fp = replace_file_extension(data_csv, 'hdf5')
+            data_utils.save_hdf5(data_hdf5_fp, data, train_set_metadata)
+            logging.info('Writing train set metadata with vocabulary')
+
+            train_set_metadata_json_fp = replace_file_extension(data_csv,
```
ydudin3: Nit: same as the comment below.
We should have either
x = foo(
    y,
    z
)

or 
x = foo(y,
             z)

everywhere.

I would opt for option 1 since it results narrower lines.

ludwig/data/preprocessing.py
```diff
@@ -382,139 +393,31 @@ def preprocess_for_training(
...
-        )
-        if not skip_save_processed_input:
-            logging.info('Writing dataset')
-            data_utils.save_hdf5(
-                data_train_hdf5_fp,
-                training_set,
-                train_set_metadata
+    if data_df is not None or data_train_df is not None:
+        # Preprocess data frames
+        training_set, test_set, validation_set, train_set_metadata = \
```
ydudin3: Nit: don't use backslashes. Can do this instead
(
    training_set,
    test_set,
    validation_set,
    train_set_metadata
) = (
    ....
)

Issue #293, opened 2019-04-17
KeyError: 'tied_weights'


Rhasz:
**Describe the bug**
Using API to load a pre trained model and getting the error
KeyError: 'tied_weights'

**To Reproduce**
Run in Google Colab
Python code so indents may need to be added for loops and conditionals for code to work.

**Imports and Installations**
_from __future__ import print_function
import math
import time
from IPython import display
from matplotlib import cm
from matplotlib import gridspec
from matplotlib import pyplot as plt
import numpy as np
import pandas as pd
from sklearn import metrics
import tensorflow as tf
from tensorflow.python.data import Dataset
tf.logging.set_verbosity(tf.logging.ERROR)
pd.options.display.max_rows = 10
pd.options.display.float_format = '{:.2f}'.format
!pip3 install ludwig
import yaml
from ludwig import LudwigModel_

**ModelDefinition:**
_dict = {'input_features': [{'name': 'Open_Price', 'type': 'numerical'}], 'output_features': [{'name': 'Closing_Price', 'type': 'numerical'}], 'training':{'epochs': 3000}}
model_definition = dict 
ludwig_model = LudwigModel(model_definition)_

**Sample Data:**
The full code pulls data from an exchange through their API, the below is a modified version to generate a sample.
Data is generated in code
Create a data frame with first series having numerical and a second series having numerical...
Series are generated in a loop using append

Replace stock_bars[i].o and stock_bars[i].o with a constant like 200.00 in following code to generate sample series

_limit = 1000
stock_open = []
stock_close = []
if len(stock_open) == 0:
   for i in range(limit):
       stock_open.append(stock_bars[i].o)_

_if len(stock_close) == 0:
   for i in range(limit):
          stock_close.append(stock_bars[i].c)_

_df = pd.DataFrame({ "Open_Price": pd.Series(stock_open), "Closing_Price": pd.Series(stock_close)})
print(df)
df.describe()_

**Training**
_train_stats = ludwig_model.train(df)_

**Saving** (Drive needs to be mounted before saving, will need to also replace stock_symbol)

_save_path = "gdrive/My Drive/Colab Notebooks/" + stock_symbol
ludwig_model.save(save_path)_

**Loading**

_model_dir = "gdrive/My Drive/Colab Notebooks/" + stock_symbol
ludwig_model = []
ludwig_model = LudwigModel.load(model_dir)_

Please provide code, yaml definition file and a sample of data in order to entirely reproduce the issue.
[...]

PR #294, opened 2019-04-18
Fixes a minor bug with tied_weights.

msaisumanth:
For some features (ex: numerical) there is no default value for tied_weights. So inputs.py throws an error when we try to access the tied_weights feature.

fixes issue #293


msaisumanth:
the default value for .get is None. So the extra argument is not really necessary, but I can add it if you think it improves readability.

w4nderlust:
Yes please, add the None to avoid confusion

msaisumanth:
done

PR #298, opened 2019-04-20
Refactors experiment.py, api.py to use the full_train method from train.py

msaisumanth:
1. experiment.py, api.py and train.py all use more or less the exact same code for model training. This commit removes the redundancy. All of them now use the full_train method from train.py
2. Adds a test case for api.py. I modified one of the test cases for experiment to use api.py
3. Refactored some of the integration tests

All tests pass.


w4nderlust:
Great jos @msaisumanth ! It also solves one bug (the use of max instead of best function in train.py). Merging this, we should do the same for predict now.

PR #303, opened 2019-04-24
Replaces only_predict with evaluate_performance in predict.py, apy.py  and in cascade everywhere needed

w4nderlust:


msaisumanth:
Looks good! You will have many conflicts with my preprocessing cleanup diff. So lookout for that.

PR #311, opened 2019-04-26
Fixes a model definition bug causing failures when using model_definition_file

msaisumanth:
1. currently model_definition is missing in experiment.py, api.py if a model_definition_file was used. Fixing it by having full_train return the updated model_definition2
2. Modifies one of the integration tests to use a model_definition file instead of a dictionary.
     >> I re-used one of the existing tests, so as to not increase the testing time.

### DIFF
--- a/tests/integration_tests/test_api.py
+++ b/tests/integration_tests/test_api.py
@@ -39,7 +39,7 @@ def run_api_experiment(input_features, output_features, data_csv):
         output_name=output_features
     )
 
-    model = LudwigModel(yaml.load(model_definition))
+    model = LudwigModel(yaml.safe_load(model_definition))
 
     # Training with csv
     model.train(
--- a/tests/integration_tests/test_api.py
+++ b/tests/integration_tests/test_api.py
@@ -22,6 +22,7 @@ from ludwig.utils.data_utils import read_csv
 from tests.integration_tests.utils import ENCODERS
 from tests.integration_tests.utils import generate_data
 from tests.integration_tests.utils import model_definition_template
+from tests.fixtures.filenames import csv_filename
 
 
 def run_api_experiment(input_features, output_features, data_csv):
--- a/tests/integration_tests/test_api.py
+++ b/tests/integration_tests/test_api.py
@@ -22,6 +22,7 @@ from ludwig.utils.data_utils import read_csv
 from tests.integration_tests.utils import ENCODERS
 from tests.integration_tests.utils import generate_data
 from tests.integration_tests.utils import model_definition_template
+# The following imports are pytest fixtures, required for running the tests
 from tests.fixtures.filenames import csv_filename
 
 
--- a/tests/integration_tests/test_experiment.py
+++ b/tests/integration_tests/test_experiment.py
@@ -47,7 +47,7 @@ def run_experiment(input_features, output_features, data_csv):
     )
 
     experiment(
-        yaml.load(model_definition),
+        yaml.safe_load(model_definition),
         skip_save_processed_input=True,
         skip_save_progress=True,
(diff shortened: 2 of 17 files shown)


tests/integration_tests/test_experiment.py
```diff
@@ -28,6 +28,9 @@
 from tests.integration_tests.utils import ENCODERS
 from tests.integration_tests.utils import generate_data
 from tests.integration_tests.utils import model_definition_template
+# The following imports are pytest fixtures, required for running the tests
+from tests.fixtures.filenames import csv_filename
+from tests.fixtures.filenames import yaml_filename
```
ydudin3: Is this used anywhere?

tests/integration_tests/test_experiment.py
```diff
@@ -28,6 +28,9 @@
 from tests.integration_tests.utils import ENCODERS
 from tests.integration_tests.utils import generate_data
 from tests.integration_tests.utils import model_definition_template
+# The following imports are pytest fixtures, required for running the tests
+from tests.fixtures.filenames import csv_filename
```
ydudin3: Is this used anywhere?

tests/integration_tests/test_api.py
```diff
@@ -22,6 +22,8 @@
 from tests.integration_tests.utils import ENCODERS
 from tests.integration_tests.utils import generate_data
 from tests.integration_tests.utils import model_definition_template
+# The following imports are pytest fixtures, required for running the tests
+from tests.fixtures.filenames import csv_filename
```
ydudin3: Is this used anywhere?

ydudin3: Or I guess the comment above says so.

tests/integration_tests/test_experiment.py
```diff
@@ -77,6 +80,34 @@ def test_experiment_seq_seq1(csv_filename):
...
+    output_features = ('[{name: iob, type: text, reduce_input: null, '
+                       'vocab_size: 3, min_len: 10, max_len: 10,'
+                       ' decoder: tagger}]')
+
+    # Save the model definition to a yaml file
+    model_definition = yaml.load(model_definition_template.substitute(
+        input_name=input_features,
+        output_name=output_features
+    ))
+    yaml.dump(model_definition, open(yaml_filename, 'w'))
```
ydudin3: Yep, I think it'd be safer to use `with`

tests/integration_tests/test_experiment.py
```diff
@@ -77,6 +80,34 @@ def test_experiment_seq_seq1(csv_filename):
         run_experiment(input_features, output_features, data_csv=rel_path)
 
 
+def test_experiment_seq_seq1_model_def_file(csv_filename, yaml_filename):
+    # seq-to-seq test to use model definition file instead of dictionary
+    input_features = ('[{name: utt, type: text, reduce_output: null, '
```
ydudin3: Would it make sense to create a util to convert a dictionary to a yaml string so instead of declaring strings we can declare dictionaries (more readable) and then convert them to strings using a util function.

tests/integration_tests/test_experiment.py
msaisumanth: yeah, every test is using the csv_filename (pytest fixture). Essentially a random filename is generated for each test and the file is deleted after the test runs.

tests/integration_tests/test_experiment.py
msaisumanth: will fix this

tests/integration_tests/test_experiment.py
msaisumanth: great point. let me try.

PR #316, opened 2019-04-30
Refactor tests

msaisumanth:
1. Now the tests use dictionaries for model definition instead of yaml files
2. Helper methods with default values for each feature type for the tests
3. Tests are faster because all the parameters are chosen as small as possible
4. All tests delete the intermediate data saved to disk -> model, predictions etc.
5. One of the new tests saves processed input data (and deletes after). Another new test uses train/test/valid files instead of just using data_csv. 

All tests pass.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -138,6 +138,7 @@ class LudwigModel:
             self.model_definition = merge_with_defaults(model_definition)
         self.train_set_metadata = None
         self.model = None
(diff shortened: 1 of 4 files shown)


msaisumanth:
> looks good to me.
> Little notes: make sure to run pycharm's cleanup, import optimization etc + don't forget about the header in all the files

Done. Thanks for the headsup

PR #320, opened 2019-05-02
Update user_guide.md

carlogrisetti:
Fix copy\paste typo on numerical features


w4nderlust:
Thank you! Can you please update `docs/userg_guide/index.html`? I prefer not to have two separate commits, so if you do it here i can squash and merge them.

PR #322, opened 2019-05-03
Fix copy\paste typos on binary feature in place of numerical\set\category feature

carlogrisetti:
Fix copy\paste typos on binary feature in place of numerical\set\category feature


carlogrisetti:
Hah... not used to GitHub :)
I looked for a way to squash commits but either is something only you can do or (most probably) something I should have done before pulling.
In any case, here it is. I got some more looking carefully.

w4nderlust:
I can do it as long as they are in the same PR, don't worry ;)

PR #326, opened 2019-05-06
Add GCS support to read files via: data_csv, data_hdf5.

gogasca:
Training data may exists in Google Cloud Bucket

- This PR adds GCS support to read files via: data_csv, data_hdf5 train_set_metadata_json. 
- Use gcsfs library to read files. Since gcsfs does not support full hdf5 we copy files to /tmp folder and read from there. 
- Minor refactor for requirements.txt and setup.py
- Add requirements-gcs to support GCS

```
ludwig train --data_csv gs://ludwig/Tweets.csv --model_definition_file model_definition.yaml --output_directory /tmp/
```
```
ludwig train --data_hdf5 gs://ludwig/Tweets.hdf5 --train_set_metadata_json gs://ludwig/Tweets.json --model_definition_file model_definition.yaml --output_directory /tmp/
```
https://pastebin.com/Z5t4JMWZ

model_definition.yaml

```
input_features:
    -
        name: content
        type: text

output_features:
    -
        name: airline_sentiment
        type: category

training:
        epochs: 2
```

Feature request:
https://github.com/uber/ludwig/issues/327


gogasca:
Build reports error
```When a method is defined in a child class which has the same name as an attribute in the ancestor class, the attribute in the ancestor class takes precedence and hides the method.```
which is part of original code.

msaisumanth:
@gogasca Thanks a lot for creating this pr!! Could you please check if your change is breaking the tests? Run `python -m pytest` from the ludwig main directory.

gogasca:
> python -m pytest
@msaisumanth I run the command above and I get this:
21 passed, 160 warnings in 153.95 seconds 
Here is the result: https://pastebin.com/fNeF5rBK

When I submitted this PR, DeepSource keep throwing an error on existing code before this PR was created. (Correct me if I'm wrong)

w4nderlust:
Thank you, it will take a little wile to evaluate the pr as it is pretty big, but be sure we will pay close attention to it.

gogasca:
> Thank you, it will take a little wile to evaluate the pr as it is pretty big, but be sure we will pay close attention to it.

Thanks for taking a look, -Gonzalo

gogasca:
@w4nderlust wondering if I should reduce this PR to make it more easy to review? Thanks

w4nderlust:
@gogasca sorry for the long time this is taking. The reason is that we are considering a different way of loading data entirely at the moment https://eng.uber.com/petastorm/ and trying to think of ways to make it and gcp work together is a bit trickier than originally imagine. We really appreciate your contribution anyway, it's really good stuff, we are just not sure yet if merging this pr could actually make things harder if we decide to move to petastorm. And at the same time we are still putting out smaller fires :) Hope you would understand.

PR #332, opened 2019-05-09
Fix visualize parameter naming

carlogrisetti:
Renaming Prediction Stats to Test Stats (i'm taking the published website documentation as the most valid)

Inverting "trs" and "tes" in the user guide on the idea that TRS->TRainingStatistics and TES -> TEstStatistics

Updating mkdocs to reflect published html, since it seems to still contain the "old" name (prediction statistics).


w4nderlust:
thanks for contributing this

PR #334, opened 2019-05-09
 Import contribs from api and cli

dsblank:
This PR does three things:

* adds contrib_import() to cli - ludwig.cli used to import ludwig.api, but now doesn't so need additional call
* checks so that calling contrib_import() twice doesn't doubly instantiate contrib
* surround instantiation of contrib with try/except to protect from bad constructors

After PR:

* contribs will be instantiated in api and cli
* multiple calls of contrib_import() will only create one instance
* won't crash with bad contrib constructor


w4nderlust:
Thank you @dsblank !

PR #338, opened 2019-05-13
Update visualize.py

carlogrisetti:
match prediction_statistics -> test_statistics to have the same parameter name within functions

PR #332 introduced a misalignment between parameter calling and actual parameter name in the called function. This aligns that all to "test_statistics".

There's one more place where "prediction_statistics" is used, and it's in the evaluate performance part of predict.py. Do you want to change that naming to "test_statistics" also?
I don't know how to proceed there, because leaving "prediction_statistics.json" there could lead to confusion on how to ingest it, as you then require some test statistics in visualize.py, but on the other hand that is a prediction statistic (like we are running predictions and optionally evaluating performances), not a test one (visualize is ingesting prediction-data, so you may want to explicit this duality in the docs?)

### DIFF
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -1448,7 +1448,7 @@ def calibration_multiclass(
 
 
 def confusion_matrix(
-        prediction_statistics,
+        test_statistics,
         ground_truth_metadata,
         field,
         top_n_classes,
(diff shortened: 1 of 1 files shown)


w4nderlust:
I did some thinking about it. I'm thinking about introducing a `test.py` script that does exactly the same thing as predict, but with the default behavior of evaluating the quality of the predictions. i will rename a bunch of parameters accordingly, so let's keep this PR on hold for a little and let's reconsider after i make the change. Thank you for submitting it anyway!

carlogrisetti:
You should revert the other one, or merge this one.
At the current state there's no way to pass test statistics / prediction statistics between the arg parser and the methods called, and Ludwig visualize fails

w4nderlust:
Got it, so I'm merging this, but the jury is still out on what's the best naming convention :)

carlogrisetti:
Yeah, I wouldn't be sure either,... I was before, since I saw the intent to standardize under test\training in place of prediction, but now...
Sorry for the half-assed PR that caused this empasse :)

w4nderlust:
No worries, I should have checked more thoroughly.

PR #342, opened 2019-05-14
Refactors the sequence encoders code and adds some unittests

msaisumanth:
1. Refactors the sequence_encoders module. The handling of num_fc_layers and fc_layers was not clean.
2. Adds some tests for sequence encoders. I was struggling with ParallelConv1D. I need to add a test for that. 
3. Removes a duplicate test file

All tests pass.


w4nderlust:
lgtm, wait on @ydudin3 's review to merge

tests/ludwig/models/modules/test_encoder.py
```diff
@@ -38,6 +47,25 @@ def generate_images(image_size, num_images):
     return np.array([_generate_image(image_size) for _ in range(num_images)])
 
 
+def _generate_sentence(vocab_size, max_len):
+    sentence = np.zeros(max_len, dtype=np.int32)
+    random_length = random.randint(1, max_len)
+    sentence[:random_length] = [random.randint(0, vocab_size-1) for _ in
```
ydudin3: Can just use np.random of a given shape instead of for loop?

tests/ludwig/models/modules/test_encoder.py
```diff
@@ -171,3 +196,108 @@ def test_image_encoders_stacked_2dcnn():
...
+            )
+
+            embed = encoder.embed_sequence.embed
+            assert embed.representation == 'dense'
+            assert embed.embeddings_trainable == trainable
+            assert embed.regularize is True
+            assert embed.dropout is False
+
+
+def test_sequence_encoders():
```
ydudin3: Add a test for unhappy paths - bad encoder parameter initialization and make sure they return errors?

Issue #331, opened 2019-05-09
Single object detection and boxing


carlogrisetti:
I'm having a use case where I do not only have to classify an image on a (small) set of categories, but I also need to do a really basic object detection, that can be accomplished via the image classification itself, since it will always be a single object in the image, but I will also need to have a rough estimate on where a specific feature of this object is, in order to perform further analysis on that part of the image.
I cannot complete this using other techniques, because the object I'm focusing on can be easily identified by a human, but has subtle variances between each instance that forces the solution into an AI-driven approach.

Is there some simplification in implementing some sort of single object detection into Ludwig, as opposite to a multi-object detection, or can that be naively trained with an image input and adding some numeric outputs (for ROI coordinates) ?

In the latter, a simple custom "box" output type extension, that wraps around existing numerical output features could be useful in abstracting this to a higher level.

I know you are focused on bugs, and I could take on this extension, but I wanted to pick your mind in the feasibility of such an approach.

Bye!

w4nderlust:
A box output type (or more generic vector potentially) would be a nice addition.
One could train it with the sum of mse on each dimension and use iou or jaccard index as a measure of quality.
It shouldn't be hard to implement, if you are interested we can chat about it.
Moreover, I believe using [CoordConv](https://arxiv.org/abs/1807.03247) would be greatly beneficial for the quality of the results. It will be super simple to implement it.

As a REALLY SUBOPTIMAL hacky solution, you can have 4 numerical output features for x, y, h and w respectively and see how far can you get with it.

carlogrisetti:
Really suboptimal but i'm having great results with a simple numerical output, with initializer: ones
I really hope that's not only due of my net having memorized the inputs, because images are so similar, so I'm waiting for a bigger dataset to be ready, and in the meantime I'm playing with hyperparameters.

I took a look at what you proposed, but it's really out of my league to implement :)

Pull the "help wanted" tag, and just leave this as enhancement if you want

nitrofortin:
I would be interested in implementing this enhancement. Could you just point me where the solution would live? Thanks!

w4nderlust:
@nitrofortin  thanks for offering to help!
The ideal place to put it would be in the `features` package, something like `boundingbox_feature.py`.
It would have to contain both a class improting from `BaseFeature` and one importing from `OutputFeature` and should implement all the methods that are defined in those objects.
[...]

PR #344, opened 2019-05-15
[WIP] Add bounding box feature

jstpierref:
# PR details
This PR refers to issue #331. It adds a `bounding_box` output feature to the library. This way the user can train models with `bounding_box` output like it is encountered in object detection. 

Currently, the only implemented measure is IOU (Intersection-Over-Union). The current default loss is the smoothed L1 loss (also called Huber loss) used by the Fast R-CNN algorithm. I plan to add the YOLO v3 localization loss as well. The following snippet runs flawlessly, but I might have forgotten key additions to make this feature usable so feel free to change it/let me know!

Basic test that I ran:
```python
from sklearn import datasets
import pandas as pd 
import numpy as np

iris = datasets.load_iris()
training_dataframe = pd.DataFrame(iris['data'], columns=['a','b','c','d'])
training_dataframe['target'] = list(map(lambda x: np.array([x,x,x,x]), iris['target']))

# train a model
model_definition = {
	'input_features': [
		{'name': 'a', 'type': 'numerical'},
		{'name': 'b', 'type': 'numerical'},
		{'name': 'c', 'type': 'numerical'},
		{'name': 'd', 'type': 'numerical'}
	], 
	'output_features': [{'name': 'target', 'type': 'bounding_box'}]
}
model = LudwigModel(model_definition)
train_stats = model.train(training_dataframe)
predictions = model.predict(training_dataframe)
model.close()
```

# Documentation 
Doc not updated yet


w4nderlust:
@nitrofortin did a first pass, you're definitely on the right track, thank you for contributing this!
Will do a second pass and leave some specific comments.
One generic thing: usually before committing we run pycharm code refactoring for optimizing imports and whitespaces etc, could you please do the same (in pycharm's ui when you commit there's a box in the top right of the widnow that does all the clode cleanup and checks).

nitrofortin:
Yes sure!

ludwig/features/boundingbox_feature.py
```diff
@@ -0,0 +1,333 @@
...
+
+    @staticmethod
+    def add_feature_data(
+            feature,
+            dataset_df,
+            data,
+            metadata,
+            preprocessing_parameters,
+    ):
+        data[feature['name']] = np.stack(dataset_df[feature['name']],
```
w4nderlust: How do you expect the content of the column containing the bounding box to look like? I imageine something like this:
```
image_path,bbox
path.jpeg,10 20 15 15
```
so the bbox column should contain 4 values divided by space.
In this case, I believe the processing should be different from what you are doing, but maybe you are thinking about another format.

ludwig/features/boundingbox_feature.py
```diff
@@ -0,0 +1,333 @@
...
+                self.name
+            )
+
+            iou_val = get_bbox_iou(
+                targets,
+                predictions,
+                self.name
+            )
+
+            mean_average_precision = None
```
w4nderlust: usually in ludwig we don't compute means in the graph because when we evaluate we want the means to be computed on the whole set rather than on the specific batch. There is a mechanism for collecting all the outputs of the batches when evaluating and then performa the average just one time. look at the code in model.py that uses OutputConfig for more details.

ludwig/features/boundingbox_feature.py
```diff
@@ -0,0 +1,333 @@
...
+
+    output_config = OrderedDict([
+        (LOSS, {
+            'output': EVAL_LOSS,
+            'aggregation': SUM,
+            'value': 0,
+            'type': MEASURE
+        }),
+        ('mean_iou', {
+            'output': 'iou',
```
w4nderlust: 'iou' should be a constant

ludwig/features/boundingbox_feature.py
```diff
@@ -0,0 +1,333 @@
...
+        if PREDICTIONS in result and len(result[PREDICTIONS]) > 0:
+            postprocessed[PREDICTIONS] = result[PREDICTIONS]
+            if not skip_save_unprocessed_output:
+                np.save(
+                    npy_filename.format(name, PREDICTIONS),
+                    result[PREDICTIONS]
+                )
+            del result[PREDICTIONS]
+
+        if PROBABILITIES in result and len(result[PROBABILITIES]) > 0:
```
w4nderlust: this can probably be removed

ludwig/models/modules/measure_modules.py
```diff
@@ -158,3 +158,25 @@ def r2(targets, predictions, output_feature_name):
     r2 = tf.subtract(1., res_ss / tot_ss,
                      name='r2_{}'.format(output_feature_name))
     return r2
+
+
+def bbox_iou(targets, predictions, output_feature_name):
+    x11, y11, x12, y12 = tf.split(targets, 4, axis=1)
```
w4nderlust: Not sure why are you transposing few lines from here, can you add a comment on what shapes do you expcet the inputs to have?

ludwig/features/boundingbox_feature.py
nitrofortin: It is currently accepting bounding boxes as np arrays, but I agree that 4 values divided by space  is more convenient especially when dealing with files. I guess it means the bbox values would be handled as strings in python, right?

ludwig/features/boundingbox_feature.py
nitrofortin: Done

ludwig/features/boundingbox_feature.py
nitrofortin: Done

ludwig/models/modules/measure_modules.py
nitrofortin: I removed the transposed variables.

nitrofortin:
Pushed some changes following your comments! Let me know if there are other requirements you think this PR should match.

ludwig/models/modules/measure_modules.py
w4nderlust: Was looking into literature, it seems like it is more common to provide bounding boxes as: x, y h, w where x and y are the coordinates of the center, h w are height and width respectively.
Can you please doublecheck a couple implementations like faster rcnn and yolo to make sure we are doing the most common thing?

nitrofortin: Yes, I will have look over the week

w4nderlust:
> Pushed some changes following your comments! Let me know if there are other requirements you think this PR should match.

Looks good, thank you!
I would doublecheck the format of the bounding boxes (x y h w versus x1 y1 x1 y2) in the most common datasets for object detection. You could also potentially add it as a parameter of the decoder and implement both functions, but that we can do later.
I guess now a good test would be to test this on a simple dataset to confirm it works, and add that test to the test suite.
Finally documentation needs to be added in the user guide.
Thank you so much again for the work your are putting into this!

w4nderlust:
@nitrofortin any update on this?

nitrofortin:
@w4nderlust Still need to test it on a real example. And you are right, (x y h w) seems more common. To make it work we will need a feature extractor, and I am still looking into how Ludwig does this automatically. If you have any insights, let me know!

nitrofortin:
@w4nderlust Still need to test it on a real example. And you are right, (x y h w) seems more common. To make it work we will need a feature extractor, and I am still looking into how Ludwig does this automatically. If you have any insights, let me know!

ifokeev:
@nitrofortin up, please

carlogrisetti:
I could have some time to spare in trying to pull this in... Do you (@nitrofortin \ @w4nderlust) know what's still missing?
I will be doing another boxing network in the near future (I hope it's days from now) and would really like to leverage this WIP PR to find out if this is a viable way to go.

I'll try to tinker with it, but any help is appreciated

ludwig/models/modules/measure_modules.py
carlogrisetti: Moreover, it seems to me that the bounding box output feature has been placed in the input features. Am I reading that correctly?

nitrofortin:
Been a while since I last worked on this. I'll check and let you know.

w4nderlust:
> I could have some time to spare in trying to pull this in... Do you (@nitrofortin \ @w4nderlust) know what's still missing?
> I will be doing another boxing network in the near future (I hope it's days from now) and would really like to leverage this WIP PR to find out if this is a viable way to go.
> 
> I'll try to tinker with it, but any help is appreciated

It would be great if you could help @carlogrisetti , but if you can I would hold on until we complete the TF2 porting, as things may look different then. We are almost there, probably a couple more weeks. Is that a good time for you?

carlogrisetti:
Oh great news regarding the TF2 porting... I can surely see the value in doing this after the fact :)
It might not be aligned with my timetable but I think TF2 porting is more important than a minor rework on my side.

w4nderlust:
If you have time snd willingness can I suggest you to help with the tf2
porting instead? I can point you to specific self contained tasks that
would be easy but require care.

On Fri, 29 May 2020 at 00:16, Carlo Grisetti <notifications@github.com>
wrote:

> Oh great news regarding the TF2 porting... I can surely see the value in
> doing this after the fact :)
> It might not be aligned with my timetable but I think TF2 porting is more
> important than a minor rework on my side.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_uber_ludwig_pull_344-23issuecomment-2D635805046&d=DwMCaQ&c=r2dcLCtU9q6n0vrtnDw9vg&r=RrWxefUdVFeb7Dpg6EF4Ug&m=mJiXFn31X7CFFdi5o6KA6xYleI6MiA4gb0V_zEGbCU4&s=dGAyO56M-cJAvmKBuBj9stNQR6ytVRHKXJhVLXl63vY&e=>,
> or unsubscribe
> <https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_notifications_unsubscribe-2Dauth_AACVISG6YSYRIHEBW6PLTBLRT5OMZANCNFSM4HNFYAYA&d=DwMCaQ&c=r2dcLCtU9q6n0vrtnDw9vg&r=RrWxefUdVFeb7Dpg6EF4Ug&m=mJiXFn31X7CFFdi5o6KA6xYleI6MiA4gb0V_zEGbCU4&s=z7Bj4L8E5LX6k463aKzh3aeD6AF4L5SOA026A-HcYqs&e=>
> .
>
-- 
----
Piero Molino
Sr Research Scientist
UBER AI Labs

carlogrisetti:
Willingness, yes... time, maybe (it really depends upon daily tasks)... skill... well let's see...
The only concern is that it might be way over my level, since I have really limited experience with TF\Keras (hence my love for ludwig), but I can always try. Maybe the more tedious tasks can be the best suited ones...

Throw something at me and we shall see :)

w4nderlust:
> Throw something at me and we shall see :)

Give me an email or other means for reaching out to you and I can give you all the tedious tasks you want :)

gustavorps:
Any updates about this feature?

carlogrisetti:
I focused elsewhere, since I was able to effectively obtain the same result using an output feature in the form of a vector with TRBL pixel positions (but I think any notation would have been effective), since I start from the idea that the object I am looking for is always present and it's only matter of determining where

PR #348, opened 2019-05-19
Add TF summary logging for validate loss

carlogrisetti:
I know it's ugly but this serves the purpose I need. Having few data to work on, training loss and validate loss do not always go hand-in-hand, and this gives me an easy way to have an overview on how validation loss is performing (and infer how likely the model is to be successfully updated in the near future).

### DIFF
--- a/ludwig/models/model.py
+++ b/ludwig/models/model.py
@@ -202,7 +202,7 @@ class Model:
             )
 
             tf.summary.scalar('train_reg_mean_loss', self.train_reg_mean_loss)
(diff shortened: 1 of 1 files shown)


carlogrisetti:
However, I'm starting to suspect that the eval_combined_loss I used is not what I think it is, but just the evaulation of the loss on the training dataset, hence defies my purpose :(

I'm not seeing any real difference in the cases i'm trying to reproduce

carlogrisetti:
Nevermind, i'm adding a feature request since I'm not able to find the real value...

Issue #340, opened 2019-05-14
Logging issue when activating Comet contrib


Lothiraldan:
**Describe the bug**

When activating the Comet contrib, most of Ludwig log message disappears.

**To Reproduce**

Launch: `ludwig experiment --data_csv reuters-allcats.csv --model_definition_file model_definition.yaml -l info --comet`

You won't see the following output:
```
 _         _        _      
| |_  _ __| |_ __ _(_)__ _ 
| | || / _` \ V  V / / _` |
|_|\_,_\__,_|\_/\_/|_\__, |
                     |___/ 
ludwig v0.1.2 - Experiment

Experiment name: experiment
Model name: run
Output path: results/experiment_run_43


ludwig_version: '0.1.2'
```

**Expected behavior**

The log messages should be displayed when the Comet contrib is activated.

**Environment (please complete the following information):**
 - OS: Fedora
 - Version 28
- Python version: 3.6.8
- Ludwig version: 0.1.2

**Additional context**

I think the issue is that ludwig is using the root-level logger configured through `logging.basicConfig`. The comet contrib integration contains some logging calls, for example, https://github.com/uber/ludwig/blob/master/ludwig/contribs/comet.py#L56.

Those calls happen before any `basicConfig` call https://github.com/uber/ludwig/blob/master/ludwig/experiment.py#L461.

The issue with calling the root-level `logging.info`, `logging.error` and so on is that they will call `logging.basicConfig` on their own if the root logger is not configured yet https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L2065. The direct effect is that the first call to `logging.info` will configure the root logger with no configuration which will create a StreamHandler pointing to `/dev/stderr`.

The unfortunate side-effect is that calling `basicConfig` will do nothing as the root handler as already a handler so the root logger will not be set to the right log level and the stream handler will not point to the right device.

I would recommend moving from using the root logger and configure the logger through `basicConfig` to using a `ludwig` logger and configure it manually, it's not that more complex. I can help if wanted.

One last issue with using the root logger is when configuring the root logger to the debug level, all libraries which are logging will start displaying their log messages. That includes requests and is polluting the output. Using a separate logger would also solve this issue.

w4nderlust:
@Lothiraldan thanks for reporting this. It is indeed a bug to fix. Also notifying @dsblank as he is the main contributor of the comet integration.
[...]

PR #352, opened 2019-05-20
Configures logging in ludwig/__init__

msaisumanth:
issue #340


ludwig/__init__.py
```diff
@@ -13,4 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import sys
+import logging
+
 from ludwig.globals import LUDWIG_VERSION as __version__
+
+
+logging.basicConfig(level=logging.INFO, stream=sys.stdout, format='%(message)s')
```
Lothiraldan: You should configure the top-level ludwig logger instead of the root level logger, this way setting info or debug won't activate info or debug logs for other libraries. The gist of `basicConfig` should be quite easy to extract for your need: https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L1893

ludwig/api.py
```diff
@@ -60,6 +60,10 @@
 from ludwig.utils.print_utils import logging_level_registry
 
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
Lothiraldan: You don't need to set the level on each individual logger as log message propagate to their parent by default. So a logger `ludwig.api` will sends received messages to the logger `ludwig` that is configured on the right right log level.

ludwig/collect.py
```diff
@@ -273,11 +277,7 @@ def cli_collect_activations(sys_argv):
 
     args = parser.parse_args(sys_argv)
 
-    logging.basicConfig(
-        stream=sys.stdout,
-        level=logging_level_registry[args.logging_level],
-        format='%(message)s'
-    )
+    logger.setLevel(logging_level_registry[args.logging_level])
```
Lothiraldan: Do you want to set the log level for the whole package or only for the logger of this file? I think you might want to do something like: `getLogger("luwdig").setLevel(...)`

dsblank:
@msaisumanth @w4nderlust You probably need to address the above three points made by @Lothiraldan

ludwig/api.py
msaisumanth: In this case, we wanted to set specific levels for different modules.

ludwig/collect.py
msaisumanth: yeah, you're right. Will update.

ludwig/__init__.py
msaisumanth: Yeah, makes sense. Will update this as well.

PR #361, opened 2019-05-30
randomly sample from the generated beams

aan1092j:
This adds code to return a random result from the beam search results, instead of the top result.


w4nderlust:
Thanks @aan1092j for your contribution.
Two little things:
- can you please sign the CLA?
- can you add a binary parameter to decide if to return a random one or the top one? The parameter should be added also in the __init__ of the sequence decoder so that users can set it in the yaml file.

PR #363, opened 2019-06-04
[WIP] use a random index value

aan1092j:
--> changed the code so that a random index is sampled every time
--> will aadd another commit on top to make it a function parameter

### DIFF
--- a/ludwig/models/modules/recurrent_modules.py
+++ b/ludwig/models/modules/recurrent_modules.py
@@ -466,6 +466,13 @@ def recurrent_decoder(encoder_outputs, targets, max_sequence_length, vocab_size,
                 projection_layer=projection_layer)
 
             if beam_width > 1:
+                # sample a random index
+                # note: this will return the same i'th beam prediction for all
+                # the inputs, but if called more than once, it will return different
+                # indices.
+                random.seed(int(time.time()))
+                chosen_sequence_index = random.randint(0, beam_width-1)
+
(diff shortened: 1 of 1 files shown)


ludwig/models/modules/recurrent_modules.py
```diff
@@ -466,6 +466,13 @@ def decode(initial_state, cell, helper, beam_width=1,
                 projection_layer=projection_layer)
 
             if beam_width > 1:
+                # sample a random index
+                # note: this will return the same i'th beam prediction for all
+                # the inputs, but if called more than once, it will return different
+                # indices.
+                random.seed(int(time.time()))
```
w4nderlust: don't set the random seed here, Ludwig already has a global mechanism for managing the random seed

aan1092j: Currently, the seed is fixed for predict(). 
Subsequent calls to tf.random.uniform() or random.randint(), when calling predict() always return the same index value, which I think is due to the global seed, so I used time.Now() to add randomness.

I could change it to something like md5(time.Now()) (which is kinda hacky) to have the sampling be random.

w4nderlust: There are two problems with that: first you have to set the tf.random_seed not the random.seed, second you have to do it before running predict, but within a session. So you probably want to do it withing Model.predict(), around line 1226 and add a parameter `use_now_as_seed=False` in the function signature.

PR #370, opened 2019-06-07
Fixes links to edit documentation in github.io

IgorWilbert:
# Documentation Pull Request

## Description
Fixed links to "Edit this page"  in github.io. They gave a 404 not found error because they were pointing to the wrong directory.

## Reproduction Steps
Open any page in https://uber.github.io/ludwig/ (Getting Started, Examples ...) and click on the pencil next to the title of the page.

### DIFF
--- a/docs/api/index.html
+++ b/docs/api/index.html
@@ -5,88 +5,88 @@
 <!doctype html>
 <html lang="en" class="no-js">
   <head>
-    
+
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width,initial-scale=1">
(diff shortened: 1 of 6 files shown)


w4nderlust:
@IgorWilbert  thanks for your PR. The problem is that the html pages are generated automatically by mkdocs (look at the developers guide for details), so even if I accept this, next time we are going to generate them they will be generated with the wrong links.
I figured out you can change it through this: https://www.mkdocs.org/user-guide/configuration/#edit_uri so I'm closing this PR and solving the issue. Thank you for pointing it out anyway!

PR #376, opened 2019-06-14
Refactor text formatters

ydudin3:
Have tokenizer objects instead of functions so we can have encoders which require initialization to follow similar interface.

Test plan: unit and integration tests.

### DIFF
--- a/ludwig/features/feature_utils.py
+++ b/ludwig/features/feature_utils.py
@@ -42,6 +42,6 @@ def set_str_to_idx(set_string, feature_dict, format_func):
         raise Exception('Format {} not supported'.format(format_func))
 
     out = [feature_dict.get(item, feature_dict[UNKNOWN_SYMBOL]) for item in
-           tokenizer.tokenize(set_string)]
+           tokenizer(set_string)]
 
     return np.array(out, dtype=np.int32)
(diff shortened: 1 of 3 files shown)


msaisumanth:
The code is not that different for different languages. Why do we need a class for each language?

msaisumanth:
We can use python closures or functools.partial to make it cleaner, unless we need classes so that people can extend it etc.

w4nderlust:
> The code is not that different for different languages. Why do we need a class for each language?

This is the first refactoring stage, replicating 1:1 the functionality from before. I think this is good for now, later on we can improve it (of instance having one single class and passing the parameters coming from the preprocessing dictionary as parameters. Also at the moment we use a singleton nlp pipeline, something that can be replaced by a better mechanism now that instead of using functions we are using objects with separated call and init methods.

w4nderlust:
I think this is fine @ydudin3 . The only thing i may consider doing differently is instead of creating a `tokenize()` method, maybe we can just use `__call__()` like we do for the encoders.
The other point, for the future, is that we may want to add empty init functions that collect kwargs for all classes, this to make sure that then we can pass the different preprocessing parameters that each of them need.

Issue #371, opened 2019-06-07
Unnecessary spacing in github.io footer


IgorWilbert:
**Describe the bug**
The style of the footer on github.io pages does not look right

**To Reproduce**
1. Go to https://uber.github.io/ludwig/ and https://uber.github.io/ludwig/examples/
2. Scroll to the bottom
3. See error

**Expected behavior**
They should have the same style.

**Screenshots**
Wrong:
![wrong](https://user-images.githubusercontent.com/42987854/59105866-51b56a00-890b-11e9-82ba-804482eb4024.JPG)
Right:
![right](https://user-images.githubusercontent.com/42987854/59105878-5548f100-890b-11e9-8b45-2983d7188a96.JPG)

**Environment (please complete the following information):**
 - Browser: Firefox 67.0.1 (64-bits) (also happens in MS Edge and Google Chrome)
 - Screen: 1440 x 600
- OS: Windows 10

**Additional context**
This was probably optimized for another device/screen size. I would like to know if would be useful to implement some responsive styles or just change padding/font size etc.

IgorWilbert:
If this issue is valid I'd like to work on it. Thanks!

w4nderlust:
@IgorWilbert thanks for letting us know!
Actually the problem is a little bit more complex: the landing page has a custom css and html that doesn't really fit with the rest of the website, which is built with mkdocs-material.
So at the moment the footer fits well with the style of the homepage, but as you noticed it is different from the style of the other pages because of spacing.
You are welcome to give it a try to make them behave in the same way, but in general it would be better if someone with frontend skills would be willing to undertake the task of making the landing and the other pages use the same css and look more similar.
Unfortunately this is as far as my frontend skills go, so if there's someone else who wants to undertake the task i would be really glad to support their work as much as possible.

IgorWilbert:
Thank you @w4nderlust ! To be honest, I believe I have more experience with front-end than back-end, so I'd be glad to help you with this task. However, I've some doubts:

- Would you like them all to have the style of the homepage, the other pages or something in between?
- What are the steps to run the documentation website? (I could not find them) I have `python` and `pip` installed, so I guess I only need to run `pip install mkdocs`, `cd` into some part of the project and run `mkdocs serve`? Thanks!

w4nderlust:
The instructions are here: https://github.com/uber/ludwig/tree/master/mkdocs
To begin with I believe making the mkdocs and the landing page parts more similar would be great.
[...]

PR #378, opened 2019-06-16
Fixes style in landing page footer

IgorWilbert:
## Summary:
For #371. Fixes the `css` style of the landing page, to be similar to the others, as can be seen below .
- How it is now:
result

- How it should look like:
target

## Important notes:

- Most critical part is the arrow, which looks too bold. It turns out all other pages except the landing page use `mkdocs-material`, that has Font Awesome arrow right weight **REGULAR**  as standard. However, if I want to use it in the html directly, only  Font Awesome arrow-right weight **SOLID** is free, the **REGULAR** needs a paid account. I tried to just copy and paste the arrow from the `css` already minified for the other pages, but then the icon breaks
- I still did **not** run `mkdocs build` because there might be more style changes.

### DIFF
--- a/mkdocs/custom/landing.html
+++ b/mkdocs/custom/landing.html
@@ -384,25 +384,25 @@ predictions = model.predict(test_dataframe)</code></pre>
 
     <!-- Link to previous and/or next page -->
 
-    <div class="md-footer-nav">
+
+    <div class="md-footer-nav md-footer-nav-wrapper">
         <nav class="md-footer-nav__inner md-grid">
 
             <!-- Link to next page -->
 
-            <a class="md-flex md-footer-nav__link md-footer-nav__link--next" href="getting_started/" rel="next"
+            <a class="md-flex md-footer-nav__link md-footer-nav__link--next md-footer-nav__link-wrapper" href="getting_started/" rel="next"
                title="Examples">
                 <div class="md-flex__cell md-flex__cell--stretch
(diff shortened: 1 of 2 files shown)


w4nderlust:
Thank you so much @IgorWilbert !

w4nderlust:
after merging, this is actually what i get when I run mkdocs serve:
Screen Shot 2019-06-16 at 3 19 38 PM
Can you please fix it and make sure it works correctly?

IgorWilbert:
@w4nderlust sure! I did not know you would merge it now, since I did not run `mkdocs build` to update `github.io` yet. Sorry, next time I'll put `[WIP]` in the title. 
What is the browser and screen size you are using? This can help me a lot fixing it.

w4nderlust:
My bad, I should have checked out it before and run mkdocs build serve. No problem anyway, we'll fix it :) 
The screenshot is from the latest safari on osx mojave, but chrome was the same.
Latest firefox on linux does the same, while Chromium does this:
image

IgorWilbert:
@w4nderlust sorry about the late reply. I tested it on Chrome, Firefox and Edge, and the bug you reported seems to happen only in `github.io` pages. It does not happen locally when I run `mkdocs serve` nor `mkdocs build`. I think some extra styles are being applied when the files in the `docs` folder are updated. I'll send a PR to fix it soon.

PR #390, opened 2019-06-23
Fixes header style in landing page

IgorWilbert:
## Description
Fixes header `css` style in `mkdocs` and `github.io` pages, so that it occupies the whole width with the right spacing between the elements.
- Before:
before

- After:
after


IgorWilbert:
@w4nderlust I ran `mkdocs build` already, so this should be ready to merge. Thanks!

w4nderlust:
Thank you @IgorWilbert !

PR #397, opened 2019-06-24
add savefig for roc curve

patrickvonplaten:
# Code Pull Requests

ROC curve plot is saved in same dir where ludwig visualize ... command is executed.

Please provide the following:

# Documentation Pull Requests


w4nderlust:
#403 addresses this already

PR #399, opened 2019-06-25
adds support for data_train_dict, data_test_dict etc. in api.py

msaisumanth:
Adds support for data_(train/test/validation)_dict in api.py.

(Piero asked to include this for the sake of completeness)

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -271,6 +271,9 @@ class LudwigModel:
             data_validation_hdf5=None,
             data_test_hdf5=None,
             data_dict=None,
+            data_train_dict=None,
+            data_test_dict=None,
+            data_validation_dict=None,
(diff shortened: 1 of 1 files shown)


ydudin3:
Lgtm unless you want to add any tests for this.

ludwig/api.py
```diff
@@ -271,6 +271,9 @@ def train(
             data_validation_hdf5=None,
             data_test_hdf5=None,
             data_dict=None,
+            data_train_dict=None,
+            data_test_dict=None,
```
w4nderlust: can you invert the order here, in the docstring and in the rest of the file? as we always use the convention: train, validation, test ;)

msaisumanth: sure. makes sense

PR #401, opened 2019-06-27
add optional arg model_definition_file for LudwigModel

Oceania2018:


w4nderlust:
Thank you for contributing, but sorry I'm not sure about the need for this pr. The main is there just for test purposes, it's not really needed, plus the new parameters are in one case redundant, in the other case have a wrong description.

Issue #87, opened 2019-02-14
Predict from stdin or via the programmatic api


loretoparisi:
As soon as my brand new model is ready, I would like `ludwig` the enable predictions via the command line, so that one could do like

```sh
echo "call me at +3912345679 for offers" | ludwig predict --only_prediction --model_path /path/to/model -
```

where the `-` may indicate the `stdin` (as an example) and get the model predictions directly. This would help to integrated `ludwing` command in a inference pipeline.

If I have understood well the api (my model is not ready yet...) it should be possibile via the programmatic api like that:

```python
from ludwig import LudwigModel

# load a model
model = LudwigModel.load(model_path)

# obtain predictions
myDict = { 'text': ['call me at +3912345679 for offers'] }
predictions = model.predict(data_dict=myDict,
  return_type=dict
  batch_size=128,
  gpus=None,
  gpu_fraction=1,
  logging_level=logging.DEBUG)

# close model (eventually)
model.close()
```

Is that correct?

w4nderlust:
I see your usecase, it is interesting as it can be used inside bash scripts.
it has a big downside I believe: you load the model every single time you call it, which is really slow.
An alternative could be to start a Rest server (something we already planned to add), and then you can just curl that endpoint in your bash line. It has the overhead of calling through rest, but it's nothing compared to the overhead of loading the model every single time.
What's your take on it?

loretoparisi:
@w4nderlust thanks! yes that is true, the downside of loading the model must be addressed. This is I what I'm doing in [fasttext.js](https://github.com/loretoparisi/fasttext.js), where I wrap fastText binary into a node.js library: The model is loaded once through the `udwig predict --only_prediction --model_path /path/to/model - `. This because the `predict` (with that option) would just open the `stdin` and waits for line feed. Then I send to the forked process the text writing to the `stdin` (hence like`call me at +3912345679 for offers\r\n`).
In this way `ludwig predict` would be used in several ways: REST servers with a process fork, attached to bash process, etc.
In my experience of FastText, the execution overhead is minimal (process fork and pipe to `stdin`) and it's faster than running programmatically, without losing the functionality. Since `ludwig` has a great command line, it would be a useful addon.

w4nderlust:
[...]

PR #414, opened 2019-07-07
add server to cli

johnwahba:
# Code Pull Requests
This code adds an inference server to models trained by ludwig. https://github.com/uber/ludwig/issues/87
`ludwig serve --model_path path/to/model --port 8000 --host 0.0.0.0`
` curl http://localhost:8000/predict -X POST -d "[{\"text\": \"test\"}]"`
I went with Starlette over something like Flask because it has a more modern API that supports async await which will come in handy for file uploads later. 
Still haven't implemented image inference. Want to get some feedback on the approach before moving forward with that. 
Could either do multipart form uploads or base64 encoding in the request body. I imagine the latter would be more costly but form uploads are generally a worse API for inference on tabular data and NLP. 

# Documentation Pull Requests
Will do after settling on an approach.
<!--

-=>

### DIFF
--- a/ludwig/serve.py
+++ b/ludwig/serve.py
@@ -44,9 +44,7 @@ COULD_NOT_RUN_INFERENCE_ERROR = {
     "error": "Unexpected Error: could not run inference on model"}
 
 
-def server(
-    model,
-):
+def server(model):
     app = FastAPI()
 
     input_features = {
@@ -95,6 +93,12 @@ def convert_input(form):
             new_input[k] = v
 
     return (files, new_input)
+
+
+def run_server(model_path, host, port):
+    model = LudwigModel.load(model_path)
+    app = server(model)
+    uvicorn.run(app, host=host, port=port)
 
 
 def cli(sys_argv):
@@ -146,9 +150,7 @@ def cli(sys_argv):
         logging_level_registry[args.logging_level]
     )
 
-    model = LudwigModel.load(args.model_path)
-    app = server(model)
-    uvicorn.run(app, host=args.host, port=args.port)
+    run_server(args.model_path, args.host, args.port)
 
 
 if __name__ == '__main__':
--- a/ludwig/serve.py
+++ b/ludwig/serve.py
@@ -57,6 +57,7 @@ def server(
     async def predict(request: Request):
         form = await request.form()
         files, entry = convert_input(form)
+
         try:
             if (entry.keys() & input_features) != input_features:
                 return JSONResponse(ALL_FEATURES_PRESENT_ERROR,
--- a/ludwig/serve.py
+++ b/ludwig/serve.py
@@ -1,0 +1,129 @@
+#! /usr/bin/env python
+# coding=utf-8
+# Copyright (c) 2019 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import logging
+import sys
+
+from ludwig.contrib import contrib_command
+from ludwig.utils.print_utils import logging_level_registry
+import pandas as pd
+import json
+
+from ludwig.api import LudwigModel
+
+
+logger = logging.getLogger(__name__)
+
+
+def start_server(
+    model_path,
+    logging_level,
+    host,
+    port
+):
+    from starlette.applications import Starlette
+    from starlette.responses import JSONResponse
+    import uvicorn
+
+    app = Starlette(debug=True)
+
+    global model
+    model = LudwigModel.load(model_path)
+
+    global input_features
+    input_features = set([
+        f['name'] for f in model.model_definition['input_features']
+    ])
+
+    @app.route('/predict', methods=["POST"])
+    async def endpoint(request):
+        data_json = await request.body()
+        entries = json.loads(data_json)
+
+        for entry in entries:
+            if (entry.keys() & input_features) != input_features:
+                return JSONResponse({"error": "entries must contain all input features"}, status_code=400)
+
+        try:
+            resp = model.predict(data_dict=entries).to_dict('records')
+            return JSONResponse(resp)
+        except Exception as e:
+            logger.error("Error: {}".format(str(e)))
+            return JSONResponse({"error": "Unexpected Error: could not run inference on model"}, status_code=500)
+
+    uvicorn.run(app, host=host, port=port)
+
+
+def cli(sys_argv):
+    parser = argparse.ArgumentParser(
+        description='This script serves a pretrained model',
+        prog='ludwig serve',
+        usage='%(prog)s [options]'
+    )
+
+    # ----------------
+    # Model parameters
+    # ----------------
+    parser.add_argument(
+        '-m',
+        '--model_path',
+        help='model to load',
+        required=True
+    )
+
+    parser.add_argument(
+        '-l',
+        '--logging_level',
+        default='info',
+        help='the level of logging to use',
+        choices=['critical', 'error', 'warning', 'info', 'debug', 'notset']
+    )
+
+    # ----------------
+    # Server parameters
+    # ----------------
+    parser.add_argument(
+        '-p',
+        '--port',
+        help='port for server',
+        default=8000,
+        type=int,
+    )
+
+    parser.add_argument(
+        '-H',
+        '--host',
+        help='host for server',
+        default='0.0.0.0'
+    )
+
+    args = parser.parse_args(sys_argv)
+
+    logging.getLogger('ludwig').setLevel(
+        logging_level_registry[args.logging_level]
+    )
+
+    start_server(**vars(args))
+
+
+if __name__ == '__main__':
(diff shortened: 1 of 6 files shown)


w4nderlust:
Thank you @johnwahba for this PR, it adds a really useful functionality.

I have three discussion points, that are the main reason why this functionality was not there to begin with:
1. I was considering [FastAPI](https://fastapi.tiangolo.com) instead of alternatives. You can see here what it provides on top of Starlette: https://fastapi.tiangolo.com/alternatives/#starlette . I'm by no menas an expert of webservices and these kind of frameworks, but I'm curious about your opinion. In case, I believe porting this code to use FastAPI instead wouldn't be hard.
2. As you noted, images and in general blob datatypes are the main issue. There's a PR for audio features for instance, and in the future others like video will likely be added, so it's not just for images, an ideal solution will be general and flexible to accommodate these datatypes too. The issue stems from the fact that currently there's no simple way to pass to Ludwig an image from memory, but they have to be loaded from a path. It's a limitation I'm going to fix in the future, but I'm considering a munch of design options still. So for the moment the (suboptimal) solution would be to receive an image, save it to a temp directory and provide in the input dataframe to the `predict()` function the path generated. Ideally the image should be deleted after obtaining the prediction. Anyway, what you have done so far would work great in cases where images / blob features are not present, so I think it's a great starting point to refine on.
3. can you give me some pointers to understand pros and cons of form upload and base64 encoding? Why are form uplaods worse for tabular data?
4. Final consideration, that is probably currently out of the scope, but it would be really cool to have also a cute little client interface so that people can test their models without having to use curl (which may come handy for images, as I believe the base64 encoding for instance would make it trickier to use curl)

This is really really much appreciated, thank you again.

johnwahba:
Thanks for the feedback!

My thinking right now is to also support multipart forms as an alternative API that works with images/video/audio. Form uploads tend to be worse for tabular data than body JSON because most net-http clients have better support for the latter. I do think it's pretty minor, and would be happy to kill standard JSON api in favor of multi-part form upload.

Forms have a few advantages over a custom base64 approach:
1. Most web frameworks already have good semantics for creating tempfiles and cleaning them up in the request lifecycle

2. It'd be straightforward to plumb into a frontend since it would be able to talk to native html forms. 

The curl would be pretty straight forward to work with as well
`curl http://localhost:8000/ -X POST -F 'entries[][input_image]=@file-1.jpg' -F 'entries[][input_image]=@file-2.jpg'`

While Base64 encoding would work there's a decent chance that it wouldn't be performant enough. Here's a benchmark that tries to encode and decode to base64 https://github.com/mayeut/pybase64#benchmark. 
Once you get to video, you might start spending a second of CPU time just to decode the image. 

In light of the above I actually think porting this to fastapi would actually be very hard if we went down the road of supporting image uploads through multi-part upload. 

My main concern is that we'd need to dynamically define route functions to support different models. 
For example a model that took in an image `photo` and text `caption` to return a categorical variable would need to look like
```python
@app.post("/predict")
async def predict(
    photo: UploadFile = File(...), 
    caption: str = Form(...)
):
...
    return model.predict({"photo": photo, "caption": caption})
```
It might be a bit challenging to set the function header dynamically to support different models. There might be a way to do it https://pydantic-docs.helpmanual.io/#dynamic-model-creation but might involve a decent amount of annotation in ludwig for each input type which is probably out of scope of what I'm trying to do. Luckily it looks like fastapi uses starlette under the hood (https://github.com/tiangolo/fastapi/blob/master/Pipfile#L28) so it should be straight forward to port later.

I'll proceed to switch the current implementation to use form upload but will stick with starlette unless you want to require fastapi to merge. Will spinup a small client cli as well :).

johnwahba:
On second thought I'll spike with fastapi and see how far I get.

w4nderlust:
> On second thought I'll spike with fastapi and see how far I get.

That sounds good too. Didn't want to put a bigger burden on your shoulders, so if it becomes too complicated using vanilla Starlette if fine too for the moment.

johnwahba:
So one challenge is that there isn't really a spec for how to handle arrays of objects for forms.
The ideal format would look like:
```
[
  {photo: <file_object>, capiton: <str>},
  {photo: <file_object>, capiton: <str>}
]
```
but to get something like that the curl would look like
` -F 'entries[0][photo]=@file-1.jpg'  -F 'entries[0][caption]=test'`

If we defer on batch inference the curl becomes a lot cleaner:
` -F 'photo=@file-1.jpg'  -F 'caption=test'`

I'll try moving forward with that since image/video/audio support is more valuable than batch inference imo.

johnwahba:
Got it working with mnist! 
`curl http://localhost:8000/predict -X POST -F 'image_path=@path/to/file/0/12004.png'`

Given how simple it is to curl, I think I'll punt on the client for this PR.

w4nderlust:
Yeah that sounds great! Batch inference rarely happens in reality as in most cases you would need something like a queue and a throttler, in most cases single datapoint inference is what you really want.

I also believe this PR may solve also a longstanding feature request: https://github.com/uber/ludwig/issues/87

johnwahba:
Tell me if you'd like anything else before merging :)

w4nderlust:
The boring things :)
1. please add one integration test, so that we know that if we make changes elsewhere we are not breaking the serve command.
2. write some documentation in the user guide (`mkdocs/docs/user_guide.md`), both documenting the command and its parameters and also giving an example of usage that loads a model and where you curl it (maybe include both a blob like input like an image and a more tabular input).
If you have time to this quickly, we can include it in in the v0.2 release!
Finally, I would want @ydudin3 and @msaisumanth to take a look.

johnwahba:
Will do. What time frame are you looking at for v0.2?

w4nderlust:
Ideally this week

ludwig/serve.py
```diff
@@ -0,0 +1,164 @@
...
+    input_features = {
+        f['name'] for f in model.model_definition['input_features']
+    }
+
+    @app.post('/predict')
+    async def predict(request: Request):
+        form = await request.form()
+        files, entry = convert_input(form)
+        try:
+            if (entry.keys() & input_features) != input_features:
```
ydudin3: Nit: stick to 79 characters line limit.

ludwig/serve.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def start_server(
+    model_path,
+    logging_level,
+    host,
+    port
+):
+
+    from fastapi import FastAPI
```
ydudin3: Since this is the main functionality anyway, why not do imports at the top of the file?

ludwig/serve.py
```diff
@@ -0,0 +1,164 @@
...
+    from starlette.requests import Request
+    from starlette.responses import JSONResponse
+    import uvicorn
+
+    app = FastAPI()
+
+    global model
```
ydudin3: Just curious, why do these (model and input_features) need to be global?

w4nderlust: Don't know well how FastAPI works, but an alternative could be to create a class and have model and input_features assigned as instance variables?

johnwahba:
Alright, implemented tests which prompted a refactor that cleaned up the globals. Next will add docs.

johnwahba:
Ready for another pass :)

ludwig/serve.py
```diff
@@ -0,0 +1,156 @@
...
+logger = logging.getLogger(__name__)
+
+ALL_FEATURES_PRESENT_ERROR = {"error": "entry must contain all input features"}
+
+COULD_NOT_RUN_INFERENCE_ERROR = {
+    "error": "Unexpected Error: could not run inference on model"}
+
+
+def server(
+    model,
```
ydudin3: Nit: keep in one line.

tests/integration_tests/test_server.py
```diff
@@ -0,0 +1,164 @@
...
+        skip_save_progress=True,
+        skip_save_unprocessed_output=True
+    )
+    model.predict(data_df=data_df)
+    return model
+
+
+def example_input(input_features):
+    data = {}
+    for i in input_features:
```
ydudin3: Nit: for feature in input_features - for better readability.

tests/integration_tests/test_server.py
```diff
@@ -0,0 +1,164 @@
...
+        if i['type'] in ['image', 'video', 'audio']:
+            data[name] = random_string()
+        else:
+            data[name] = random_string()
+    return data
+
+
+def output_keys_for(output_features):
+    keys = []
+    for f in output_features:
```
ydudin3: Same here.

tests/integration_tests/test_server.py
```diff
@@ -0,0 +1,164 @@
...
+            raise NotImplementedError
+    return keys
+
+
+def convert_to_form(entry):
+    data = {}
+    files = []
+    for k, v in entry.items():
+        if type(v) == str and os.path.exists(v):
+            files.append((k, (v, open(v, 'rb'), 'image/jpeg')))
```
w4nderlust: these files are open but never closed. They are actually removed inside the `/predict` but I'm not sure, probably we should be closing them too, ideally right before removing them in serve.py:74. What do you think?

ludwig/serve.py
```diff
@@ -0,0 +1,156 @@
...
+
+    args = parser.parse_args(sys_argv)
+
+    logging.getLogger('ludwig').setLevel(
+        logging_level_registry[args.logging_level]
+    )
+
+    model = LudwigModel.load(args.model_path)
+    app = server(model)
+    uvicorn.run(app, host=args.host, port=args.port)
```
w4nderlust: Maybe these 3 lines can be packaged in a single `serve_model(model_path, host, port)` function, so that If someone wants to start a server programmatically, or maybe starting more than one server programmatically on different ports they can do it.

tests/integration_tests/test_server.py
johnwahba: These files are consumed by the test client and get converted to `starlette.UploadFile`s by the time they get to serve.py. Will close inline

johnwahba:
Addressed! Ready to merge

w4nderlust:
THank you for this great PR!

PR #417, opened 2019-07-09
Added audio features documentation

patrickvonplaten:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/mkdocs/docs/user_guide.md
+++ b/mkdocs/docs/user_guide.md
@@ -2150,6 +2150,60 @@ There are no time series decoders at the moment (WIP), so time series cannot be 
 
 As no time series decoders are available at the moment, there are also no time series measures.
 
+Audio Features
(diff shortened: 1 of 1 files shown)


mkdocs/docs/user_guide.md
```diff
@@ -2150,6 +2150,60 @@ There are no time series decoders at the moment (WIP), so time series cannot be
...
+    type: stft
+    window_length_in_s: 0.04
+    window_shift_in_s: 0.02
+    num_fft_points: 800
+    window_type: boxcar
+```
+
+### Audio Input Features and Encoders
+
+Audio files are transformed into one of the following types according to `type` in `audio_feature` in `preprocessing`.
```
w4nderlust: this sentence and the following list can be moved inside the preprocessing section.

PR #430, opened 2019-07-15
fix(doc): fix parameters name mismatch

revolunet:
Looks like https://github.com/uber/ludwig/pull/332/files was missing one entry


w4nderlust:
thank you for your PRs!

PR #456, opened 2019-07-25
fixed AttributeError: 'Model' object has no attribute 'model'

Athanaziz:
# Code Pull Requests

Please provide the following:

- a clear explanation of what your code dose - it fixes the AttributeError: 'Model' object has no attribute 'model'
- if applicable, a reference to an issue - no reported issue
- a reproducible test for your PR (code, model definition and data sample) - ludwig_model.save_for_serving(path) fails with an error mentioned above

# Documentation Pull Requests

### DIFF
--- a/ludwig/models/model.py
+++ b/ludwig/models/model.py
@@ -1362,7 +1362,7 @@ class Model:
                     input_tensors, output_tensors)
             },
             strip_default_attrs=True,
-            saver=self.model.saver,
+            saver=self.saver,
         )
         builder.save()
 


w4nderlust:
@Athanaziz thank you for the PR, please sign the CLA and I can merge.

Athanaziz:
This CLA check is a bit glitchy. I've signed it already but it's still in pending status

Athanaziz:
Oh, now we're good to go :)

PR #464, opened 2019-07-29
Vector feature

msaisumanth:
Adds Vector input and output features.

1. Input data is expected as space separated floating point numbers
2. Only fc_stack is supported as encoder and decoder
3. Errors: either mean squared or mean absolute error
4. Integration tests for both input and output features. 

Wrote bare minimum functionality everywhere.


ludwig/features/base_feature.py
```diff
@@ -307,6 +307,8 @@ def concat_dependencies_and_build_output(
         final_hidden[self.name] = (feature_hidden, feature_hidden_size)
 
         # ================ Outputs ================
+        kwargs['is_training'] = is_training
+        kwargs['dropout_rate'] = dropout_rate
```
w4nderlust: this is suspicious, need to investigate, why do you need it?

ludwig/features/vector_feature.py
```diff
@@ -0,0 +1,445 @@
...
+    @staticmethod
+    def read_single_vector(row):
+        return [float(x) for x in row.split()]
+
+    @staticmethod
+    def feature_data(column, metadata):
+        vectors = [VectorBaseFeature.read_single_vector(x) for x in column]
+        for v in vectors:
+            if len(v) != metadata['vector_size']:
+                raise ValueError(
```
w4nderlust: here if you enumerate vectors you can also report the line of the malformed vector

ludwig/features/vector_feature.py
```diff
@@ -0,0 +1,445 @@
...
+            squared_error_val = tf.reduce_sum(
+                get_squared_error(targets, predictions, self.name), axis=1
+            )
+
+            # TODO - not sure if this is correct
+            r2_val = tf.reduce_sum(get_r2(targets, predictions, self.name))
+
+        return error_val, squared_error_val, absolute_error_val, r2_val
+
+    def vector_loss(self, error_val, squared_error_val):
```
w4nderlust: other losses may not use the errors but do something different, so it's better to have gt and predictions here

ludwig/features/base_feature.py
msaisumanth: What's happening is most features don't require the flag is_training in the build_output method. But if you use a Dense Decoder, you need that flag. concat_dependencies_and_build_output is passing the kwargs but it's not passing is_training and dropout_rate to build_output. 

Since build_output needs to have the same signature for all the features, is_training and dropout_rate need to be in the kwargs. I didn't see any way around it.

ludwig/features/vector_feature.py
msaisumanth: Makes sense. Will do

ludwig/features/vector_feature.py
msaisumanth: Got it. I wanted to avoid the repetitive computations.

ludwig/features/base_feature.py
w4nderlust: Got it, then the cleaner solution is to add those two parameters to self.build_output

ludwig/features/vector_feature.py
w4nderlust: Makes sense, but itdeally TF will optimize the computation graph for that.

ludwig/features/base_feature.py
msaisumanth: yes, It will require changes in a few more places but I can do that.

Issue #480, opened 2019-08-06
Adding routes to serve.py


areeves87:
**Is your feature request related to a problem? Please describe.**
I'd like to check whether the server is up or down (ie a healthcheck) with a simple GET request to the root path `/` of the ludwig server.

In general it would be nice to have some instructions/examples for how to extend the functionality of serve.py.

**Describe the use case**
This new route would help run health checks on a remote instance of the ludwig server. In my case it would be for an aws ecs cluster, which requires a valid URI with no spaces for healthchecks. It is not straight forward to satisfy this criteria with the `/predict` route.

**Describe the solution you'd like**
Add a simple `GET` route in serve.py's fastApi for the root path `/`, but have not had any success yet.

**Describe alternatives you've considered**

**Additional context**
Here's what I've edited in `ludwig/serve.py` before building the docker image. 

```python
def server(model):
    app = FastAPI()

    input_features = {
        f['name'] for f in model.model_definition['input_features']
    }

    @app.get('/')
    def root():
        return JSONResponse({"message": "ludwig is up!"})

    @app.post('/predict')
    async def predict(request: Request):
        form = await request.form()
        files, entry = convert_input(form)

        try:
            if (entry.keys() & input_features) != input_features:
                return JSONResponse(ALL_FEATURES_PRESENT_ERROR,
                                    status_code=400)
            try:
                resp = model.predict(data_dict=[entry]).to_dict('records')[0]
                return JSONResponse(resp)
            except Exception as e:
                logger.error("Error: {}".format(str(e)))
                return JSONResponse(COULD_NOT_RUN_INFERENCE_ERROR,
                                    status_code=500)
        finally:
            for f in files:
                os.remove(f.name)


    return app
```

And what I changed in the Dockerfile to make sure it uses the updated serve.py 
```
FROM tensorflow/tensorflow:latest-py3

RUN apt-get install -y --no-install-recommends \
        git libgmp3-dev
RUN git clone --depth=1 https://github.com/uber/ludwig.git \
    && cd ludwig/ \
    && pip install -r requirements.txt \
    && python -m spacy download en \
    && python setup.py install

COPY ludwig/serve.py ludwig/build/lib/ludwig/serve.py
COPY ludwig/serve.py ludwig/ludwig/serve.py

WORKDIR /data

ENTRYPOINT ["ludwig"]
```

msaisumanth:
@johnwahba FYI

wahba-stripe:
[...]

PR #487, opened 2019-08-13
Healthchecking server

areeves87:
Now we can perform server healthchecks by sending `GET` requests to its root path `/`. We also test that a get request to the `/` path returns status code 200.

Addresses #480 .

### DIFF
--- a/ludwig/serve.py
+++ b/ludwig/serve.py
@@ -52,8 +52,8 @@ def server(model):
     }
 
     @app.get('/')
-    def _unused_root():
-        return JSONResponse({"message": "ludwig is up!"})
+    def check_health():
+        return JSONResponse({"message": "Ludwig server is up"})
 
     @app.post('/predict')
     async def predict(request: Request):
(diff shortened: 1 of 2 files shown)


areeves87:
The deepsource check doesn't like the new route in `serve.py`:
```python
@app.get('/')
    def root():
        return JSONResponse({"message": "ludwig is up!"})
```
It complains that `root()` is defined but never used. Is there an easy way to resolve this check failure?

ludwig/serve.py
```diff
@@ -51,6 +51,10 @@ def server(model):
         f['name'] for f in model.model_definition['input_features']
     }
 
+    @app.get('/')
+    def _unused_root():
```
w4nderlust: call this `check_health`. If deepsource complains, don-t worry.

ludwig/serve.py
```diff
@@ -51,6 +51,10 @@ def server(model):
...
+        return JSONResponse({"message": "ludwig is up!"})
```
w4nderlust: Change the message to "Ludwig server is up"

PR #512, opened 2019-09-04
Fixed horovodrun commands to remove extra mpirun args

tgaddair:


w4nderlust:
Thanks Travis! That actually was all the reason to use horovodrun in the first place :)

PR #529, opened 2019-10-04
Reorganized dependencies

w4nderlust:
Now can run `pip install ludwig` for base packages or `pip install ludwig[text,serve]` for only the additional packages in those categories and `pip install ludwig[full]` for all the packages.


ydudin3:
Should we update the docs in the same PR?

w4nderlust:
> Should we update the docs in the same PR?
Yeah good point. We should also add messages when importing missing things in ImageFeature AudioFeature etc.

Issue #491, opened 2019-08-13
Option to not save any files to disk


BeWe11:
I am building a small server around ludwig that receives model definitions + data and stores the resulting model for future predictions. I have turned on all `skip_save` parameters in the `train` method, but some files are still saved to disk.

In particular, the `results` directory is always created as well as some `.npy` files.

Is there any way to configure ludwig to not save anything to disk at all and handle all model persistence myself?

Edit:
I've seen that the prediction script offers a flag to skip saving of the `.npy` files, I've created a separate issue for adding this flag to the python API https://github.com/uber/ludwig/issues/492

So the remaining question is whether it's possible to not create a directory containing experiment results.

w4nderlust:
Let me try to understand exactly the problem: the directory is created when running `LudwigModel.predict()`? Even when everything is skipped?
In that case, we'll certainly need to fix this.

BeWe11:
Thanks for your response, I'll try to clarify what's happening:

The `LudwigModel.train` method accepts four parameters that allow to skip saving of files. These are
```
skip_save_mode
skip_save_progress
skip_save_log
skip_save_processed_input
```
I've set all of these to `True`, but the `results` directory is still being created. The parameter that sets the name of this directory is `output_directory`. I've tried setting this parameter to an empty string or to `None`, but both of these values lead to exceptions. There seems to be no way to avoid the creation of this directory.

Regarding `LuwdigModel.predict`, please take a look at my other issue https://github.com/uber/ludwig/issues/492. The `skip_save_unprocessed_output` parameter, which is available for the command line prediction script, is simply not exposed in the python API, so that the `*.npy` files are always being created and there is no way to turn that off.

msaisumanth:
@w4nderlust this is why results is always being created:
https://github.com/uber/ludwig/blob/48acbf6cb63a859d06eb39b2f7ec27577394251d/ludwig/train.py#L525

We could probably delete all the intermediate data at the end of full_train (if the skip variables were True)

w4nderlust:
Avooiding creating the dir to begin with is probably a better idea. We can
either remove the dir creation from within the function and check if at
least one skip is false and create the dir outside or pass a boolean
“should_create_dir” to this function.

On Mon, 19 Aug 2019 at 15:03, Sai Sumanth Miryala <notifications@github.com>
wrote:

> @w4nderlust
[...]

PR #530, opened 2019-10-07
Improving skip parameters

w4nderlust:
Improving skips by adding two new skip_save_training_description and skip_save_training_statistics. Also now full_training will not create a directory if all skips are true.
This solves #491.

### DIFF
--- a/ludwig/predict.py
+++ b/ludwig/predict.py
@@ -34,7 +34,8 @@ from ludwig.globals import TRAIN_SET_METADATA_FILE_NAME
 from ludwig.models.model import load_model_and_definition
 from ludwig.utils.data_utils import save_csv
 from ludwig.utils.data_utils import save_json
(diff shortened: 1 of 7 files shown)


ludwig/predict.py
```diff
@@ -107,7 +101,22 @@ def full_predict(
     model.close_session()
 
     if is_on_master():
-        os.mkdir(experiment_dir_name)
+        # setup directories and file names
+        experiment_dir_name = output_directory
+        suffix = 0
+        while os.path.exists(experiment_dir_name):
+            experiment_dir_name = output_directory + '_' + str(suffix)
+            suffix += 1
```
msaisumanth: We have this logic repeated in a few places. Should we write a util method?

w4nderlust: good idea

w4nderlust: done

PR #567, opened 2019-11-12
Bypass multiprocessing.Pool.map for faster execution

carlogrisetti:
After the introduction of multiprocessing management, inferring time has been really slow when working with a single image.
This workaround bypasses the pooling when inferring on a single image and using only one process.

There is for sure a way to implement this to cover more cases (single process, multiple images). But this is way less invasive and immediately effective.

### DIFF
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -313,7 +313,7 @@ class ImageBaseFeature(BaseFeature):
             # image just use this faster shortcut, bypassing multiprocessing.Pool.map
             else:
                 logger.warning(
-                        'No pooling applied. Using one process for preprocessing images'
+                        'No process pool initialized. Using one process for preprocessing images'
                 )
                 img = read_image_and_resize(all_file_paths[0])
                 data[feature['name']] = np.array([img])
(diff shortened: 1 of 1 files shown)


msaisumanth:
This is good idea. Thanks @carlogrisetti 

@w4nderlust

ludwig/features/image_feature.py
```diff
@@ -296,15 +296,27 @@ def add_feature_data(
...
+                        )
                     )
+                    data[feature['name']] = np.array(
+                        pool.map(read_image_and_resize, all_file_paths)
+                    )
+            # If we're not running multiple processes and we are only processing one
+            # image just use this faster shortcut, bypassing multiprocessing.Pool.map
+            else:
+                logger.warning(
+                        'No pooling applied. Using one process for preprocessing images'
```
w4nderlust: replace "No pooling applied." with "No process pool initialized."

w4nderlust:
lgtm, just added a minor comment, please address it and I'll merge

carlogrisetti:
I think you might have missed the label fix. Just tell me if you want to generalize that to cover the single process, multiple images case, and I'll try to look at it as soon as I have time (right as it is it solves my specific issue, so it's on a lower priority now)

w4nderlust:
That may be useful in the future, but we are in the process of changing the preprocessing pipeline quite substantially, which means i would postpone any effort after we are done ;)

Issue #350, opened 2019-05-20
Feature Request: add validation loss to tf.summary outputs


carlogrisetti:
**Is your feature request related to a problem? Please describe.**
I am not able to evaluate a model validation loss history and\or track the model validation loss evolution when not having access to the CLI and have to rely to the train_reg_mean_loss graph to deduce what the validation loss could be trending to. I often find myself with a continuously improving training loss, but a stuck validation loss (hence, the model is not getting better and not being saved).
I can only find out about this when looking at the CLI and finding out that there has not been an improvement in the last X epochs. Being able to see the validation loss graph not having improved in the last X epochs could lead me to stop the training and try a different approach, without having to in any case remote into the training machine, but allowing for a web-based "diagnostic".

**Describe the use case**
Being able to track validation loss from the existing TensorBoard interface, since when having small datasets and not having the capability to augment input data, the validation set (thus the validation loss) could become a very separate thing from the training set (and its loss value).

**Describe the solution you'd like**
Adding the validation loss measure already calculated in the CLI as a tf.summary output, alongside the existing train_reg_mean_loss

w4nderlust:
While training Ludwig keeps track of loss and measures and at the end it saves them in `training_statistics.json` so you can find there what you are looking for. Moreover there is also a  visualization called ` learning_curves` that show you losses and measures on training and validation over the course of the training.
So I believe this use case is already covered.

carlogrisetti:
Think about a long running training. I should remote into the machine, stop the training, run the visualization scripts and then decide to resume or not the training.
My intent is to monitor the training whilist doing other things (like in meetings or away from work), and I can only write some sort of parser for the training_statistics.json file, but having tf.summary available (and already used for the training statistics) seems really to be a waste...
I would like to open a remote session only to actually work on that (even just for stopping the training) but not each time I want to see how that's going.

If this doesn't cut it, ok for me, I just wanted to better explain the use case

w4nderlust:
If you are keeping track of progress, you can you the training_rpogress.json file in the model dir that gets updated every epoch. For reading the validation loss you just read the json and then `loaded_json['vali_stats'][your_output_feature_name]['loss']`.

carlogrisetti:
If and when I'm able to sort out how to eventually implement this, would you agree in having that value as an additional output in the existing logfiles alongside the train_reg_mean_loss ?

w4nderlust:
[...]

PR #571, opened 2019-11-14
Add tracker measures (combined train\validation\test loss and accuracy) in output towards TensorBoard

carlogrisetti:
This is the implementation of my own feature request #350 
"works fine on my machine" as usual, and since it's based on the same value calculated for the CLI output, treating it just as a number and adding it to the same step, it has no impact on any of the calculations. It's using the same file_writer, appending a "generated" summary containing only that one value to the corresponding step.

### DIFF
--- a/ludwig/models/model.py
+++ b/ludwig/models/model.py
@@ -264,6 +264,19 @@ class Model:
                     output_feature['name']]
         return feed_dict
 
+    @classmethod
+    def add_tensorboard_combined_summary(cls, stats, prefix, train_writer, step):
+        summaries = []
+        combined_output_f = stats['combined']
+        for metric in combined_output_f:
+            metric_tag = "evaluation/{}_{}".format(prefix, metric)
+            metric_val = combined_output_f[metric][-1]
+            summaries.append(
+                tf.compat.v1.Summary.Value(tag=metric_tag, simple_value=metric_val)
+            )
+        test_summary = tf.compat.v1.Summary(value=summaries)
+        train_writer.add_summary(test_summary, step)
+
     def train(
             self,
             training_set,
@@ -581,13 +594,12 @@ class Model:
             # the same way as in the CLI. For each one, progress_tracker.steps has already
             # been incremented before, so in order to write on the previous summary, we need
             # to use -1
-            training_loss = progress_tracker.train_stats['combined'][LOSS][-1]
-            training_accuracy = progress_tracker.train_stats['combined'][ACCURACY][-1]
-            training_summary = tf.compat.v1.Summary(value=[
-                tf.compat.v1.Summary.Value(tag="evaluation/training_loss", simple_value=training_loss),
-                tf.compat.v1.Summary.Value(tag="evaluation/training_accuracy", simple_value=training_accuracy)
-            ])
-            train_writer.add_summary(training_summary, progress_tracker.steps - 1)
+            self.add_tensorboard_combined_summary(
+                progress_tracker.train_stats,
+                "training",
+                train_writer,
+                progress_tracker.steps - 1
+            )
 
             if validation_set is not None and validation_set.size > 0:
                 # eval measures on validation set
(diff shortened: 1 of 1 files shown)


w4nderlust:
lgtm thank you Carlo. Will test it out soon and. then merge

w4nderlust:
Tested, looks fine, but actually why not adding all the values in the tracker instead of only the validation loss? :) Let me know if you are willing to add them, otherwise I will do it myself after merging this.

carlogrisetti:
Oh well... That was the only useful value to me at the time (and still now it is). You mean validation accuracy, test loss, and test accuracy, right?
Since there's a different granularity on training metrics acquired via the usual training loop and the ones calculated at the end of the epoch, do you want training statistics also?

I wouldn't think of an use case for any "live" metric other than validation loss/accuracy, but you surely know better :)

carlogrisetti:
Ok so... I've added everything (I just thought... it's free, and you just can not look at that if you don't need those).
The only "open point" I see is with the formatting we want to give to that.

- All values under one "evaluation" group (as it is now with this latest commit)
- Separate "evaluation_training" \ "evaluation_validation" \ "evaluation_test" groups?

I can't make my mind up on how to display those... also since this is an ubiquitous metric that will always be present in TensorBoard, it might be useful to force its ordering to be the first one (ie: the old awful "!" to force it to be alphabetically sorted as first).

Also, it might be useful to move train_reg_mean_loss under the same metrics group, since it's almost the same measure (or at least it means almost the same thing) as the evaluation\training_loss, with a different granularity.

w4nderlust:
Hey @carlogrisetti , the tests don;t pass and I undesrand why: you assume that there's an accuracy measure, but in some cases there is not. Instead of assuming what measures are there you can do something as simple as cycling through the ones available in the progress tracker, something like:
```
def add_tensorboard_summary(stats, prefix, train_writer, step):
    summaries = []
    for output_f in stats:
        for metric in output_f:
            metric_tag = "evaluation/{}_{}_{}".fomart(prefix, output_f, metric)
            metric_val = test_stats[output_f][metric][-1]
            summaries.append(
                tf.compat.v1.Summary.Value(tag=metric_tag, simple_value=metric_val)
             )
     test_summary = tf.compat.v1.Summary(value=summaries)
     train_writer.add_summary(test_summary, step)

add_tensorboard_summary(progress_tracker.train_stats, "train", train_writer, progress_tracker.steps - 1)
add_tensorboard_summary(progress_tracker.vali_stats, "validation", train_writer, progress_tracker.steps - 1)
add_tensorboard_summary(progress_tracker.test_stats, "test", train_writer, progress_tracker.steps - 1)
```
This will dinamically adapt to whatever is in the tracker.

Regarding "evaluation/" I think  it's fine, we can change later the older summaries to have a "train_batch" prefix, so that it's clear that evaluation is for each epoch, train batch for each batch.

carlogrisetti:
... couldn't have done it better... and in fact I didn't :)

That makes totally sense... I just reformatted and declared that as a classmethod since it has all external parameters passed to it and uses no class objects implicitly. I hope in turn it makes sense to you. I am targeting only the combined measures, since they're the only ones having a "global" meaning in regards to the training.

On the naming convention maybe we could go this way:
- `evaluation_epoch` (or `evaluation_by_epoch`, or even `Evaluation (by epoch)`) under which to move everything i've added here)`
- `evaluation_batch` (or `evaluation_by_batch`, or even `Evaluation (by batch)`) under which to move train_reg_mean_loss only, since all the others are grouped by feature)
I wouldn't call them batch_evaluation and epoch_evaluation (prefixing the granularity), just for the sake of alphabetical ordering and thus graphical output in TensorBoard.

If that's ok, I could also edit the train_reg_mean_loss to prefix with the `evaluation_batch/` tag (or whatever you choose to be the best fit) and push it alltogether.

w4nderlust:
I'm not sure if one could do nested groups like "evaluation/feature_name/measure" and "batch/feature_name/measure", but in case we probably wnat to do that.

carlogrisetti:
It would be the ideal arrangement, but unfortunately you can't do that and are limited to one level of grouping, so batch/epoch "tag" and feature name should be in some way combined.
We could also prepend "batch_" to any batch-level measure and "epoch_" to the tracker ones I've added (as epoch_combined)

w4nderlust:
> It would be the ideal arrangement, but unfortunately you can't do that and are limited to one level of grouping, so batch/epoch "tag" and feature name should be in some way combined.
> We could also prepend "batch_" to any batch-level measure and "epoch_" to the tracker ones I've added (as epoch_combined)

Got it, yes that makes sense!

carlogrisetti:
I'm having some trouble finding where to inject the "batch_" prepend in the single measure tag for all the other features, since it seems this is handled autonomously by TF, and I don't want to mess too much with the inner workings of Ludwig (I don't really have the expertise to do that). Any help you can provide on that?

However, I think this could already be worthy of a merge at the current state. If all the already existing measures get a "batch_" prepend, this could be also worthy of a changelog mention in the next release, not to get confusing when comparing old training results with a new one.

w4nderlust:
You can search for "summary" and see where they are, but I can already tell you there is one in model.py (train_reg_mean_loss) and the other ones are in the output feature objects of each datatype. For instance, for numerical, you have to look in `features/numerical_feature.py` and summaries definitions are in `NumericalOutputFeature.build_output()`. The same is true for all other feature types.
If it's too much work don't worry, I can do it myself after we merge your PR.

carlogrisetti:
I already patched the `train_reg_mean_loss` as `epoch_combined/train_reg_mean_loss` in the last commit.

I understood the other values are built in each feature, but couldn't understand how they are given a name. Maybe, now knowing I was looking in the right spot, I'll try to have a look at it tomorrow and otherwise will defer to you for fixing that last thing.

Have a good one

carlogrisetti:
I never got around to do this, and I don't know when I'll have time to look deeper into that.
Up to you to choose to wait, fix yourself, or merge as is

w4nderlust:
No worries, will merge and work on the naming convention.

PR #564, opened 2019-11-09
Initial api example

jimthompson5802:
# Code Pull Requests

@msaisumanth This PR provides an example of addressing #560.  

Summary of changes:
* added  `examples` sub-directory. 
* added `examples/titanic` illustrating using the Ludwig Api for
   * model training
   * generating visualizations
   * demonstrating how to extract training statistics for custom analysis

PR #578, opened 2019-11-21
Mnist api example

jimthompson5802:
# Code Pull Requests
Edit:
@w4nderlust   Ready to merge.  I'm submitting the PR so that I can get some feedback.  This example contains the following examples of using the Ludwig api:

|File|Description|
|----|-----------|
|create_training_test_data.sh|Bash script to create MNIST training and test data sets.|
|simple_model_training.py|Demonstrates using Ludwig api for training a model.|
|advance_model_training.py|Demonstrates a method to assess alternative model architectures.|
|assess_model_performance.py|Assess model performance on hold-out test data set.  This shows how to load a previously trained model to make predictions.|
|visualize_model_test_results.ipynb|Example for extracting training statistics and generate custom visualizations.|

My question is related to the visualizations in the Jupyter notebook `visualize_model_test_results.ipynb`.  I'm not sure if the output for  `ludwig.vlsualize.compare_classifiers_performance_from_pred` is correct.  I was expecting the `accuracy` measure to match in both set of charts.  But this does not seem to be the case.

**compare_performance chart**
compare_performance_label

**compare_classifiers_performance_from_pred**
compare_classifiers_performance_from_pred

In addition, I'm not clear on the purpose for the `labels_limit` in `compare_classifiers_performance_from_pred()` function.  According to the docstring:
```
labels_limit: Maximum numbers of labels.
             If labels in dataset are higher than this number, "rare" label
```

To recreate this situation, run the following:
```
# create training and test data sets
./create_training_test_data.sh

# create three neural network models
python advance_model_training.py

# Assess performance of the three neural network models
python assess_model_performance.py
```
Open the Jupyter notebook `visualize_model_test_results.ipynb`

Any guidance will be appreciated.


jimthompson5802:
Resolved issue with `accuracy` score in `compare_classifiers_performance_from_pred()`.  Resolution was to make sure predicted class and ground truth are stored as integers.  With this change `compare_classifiers_performance_from_pred()` chart is now consistent with the `compare_performance()` chart.
compare_classifiers_performance_from_pred

jimthompson5802:
@w4nderlust   This the second Python API example.  If you'd like to see changes, let me know.

This PR depends on PR #564 being merged first.

examples/mnist/advance_model_training.py
```diff
@@ -0,0 +1,92 @@
...
+    os.remove(file_list)
+except:
+    pass
+
+# read in base model definition
+with open('./model_definition.yaml', 'r') as f:
+    base_model = yaml.safe_load(f.read())
+
+# Specify named tuple to keep track of training results
+TrainingResult = namedtuple('TrainingResult', ['name', 'train_stats'])
```
w4nderlust: Could you please change the names of variables (everywhere) to be lowercased and use _ as separators? like `training_results`.

examples/mnist/simple_model_training.py
```diff
@@ -0,0 +1,43 @@
...
+
+# clean out prior results
+try:
+    shutil.rmtree('./results')
+except:
+    pass
+
+
+# set up Python dictionary to hold model training parameters
+model_definition = {...}
```
w4nderlust: here you can load the YAML with `yaml.safe_load(path)` and then use the loaded dict when initializing the model.

jimthompson5802: made the requested change

examples/mnist/advance_model_training.py
jimthompson5802: I understand you'd like me to change `TrainingResult` to be `training_result`.  

Before making this change, I'd like to explain the rationale for `TrainingResult`.  If after reading this rationale you still want the change, I'll make the change.

`TrainingResult` is defined by the factory function `collections.namedtuple()`.   I viewed this a new class definition so followed a the class naming convention in PEP8.  This is similar to the example code found in the [`namedtuple()` api documentation](https://docs.python.org/3/library/collections.html#collections.namedtuple), e.g.,
```
EmployeeRecord = namedtuple('EmployeeRecord', 'name, age, title, department, paygrade')
```
This is why I used CamelCase for `TrainingResult`. 

Do you still want me to change `TrainingResult` to 'training_result`?

w4nderlust: I see, so that's a factory of constructors. Understood, let's keep it like it is.

Issue #565, opened 2019-11-11
ludwig test sometimes does not produce correct confusion matrix values


carlogrisetti:
This has happened to me multiple times, and I am now trying to get to the root cause of this.
Basically, really often, this is the confusion matrix that I generate from the model+dataset I am working on:
![image](https://user-images.githubusercontent.com/4464640/68606216-45cb6680-04ae-11ea-88ef-42ed36eae966.png)
As you can see there are some actual values that are never predicted, but they should at least be predicted as unknown (note that I have set the --normalize flag, so total sum per-row should always be 1)
I checked in the test_statistics.json (attached)
[test_statistics.zip](https://github.com/uber/ludwig/files/3832450/test_statistics.zip)
and the values reported for the confusion matrix always show a 0 in the corresponding column.
In the other result files, all source items have a corresponding prediction (right or wrong, that's not the point)
[results.zip](https://github.com/uber/ludwig/files/3832460/results.zip)

This is the commandline I use to generate the results:
ludwig test --data_csv dataset.csv --split validation --model_path path\to\model --batch_size 12
This is the commandline I use to generate the confusion matrix:
ludwig visualize --visualization confusion_matrix --normalize --test_statistics results\test_statistics.json --ground_truth_metadata dataset.json

I also tried using hdf5 in place of the .csv, but with no avail. Actual predictions return a result (of course). Given the performance values of this example it might be the wrong result, but it's none the less a result that should be mapped in the confusion matrix.

For this specific one I cannot share the dataset (personal IDs) but anything else I can do, please advise.

w4nderlust:
Not sure what is the problem here. As some labels are never predicted, the sum of the row is 0. In theory each value of the row should be 0/0 but i avoid doing division by zero so the resulting value is 0. It sounds all correct to me, what behavior would have you expected?
The relevant code is here anyway: https://github.com/uber/ludwig/blob/master/ludwig/visualize.py#L2738-L2743

carlogrisetti:
Since PP_FRONTE, PG_RETRO and CF_FRONTE (for example, bunching up the last ones) are present in the validation dataset, I expected to see a value greater than 0, in whichever predicted label have they been predicted.
I just double-checked again and purged .hdf5 and .json, they are present in the validation dataset, but they are not in output, according to the confusion matrix (and more importantly according to the confusion matrix data within test_statistics.json).

[...]

PR #579, opened 2019-11-22
Fix inverted confusion_matrix axis

carlogrisetti:
This is related to issue #565 

confusion_matrix accepts two arrays, but the first one should be the true value, the second one should be the predictions made to be compared. It was inverted right here.

### DIFF
--- a/ludwig/utils/metrics_utils.py
+++ b/ludwig/utils/metrics_utils.py
@@ -48,8 +48,8 @@ class ConfusionMatrix:
             self.idx2label = {idx: str(label) for idx, label in
                               enumerate(np.unique(
                                   [self.predictions, self.conditions]))}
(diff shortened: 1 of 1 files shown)


carlogrisetti:
The only known issue with this fix is that, when no predicted classes are found (ie: classes that have no prediction made with the test you're running), some warnings from sklearn and a bad graph gets outputted, since there are missing classes in output. I think this is a "wontfix" case, since the input data is malformed (not all classes are present), but an improvement would be to always display class name, regardless of it being present or not.

carlogrisetti:
FYI, days after this, it looks 100% good and have seen no side effects whatsoever on multiple datasets.

w4nderlust:
Thank you, this looks good!

PR #584, opened 2019-12-01
Hack to override in_memory flag for API usage

ydudin3:
Tested by initializing a pretrained model and then calling predict function of the model API with dataframe and a csv separately.

### DIFF
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -385,6 +385,14 @@ def add_sequence_feature_column(df, col_name, seq_length):
     df[new_col_name] = new_data
     df[new_col_name] = df[new_col_name].fillna(method='backfill')
 
+def override_in_memory_flag(input_features, override_value):
+    num_overrides = 0
+    for feature in input_features:
+        if 'preprocessing' in feature:
+            if 'in_memory' in feature['preprocessing']:
+                feature['preprocessing']['in_memory'] = override_value
+                num_overrides += 1
+    return num_overrides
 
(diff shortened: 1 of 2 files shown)


ludwig/utils/data_utils.py
```diff
@@ -385,6 +385,14 @@ def add_sequence_feature_column(df, col_name, seq_length):
     df[new_col_name] = new_data
     df[new_col_name] = df[new_col_name].fillna(method='backfill')
 
+def override_in_memory_flag(input_features, override_value):
+    num_overrides = 0
+    for feature in input_features:
+        if 'preprocessing' in feature:
+            if 'in_memory' in feature['preprocessing']:
+                feature['preprocessing']['in_memory'] = override_value
```
msaisumanth: @ydudin3 what if the user sets 'in_memory': True ? This is ignoring the user specified value, no?

msaisumanth: Also, why do we need to count the number of overrides, if we're just dong num_overrides > 0?

ydudin3: > @ydudin3 what if the user sets 'in_memory': True ? This is ignoring the user specified value, no?

Yep, we want to always override to true regardless of what user specified.

> Also, why do we need to count the number of overrides, if we're just dong num_overrides > 0?

Made it like this in case we want to use it later for some reason. Can change it to bool if you want, don't have a strong preference.

w4nderlust:
This looks fine. It's a temporary fix, we'll find a better solution when working on the refactoring of the preprocessing pipeline.

Issue #585, opened 2019-12-02
learning_curves() function fails with ValueError: x and y must have same first dimension


jimthompson5802:
**Describe the bug**
learning_curves() function fails when plotting two more more models at once.  Error message:
```
Traceback (most recent call last):
  File "/opt/project/sandbox/mwe.py", line 76, in <module>
    learning_curves(train_stats_list,'y', model_names=model_names, output_directory='./viz', file_format='png')
  File "/usr/local/lib/python3.6/dist-packages/ludwig-0.2.1-py3.6.egg/ludwig/visualize.py", line 740, in learning_curves
  File "/usr/local/lib/python3.6/dist-packages/ludwig-0.2.1-py3.6.egg/ludwig/utils/visualization_utils.py", line 94, in learning_curves_plot
  File "/usr/local/lib/python3.6/dist-packages/matplotlib/axes/_axes.py", line 1666, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/usr/local/lib/python3.6/dist-packages/matplotlib/axes/_base.py", line 225, in __call__
    yield from self._plot_args(this, kwargs)
  File "/usr/local/lib/python3.6/dist-packages/matplotlib/axes/_base.py", line 391, in _plot_args
    x, y = self._xy_from_xy(x, y)
  File "/usr/local/lib/python3.6/dist-packages/matplotlib/axes/_base.py", line 270, in _xy_from_xy
    "have shapes {} and {}".format(x.shape, y.shape))
ValueError: x and y must have same first dimension, but have shapes (100,) and (24,)
```
Ludwig version: 0.2.1

**To Reproduce**
This program will reproduce the error:
```
#
# Minimum Working Example for learning curves Plot error with early stopping
#


#%%
import numpy as np
import pandas as pd
import warnings
warnings.simplefilter('ignore')

import ludwig
from ludwig.api import LudwigModel
from ludwig.visualize import learning_curves

print(ludwig.__version__)

#%%
NUM_ROWS = 5000
NUM_COLS = 20

#%% create synthetic training data explanatory variables
np.random.seed(42)

# create relevant explanatory variables
X = np.random.rand(NUM_ROWS, NUM_COLS)
X_df = pd.DataFrame(X)
X_df.columns = ['X_' + '{:0>3d}'.format(i+1) for i in range(NUM_COLS)]

#%% create response variable
np.random.seed(21)
beta = -2 + 4 * np.random.rand(NUM_COLS)
[...]

PR #586, opened 2019-12-02
fix: failure in generating learning curves plot due to early stopping

jimthompson5802:
# Code Pull Requests

Fix Issue #585.

The reproducible example from #585 no longer fails.  

These are the generated learning curve plots after implementing the proposed fix:
learning_curves_y_accuracy

learning_curves_y_loss


msaisumanth:
Thanks for the fix @jimthompson5802

PR #608, opened 2020-01-09
Add k fold cv cli feature

### LINKED ISSUE #462: k-fold cross validation for training models with CLI
**Describe the use case**
Training multiple models with shuffled training data (k-folds) can reveals information about our data sets. For example, one fold may be much less accurate than other folds telling us that we might need to increase the total number of samples.

For the CLI, it would be useful if we could have a parameter to train with k-fold cross validation.

k-fold can be achieved with the python API but not as convenient as the CLI.

**Additional context**
I can provide the sample code I've been using for the API.

jimthompson5802:
# Code Pull Requests

Resolves #462.  

#### Summary of changes:
```
Added function kfold_cross_validate() function in train..py that calls full_train() k times. 

Added function generate_kfold_splits() function to utils/data_utils.py

Create `kfold_training_statistics.json` statistics file to hold results from the k-fold cv run

```
A new parameter `--k_fold` is now available for the `ludwig train` command.  This is the help output for the new parameter:
```
 -kf K_FOLD, --k_fold K_FOLD
                        number of folds for a k-fold cross validation run
``` 

#### Work to be completed:
* Add `'aggregate': {...}` node to the `fold_training_statistics.json` file
* Parallel processing for k-fold cv.  Note: I have a working version of parallel processing in a separate branch.  I figure it is easier to debug issues with the serial version.  Once we are comfortable with how k-fold cv processing occurs, I'll add the parallel processing functionality.  Parallel processing is implemented using the `joblib` package.
* add unit test for k-fold cv function
* update documentation


jimthompson5802:
@jaipradeesh @mohi7solanki @sanketsaurav  I'm reaching out to you because in the recent past you offered guidance re: issues I encountered with deepsource.io code quality checks.  I'm hoping you will be able to help me again.

When I submitted this PR, deepsource flagged three major issues with the code.  Since it was late at night I just briefly skimmed thorough the issues with the intent to work on them the following morning.  When I went to look at the issues the next day, the list of issues is no longer presented.  

This is the screenshot of what I now see.  According to this page, it says there are no issues.  The night before there were a list of issues presented.  
deepsource_missing_issues

Is there anything I need to do to make the issues visible so I can work on fixing them.

mohi7solanki:
@jimthompson5802 There's a small glitch on our platform, we are working on this, sorry for the inconvenience.

jimthompson5802:
@mohi7solanki   Thank you for the update.  I'll wait until the glitch is resolved.

mohi7solanki:
@jimthompson5802 you can view the result now.

jimthompson5802:
@mohi7solanki   Thank you for the update.  I can now see the issues.  The look for the UI changed.  It took a few seconds but I kinda like the new look and feel.   However,  I miss not having the short explanatory text.

jaipradeesh:
@jimthompson5802 We're on it. The descriptions in the run page will be back.

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
 
 
+def kfold_cross_validate(
+                model_definition,
+                model_definition_file=None,
+                data_csv=None,
+                data_train_csv=None,
+                data_validation_csv=None,
+                output_directory='results',
+                k_fold=None,
```
w4nderlust: I would make this parameter mandatory

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
+            data_df.iloc[train_index].to_csv(train_csv_fp, index=False)
+            data_df.iloc[test_index].to_csv(test_csv_fp, index=False)
+
+            # train and validate model on this fold
+            logger.info("training on fold {:d}".format(fold_num))
+            (model,
+             preprocessed_data,
+             _,
+             train_stats,
+             model_definition) = full_train({},
```
w4nderlust: the {} should not be needed

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
+                                            model_definition_file=model_definition_file,
+                                            data_train_csv = train_csv_fp,
```
w4nderlust: in this nad next line please remove the spces before and after the `=` (in general, if you run the pycharm linter with a limit of 80 characters per line, that would be great)

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
+                                            data_test_csv = test_csv_fp,
+                                            experiment_name='cross_validation',
+                                            model_name='fold_' + str(fold_num),
+                                            output_directory=os.path.join(temp_dir_name,'results'))
+
+            # score on hold out fold
+            preds = model.predict(preprocessed_data[2],
+                                  model_definition['training']['batch_size'])
```
w4nderlust: this should be `eval_batch_size`. If `==0` then `batch_size`.

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
+    if not os.path.isdir(output_directory):
+        os.mkdir(output_directory)
+
+    # read in data to split for the folds
+    data_df = pd.read_csv(data_train_csv)
+
+    # place each fold in a separate directory
+    data_dir = os.path.dirname(data_train_csv)
+    kfold_training_stats = {}
+    for train_index, test_index, fold_num in generate_kfold_splits(data_train_csv, k_fold):
```
w4nderlust: maybe an option to save the train and test indices for each fold would be useful for reproducibility.

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
+                                            experiment_name='cross_validation',
+                                            model_name='fold_' + str(fold_num),
+                                            output_directory=os.path.join(temp_dir_name,'results'))
+
+            # score on hold out fold
+            preds = model.predict(preprocessed_data[2],
+                                  model_definition['training']['batch_size'])
+
+            # augment the training statistics with scoring metric fron the hold out fold
+            train_stats['fold_metric'] = preds['combined']
```
w4nderlust: why only combined and not all output features?

ludwig/train.py
```diff
@@ -396,6 +400,68 @@ def full_train(
...
+
+            # collect training statistics for this fold
+            kfold_training_stats['fold_'+str(fold_num)] = train_stats
+
```
w4nderlust: before saving i would add an additional `'overall'` key to the kfold stats that for each output feature takes the overall and adds mean and std.
Something like:
```
overall_kfold_stats = {}
for fold_stats in kfold_training_stats:
    for of in fold_stats:
        if OVERALL in of:
            for metric in of[OVERALL]:
                if metric not in overall_kfold_stats:
                    overall_kfold_stats[metric] = []
                overall_kfold_stats[metric].append(of[OVERALL][metric])
for metric in overall_kfold_stats:
    mean = np.mean(overall_kfold_stats[metric])
    std = np.std(overall_kfold_stats[metric])
    del overall_kfold_stats[metric]
    overall_kfold_stats[metric + "_mean"] = mean
    overall_kfold_stats[metric + "_std"] = std
kfold_training_stats[OVERALL] = overall_kfold_stats
```

ludwig/utils/data_utils.py
```diff
@@ -418,3 +419,11 @@ def default(self, obj):
             return obj.tolist()
         else:
             return json.JSONEncoder.default(self, obj)
+
+def generate_kfold_splits(data_train_df_fp, k_fold):
```
w4nderlust: this should probably get the DataFrame directly as input

ludwig/train.py
jimthompson5802: Couple of questions:

1) I just want to confirm that when you write, "this parameter" you mean the "k_fold" parameter.  Is this correct?

2) Re: the function signature.  Couple of major issues called out by deepsource code quality check identified unused parameters `data_csv` and 'data_validation_csv`.  To resolve those issues, I removed them from the function definition.  Right now I'm assuming that it does not make sense to duplicate all the parameters from `full_train()` into `kfold_cross_validate()`

ludwig/train.py
jimthompson5802: removed `{}`

ludwig/train.py
jimthompson5802: Question related to this statement.  On line 450, I've coded this construct `preprocessed_data[2]`.   Is this the best way to refer to the particular data structure?  I'm usually concerned by hard coding a specific numeric value.

ludwig/train.py
jimthompson5802: Any particular preference on data format for storage?  json?  pickle?

ludwig/train.py
jimthompson5802: Picking `combined` was MVP to illustrate functionality of saving k-fold statistics to disk.  I can update to collect all the metric sets to save.

ludwig/train.py
jimthompson5802: completed changes

ludwig/train.py
jimthompson5802: Looks like I was a bit pre-mature in calling this resolved.

There may be an issue in removing `{}` in line 441.  The function definition for `full_train()` looks like this:
```
def full_train(
        model_definition,
        model_definition_file=None,
        data_df=None,
        data_train_df=None,
```

When I remove `{}` from the call to `full_train()` an error is raised.

It appears to remove `{}` from line 441, we have to modify `full_train()`.  

Do you want me to take a stab at modifying `full_train()` as well?

ludwig/utils/data_utils.py
jimthompson5802: made modification to directly use DataFram as input to the `generate_kfold_splits()`

ludwig/train.py
jimthompson5802: This is the new def function:
```
def kfold_cross_validate(
                k_fold,
                model_definition_file=None,
                data_train_csv=None,
                output_directory='results',
                **kwargs
):
```

A follow-up question, should `model_definition_file` and `data_train_csv` be mandatory as well?

ludwig/train.py
jimthompson5802: implemented `eval_batch_size if eval_batch_size != 0 else batch_size`

ludwig/train.py
jimthompson5802: Decided to use json format to make it consistent with other saved information. Saved split indices saved to a data set called `kfold_split_indices.json` in the `output_directory`.  Format of data set looks like this
```
{
   "fold_1": {"test_index": [1, 3, 10,  ...], "training_index": [2, 4, 5,...]  },
   "fold_2": {"test_index": [...], "training_index": [...]},
   "fold_3": {"test_index": [...], "training_index": [...]}
}
```

added command line parameter to skip saving kfold indices. 
```
    parser.add_argument(
        '-skfsi',
        '--skip_save_k_fold_split_indices',
        action='store_true',
        default=False,
        help='disables saving indices generated to split training data set ' 
             'for the k-fold cross validation run, but if it is not needed '
             'turning it off can slightly increase the overall speed'
    )
```

ludwig/train.py
jimthompson5802: Added `'overall'` key.  This contains the mean and standard deviation for the various metrics (loss, accuracy) from the hold out fold.  Here is an example:
```
   "overall": {
        "Survived": {
            "accuracy_mean": 0.39955106621773284,
            "accuracy_std": 0.015632310075403076,
            "loss_mean": 6.96722408684282,
            "loss_std": 0.5409600467679053
        },
        "combined": {
            "accuracy_mean": 0.39955106621773284,
            "accuracy_std": 0.015632310075403076,
            "loss_mean": 6.96722408684282,
            "loss_std": 0.5409600467679053
        }
    }

```
The structure of the `fold_training_statistics.json` file is now
```
{
  "fold_1": {...}
 "fold_2": {...}
 "fold_3": {...}
 "overall": {...}
}
```

ludwig/train.py
jimthompson5802: Ignore my earlier comment.  I just realized what you were asking me to do. This issues was resolved by this coding sequence:
```
           # train and validate model on this fold
            with open(model_definition_file, 'r') as def_file:
                model_definition = merge_with_defaults(yaml.safe_load(def_file))
            logger.info("training on fold {:d}".format(fold_num))
            (model,
             preprocessed_data,
             _,
             train_stats,
             model_definition) = full_train(
                model_definition,
                data_train_csv=train_csv_fp,
                data_test_csv=test_csv_fp,
                experiment_name='cross_validation',
```

ludwig/train.py
w4nderlust: In the current API you have model_definition and model_definition_file both None and then there's some logic to chack tha at least one of them is not none: https://github.com/uber/ludwig/blob/master/ludwig/api.py#L141-L156
You could do the same.

Regarding data_train_csv, I would actually call it data_csv. The rationale being that that data is going to be split in folds and used for both training and test, so I'd rather be more generic.

ludwig/train.py
w4nderlust: You are right, I added that check from API also in full_train: https://github.com/uber/ludwig/commit/7826a1b4f594cf760d5403929fa1ab41302b53de
now you can pass only the file path and it will work.

ludwig/train.py
w4nderlust: looks good

ludwig/train.py
w4nderlust: Yeah all of them would be better

jimthompson5802:
@w4nderlust 
This last commit updates the ludwig user guide to describe the parameters that were implemented to support the k-fold cross validation function.

Is there anything else for this this change?

This current implementation performs sequential processing for the k-fold processing.  I do have in a separate branch a working version of parallel processing.  If it is desired to provide parallel processing, I have a few questions:
* Should the parallel function be added to this PR or should I start a new PR the parallel work.
* Do we want to give the user the ability select either parallel or sequential processing
* Current parallel implementation uses the `joblib` package.  I noticed the `image_feature.py` module uses `multiprocessing` package for parallel processing.  I'm assuming you want to minimize dependencies and I should migrate from `joblib` to `multiprocessing` package.

jimthompson5802:
added pytest unit test for k-fold cv function

jimthompson5802:
@w4nderlust   As far as I can tell, I've incorporated all your comments re: the k-fold CV function.  Is there anything else I need to do?

Note: The build failure is due to issue #620, which you put in a temporary work-around and I'm looking at the longer-term fix.

w4nderlust:
> @w4nderlust As far as I can tell, I've incorporated all your comments re: the k-fold CV function. Is there anything else I need to do?

It should be fine, I just need to run it myself locally and do a final pass of checks before merging, will not take long.
 
> Note: The build failure is due to issue #620, which you put in a temporary work-around and I'm looking at the longer-term fix.

That's great thanks!

w4nderlust:
So I reviewed and tested everything, this looks great. I only have one question left:  I'm not sure why do you actually need to save the splitted data as csv files https://github.com/uber/ludwig/pull/608/files#diff-0b9593aca94215c7fe8d14388e532b52R479-R487 , why not calling `full_training` with the dataframes obtained by `iloc`? As there's the option to save the indices, I think saving the splitted data as csv may not be needed. What do you think?

w4nderlust:
I actually moved everything to experiment myself. The only open point is about saving or not the datasets in the temp file.

w4nderlust:
I replaced the temp csv part with in memory dataframes. Now I-m merging, let me know if you have a strong opinion about this change, in case we can discuss and figure out the best solution.

jimthompson5802:
@w4nderlust   Thank you for the kind comments and merging the branch.
 
**re: replacing temp csv with dataframes.**  I don't have any concerns.  Given the current sequential implementation dataframes make sense.  

The reason I originally took the approach for temp csv was in anticipation of eventually implementing parallel k-fold runs.  For this approach I anticipated we would use the` fork()` call through one of the `multiprocessing` package capabilities.  If the training data was large, I anticipated there may be problems with passing the large amount of data to the children processes in memory. I don't know this for a fact, it is just a supposition at this time.  

 I thought passing the temp file name and just the indices that make up the training fold and test fold would use less memory compared to the training data itself when creating the child process.  With each child running in parallel, I thought the parallel reading of the temp csv files would compensate for the multiple reading of the csv file.

In any event, I'm OK with the last change you made. 

Right now I believe all of your last set of comments have been addressed.  If this is not true, just let me know what else I need to do.

If everything has been taken care of, I've thought of some additional enhancements:
* Adding visualization for the k-fold results.  For this I was thinking of something like a box-whisker chart for each metric over the k-folds.  Something like this:
Screen Shot 2020-02-01 at 21 59 48 
The mean value for the metric would be added to the plot.
There would be one plot for each Model specified.

* Provide an option to run the k-fold runs in parallel.

w4nderlust:
I see your point on the parallelization, but maybe we'll be able to find a good solution for that case that does not require exchanging many object between processes/threads.

In terms of next steps, The visualization would be lovely, there is already code for a multimodel / multimetric comaprison, maybe that one can be a base for something similar but with the error bars.

Regarding parallelization, as many Ludwig models are heavy on CPU during training and / or use most of the GPU if you have one, I think in the immediate that is not a huge need, although it can definitely be useful in a cluster o multi-machine / multi-gpu setting.

What I believe is the natural next step is actually to expose the kfold command in the API. It should be pretty straightforward, but I believe it could be the best in terms of utility/effort ration, what do you think?

jimthompson5802:
got it...makes sense.  I'll work on the api and visualization capabilities.  I'll start with the api one.

PR #617, opened 2020-01-23
Change preprocessing at Siamese example

flozi00:
Change the preprocessing syntax in Siamese network example
Changing second image to the same as the first image has


w4nderlust:
Thank you!

PR #625, opened 2020-01-29
open vocab with utf-8

jenishah:
# Open vocabulary file with utf-8 encoding

[load_vocabulary](https://github.com/uber/ludwig/blob/997c6bc02dffcc726efda7ddbb5978e425ffab96/ludwig/utils/strings_utils.py#L76) gives UnicodeEncodeError in RHEL 7.6

- Open the vocab file with utf-8 encoding.


w4nderlust:
Thanks

Issue #328, opened 2019-05-06
CLI loss\accuracy output is displayed incorrectly after a training resume


carlogrisetti:
**Describe the bug**
While trying to build an image classification model, with a 2-categories output, TensorFlow (through TensorBoard) reports a gradually decreasing loss value (too good to be true, to be honest) and an accuracy approaching 1 over time, however within the Ludwig CLI the loss values are all over the place. I'm talking about <0.01 loss in tensorflow and >300 loss in CLI.
This is not happening using a stacked_cnn encoder of any size\shape
Input image is sized 300x300x1 (grayscale). I know that's not an optimal input, but that's what I'm working with.

**To Reproduce**
I don't know the specifics of how to repro the issue, since I'm not able to get a "working" example. It's only happening using resnet encoding, and I tried with the simplest one, of size 8 and with all the default values.
Here's the YAML
input_features:
    -
        name: ImageFile
        type: image
        encoder: resnet
        resnet_size: 8
        num_filters: 16
        kernel_size: 3
        conv_stride: 1
output_features:
    -
        name: Type
        type: category
training:
        batch_size: 64
        learning_rate: 0.0001
        epochs: 100
        early_stop: 30
        decay: true
        validation_measure: loss

**Expected behavior**
CLI values and TensorFlow Logs values to be at least similar (I understand there are some aggregations performed, but they're in two completely different directions)

**Screenshots**
![image](https://user-images.githubusercontent.com/4464640/57237685-02d49600-7028-11e9-81a0-989f91f77766.png)
![image](https://user-images.githubusercontent.com/4464640/57237708-0f58ee80-7028-11e9-8f68-1e410fd59172.png)
(orange is batch size 32, blue is batch size 64. Loss in the CLI goes 8,15,20,24 etc while in the graph is <1)
![image](https://user-images.githubusercontent.com/4464640/57237749-20096480-7028-11e9-8b7f-dc37093019f9.png)
(the only save I get is the first one, after that loss value goes away and model never improves)

**Environment (please complete the following information):**
 - OS: Windows 10
- Python 3.6.3
- Ludwig 0.1.2

carlogrisetti:
Oh I forgot to mention I also tried with ResNet size 14, but had the same exact "exploding" result, with no correlation in the TensorFlow Logs.
The really weird thing is that loss\accuracy values with ResNet 14 *exacly* match the ResNet 8 run
[...]

PR #636, opened 2020-02-13
Fix CLI output table column ordering

carlogrisetti:
This implements a fix to issue #328 
I know it's ugly and might be prettier, but if checks pass it's a valid solution that can be improved upon.

I already tested the cases where it happened to me, and it's not doing that anymore. Hope it passes CI tests.


w4nderlust:
Refactored the code a little bit and cleaned it up. Can you please confirm this fors in your usecase with resuming training?

PR #637, opened 2020-02-14
[WIP] TF2 porting

w4nderlust:
This PR is for porting Ludwig to TF2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,6 @@ python:
   - "3.6"
 install:
   - pip install '.[test]'
-script: pytest
+# TODO: re-enable tests/integration_tests/test_contrib_wandb.py after resolving issue with TF2
+script:
+  - pytest --ignore tests/integration_tests/test_contrib_wandb.py tests
--- a/ludwig/features/audio_feature.py
+++ b/ludwig/features/audio_feature.py
@@ -19,7 +19,7 @@ import os
 import sys
 
 import numpy as np
-import tensorflow as tf
+import tensorflow.compat.v1 as tf
 
 from ludwig.constants import AUDIO, BACKFILL
 from ludwig.features.base_feature import BaseFeature
@@ -358,7 +358,7 @@ class AudioInputFeature(AudioBaseFeature, SequenceInputFeature):
                 'check "update_model_definition_with_metadata()"')
 
     def _get_input_placeholder(self):
-        return tf.compat.v1.placeholder(
+        return tf.placeholder(
             tf.float32, shape=[None, self.length, self.embedding_size],
             name='{}_placeholder'.format(self.name)
         )
--- a/ludwig/features/bag_feature.py
+++ b/ludwig/features/bag_feature.py
@@ -18,7 +18,7 @@ import logging
 from collections import Counter
 
 import numpy as np
-import tensorflow as tf
+import tensorflow.compat.v1 as tf
 
 from ludwig.constants import *
 from ludwig.features.base_feature import BaseFeature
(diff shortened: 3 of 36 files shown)


w4nderlust:
@jimthompson5802 This is a starting point. There are many things that still don't work and that needs to be changed. Moreover I removed some functionalities because they don't mix well between TF1 and TF2, in particular: the FileWriter to save statistics, the regularizers and the saver to save weights. You can find a bunch of `todo tf2` comments, to differentiate from other todos that were already there.

jimthompson5802:
Got it...one question on mechanics.   Up to now, I've usually created the branch and PR and pushed to commits to the branch I created.  This is the first time, I'm going to contribute to a branch created by someone else.   Based on what I know about GitHub, I believe we'll interact in this manner:
* I'm assuming any commits I make must be worked through my fork `jimthompson5802/ludwig`.
* I need to create a `tt2_porting`  tracking branch in my fork and local repo.
* I'll create feature branch off of `tf2_porting` for each set of changes and push commits to my fork.
* I'll then create a pull request to merge from my feature branch to the `tf2_porting` branch.

Is this the procedure I should follow?

jimthompson5802:
In reading GitHub.com docs it appears that if I have 'write' access to `uber/ludwig`, I can push commits directly to this repo.  
Just let me know how I should submit my commits:  push directly to `uber/ludwig` or work through my fork `jimthompson5802/ludwig` with a pull request.
I've successfully created the `tf2_porting` branch in my local repo and I'll start reviewing the TODO items.

jimthompson5802:
I've looked at the TODO list and I wanted to confirm my understanding.

In an earlier exchange 
>.1. Change all the imports to compat.v1 and try to run to TF2. If that works fine (as it should), then we can change the dependency to be TF2 and forget about TF1. The caveat here is that come thing, in particular the contrib package, have been moved out of TF into the TF addons package, so there could be things that don't work immediately out of the box

I think we are in this phase of the work.  

I see you've already incorporated the `compat.v1`, i.e,.
```
import tensorflow.compat.v1 as tf
```
In theory Ludwig should work in TF2 with the above change.  However, errors occurred.  This led marking TF2 specific TODOs like this in `model.py`, such as this
```
        # todo tf2: fix this
        # summary = tf.Summary(value=summaries)
        # train_writer.add_summary(summary, step)
```
I'm testing changes using the `pytest` unit tests.  When I run the `pytest` tests in a TF2 Docker container I see many errors, some tied to the code fragments marked with the `TODO tf2` marker.

I believe my focus is to get a clean run in `pytest` with the `import tensorflow.compat.v1 as tf` construct in TF2.

Is this a correct understanding?

ifokeev:
FYI:  take a look, please. It's auto upgrade to TF2. It will not work from the box, but it's a good example how to update Ludwig scripts.
https://www.tensorflow.org/guide/upgrade
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/compatibility/tf_upgrade_v2.py

w4nderlust:
A few answers:

- work on your fork, if you are unable to push to this pr, it's fine, I can delete this PR and you can create a new one, and I can push to that PR.
- yes the todos are things I removed to make it work in one specific usecase (nlu example) although it still doesn't work fine, at least it get to the end of training and prediction. So those todo are thing one would want to bring back, but are by no means the only broken things.
- Yes ideally all tests pass and then we can re-add the things I removed (filewriter and regularizers)
- regarding @ifokeev 's suggestion: I'm pretty sure the automatic porting will not work great, but it's worth trying. The process they suggest is a bit different from what we were sketching, but it can be worth trying to follow it instead. What do you think @jimthompson5802 ?

jimthompson5802:
re: the documents pointed out by @ifokeev I understand porting to TF2 is basically a two step process:
Step 1...while in TF1 implement where applicable the compatibility apis `tf.compat.v1`.
Step 2...in TF2 convert the `tf.compat.v1` apis to the native version and recode the apis that were not compatible to the native TF2 api equivalent functions.

In looking at the code base, it looks like Step 1 has already been done. So the focus on the current work is on Step 2.  If this is the case, then I think we're pass the state where the `tf_upgrade_v2` could be useful.

Issue #650, opened 2020-03-09
Output overrided in save_savedmodel


ifokeev:
**Describe the bug**
I was testing new `save_savedmodel` implemented method and was getting response full of probabilities for `category` output feature. I checked the definition and found that outputs are overriding https://github.com/uber/ludwig/blob/4039f93e605ed929e4f4a75c725bb70e148df92b/ludwig/models/model.py#L1419-L1430.

Providing PR of the fix.

PR #651, opened 2020-03-09
fix(model/save_savedmodel): don’t override output

ifokeev:
https://github.com/uber/ludwig/issues/650

### DIFF
--- a/ludwig/models/model.py
+++ b/ludwig/models/model.py
@@ -1416,7 +1416,7 @@ class Model:
             inputs[feature['name']] = getattr(self, feature['name'])
 
         for feature in self.hyperparameters['output_features']:
-            outputs[feature['name']] = getattr(
+            outputs['predictions_' + feature['name']] = getattr(
                 self, 'predictions_' + feature['name']
             )
             if hasattr(self, 'probabilities_' + feature['name']):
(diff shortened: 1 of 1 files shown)


ludwig/models/model.py
```diff
@@ -1416,17 +1416,17 @@ def save_savedmodel(self, save_path):
             inputs[feature['name']] = getattr(self, feature['name'])
 
         for feature in self.hyperparameters['output_features']:
-            outputs[feature['name']] = getattr(
+            outputs['predictions_' + feature['name']] = getattr(
```
ifokeev: I changed the output name because of datasets compatibility. Predictions should have different name by design, to  achieve datasets merging (input + output + predictions) smoothly as possible.

w4nderlust:
Please sign the CLA and I will merge.

ifokeev:
@w4nderlust signed

w4nderlust:
Thank you! I was thinking: are you using saved models from Javascript? It would be great to have an example for that.

ifokeev:
@w4nderlust I'm using `ludwig` with `mlflow` (tracking api and model deployment) and have tests and examples regarding this stack. Really want to contribute this.

I was thinking about: is it cool to make `ludwig` serverless automl? I started to write pipelines using https://github.com/Netflix/metaflow/ and really love it. Within `metaflow` it will be very easy to build AWS batch pipeline processing with distributed store. 
What do you think?

w4nderlust:
Metaflow sounds pretty cool and integrating Ludwig with these tools sounds like a good idea to me. I haven't looked at the in detail enough to be able to differentiate well what Metaflow does differently from MLFlow and Kubeflow for instance, but will look into it. Links and suggestions would be great.

PR #655, opened 2020-03-14
Removed redundant code and modified embedding_modules file

soovam123:
I removed redundant code in embedding_modules file and added them to a new set_embed_para function.

### DIFF
--- a/ludwig/models/modules/embedding_modules.py
+++ b/ludwig/models/modules/embedding_modules.py
@@ -95,9 +95,6 @@ def embedding_matrix_on_device(
         initializer=None,
         regularize=True,
 ):
-    if not regularize:
-        regularizer = None
-
     if embeddings_on_cpu:
         with tf.device('/cpu:0'):
             embeddings, embedding_size = embedding_matrix(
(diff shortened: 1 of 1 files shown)


soovam123:
@w4nderlust Could you please clarify your first and last points? The function is still in `embedding_modules.py`. And for the regularizer part, what are you exactly suggesting?

w4nderlust:
> @w4nderlust Could you please clarify your first and last points? The function is still in `embedding_modules.py`. And for the regularizer part, what are you exactly suggesting?

1. You are right, i thought it was placed in another module, my bad.
2. The 2 lines `if not regularize: regularizer = None` should be placed before the call to `embed_matrix_on_device` not inside. The reason is that I try to avoid if i can functions with side effects like this, so that you then have to return the input.

soovam123:
Aah okay, that is indeed a good precautionary step. Thanks for the insight.

soovam123:
@w4nderlust I am afraid I am not so sure about what you are trying to convey. But, in the code, the function `embedding_matrix_on_device` already has the `regularizer` parameter which is passed only after modifying it with a value depending on the `regularize` parameter in the _if_ statement before the function call.

w4nderlust:
Ah you moved it around from the bottom, that's fine then

PR #664, opened 2020-03-26
Update model.py

JiByungKyu:
# Code Pull Requests

When retraining using mpg dataset

Following error occurred. (it probably occur another dataset)
ValueError: The specificed validation_field combined is not valid.Available ones are: ['mpg', 'combined']

So, I changed string compare operator 'is' to ==, then it works!


w4nderlust:
Thanks!

PR #669, opened 2020-03-31
Refactoring ImageFeature for TF 2.0

ydudin3:
Tested on MNIST dataset with the following model 

input_features:
    -
        name: image_path
        type: image
        preprocessing:
            in_memory: true

output_features:
    -
        name: label
        type: numerical

training:
    dropout_rate: 0.4

Using numerical outputs for now since categories are not implemented yet.

Added ImageTestEncoder which uses FCStack only (Conv stacks are not implemented yet) just to test end-to-end training.

### DIFF
--- a/ludwig/models/modules/image_encoders.py
+++ b/ludwig/models/modules/image_encoders.py
@@ -14,9 +14,42 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import tensorflow as tf
+from tensorflow.keras.layers import Layer
+
 from ludwig.models.modules.convolutional_modules import flatten, ConvStack2D, \
     ResNet, get_resnet_block_sizes
 from ludwig.models.modules.fully_connected_modules import FCStack
+
+# ImageTestEncoder is just for end-to-end testing, will remove this
+class ImageTestEncoder(Layer):
+
+    def __init__(
+            self,
+            **kwargs
+    ):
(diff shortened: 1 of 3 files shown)


ludwig/models/modules/image_encoders.py
```diff
@@ -14,10 +14,43 @@
...
+
+    def call(self, inputs, training=None, mask=None):
+        """
+            :param inputs: The inputs fed into the encoder.
+                   Shape: [batch x height x width x channels], type tf.uint8
+        """
+
+        # use FCStack only for now as it is already implemented and 
+        # ConvStack2D is not
+        fc_stack = FCStack(
```
w4nderlust, line 41: This goes into the init

ydudin3: Right.. updated.

w4nderlust:
looks good so far, next step is to implement the encoders.
You can look at FCStack as an example of an encoder that was adapted to TF2.
Also I'm pretty sure that online you can find several examples of both VGG and ResNet for TF2/Keras. I'm also pretty sure they are also available as example models already in their codebase.

It could also be a good time to look into loading pretrained models if you have time as with TF2 it should be substantially easier.

ydudin3:
Let's merge this to tf_porting and I'll follow up with new PRs? I generally try to discourage commits which modify more than a couple hundred lines of code.

PR #671, opened 2020-04-03
Refactor convolution helper fuction to work with TF 2.0

ydudin3:
Tested locally with 1D and 2D tensors.


w4nderlust:
This looks fine, but at the same time our own conv layers may be not needed anymore as now there are Conv1D and Conv2D directly in TF2: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Conv2D . I did the same thing for FCLayer, I removed our former fc_layer and repalced it with TF2's Dense. we should probably do the same for conv.

ydudin3:
Makes sense, let me look into that.

PR #673, opened 2020-04-04
Refactor Stacked2DCNN to work with tf 2.0

ydudin3:
As title. Some details are missing - will address in future PRs, left a few TODOs.

Tested end-to-end training with MNIST dataset.

### DIFF
--- a/ludwig/models/modules/convolutional_modules.py
+++ b/ludwig/models/modules/convolutional_modules.py
@@ -498,7 +498,7 @@ class ConvStack2D:
             default_stride=2,
             default_pool_stride=None,
             default_norm=None,
-            default_dropout=False,
+            default_dropout=True,
             default_initializer=None,
             default_regularize=True
     ):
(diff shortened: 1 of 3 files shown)


ludwig/models/modules/convolutional_modules.py
```diff
@@ -544,60 +544,44 @@ def __init__(
...
+            # TODO provide regularization penalty
+            # TODO provide bias and activity regularizers
+            # TODO add initializers
 
-                    logger.debug('  pool_{}: {}'.format(
-                        i, layer_output))
-
-            hidden = layer_output
-            logger.debug('  hidden: {0}'.format(hidden))
+            conv_layer = tf.keras.layers.Conv2D(
```
w4nderlust, line 560: this needs to be in the init

ludwig/models/modules/convolutional_modules.py
```diff
@@ -544,60 +544,44 @@ def __init__(
...
+                layer['num_filters'],
+                layer['filter_size'],
+                kernel_regularizer=regularizer if layer['regularize'] else None,
+            )
+            hidden = conv_layer(hidden)
+
+            if layer['dropout'] and dropout_rate > 0:
+                hidden = tf.nn.dropout(hidden, rate=dropout_rate)
```
w4nderlust, line 568: I think there's a tf2 dropout layer too https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dropout
Also consider dropout will soon be unified in one singe number, not rate and boolean as of now., so there will be no if

ludwig/models/modules/convolutional_modules.py
```diff
@@ -544,60 +544,44 @@ def __init__(
...
+            if layer['pool_size'] is not None:
+                pool_size = layer['pool_size']
+                pool_kernel_size = [1, pool_size, pool_size, 1]
+                pool_stride = layer['pool_stride']
+                if pool_stride is not None:
+                    pool_kernel_stride = [1, pool_size, pool_size, 1]
+                else:
+                    pool_kernel_stride = [1, pool_stride, pool_stride, 1]
+                
+                hidden = tf.nn.max_pool(
```
w4nderlust, line 579: https://www.tensorflow.org/api_docs/python/tf/keras/layers/MaxPool2D

ludwig/models/modules/convolutional_modules.py
ydudin3: Updated

ludwig/models/modules/convolutional_modules.py
ydudin3: Yep, updated.

ludwig/models/modules/convolutional_modules.py
ydudin3: Updated.

ludwig/models/modules/convolutional_modules.py
```diff
@@ -530,74 +535,57 @@ def __init__(
...
+                    self.stack.append(
+                        MaxPool2D(
+                            pool_size=layer['pool_size'],
+                            strides=layer['pool_strides']
+                        )
                     )
 
-                    logger.debug('  pool_{}: {}'.format(
-                        i, layer_output))
+    def __call__(
```
w4nderlust, line 578: ConvStack2D should extend Layer and because of that this should be `call()` instead of `__call__()`.
Also regularizer and dropout rate shoudl be removed and a `mask=None` parameter shoudl be added (because `Layer.call()` expects it and because it should be propagated to the internal layers calls too.

ydudin3: Updated.

ludwig/models/modules/convolutional_modules.py
```diff
@@ -530,74 +538,51 @@ def __init__(
...
 
-                    logger.debug('  pool_{}: {}'.format(
-                        i, layer_output))
+    def call(self, inputs, training=None, mask=None):
+        hidden = inputs
 
-            hidden = layer_output
-            logger.debug('  hidden: {0}'.format(hidden))
+        for layer in self.stack:
+            hidden = layer(hidden, training=training)
```
w4nderlust, line 585: Should probably add mask to this call

ludwig/models/modules/convolutional_modules.py
```diff
@@ -530,74 +538,51 @@ def __init__(
...
+                # TODO add initializers
+                self.stack.append(
+                    Conv2D(
+                        layer['num_filters'],
+                        layer['filter_size'],
+                        # kernel_regularizer=regularizer if layer['regularize'] else None,
+                    )
+                )
+
+                self.stack.append(Activation(layer['activation']))
```
w4nderlust, line 568: Should also add norm before activation

ludwig/models/modules/convolutional_modules.py
ydudin3: I'm getting this "TypeError: call() got an unexpected keyword argument 'mask'".
The docs say it should be there but maybe we need to update the version or smth.

ludwig/models/modules/convolutional_modules.py
ydudin3: Added.

ludwig/models/modules/convolutional_modules.py
w4nderlust: mmm it could be the case that some layers use it and some don't... which is weird, what's the purpuse to have an interface if you don't respect it? :D Anyway, don't pass it for now, but maybe add a todo so that we will chack this later

PR #674, opened 2020-04-07
FC user input check

yyunon:
# Code Pull Requests
I have tried to decrease code duplicates in [sequence_encoders.py](https://github.com/uber/ludwig/blob/master/ludwig/models/modules/sequence_encoders.py). StackedCNN, StackedParallelCNN, ParallelCNN classes check fc_layers and num_fc_layers each time. Therefore, a check_fc_layers helper function is written.


w4nderlust:
Thank you but we ar working on porting to TF2 and encoders are changing dramatically. After the porting is done we can revise this and see if it is still needed. Closing it for now. Thank you anyway for trying to help, very appreciated.

PR #676, opened 2020-04-08
Removed code duplications in embedding_modules file

KushalP1:
I have removed duplicates and created a new function Get_Embed_Details.

### DIFF
--- a/ludwig/models/modules/embedding_modules.py
+++ b/ludwig/models/modules/embedding_modules.py
@@ -25,376 +25,369 @@ logger = logging.getLogger(__name__)
 
 
 def embedding_matrix(
-        vocab,
-        embedding_size,
-        representation='dense',
-        embeddings_trainable=True,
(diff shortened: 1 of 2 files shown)


w4nderlust:
@KushalP1 thank you for your effort, is really appreciated.
Although we are currently porting the backend to TF2 ( https://github.com/uber/ludwig/tree/tf2_porting ) and the embedding module will change dramatically, so I will hold on to this a bit and see if it is still ended after the porting is completed.

On a side note, when opening PRs please respect python PEP8 conventions for naming and spacing. Your function contains capital letters in the name and you are using a different spacing from the default, this can be solved by running code cleanup from pycharm for instance.
Moreover, the name of the function is not meaningful and what it does is basically encapsulating an if, not sure it is really needed.

If you are willing to contribute anyway, please reach out to me as there are many tasks that can be very usefult that you could tackle!

PR #677, opened 2020-04-08
Initial Hyperparamter Optimization Functionality

kaushikb11:
##### This PR is for adding Hyperparameter Optimization to Ludwig with a new command `ludwig hyperopt`

It would be divided into the following components:
- [ ] Logic for specifying parameters in the YAML model definition file
- [ ] Logic for using sampled parameters hyperparameters
- [ ] Sampling hyper-parameters
- [ ] Mechanism for performing the hyperparameter optimization outer loop
- [ ] Tests
- [ ] Documentation

Specifying hyper-parameters
---------------------------

Hyper-parameters will be specified inside the Ludwig model definition as a root key.
Each hyper-parameter parameter will have its type, range and additional parameters.

In order to specify hyper-parameters to optimize, one needs to add a `hyperopt` section to the YAML that contains all hyper-parameters to optimize. If the parameters are nested, `.` is used to reference a parameter nested one layer below. For instance, to reference the `learning_rate`, one would have to use the name `training.learning_rate`. If the parameter to reference is nested inside in an input or output feature, the name of that feature will be be used as a starting point, for instance, for referencing the cell_type of the utterance specify `utterance.cell_type`.
here is an example:

```yaml
hyperopt:
  strategy:
    type: grid  # random, bayesian
    # strategy parameters...
  executor:
    type: serial  # parallel, remote, cluster
    # executor parameters...
  parameters:
    utterance.cell_type: ...
    utterance.num_layers: ...
    combiner.num_fc_layers: ...
    section.embedding_size: ...
    preprocessing.text.vocab_size: ...
    training.learning_rate: ...
    training.optimizer.type: ...
```

The `strategy` section contains which strategy to use to perform hyper-parameter optimization. Different types are grid, random and bayesian. The strategy parameters are used to modify the strategy, for instance for random, how many random samples to draw.

The `executor` section specifies how to execute hyperparameter optimization.
The Implementation could happen locally in a serial manner, parallelly or remotely on multiple machines within a cluster.

In the `parameters` section, the values associated with those keys vary depending on the type of the parameter. Types can be `float`, `int` and `category`.

### Strategy

Example of grid strategy:
```yaml
strategy:
  type: grid
```

Example of random strategy:
```yaml
strategy:
  type: random
  num_samples: 20
```

### Executor

Example of serial executor:
```yaml
executor:
  type: serial
```

Example of parallel strategy:
```yaml
executor:
  type: parallel
  workers: 4
```
Ps: Thank you Piero and to the team for building such a great product and for inspiring us. :blush: :rocket:


ludwig/cli.py
```diff
@@ -35,6 +35,8 @@ class CLI(object):
       representation
     - collect_activations - For each datapoint, there exists a corresponding
       tensor representation which are collected through this method
+    - hyperopt - Hyperparameter search for optimal parameters
```
w4nderlust, line 38: replace with:
Performs an hyper-parameter search

ludwig/cli.py
```diff
@@ -93,6 +95,11 @@ def visualize(self):
         ludwig.contrib.contrib_command("visualize", *sys.argv)
         visualize.cli(sys.argv[2:])
 
+    def hyperopt(self):
+        from ludwig import hyperopt
+        ludwig.contrib.contrib_command("visualize", *sys.argv)
```
w4nderlust, line 100: "visualzie" -> "hyperopt"

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+
+class HyperoptStrategy:
+    def __init__(self, strategy: Dict[str, Any], parameters: Dict[str, Any]) -> None:
+        self.strategy = strategy
+        self.parameters = parameters
+        self.build_hyperopt_strategy = get_build_hyperopt_strategy(
+            strategy["type"]
+        )(**strategy)
+
+    def sample(self) -> Dict[str, Any]:
```
w4nderlust, line 48: make this an abstract method

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+        # TODO: Yields a set of parameters names and their values.
+        # Define `build_hyperopt_strategy` which would take paramters as inputs
+        yield {}
+
+    def update(self, sampled_parameters: Dict[str, Any], statistics: Dict[str, Any]):
```
w4nderlust, line 53: make this an abstract method

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+
+class HyperoptExecutor:
+    def __init__(self, hyperopt_strategy: HyperoptStrategy, executor: Dict[str, Any]) -> None:
+        self.hyperopt_strategy = hyperopt_strategy
+        self.executor = executor
+        self.build_hyperopt_executor = get_build_hyperopt_executor(
+            executor["type"]
+        )(**executor)
+
+    def execute(
```
w4nderlust, line 68: make this method abstract. Create a class that extends `HyperoptExecutor` called `SerialExecutor` and paste there the current inner implementation of the `ecevute` method.

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def update(self, sampled_parameters: Dict[str, Any], statistics: Dict[str, Any]):
+        # TODO: Given the results of previous computation, it updates
+        # the strategy (not needed for stateless strategies like "grid"
+        # and random, but will be needed by bayesian)
+        pass
+
+
+class HyperoptExecutor:
+    def __init__(self, hyperopt_strategy: HyperoptStrategy, executor: Dict[str, Any]) -> None:
```
w4nderlust, line 61: the `executor: Dict[str, Any]` is not needed. Different objects implementing this abstract class will have different parameters.

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+from ludwig.utils.defaults import default_random_seed, merge_with_defaults
+from ludwig.utils.misc import set_default_value, set_default_values
+from ludwig.utils.print_utils import logging_level_registry, print_ludwig
+
+
+logger = logging.getLogger(__name__)
+
+
+class HyperoptStrategy:
+    def __init__(self, strategy: Dict[str, Any], parameters: Dict[str, Any]) -> None:
```
w4nderlust, line 41: the `strategy: Dict[str, Any]` is not needed. Different objects implementing this abstract class will have different strategy parameters.

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+        )
+
+    # merge with default model definition to set defaults
+    if model_definition_file is not None:
+        with open(model_definition_file, "r") as def_file:
+            model_definition = merge_with_defaults(yaml.safe_load(def_file))
+    else:
+        model_definition = merge_with_defaults(model_definition)
+
+    if HYPEROPT not in model_definition:
```
w4nderlust, line 174: Move this check at the beginning of the function

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+        raise ValueError("Hyperopt Section not present in Model Definition")
+
+    hyperopt_params = model_definition["hyperopt"]
+    update_hyperopt_params_with_defaults(hyperopt_params)
+
+    strategy = hyperopt_params["strategy"]
+    executor = hyperopt_params["executor"]
+    parameters = hyperopt_params["parameters"]
+
+    hyperopt_strategy = HyperoptStrategy(strategy, parameters)
```
w4nderlust, line 184: for both `HyperoptStrategy` and `HyperoptExecutor` here we need to get the class from a registry of name -> class (look at `sequence_encoder_registry` in `features/sequence_feature.py` for instance, use the `get_sequence_encoder` function in the same module as an example of how to use the `get_from_registry` function). After we get the class based on the name, we can initialize it with the parameters in the strategy / executor dictionaries.

ludwig/hyperopt.py
```diff
@@ -0,0 +1,512 @@
...
+
+    set_default_values(
+        hyperopt_params[STRATEGY],
+        {
+            "type": "random",
+            "num_samples": 12
+        }
+    )
+
+    if hyperopt_params[STRATEGY]["type"] == "grid":
```
w4nderlust, line 240: fr now this is ok, later we can change it so that each HyperoptStrategy contains its default parameters and this function reads the defaults from there.

ludwig/models/hyperopt.py
```diff
@@ -0,0 +1,76 @@
...
+# limitations under the License.
+# ==============================================================================
+import logging
+from ludwig.utils.misc import get_from_registry
+
+
+logger = logging.getLogger(__name__)
+
+
+class RandomStrategy:
```
w4nderlust, line 24: Some of the previous comments were about these objects, i did not see them before doing the comments, sorry. So, instead of "reating" thsoe objects, they should be modified according to the previous indications, so they should extend the base strategy and executor class.

w4nderlust:
Thank you so much, this is a great start!

Couple things @kaushikb11 :
- please login from a github account in order to sign the CLA otherwise I can't use your code.
- thanks for pasting the design document I gave you. In the meantime I improved it a little bit, so I will modify your post to reflect that
- I'm adding some comments, did not see the last file `models/hyperopt.py` when I was writing some  of the comments, so that file alrady anwers some of my questions there.

w4nderlust:
Some python code to clarify my suggestions:
```python
class HyperoptStrategy:
    def __init__(self, parameters: Dict[str, Any]) -> None:
        self.parameters = parameters

    @abstractmethod
    def sample(self) -> Dict[str, Any]:
        pass

    @abstractmethod
    def update(self, sampled_parameters: Dict[str, Any], statistics: Dict[str, Any]):
        pass


class RandomStrategy(HyperoptStrategy):
    def __init__(self, parameters: Dict[str, Any], num_samples=10) -> None:
        HyperoptStrategy.__init__(self, parameters)
        self.num-samples = num_samples

    def sample(self) -> Dict[str, Any]:
        # actual implementation ...
        pass

    def update(self, sampled_parameters: Dict[str, Any], statistics: Dict[str, Any]):
        # actual implementation ...
        pass

class HyperoptExecutor:
    def __init__(self, hyperopt_strategy: HyperoptStrategy) -> None:
        self.hyperopt_strategy = hyperopt_strategy

    @abstractmethod
    def execute(...):
        pass

class SerialExecutor(HyperoptExecutor):
    def __init__(self, hyperopt_strategy: HyperoptStrategy, ... other parameters ...) -> None:
        HyperoptStrategy.__init__(self, hyperopt_strategy)
        self.other_param = other_param

    def execute(...):
        # actual implementation ...
        pass


# in the hyperopt function:

hyperopt_strategy = get_build_hyperopt_strategy(strategy["type"])(parameters, **strategy)
hyperopt_executor = get_build_hyperopt_executor(executor["type"])(hyperopt_strategy, **executor)
results = hyperopt_executor.execute(...)
```

ludwig/models/hyperopt.py
kaushikb11: Makes sense. Noted! @w4nderlust

ludwig/hyperopt.py
kaushikb11: Should I move these Base Classes to `models/hyperopt.py` as well? @w4nderlust

w4nderlust: I think for now you can put everything, including registries, in `ludwig/hyperopt.py`, later we can decide if / how to split the code in modules that make sense, if needed.

w4nderlust:
One additional consideration: each hyperopt strategy can work for minimizing or maximizing a measure or a loss of any of the output features (as Ludwig supports more than one) on any of the splits, so we want the user to be able to specify that.

So let's add new parameters:
```
hyperopt:
  output_feature: combined  # this is the default, but can be the name of any output feature
  measure: loss  # but can be anything returned by the training
  split: validation  # can also be training or test
  goal: minimize  # or maximize
```

This information will allow few things things:
- extract the right value from the training results
- sort the results according to such value
- tell the bayesian optimization what to do

So i would update the example code to the following:
```python
class HyperoptStrategy:
    def __init__(self, goal:str, parameters: Dict[str, Any]) -> None:
        self.goal = goal  # useful for bayesian stratiegy
        self.parameters = parameters

    @abstractmethod
    def sample(self) -> Dict[str, Any]:
        pass

    @abstractmethod
    def update(self, sampled_parameters: Dict[str, Any], statistics: Dict[str, Any]):
        pass


class RandomStrategy(HyperoptStrategy):
    def __init__(self, parameters: Dict[str, Any], num_samples=10) -> None:
        HyperoptStrategy.__init__(self, parameters)
        self.num_samples = num_samples

    def sample(self) -> Dict[str, Any]:
        # actual implementation ...
        pass

    def update(self, sampled_parameters: Dict[str, Any], statistics: Dict[str, Any]):
        # actual implementation ...
        pass

class HyperoptExecutor:
    def __init__(self, hyperopt_strategy: HyperoptStrategy, output_feature: str, measure: str, split: str) -> None:
        self.hyperopt_strategy = hyperopt_strategy
        self.output_feature = output_feature
        self.measure = measure
        self.split = split

    def get_measure(training_results):
        return results[self.split][self.output_feature][self.measure]

    def sort_hyperopt_results(hyperopt_results):
        # sorts based on self.strategy.goal and get_measure
        return sorted hyperopt_results

    @abstractmethod
    def execute(...):
        pass

class SerialExecutor(HyperoptExecutor):
    def __init__(self, hyperopt_strategy: HyperoptStrategy, ... other parameters ...) -> None:
        HyperoptStrategy.__init__(self, hyperopt_strategy)
        self.other_param = other_param

    def execute(...):
        # actual implementation ...
        # uses self.get_measure(training_results) to get the measure to optimize
        # at the end sorts results using self.sort_hyperopt_results
        pass


# in the hyperopt function:

hyperopt_strategy = get_build_hyperopt_strategy(strategy["type"])(parameters, **strategy)
hyperopt_executor = get_build_hyperopt_executor(executor["type"])(hyperopt_strategy, output_feature, measure, split, goal, **executor)
results = hyperopt_executor.execute(...)
```

w4nderlust:
I did a pass in which I integrated my suggested modification in the code, please pull a continue working from there ;)

kaushikb11:
@w4nderlust Haha Thanks! Will start with the required fixes and implementing the Strategies.

ludwig/utils/hyperopt_utils.py
```diff
@@ -0,0 +1,333 @@
...
+        self.goal = goal  # useful for bayesian stratiegy
+        self.parameters = parameters
+
+    @abstractmethod
+    def sample(self) -> Dict[str, Any]:
+        # Yields a set of parameters names and their values.
+        # Define `build_hyperopt_strategy` which would take paramters as inputs
+        pass
+
+    @abstractmethod
```
kaushikb11, line 39: Hi @w4nderlust, Would it be necessary for the `update` method to be an `abstract` method? As stateless strategies like Grid and Random won't need it.

w4nderlust: As there will be statefull strategies, let's keep it, so that the executor can always call it *otherwise there will need to be a check in the excutor like `if hasattr(strategy, update)` which is not super nice. Strategies that make no use of it will just not implement it / implement it with `pass`.

w4nderlust:
Merging this into a new hyperopt branch, so that other people can work on it too. To keep refining it, just create a new pr from the hyperopt branch

PR #684, opened 2020-04-11
Fix documentation, wrong parameter name

martinremy:
The parameter name `field` in the docs didn't match the [actual parameter name](https://github.com/uber/ludwig/blob/ec43b84745067bc42b4f0e61f4a168f7bb0381e8/ludwig/visualize.py#L2513) for the viz `roc_curves_from_test_statistics`.


w4nderlust:
Thank you for catching this

PR #689, opened 2020-04-15
Add Local Parallel Executor to Hyperopt with GPU Support

kaushikb11:
### Executor

Example of parallel strategy:
```yaml
executor:
  type: parallel
  num_workers: 4
```


w4nderlust:
This looks good @kaushikb11 ! Thank you! i may do some little cleanup on top of it, but it looks fine.
Now I guess there are only few things missing:
1. the management of gpus / gpu_fraction
2. writing some tests
3. writing some docs

I guess 1. should be part of this PR so that it is self contained, while 2. and 3. can be their own PRs.
How does that sound?

kaushikb11:
Sounds good! Will start looking into the GPUs part and will probably implement it by tomorrow.

kaushikb11:
I was also thinking about adding logic for Changing `device_type` to `CPU` if all the GPUs are occupied. How does it sound?

w4nderlust:
I'm not 100% sure you would be able to actually figure out if the GPUs are used or not, as TF bgrabs all it can anyway.

Regarding the logi what I was thinking it that `gpus` contains the list of GPUs specified by the user. If there are more workers than GPUs, then you can modify the gpu_fraction accordingly.

Examples:
- user does no specify GPUs -> do nothing
- user specifies one gpu and 4 workers, does not specify fraction (or fraction is 1) -> you set gpu fraction to be 0.25 (or 0.25 - epsilon) and send a warning to the user you are doing it
- user specifies 2 gpus and two workers -> you assign one gpu per worker
- user specifies 2 gpu and 4 workers -> you assign 1 gpu to two workers and another gpu to other two workers and set the gpu fraction to 0.2 and warn the user if their specified fraction is > 0.5, if they specified less than 0.5 than that's fine
- user specifies more gpus than workers -> this is easy :)

Does this make sense to you?

kaushikb11:
Yeah, thank you! It's just that I worked on a framework recently which did this.
Ref: https://github.com/videoflow/videoflow/blob/743af6905e/videoflow/engines/batch.py#L208

But obviously `gpu_fraction` part couldn't be figured out.

ludwig/utils/hyperopt_utils.py
```diff
@@ -472,6 +472,21 @@ def execute(
...
+
+        if gpus is not None:
+            if isinstance(gpus, int):
+                gpus = str(gpus)
+            gpus = gpus.strip()
+            total_gpus = len(gpus.split(','))
+
+            if total_gpus < self.num_workers:
+                fraction = total_gpus / self.num_workers
+                if fraction < 1:
```
w4nderlust, line 484: isnt this always true because fo line 482?

w4nderlust:
A comment on the last commit: it looks fine, it fixes the fraction. I guess the thing to figure out now is how to change the gpus parameter for each task so that it is set to the available gpu right now, which i guess is slightly more complicated and may require you to change the pre-generation of all tasks to something that looks like:
1. I generate all the tasks in a list of tasks
2. I keep track of the worker -> gpu assignment in a tracker dictionary initialize it to empty
3. I set the `gpus` and `gpu_fraction` depending on what's available in the tracker value for the first `num_workers - len(tracker_dict)` tasks in the task list
4. remove those tasks from the task list and run them
5. when a task completes, I know which gpu was freed from my tracking dictionary, remove the task it from tracking
6. goto 3.

If you have a better solution in mind, that's great too, this is just a suggestion of a possible implementation.

kaushikb11:
@w4nderlust I decided to go with the Queue strategy to manage the available GPUs and defined an `epsilon` to remove from `gpu_fraction` as a safeguard. I was wondering what would be the right way for the user to change the value if needed.

w4nderlust:
I would add it as a optional executor parameter.

w4nderlust:
lgtm! Have you tested it in a multiple gpu environment by any chance? Ig ti works we can merge and move forward with tests and documentation

kaushikb11:
Yes, I have tested it with multiple `gpus` on Sagemaker. Also, I think we could add more condition when our calculated `fraction` is greater than 0.5 and `fraction` < `gpu_fraction`, we could go with the `gpu_fraction` value because other processes won't be able to use that GPU anyway.

w4nderlust:
> Yes, I have tested it with multiple `gpus` on Sagemaker.

Sounds good. It would be great if you could write little instructions for me to run the same if it's not too complicated.

> Also, I think we could add more condition when our calculated `fraction` is greater than 0.5 and `fraction` < `gpu_fraction`, we could go with the `gpu_fraction` value because other processes won't be able to use that GPU anyway.

Makes sense.For clarity I'm calling fraction as `calc_fraction`. The logic should be:
- if `calc_fraction > 0.5` we set `calc_fraction = 1` and we warn the user about it if they specified `gpu_fraction != 1`, also explaining why
- if `calc_fraction < gpu_fraction` we use `calc_fraction` and we warn the user about the change (explaining why
- if `gpu_fraction < calc_fraction` we use `gpu_fraction` as we don't know the motivations for the user to specify a smaller fraction, and also because those fractions are elastic anyway, but we warn the user about the fact that given the number of gpus available and the number of workers selected, they can increase the `gpu_fraction` up to `calc_fraction`

kaushikb11:
Also suppose User has a Machine with 2 cores. And has set `num_workers` = 4 as an optional param to Executor, if there's one GPU, the `gpu_fraction` would be 0.25 excluding epsilon. We wouldn't be using the GPU at full utilization.
Is this something we should look into as well?

kaushikb11:
> Sounds good. It would be great if you could write little instructions for me to run the same if it's not too complicated.

I set `gpus` to the syntax of CUDA_VISIBLE_DEVICES, and watch the GPU processes and memory usage using `nvidia-smi` and verify the multiple python processes get created and the Execution time. I experimented with multiple sets of `gpus` and `num_workers` with 7 gpus. I have come across an Error whenever `gpu_fraction` is greater than 0.5:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Invalid device ordinal value (2). Valid range is [0, 1].
	while setting up XLA_GPU_JIT device number 2
```
I haven't written any kind of script for testing. Let me know if anything else is required.

> Makes sense.For clarity I'm calling fraction as calc_fraction. The logic should be:
> .....

Noted! That's how I was thinking about it. Will add the required warnings with an explanation. Thanks

---------------------------------------------------------------------------
When `gpus`="0,1" & `num_workers`=6, 3 processes per gpu is created:

gpu_proc

w4nderlust:
> Also suppose User has a Machine with 2 cores. And has set `num_workers` = 4 as an optional param to Executor, if there's one GPU, the `gpu_fraction` would be 0.25 excluding epsilon. We wouldn't be using the GPU at full utilization.
> Is this something we should look into as well?

I see the problem. We can do the following:
- get the number of cpus with `os.cpu_count()` or `multiprocess_cpu_count()` (I-m not sure what's the difference to be honest, maybe check into it) . If the `num_workers > num_available_cpus` print a warning to the user that setting the `num_workers < num_available_cpus` is suggested.

w4nderlust:
> I set `gpus` to the syntax of CUDA_VISIBLE_DEVICES, and watch the GPU processes and memory usage using `nvidia-smi` and verify the multiple python processes get created and the Execution time. 

I actually meant how can i est by setting up an instance on SageMaker, but that's fine, I can figure out on my own I guess :)

> I experimented with multiple sets of `gpus` and `num_workers` with 7 gpus. I have come across an Error whenever `gpu_fraction` is greater than 0.5:
> 
> ```
> tensorflow.python.framework.errors_impl.InvalidArgumentError: Invalid device ordinal value (2). Valid range is [0, 1].
> 	while setting up XLA_GPU_JIT device number 2
> ```

This is weird. We should check if that is always true independently of hyperopt or if it's a hyperopt related thing.

> I haven't written any kind of script for testing. Let me know if anything else is required.

No problem, we can do another PR for test and another one for documentation.
 
> When `gpus`="0,1" & `num_workers`=6, 3 processes per gpu is created:
> 
> gpu_proc

Pretty neat :)

kaushikb11:
> I see the problem. We can do the following:
> 
> * get the number of cpus with `os.cpu_count()` or `multiprocess_cpu_count()` (I-m not sure what's the difference to be honest, maybe check into it) . If the `num_workers > num_available_cpus` print a warning to the user that setting the `num_workers < num_available_cpus` is suggested.

`multiprocessing.cpu_count` is a wrapper around `os.cpu_count` with added Exception Handling. But those functionalities return (physical cores + logical cores). We need only physical cores for `Multiprocessing` as it couldn't take advantage of the logical cores. `psutil` helps to just get the num of physical cores.

```
import psutil 
psutil.cpu_count(logical=False)
```

w4nderlust:
Got it. Not sure why you say we couldn't take advantage of logical cores. Whay is that the case?

kaushikb11:
I have an understanding that the logical cores don't have their memory access and it relies on the memory & cache of its host physical processor. As OS allocates CPU-consuming processes on different physical cores and the heavy tasks we have at hand, logical cores won't be much helpful.

If we use more cores than it's available, there would be high-performance drops as it would be dependent on the hardware & OS scheduler on how they handle the tasks, which could lead to chaotic and unpredictable behavior.

w4nderlust:
Goy t it. Makes sense. Reading the dpcs here https://psutil.readthedocs.io/en/latest/#psutil.cpu_count it looks like that in some OSs `psutils.cpu_count(logical=False)` can return `None`, if that happens I would call the function again without the logical parameters and use the number of logical cores warning the user that we couldn't get the number of physical cores from their os and that we are using the number of logical instead, but to avoid processing slowdowns we suggest to set the number of workers to be no greater than the number of physical cores.
Does that sound fine?

kaushikb11:
@w4nderlust Yes, that makes sense. Did the required changes. Thanks!

w4nderlust:
Perfect, let me merge this then! Now let's move to tests and documentation. Great job :)

PR #695, opened 2020-04-22
Added Fiber executor to allow large scale parallel hyperparameter opt…

calio:
…imization

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests


ludwig/utils/hyperopt_utils.py
```diff
@@ -23,6 +24,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List
 
+import fiber
```
w4nderlust, line 27: Can yu move this import in the init of the FiberExecutor? This way if one does not need to use the fiber executor there will be no import error. I believe you refer to fiber directly only in the init to create the pool

w4nderlust:
This looks great @calio thank you so much.
Could you please provide some instructions for me to test it both locally and remotely?

ludwig/utils/hyperopt_utils.py
calio: sure, will do.

w4nderlust:
This works like a charm :)

PR #696, opened 2020-04-22
ResNet TF2 porting

ydudin3:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/models/modules/convolutional_modules.py
+++ b/ludwig/models/modules/convolutional_modules.py
@@ -16,6 +16,7 @@
 import logging
 
 import tensorflow.compat.v1 as tf
+from tensorflow.keras.initializers import VarianceScaling
 from tensorflow.keras.layers import Activation
 from tensorflow.keras.layers import AveragePooling1D
 from tensorflow.keras.layers import AveragePooling2D
@@ -27,6 +28,9 @@ from tensorflow.keras.layers import Layer
 from tensorflow.keras.layers import LayerNormalization
 from tensorflow.keras.layers import MaxPool1D
 from tensorflow.keras.layers import MaxPool2D
+from tensorflow.keras.layers import ZeroPadding2D
+
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


ludwig/models/modules/image_encoders.py
```diff
@@ -187,36 +189,21 @@ def __init__(
...
         # ================ Conv Layers ================
-        hidden = self.resnet(
-            input_image,
-            regularizer,
-            dropout,
-            is_training=is_training
-        )
-        hidden = self.flatten(hidden)
+        hidden = self.resnet(inputs, training)
+        hidden = tf.reshape(hidden, [hidden.shape[0], -1])
```
w4nderlust, line 204: there's a tf.keras.layers.Flatten if you prefer to use that, but it does the same thing really

ludwig/models/modules/convolutional_modules.py
```diff
@@ -722,6 +723,152 @@ def call(self, inputs, training=None, mask=None):
...
+        weights_regularizer=None,
+        batch_norm_momentum=0.9,
+        batch_norm_epsilon=0.001
+    ):
+        super(ResNet2, self).__init__()
+        self.resnet_size = resnet_size
+
+        self.is_bottleneck = is_bottleneck
+        if is_bottleneck:
+            self.block_fn = ResNetBottleneckBlock
```
w4nderlust, line 801: it seems that \block_fn\ is used only in this function so maybe we don't neeed to assign it to self. Also, as it's not a function anymore but a class, maybe let's cal it \block_class\?

w4nderlust:
Thanks, this looks good! Now just make sure all dimensions and all inputs and outputs of every step is the right shape by comparing 1-1 lines of code with the previous implementantation. If everything is correct, we can move forward and merge.

ludwig/models/modules/convolutional_modules.py
```diff
@@ -519,6 +524,349 @@ def call(self, inputs, training=None, mask=None):
...
+            filter_size=3,
+            strides=1,
+            weights_regularizer=None
+    ):
+        super(Conv2DLayerFixedPadding, self).__init__()
+
+        self.layers = []
+
+        if strides > 1:
+            self.layers.append(ZeroPadding2D(padding=((filter_size - 1) // 2)))
```
ydudin3, line 541: @w4nderlust have a look at this one, does this padding size make sense compared to the previous implementation?

PR #697, opened 2020-04-23
TF2 porting: tagger decoder cleanup

jimthompson5802:
# Code Pull Requests

This should finish off the Tagger decoder
* todo for feedforward attention
* add time series support
* add regularizer specification

### DIFF
--- a/ludwig/models/modules/sequence_decoders.py
+++ b/ludwig/models/modules/sequence_decoders.py
@@ -148,33 +148,29 @@ class SequenceGeneratorDecoder(Layer):
 class SequenceTaggerDecoder(Layer):
     def __init__(
             self,
-            initializer=None,
+            num_classes,
             use_bias=True,
-            kernel_initializer='glorot_uniform',
+            weights_initializer='glorot_uniform',
             bias_initializer='zeros',
-            kernel_regularizer=None,
+            weights_regularizer=None,
             bias_regularizer=None,
             activity_regularizer=None,
             attention=False,
-            num_classes=0,
-            is_timeseries=None,
+            is_timeseries=False,
             **kwargs
     ):
         super(SequenceTaggerDecoder, self).__init__()
-        self.initializer = initializer
         self.attention = attention
 
         if is_timeseries:
-            units = 1
-        else:
-            units = num_classes
+            num_classes = 1
 
         self.decoder_layer = Dense(
-            units,
+            units=num_classes,
             use_bias=use_bias,
-            kernel_initializer=kernel_initializer,
+            kernel_initializer=weights_initializer,
             bias_initializer=bias_initializer,
-            kernel_regularizer=kernel_regularizer,
+            kernel_regularizer=weights_regularizer,
             bias_regularizer=bias_regularizer,
             activity_regularizer=activity_regularizer
         )
@@ -185,7 +181,6 @@ class SequenceTaggerDecoder(Layer):
             training=None,
             mask=None
     ):
-        logger.debug('  hidden shape: {0}'.format(inputs.shape))
         if len(inputs.shape) != 3:
             raise ValueError(
                 'Decoder inputs rank is {}, but should be 3 [batch x sequence x hidden] '
--- a/ludwig/models/modules/sequence_decoders.py
+++ b/ludwig/models/modules/sequence_decoders.py
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
     def __init__(
             self,
             initializer=None,
-            regularize=True,
+            use_bias=True,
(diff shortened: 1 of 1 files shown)


ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
     def __init__(
             self,
             initializer=None,
```
w4nderlust, line 151: this doesn't seem to be used, can remove

ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
...
-            regularize=True,
+            use_bias=True,
+            kernel_initializer='glorot_uniform',
```
w4nderlust, line 153: although Dense uses ths name, throughout the rest of the code we are always calling this weights_initializer. The same  applies to kernel_regularizer too.

ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
...
+            bias_initializer='zeros',
+            kernel_regularizer=None,
+            bias_regularizer=None,
+            activity_regularizer=None,
             attention=False,
             num_classes=0,
```
w4nderlust, line 159: this should probably be the first parameter and it should not be optional, as 0 num classes does not make sense.

ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
...
-            loss=None,
+            is_timeseries=None,
```
w4nderlust, line 160: this should be a bool, so default is False

ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
...
     ):
         super(SequenceTaggerDecoder, self).__init__()
         self.initializer = initializer
-        self.regularize = regularize
         self.attention = attention
 
-        # todo tf2 add other required parameters for Dense layer
-        self.decoder_layer = Dense(num_classes)
+        if is_timeseries:
+            units = 1
```
w4nderlust, line 168: would simplify to `num_classes = 1` and remove the else

ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
...
+        else:
+            units = num_classes
+
+        self.decoder_layer = Dense(
+            units,
```
w4nderlust, line 173: would change to `units=num_classes`

ludwig/models/modules/sequence_decoders.py
```diff
@@ -149,104 +149,56 @@ class SequenceTaggerDecoder(Layer):
...
-            # hidden_size,
-            # regularizer,
-            # is_timeseries=False
+            inputs,
+            training=None,
+            mask=None
     ):
-        logger.debug('  hidden shape: {0}'.format(hidden.shape))
-        if len(hidden.shape) != 3:
+        logger.debug('  hidden shape: {0}'.format(inputs.shape))
```
w4nderlust, line 188: I'm removing all this loggig calls from call() everywhere for now

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

ludwig/models/modules/sequence_decoders.py
jimthompson5802: done

PR #700, opened 2020-04-25
Add tests for Hyperopt Strategies & Executors

kaushikb11:
#### Add tests for Hyperopt Strategies & Executors

### DIFF
--- a/tests/ludwig/utils/test_hyperopt_utils.py
+++ b/tests/ludwig/utils/test_hyperopt_utils.py
@@ -39,9 +39,9 @@ HYPEROPT_PARAMS = {
             }
         },
         "expected_search_space": {
-            'training.learning_rate': [0.0001, 0.001, 0.01, 0.1],
-            'combiner.num_fc_layers': [0, 1, 2, 3, 4],
-            'utterance.cell_type': ['rnn', 'gru', 'lstm']
+            "training.learning_rate": [0.0001, 0.001, 0.01, 0.1],
+            "combiner.num_fc_layers": [0, 1, 2, 3, 4],
+            "utterance.cell_type": ["rnn", "gru", "lstm"]
         },
         "goal": "minimize",
         "expected_len_grids": 60,
@@ -74,7 +74,7 @@ HYPEROPT_PARAMS = {
 }
 
 
-@pytest.mark.parametrize('key', ['test_1', "test_2"])
+@pytest.mark.parametrize("key", ["test_1", "test_2"])
 def test_grid_strategy(key):
 
     hyperopt_test_params = HYPEROPT_PARAMS[key]
@@ -103,7 +103,7 @@ def test_grid_strategy(key):
         grid_strategy.samples) == hyperopt_test_params["expected_len_grids"]
 
 
-@pytest.mark.parametrize('key', ['test_1', "test_2"])
+@pytest.mark.parametrize("key", ["test_1", "test_2"])
 def test_random_strategy(key):
 
     hyperopt_test_params = HYPEROPT_PARAMS[key]
--- a/tests/ludwig/utils/test_hyperopt_utils.py
+++ b/tests/ludwig/utils/test_hyperopt_utils.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2019 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import pytest
+import numpy as np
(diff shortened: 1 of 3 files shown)


tests/ludwig/utils/test_hyperopt_utils.py
```diff
@@ -0,0 +1,117 @@
...
+                "high": 4
+            },
+            "utterance.cell_type": {
+                "type": "category",
+                "values": ["rnn", "gru", "lstm"]
+            }
+        },
+        "expected_search_space": {
+            'training.learning_rate': [0.0001, 0.001, 0.01, 0.1],
+            'combiner.num_fc_layers': [0, 4],
```
w4nderlust, line 43: this should probably be [0, 1, 2, 3, 4]

tests/ludwig/utils/test_hyperopt_utils.py
```diff
@@ -0,0 +1,117 @@
...
+    hyperopt_test_params = HYPEROPT_PARAMS[key]
+    goal = hyperopt_test_params["goal"]
+    grid_strategy_params = hyperopt_test_params["params"]
+
+    grid_strategy = GridStrategy(goal=goal, parameters=grid_strategy_params)
+
+    search_space_dict = {}
+
+    for key, values in grid_strategy.search_space.items():
+        if isinstance(values, np.ndarray):
```
w4nderlust, line 89: not sure why do you have to do this

tests/ludwig/utils/test_hyperopt_utils.py
```diff
@@ -0,0 +1,117 @@
...
+    num_samples = hyperopt_test_params["num_samples"]
+
+    random_strategy = RandomStrategy(
+        goal=goal, parameters=random_strategy_params, num_samples=num_samples)
+
+    actual_params_keys = random_strategy.sample().keys()
+    expected_params_keys = hyperopt_test_params["params"].keys()
+
+    assert actual_params_keys == expected_params_keys
+    assert len(random_strategy.samples) == num_samples
```
w4nderlust, line 117: Here you can also do a soft check of the samples: for integers and float you can check that they are withing the low and high and for categorical you can check if it is within the list (can transforms it into a set for efficiency)

tests/ludwig/utils/test_hyperopt_utils.py
kaushikb11: Using `np.linspace(start=low,stop=hight,num=steps)` Steps are 2 by default. If we go for using `arange` instead, User won't be sure how much samples could out of a particular param?

tests/ludwig/utils/test_hyperopt_utils.py
kaushikb11: Some sampling functions use numpy functionalities to calculate data. Hence, as they are of `np.array` type, we would need to convert them to `list` for assertion of lists.

tests/ludwig/utils/test_hyperopt_utils.py
kaushikb11: Got it!

tests/ludwig/utils/test_hyperopt_utils.py
w4nderlust: we can then add a check like:
```
if steps == None:
  steps = high - low + 1
```

tests/ludwig/utils/test_hyperopt_utils.py
w4nderlust: I see. This raises a question: do we want to do this conversion in the Strategy before returning the values instead. so that all values are python lists?

kaushikb11: Yup, that sounds like a better approach and also, to have consistency across sampled parameters.

tests/ludwig/utils/test_hyperopt_utils.py
kaushikb11: But for Grid Strategy, we could say this is already happening while comparing the search spaces, right? Samples are bounded by their search space.

w4nderlust: Yes you are right. Although Iguess once you implement it, adding it to the other strategy too would be straightforward.

w4nderlust:
With the last changed, I believe the tests for strategies are complete!

kaushikb11:
@w4nderlust Yes, will start with Integration tests for Executors.

w4nderlust:
Checked the last commit. It looks great. Have you tested that everything is ok and that all tests pass? In that case, I will run the tests myself, and then I believe I can merge this.

kaushikb11:
@w4nderlust Yes, I have tested it locally and tests seemed to have passed in the Travis CI as well.
But when I set `combiner.num_fc_layers` to 0, `IndexError: list index out of range` Error is raised. Should a check be added at https://github.com/uber/ludwig/blob/master/ludwig/models/combiners.py#L53 for `num_fc_layers` is 0?

w4nderlust:
Yeah there should be morechecks, but we are introducing them in TF2 porting (at least some of them). So don't worry, just change the minimum to 1 instead 0 in the test.

kaushikb11:
Noted! Did the required changes. Thanks!

w4nderlust:
Great! merging this. Now let's move to the documentation. I think the document that I sent you can be a good starting point, nu we changed a few things in the mean time so it needs to be adapted.

Ideally there are 2 pieces to the documentation, one that goes into the user guide that explains how to use it, and one that goes into the developers guide, that eplains the design and the Strategy and executor abstract classes and explains how to implement a new strategy or a new executor.

PR #706, opened 2020-05-07
Use the standard library for CPUs count instead of psutil

kaushikb11:
After running a test of training ten instances of models with Pool using 4 cores (`mp.cpu_count`) and 2 cores (`psutil.cpu_count(logical=False)`). There wasn't much difference in the task completion time for both the approaches. The CPU utilization among the cores was the difference, as the logical cores spend time swapping as well. There was no apparent degradation in performance by using logical cores. Hence, we decided to go with the standard library than using `psutil`.

### DIFF
--- a/ludwig/utils/hyperopt_utils.py
+++ b/ludwig/utils/hyperopt_utils.py
@@ -27,7 +27,6 @@ from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Tuple, Iterable
 
 import numpy as np
-import psutil
 
 from ludwig.constants import EXECUTOR, STRATEGY, MINIMIZE, COMBINED, LOSS, VALIDATION, MAXIMIZE, TRAINING, TEST
 from ludwig.data.postprocessing import postprocess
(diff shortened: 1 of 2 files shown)


ludwig/utils/hyperopt_utils.py
```diff
@@ -517,14 +516,7 @@ def execute(
...
 
-            num_available_cpus = psutil.cpu_count(logical=False)
-
-            if num_available_cpus is None:
-                num_available_cpus = psutil.cpu_count()
-                logger.warning(
-                    'WARNING: Couldn\'t get the number of physical cores '
-                    'from the OS, using the logical cores instead.'
-                )
+            num_available_cpus = multiprocessing.cpu_count()
```
kaushikb11, line 519:
```
def cpu_count(self):
    '''Returns the number of CPUs in the system'''
    num = os.cpu_count()
    if num is None:
        raise NotImplementedError('cannot determine number of cpus')
    else:
        return num
```

PR #715, opened 2020-05-31
tensorflow-addons fixed version range

carlogrisetti:
tensorflow-addons==0.10.0 is only compatible with tensorflow==2.2.0
Setting lower and higher limit as per https://github.com/tensorflow/addons

(pytest was failing hard looking for random number generator introduced in 2.2.0)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,6 @@ tabulate>=0.7
 scikit-learn
 tqdm
 tensorflow==2.1
-tensorflow-addons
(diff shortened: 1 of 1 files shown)


w4nderlust:
let's bump tensorflow to 2.2.0 instead

carlogrisetti:
As you prefer, I was doing this on the assumption to leave things as-is as much as possible.
Will try and bump it up tomorrow. In any case 0.10.0 supports 2.2.0 only (explicitly lower than 2.3.0) so a similar constraint remains

PR #699, opened 2020-04-25
TF2 porting: Sequence generator decoder

jimthompson5802:
# Code Pull Requests

Start of the sequence feature's generator decoder.  This is not a complete implementation.  I wanted to give you an early look at the approach I'm taking.

Instead of retrofitting the TF1 code, I'm using the approach described in [TFA](https://github.com/tensorflow/addons/tree/master/tensorflow_addons/seq2seq) and demonstrated in these examples: [example one](https://github.com/tensorflow/addons/blob/master/docs/tutorials/networks_seq2seq_nmt.ipynb) and [example two](https://github.com/uber/ludwig/pull/682#issuecomment-612664646)

I'm just restating the obvious to establish my starting point.  Ludwig supports one of two tensor shapes for Decoder input
* shape [b, h] w/o attention
* shape [b, s, h]  with attention

The expected decoder output: shape [b, s, c]

One question, it is not clear to me how to determine sequence length for the decoder's output tensor.  Right now I'm just using `num_classes` as a place holder so that the decoder generates a rank 3 tensor.

**State of the implementation**

* Initial commit supports decoder input shape [b, h] and generates a tensor with shape [b, s, c], implemented the basic decoder.
* No exceptions are generated in`full_experiment()`.  The loss value decreases each epoch.
* These warning messages appear during training.  Same as with category feature.
``` 
WARNING:tensorflow:Gradients do not exist for variables ['embeddings:0', 'ecd/sequence_output_feature/fc_stack/fc_layer/dense/kernel:0', 'ecd/sequence_output_feature/fc_stack/fc_layer/dense/bias:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_1/dense_1/kernel:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_1/dense_1/bias:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_2/dense_2/kernel:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_2/dense_2/bias:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_3/dense_3/kernel:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_3/dense_3/bias:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_4/dense_4/kernel:0', 'ecd/sequence_output_feature/fc_stack/fc_layer_4/dense_4/bias:0'] when minimizing the loss.
```
* If interested here is the log file for an experiment run.
[...]

PR #718, opened 2020-05-31
TF2 porting: audio feature

carlogrisetti:
Audio feature now passes pytest with 6 warnings
I'm doing pytest tests\integration_tests\test_experiment.py::test_experiment_audio_inputs

There are some things that go over my knowledge, ie: is audio input always a float32 as opposed as various datatype with the sequence feature? If so, a check\cast can be avoided.
I marked any dubious line with a TODO


w4nderlust:
Thank you! this was quick, I'll look into it tomorrow and let you know.

carlogrisetti:
Do you have some feedback on this? I suspect I'll have fewer and fewer time going forward in the near future, and want help as much as possible

ludwig/features/audio_feature.py
```diff
@@ -357,30 +364,18 @@ def __init__(self, feature):
...
 
-    def _get_input_placeholder(self):
-        return tf.placeholder(
-            tf.float32, shape=[None, self.length, self.embedding_size],
-            name='{}_placeholder'.format(self.feature_name)
-        )
+    def call(self, inputs, training=None, mask=None):
+        assert isinstance(inputs, tf.Tensor)
+        assert inputs.dtype == tf.float32  # TODO: check that only tf.float32 is correct
+        # assert len(inputs.shape) == 2  # TODO: check correct inputs.shape as per sequence feature
```
w4nderlust, line 370: This should be assert len(inputs.shape) == 2, audio features are either raw, in that case [b, l, 1] or with extracted features, in which case [b, l, f] where f is the number fo features

carlogrisetti: Actually I commented that row because I always got == 3 and wanted to check wether it was correct, since being an extension of the sequence feature I expected to have the same shape lenght (which is 2 in the sequence feature code).

w4nderlust:
This looks good to me now. @jimthompson5802 as you also worked on the sequence and text feature, and audio is very similar to them, could you also do a quick check on this?

jimthompson5802:
@carlogrisetti 
I noticed two related items.
* `AudioInputFeature` is a subclass of `SequenceInputFeature`.  
https://github.com/carlogrisetti/ludwig/blob/aca1cf63e87347a56aa6e311b48805710bc26764/ludwig/features/audio_feature.py#L345-L347
I think line 346 is redundant because the default encoder for `SequenceInputFeature` is the `embed` encoder as shown here https://github.com/uber/ludwig/blob/0c12a5e2b1d4c609033ae8a82f809840b53be464/ludwig/features/sequence_feature.py#L128-L130

* Given that `embed` encoder is `SequenceEmbedEncoder` class, then the return here https://github.com/carlogrisetti/ludwig/blob/aca1cf63e87347a56aa6e311b48805710bc26764/ludwig/features/audio_feature.py#L367-L377 should just be `return encoder_output`. 

@w4nderlust There is a "timing issue" with the second point.  The baseline that @carlogrisetti  started with is out-of-date from the recent work with PR https://github.com/uber/ludwig/pull/699.  My version of `SeqenceEmbedEncoder` looks like this https://github.com/jimthompson5802/ludwig/blob/00e92731ee370bfcc54f68c8b11cb15a73dd07f3/ludwig/models/modules/sequence_encoders.py#L190-L206.  In this version, the return is `        return {'encoder_output': hidden}`.  I made changes to all the sequence encoders to return the dictionary.
In @carlogrisetti fork, his version of `SequenceEmbedEncoder` looks like this https://github.com/carlogrisetti/ludwig/blob/aca1cf63e87347a56aa6e311b48805710bc26764/ludwig/models/modules/sequence_encoders.py#L190-L206, where the return is `return hidden`  

Once the review for PR #699 completes and the merge occurs for the PR, there will be a breaking api change that will need resolution.

I hope this makes sense.

w4nderlust:
@jimthompson5802 got it, yes this is a good point. I guess @carlogrisetti can nor merge from `tf2_porting` into his for, because I actually merged #699 into `tf2_porting`. After the merging he can apply your suggestions, and it should work fine.

Dorry to both of you about this, I wasn't expecting the work on the sequence decoder to impact the returns of the sequence encoder so I thought these two PRs could be done in parallel.
Hopefully the changes are small enough that adapting should be simple and straightforward.

carlogrisetti:
@jimthompson5802 I tought the same on the redundancy of the embed variable, since it should be inherited from their parents, but I haven't been able to pass tests removing that, since it's missing an argument in the init function.
Honestly I'm not that good ad tracing that issue back, and since it was simply redundant and not harmful, I lazily left it there :)

w4nderlust:
Looks good to me, thank you!

PR #719, opened 2020-06-01
Cache pre-trained embeddings

jmizgajski:
# Code Pull Requests

Please provide the following:

When multiple input features are using the same pre-trained embeddings ludwig loads the embedding from scratch for each feature - thus it takes a lot of time to even begin training. This change prevents that, at the cost of keeping the most recent embedding in memory.

# Documentation Pull Requests

### DIFF
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -22,6 +22,7 @@ import os.path
 import pickle
 import random
 import re
+import functools
 
 import h5py
 import numpy as np
@@ -204,6 +205,7 @@ def load_pretrained_embeddings(embeddings_path, vocab):
     return embeddings_matrix
 
 
+@functools.lru_cache(1)
 def load_glove(file_path):
     logger.info('  Loading Glove format file {}'.format(file_path))
     embeddings = {}


w4nderlust:
This is interesting, I didn't know about `lru_cache()` and it's a pretty smart solution!
The only concern I have is: is there a mechanism for flushing the cache? The reason is that, after the embeddings have been assigned to the weight matrix, this cache will likely keep a reference to them. But after model building time, this function will never be called again, so that huge loaded embedding matrix will occupy RAM while it could have been freed. Ideally i flush should be called at the end of model building or something like it.
Or am I misunderstanding? It may be the case as I'm not familiar with this function.

jmizgajski:
@w4nderlust yes there is a simple way, you just call `load_glove.cache_clear()` - see here https://stackoverflow.com/questions/37653784/how-do-i-use-cache-clear-on-python-functools-lru-cache

Alternatively we could use `joblib.Memory` https://joblib.readthedocs.io/en/latest/auto_examples/memory_basic_usage.html
to cache in-between runs - but this could be problematic in cases when the user changes the embedding but keeps it under the same path, so would probably require an extra option.

w4nderlust:
I would go with the `cache_clear` option. Let's figure out where is the best place to put it. Probably at the end of `ECD.__init__()` in the new TF2 implementation.

jmizgajski:
@w4nderlust I would leave that to you, as I am not yet sufficiently familiar with the control-flow in ludwig. :)

w4nderlust:
> @w4nderlust I would leave that to you, as I am not yet sufficiently familiar with the control-flow in ludwig. :)

Sounds good :) So I'll probably merge this now and add adding cache clear as a task for after the TF2 posrting is done.

PR #721, opened 2020-06-02
Bag input feature tf2 porting

ydudin3:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/models/modules/bag_encoders.py
+++ b/ludwig/models/modules/bag_encoders.py
@@ -41,7 +41,7 @@ class BagEmbedWeightedEncoder(Layer):
     ):
         super(BagEmbedWeightedEncoder, self).__init__()
 
-        self.embed_sparse = EmbedWeighted(
+        self.embed_weighted = EmbedWeighted(
             vocab,
             embedding_size,
             representation=representation,
@@ -61,5 +61,5 @@ class BagEmbedWeightedEncoder(Layer):
 
             :param return: embeddings of shape [batch x embed size], type tf.float32
         """
-        embedded = self.embed_sparse(inputs, training=None, mask=None)
+        embedded = self.embed_weighted(inputs, training=None, mask=None)
         return embedded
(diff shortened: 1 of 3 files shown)


ludwig/models/modules/bag_encoders.py
```diff
@@ -0,0 +1,65 @@
...
+            force_embedding_size=False,
+            embeddings_on_cpu=False,
+            dropout_rate=0.0,
+            initializer=None,
+            regularizer=None,
+            **kwargs
+    ):
+        super(BagEmbedWeightedEncoder, self).__init__()
+
+        self.embed_sparse = EmbedWeighted(
```
w4nderlust, line 44: self.embed_sparse -> self.embed_weighted ?

ludwig/models/modules/embedding_modules.py
```diff
@@ -163,6 +163,8 @@ def __init__(
     ):
         super(EmbedWeighted, self).__init__()
 
+        self.vocab = vocab
```
w4nderlust, line 166: not super sure why you need this, is self.vocab used somewhere else?

w4nderlust:
Left a couple comments, minor stuff.
Other than that, does this pass the tests? If yes I'll merge once the comments are addressed.

ludwig/models/modules/embedding_modules.py
ydudin3: It's referenced in line 203.

ludwig/models/modules/bag_encoders.py
ydudin3: Yep, fixed that.

ydudin3:
@w4nderlust `test_experiment_various_feature_types` which uses bag feature as input passes locally.

ludwig/models/modules/embedding_modules.py
w4nderlust: You are right. But there we are only using the length, so maybe it's worth just doing:
```
self.vocab_length = len(vocab)
```
and then using it in line 203.

ydudin3: Makes sense, updated.

PR #726, opened 2020-06-07
TF2 porting: text feature

carlogrisetti:
Text feature now passes pytest with 2 warnings on precision\recall ill defined (I suppose due to demo\test data)
I'm doing pytest tests\integration_tests\test_experiment.py::test_experiment_seq_seq (which contains a text input and text output)

This seems to be specular to the audio feature, although I'm mentally missing the output feature call method, along with some doubts on the datatypes (I did not expect text to be loaded as int8, might need to widen the assert to allow other datatypes if necessary)

As usual, let me know if this is useful at all :)


w4nderlust:
I guess both you and @jimthompson5802 worked on the same, my bad for not syncying the work of both. Jim could you please check this and see if it is the same code you did? If not could you post here the differences?

jimthompson5802:
@w4nderlust 
re:
> Jim could you please check this and see if it is the same code you did?

Yes, @carlogrisetti and I worked on the same module: `text_feature.py`.  Here is a diff showing the changes I made.
[text_feature_diff.txt](https://github.com/uber/ludwig/files/4742008/text_feature_diff.txt)

Since Text is a subclass of Sequence and dependent the recent changes, I'm just waiting for the clean-up work on Sequence classes before pushing the Text Feature in case there are last minute breaking api changes to work out.

w4nderlust:
now that I merged the sequence decoder PR, could you two merge `tf_porting` in this and then merge both your  efforts into a single PR?

jimthompson5802:
@carlogrisetti if you merge the updated `tf2_porting` branch into your branch should bring you up-to-date on the other changes and allow you to finish this PR.  If you need help or have questions, let me know.

carlogrisetti:
And this passes tests again... seems even too simple to be true

ludwig/features/text_feature.py
```diff
@@ -183,42 +184,30 @@ def add_feature_data(
...
+        else:
+            self.encoder_obj = self.initialize_encoder(feature)
 
-        self.encoder = 'parallel_cnn'
-
-        encoder_parameters = self.overwrite_defaults(feature)
-        self.encoder_obj = self.get_sequence_encoder(encoder_parameters)
+    def call(self, inputs, training=None, mask=None):
+        assert isinstance(inputs, tf.Tensor)
+        assert inputs.dtype == tf.int8
```
w4nderlust, line 202: this one should be any \int\, 8 16 32 64.

ludwig/features/text_feature.py
```diff
@@ -247,12 +236,12 @@ def populate_defaults(input_feature):
...
 class TextOutputFeature(TextBaseFeature, SequenceOutputFeature):
-    def __init__(self, feature):
-        super().__init__(feature)
-        self.type = TEXT
+    decoder = 'generator'
 
+    def __init__(self, feature, decoder_obj=None):
+        TextBaseFeature.__init__(self, feature)
+        SequenceOutputFeature.__init__(self, feature)
         self.level = 'word'
```
w4nderlust, line 244: `level` `max_sequence_length` and `level` should all be class variables like decoder

ludwig/features/text_feature.py
carlogrisetti: That was my doubt, but couldn't generate anything else than 8. Thanks

ludwig/features/text_feature.py
carlogrisetti: Missed those!

w4nderlust:
Thank you!

PR #730, opened 2020-06-09
Convert neuropod output float32 np array to float64

linwu-zju:
convert the float32 np ndarray to float64 to make it compatible with neuropod and spark.


w4nderlust:
Thank you Lin! A couple thngs, I moved the neuropod code to a different directory, can you please merge from master and override the moved file? It's `ludwig_utils/neuropod_utils.py`.
Also, this change solves the problem you were having with Spark, right?

w4nderlust:
I committed it myself https://github.com/uber/ludwig/commit/9b34c75146fb62a56fd8af49b008f9e0832cfa16 because I changed a few thing in the meantime and there were conflicts, but thank you for your contribution, much appreciated!

PR #738, opened 2020-06-13
TF2 porting: Date encoder method docstring update, unit test refinement

jimthompson5802:
# Code Pull Requests

From what I can tell the underlying class parameters are surfaced.  I noticed a mis-match between parameters defined in `__init__()` method and docstring.  This PR updates docstring to match the `__init()__` parameters. 

I updated the `test_experiment_date` unit test to make use of parameterize decorator.

### DIFF
--- a/ludwig/models/modules/date_encoders.py
+++ b/ludwig/models/modules/date_encoders.py
@@ -64,28 +64,38 @@ class DateEmbed(Layer):
                    of the embedding matrix in regular memroy and the CPU is used
                    to resolve them, slightly slowing down the process
                    as a result of data transfer between CPU and GPU memory.
-            :param dropout: determines if there should be a dropout layer before
+            :param fc_layers: list of dictionaries containing the parameters of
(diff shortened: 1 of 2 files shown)


jimthompson5802:
Assuming this PR is OK, per your earlier suggestion, I'll look at this as the next task:
> I think the highest priority is to restore the functionality of the training loop, like validation, early stopping and the like, that we removed. Luckily now it will be much simplified because collecting metrics is much cleaner and simpler now.

w4nderlust:
looks good, thank you!

jimthompson5802:
I noticed this PR was `Closed` instead of `Merged`.     This means the proposed docstring updates and change to `test_experiment_date` are not going to be incorporated.  What were the issues with the proposed changes.

Before I delete the branch on my side, I wanted to confirm that is intent.

w4nderlust:
Just wrong button

jimthompson5802:
np...thank you.

Issue #740, opened 2020-06-14
Fix broken links to uber.github.io/ludwig


vkuzmin-uber:
I started to look at docs, examples and found that links to uber.github.io/ludwig are broken and it moved to ludwig-ai.github.io/ludwig-docs

I forked and cloned repo and so I am willing to provide PR with fix.

w4nderlust:
Thanks for your contribution!

PR #741, opened 2020-06-14
Fix broken links in docs and comments #740

vkuzmin-uber:
Links to uber.github.io/ludwig are broken and should be replaced with ludwig-ai.github.io/ludwig-docs


w4nderlust:
Thank you for this! We are transitioning to a new org an these things ore the side product of it, thank for helping fixing them!

PR #747, opened 2020-06-24
Fix frequency_vs_f1 graph labels

JiByungKyu:
# Code Pull Requests
For categorical output features, I try frequency_vs_f1 but result is weird.(it is weird in ludwig's docs result pictures.)

I use titanic data, below picture is the result.
image

What I found is in visualize.py, frequency_vs_f1 function is not correct.

Because labels are not matching with value of frequency and f1 score when plotting graph.

So I add logic which is matching label to value.

Sorry for not good at English, and thank you for reading pull request.

### DIFF
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -3017,6 +3017,10 @@ def frequency_vs_f1(
             f1_reordered = f1_np[f1_sorted_indices[::-1]][
                            :len(f1_sorted_indices)]
 
+            labels_np = np.array(labels)
+            labels_reordered = labels_np[f1_sorted_indices[::-1]][
+                           :len(f1_sorted_indices)]
+
             filename = None
             if output_directory:
                 os.makedirs(output_directory, exist_ok=True)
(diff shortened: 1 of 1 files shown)


w4nderlust:
Thanks for the contribution. I'm not really sure what you mean by weird, can you explain what the problem is and how are you solving it? Can you also provide come code to reproduce the problem so that i can both see it and see the solution?

JiByungKyu:
Sure, actually the origin problem is in pictures https://ludwig-ai.github.io/ludwig-docs/user_guide/#frequency_vs_f1.
In two pictures, frequency and f1 score are plotted differently but in each picture's labels(category class) are same(and the sequence).
In second graph, what I thought is if plotting values are resorted by frequency, labels are reordered too. 

In my pull request, I train on titanic data and using ludwig's save_test_statistics function, save the test_statistics result.
And call ludwig.visualize.frequency_vs_f1() function with the saved test_statistics result then the pictures are results.
In titanic data, there are no <UNK> data, but in the picture, <UNK> frequency are almost 600 and it is for 0(not survived).

I hope this will improve ludwig or may be i'm wrong... 
Thanks for answering :)

w4nderlust:
I inspected the code and i understand what the problem was and how the solution works. Will do a minor cleanup on top of your submission and then merge. Thank you for your contribution!

PR #750, opened 2020-07-04
Port Horovod distributed training support to TensorFlow 2

tgaddair:
This PR does the following:

- Replace TensorFlow 1.X usage of `hvd.DistributedOptimizer` and `hvd.broadcast_global_variables` with TensorFlow 2.X `hvd.DistributedGradientTape` and `hvd.broadcast_variables`, respectively.
- Fixed GPU / TensorFlow configuration to use the new syntax in TF 2.X.
- Replaced the removed `gpu_fraction` option with `allow_parallel_threads` as a user-facing config to enable deterministic training.
- Added ability to detect when Ludwig is being run by `horovodrun` so the user does not need to manually specify `--use_horovod`, and updated documentation.
- Replaced `mpi4py` with custom `allgather_object` function to aggregate metrics across workers.
- Added back learning rate warmup by directly setting `lr` property on the optimizer.
- Updated documentation to point to updated URLs.
- Added extra for Horovod when installing Ludwig: `pip install ludwig[horovod]`.


tgaddair:
cc @w4nderlust @jimthompson5802

ludwig/api.py
```diff
@@ -308,8 +308,8 @@ def close(self):
         """Closes an open LudwigModel (closing the session running it).
         It should be called once done with the model to release resources.
         """
-        if self.model is not None:
-            self.model.close_session()
+        # TODO(travis): determine whether this method is still needed in TF2
```
w4nderlust, line 311: Yes all the API still needs to be adapted to TF2, close on and train online and other things may not be needed or may work differently

ludwig/api.py
```diff
@@ -342,8 +342,8 @@ def train(
             skip_save_processed_input=False,
             output_directory='results',
             gpus=None,
-            gpu_fraction=1.0,
```
w4nderlust, line 345: The presence of gpu_graction was no for reproducibility purposes, but for allowing for the use of a single gpu for multiple trainings. It is used by the hyperopt module quite heavily for instance and is needed together with allow_growth for some problems TF has with RTX cards.

mkdocs/docs/user_guide.md
```diff
@@ -2981,38 +2981,45 @@ encoder: parallel_cnn
...
 horovodrun -np 16 \
     -H server1:4,server2:4,server3:4,server4:4 \
-    ludwig train --use_horovod ...other Ludwig parameters...
+    ludwig train ...Ludwig parameters...
 ```
 
 The same applies to `experiment`, `predict` and `test`.
 
-More details on Horovod installation and run parameters can be found in [Horovod's documentation](https://github.com/uber/horovod).
+When using `horovodrun`, Ludwig will automatically use Horovod during to conduct the model building, training and prediction phases in a distributed way.
```
w4nderlust, line 3013: Horovod _during_ to conduct -> Horovod to conduct

ludwig/api.py
tgaddair: I suppose what's not clear to me is: since there's no equivalent of a session in TensorFlow 2, is there still some kind of global state manager that needs to be manually closed, or can we just re-init TensorFlow through `tf.config`.  I think this will require some investigation to determine.

w4nderlust: My best guess: if a model is loaded for prediction purposes, there should be no problem of global state, while for training there's the optimizer state etc. So maybe we could consider having two methods, one that's actually `load` and one that is `resume_training` which takes care of recreating the global state.

ludwig/api.py
tgaddair: Yes, this was intended to be two changes rolled into one:

1. Expose `allow_parallel_threads` to the user, since they had no way to set it previously.
2. Remove `gpu_fraction` because it was removed in TensorFlow 2.

Since there's a need for something like it, we can do this in TensorFlow 2, which is roughly equivalent:

```
gpus = tf.config.experimental.list_physical_devices('GPU')
if gpus:
  try:
    tf.config.experimental.set_logical_device_configuration(gpus[0], [tf.config.experimental.LogicalDeviceConfiguration(memory_limit=1024)])
  except RuntimeError as e:
    print(e)
```

That limits the GPU memory in absolute terms.  If that sounds good, I can add back a parameter like `gpu_memory_limit`.

mkdocs/docs/user_guide.md
tgaddair: Good catch!

tgaddair:
@w4nderlust a few changes since you last reviewed:

1. Added `gpu_memory_limit` param to replace `gpu_fraction`.
2. Added Horovod integration test.
3. Fixed usage of `exp_dir_name` so that non-zero ranks do not attempt to create output files.

The third change was caught when putting together the integration test.  The new behavior of setting `exp_dir_name` to `None` instead of `'.'` makes it easier to catch bugs where we would accidentally write output files more than once to the sam location (race condition).  Let me know what you think.

PR #757, opened 2020-07-10
Prevent TensorFlow from being initialized more than once

tgaddair:
As part of this PR, move GPU and other TF config settings to Model constructor to ensure it occurs before the TensorFlow model is constructed.


w4nderlust:
Looks great, I already wanted to move initialize_tensorflow in tf_utils.py too, so this is 2 for 1 :)

PR #761, opened 2020-07-11
TF2 porting Fix visualization api unit test

jimthompson5802:
# Code Pull Requests

This is the start of fixing the unit test for visualization.  More work is needed.  

I want to give you a look and point out some of the changes.

A new custom metric class `HitsAtKMetric` was created.  This metric is a subsclass of `tf.kears.metrics.SparseTopKCategoricalAccuracy`  As I understand the documentation `SparseTopKCategoricalAccuracy` is equivalent to the `tf.nn.hit_at_k()` function used in TF1.  By subclassing, I can override the `update_state()` method to handle Ludwig's `predictions` data structure, as noted in the next paragraph.

I had to modify `OutputFeature.update_metrics()` method to account for special handling of `hits_at_k` metric.
```
    def update_metrics(self, targets, predictions):
        for metric, metric_fn in self.metric_functions.items():
            if metric == LOSS or metric == HITS_AT_K:           #<================
                metric_fn.update_state(targets, predictions)
            else:
                metric_fn.update_state(targets, predictions[PREDICTIONS])
```  

The modification is needed because `hits_at_k` metric require `logits`, just like the `loss` metric.  All the other metrics require `predictions`.   

Something for consideration, since `predictions` structure passed to `update_metrics()` is a dictionary, i.e.,
``` 
{
    logits: <tensor>
    predictions: <tensor>
    probabilities: <tensor>
}
```

If we defer to the metric function responsibility to pull out of the dictionary the required values, we can simplify the `update_metrics()` method to something like this
``` 
    def update_metrics(self, targets, predictions):
        for metric, metric_fn in self.metric_functions.items():
            metric_fn.update_state(targets, predictions)
```

Finally, and this one I need some guidance. This took a while for me to understand the situation.   

Depending how `CategoryOutputFeature.calculate_overall_stats()` method is called, the `test_stats` parameter comes in two forms.

If Ludwig's `test()` api is used the `test_stats` structure passed to `calculate_overall_stats()` is a tuple of length 2, e.g.,
**Form 1**
``` 
(
    {
        'category_5A13F': {'loss':....., 'accuracy':....., 'hits_at_k':.....}
        'combined': {'loss':....}   
    },
    {
        'category_5A13F': {
                             'predictions':  <tensor>,
                             'probabilities': <tensor>,
                             'logits': <tensor>
                           }
    }
)

```

OTOH, if `full_experiment()` is used, then the `test_stats` structure passed to `calculate_overall_stats()` is a dictionary
**Form 2**
``` 
{
    'y': {'loss':..., 'accuracy':...., 'hits_at_k': ....., 'predictions': ....., 'probabilities': ....},
    'combined': {'loss'....}
}
```

Given current structure of `calculate_overall_stats()` it appears that **Form 1** is the desired structure.  Is this correct?


jimthompson5802:
RE:
> Depending how CategoryOutputFeature.calculate_overall_stats() method is called, the test_stats parameter comes in two forms.

After digging into the history of changes, I understand what needs to be fixed.  In early days of TF2 porting, I implemented this change:
https://github.com/uber/ludwig/blob/3241911de80747c4b467220a7bb98156f5b36c15/ludwig/predict.py#L184-L188.  This was made to accommodate the changes in calculating overall stats in Output feature.  This change was made only in the `predict()` method of `predict.py` module.  The visualization test makes use of the `_predict()` method in `api.py` to calculate the overall stats.  The above code sequence is missing from `api._predict()` method.

This means **Form 2** is the correct structure for `test_stat`.

So I'm proceeding with the approach of implementing the reformatting of `test_stats` in `api._predcit()`.

jimthompson5802:
@w4nderlust 

I want to confirm my understanding of the `COMBINED` metrics.  Currently the `ACCURACY` metric is missing from the `COMBINED` metric dictionary, which is one of the issues with the visualization unit test.

In looking at the `master` branch, `COMBINED` metrics are `LOSS` and `ACCURACY`. https://github.com/uber/ludwig/blob/9b34c75146fb62a56fd8af49b008f9e0832cfa16/ludwig/models/model.py#L1127-L1132
Both of these are the sum of the respective metric for each output feature.  In the event `ACCURACY` is not relevant to a specific output feature, the accuracy measure for that feature is taken to be zero.

First question, "Is `ACCURACY` still relevant in the TF2 port?"  

If it is, then I have to add it back in, which then brings up other questions.

In the context of TF2 porting, this looks the place where I can add the missing `ACCURACY` combined metric.
https://github.com/uber/ludwig/blob/3241911de80747c4b467220a7bb98156f5b36c15/ludwig/models/ecd.py#L172-L179

One thing I noticed, the 'COMBINED` `LOSS` in TF2 is now set to be the average of the output features' losses not the sum.
https://github.com/uber/ludwig/blob/3241911de80747c4b467220a7bb98156f5b36c15/ludwig/models/ecd.py#L48

If this is a correct understanding, do you want the `COMBINED` `ACCURACY` to also be an average?  

If average is desired in TF2, then how would you like me to handle the case when `ACCURACY` is not relevant to a specific output feature?   

I propose not including that output feature in calculating the `COMBINED ACCURACY`.  Here is an example: three output features with `ACCURACY` measures of 0.9, N/A, 0.8.   In this situation the `COMBINED ACCURACY` will 0.85, the average of 0.8 and 0.9, the N/A is ignored.

The alternative is to set the missing value to a known value like 0, such as currently done in TF1.  In this case the `COMBINED ACCURACY` will be 0.566, the average of 0.8. 0.0 and 0.9.

For the moment, **I'll proceed with the assumption that `ACCURACY` is not needed** and make changes to accommodate this approach.  If 'ACCURACY` is needed, the re-work should be minimal.

jimthompson5802:
Looks like my assumption of `ACCURACY` is not needed is wrong.  So I'm going to implement a `COMBINED ACCURACY` mechanism.

I'm proceeding as follows:
* `COMBINED ACCURACY` is the average of individual output feature's accuracy.  This is similar to the current approach for `COMBINED LOSS`.
* If a particular output feature does not have an accuracy measure, zero will be used. 

If either of the above needs to change, let me know.

w4nderlust:
So the removal of COMBINED ACCURACY was deliberate, you shouldn't reimplement it, but work on the accumptio that, at least for now there's not going to be such score.

The rationale for that is twofold.

1. It was confusing (there's a bunch of opened issue about that) because it applies only on output features that had accuracy as metric, but not to numerical for instance, so seeing accuracy 0 if I had 2 numerical features may have been misleading. 
2. Moreover, with the new metrics functions, it was difficult to compute, because it was not the average of the accuracies, but was the logical AND of all the accuracy preditions (true or false for each row in the evaluation set) of all the provided output features that have an accuracy score. Before we were collecting such binary vectors (they were called CORRECT_PREDICTIONS) to compute this combined accuracy, now we are not doing it anymore, so it would be much more complicated to compute it.

The combinations of those two factors, plus the fact that it was useful only in some narrow cases, convinced me to remove it.

w4nderlust:
> # Code Pull Requests
> This is the start of fixing the unit test for visualization. More work is needed.
> 
> I want to give you a look and point out some of the changes.
> 
> A new custom metric class `HitsAtKMetric` was created. This metric is a subsclass of `tf.kears.metrics.SparseTopKCategoricalAccuracy` As I understand the documentation `SparseTopKCategoricalAccuracy` is equivalent to the `tf.nn.hit_at_k()` function used in TF1. By subclassing, I can override the `update_state()` method to handle Ludwig's `predictions` data structure, as noted in the next paragraph.
> 
> I had to modify `OutputFeature.update_metrics()` method to account for special handling of `hits_at_k` metric.
> 
> ```
>     def update_metrics(self, targets, predictions):
>         for metric, metric_fn in self.metric_functions.items():
>             if metric == LOSS or metric == HITS_AT_K:           #<================
>                 metric_fn.update_state(targets, predictions)
>             else:
>                 metric_fn.update_state(targets, predictions[PREDICTIONS])
> ```
> 
> The modification is needed because `hits_at_k` metric require `logits`, just like the `loss` metric. All the other metrics require `predictions`.
> 
> Something for consideration, since `predictions` structure passed to `update_metrics()` is a dictionary, i.e.,
> 
> ```
> {
>     logits: <tensor>
>     predictions: <tensor>
>     probabilities: <tensor>
> }
> ```
> 
> If we defer to the metric function responsibility to pull out of the dictionary the required values, we can simplify the `update_metrics()` method to something like this
> 
> ```
>     def update_metrics(self, targets, predictions):
>         for metric, metric_fn in self.metric_functions.items():
>             metric_fn.update_state(targets, predictions)
> ```
> 
> Finally, and this one I need some guidance. This took a while for me to understand the situation.
> 
> Depending how `CategoryOutputFeature.calculate_overall_stats()` method is called, the `test_stats` parameter comes in two forms.
> 
> If Ludwig's `test()` api is used the `test_stats` structure passed to `calculate_overall_stats()` is a tuple of length 2, e.g.,
> **Form 1**
> 
> ```
> (
>     {
>         'category_5A13F': {'loss':....., 'accuracy':....., 'hits_at_k':.....}
>         'combined': {'loss':....}   
>     },
>     {
>         'category_5A13F': {
>                              'predictions':  <tensor>,
>                              'probabilities': <tensor>,
>                              'logits': <tensor>
>                            }
>     }
> )
> ```
> 
> OTOH, if `full_experiment()` is used, then the `test_stats` structure passed to `calculate_overall_stats()` is a dictionary
> **Form 2**
> 
> ```
> {
>     'y': {'loss':..., 'accuracy':...., 'hits_at_k': ....., 'predictions': ....., 'probabilities': ....},
>     'combined': {'loss'....}
> }
> ```
> 
> Given current structure of `calculate_overall_stats()` it appears that **Form 1** is the desired structure. Is this correct?

Sounds fine.



> If we defer to the metric function responsibility to pull out of the dictionary the required values, we can simplify the `update_metrics()` method to something like this

That would make sense. At the same time would require us (and whoeven implements a new output feature) to create all the losses and metrics, just for wrapping existing metrics and losses, which is a bit tedious. I would say let's stick with what we have for now, but let's think about this later ok?

> # Code Pull Requests
> This is the start of fixing the unit test for visualization. More work is needed.
> 
> I want to give you a look and point out some of the changes.
> 
> A new custom metric class `HitsAtKMetric` was created. This metric is a subsclass of `tf.kears.metrics.SparseTopKCategoricalAccuracy` As I understand the documentation `SparseTopKCategoricalAccuracy` is equivalent to the `tf.nn.hit_at_k()` function used in TF1. By subclassing, I can override the `update_state()` method to handle Ludwig's `predictions` data structure, as noted in the next paragraph.
> 
> I had to modify `OutputFeature.update_metrics()` method to account for special handling of `hits_at_k` metric.
> 
> ```
>     def update_metrics(self, targets, predictions):
>         for metric, metric_fn in self.metric_functions.items():
>             if metric == LOSS or metric == HITS_AT_K:           #<================
>                 metric_fn.update_state(targets, predictions)
>             else:
>                 metric_fn.update_state(targets, predictions[PREDICTIONS])
> ```
> 
> The modification is needed because `hits_at_k` metric require `logits`, just like the `loss` metric. All the other metrics require `predictions`.
> 
> Something for consideration, since `predictions` structure passed to `update_metrics()` is a dictionary, i.e.,
> 
> ```
> {
>     logits: <tensor>
>     predictions: <tensor>
>     probabilities: <tensor>
> }
> ```
> 
> If we defer to the metric function responsibility to pull out of the dictionary the required values, we can simplify the `update_metrics()` method to something like this
> 
> ```
>     def update_metrics(self, targets, predictions):
>         for metric, metric_fn in self.metric_functions.items():
>             metric_fn.update_state(targets, predictions)
> ```
> 
> Finally, and this one I need some guidance. This took a while for me to understand the situation.
> 
> Depending how `CategoryOutputFeature.calculate_overall_stats()` method is called, the `test_stats` parameter comes in two forms.
> 
> If Ludwig's `test()` api is used the `test_stats` structure passed to `calculate_overall_stats()` is a tuple of length 2, e.g.,
> **Form 1**
> 
> ```
> (
>     {
>         'category_5A13F': {'loss':....., 'accuracy':....., 'hits_at_k':.....}
>         'combined': {'loss':....}   
>     },
>     {
>         'category_5A13F': {
>                              'predictions':  <tensor>,
>                              'probabilities': <tensor>,
>                              'logits': <tensor>
>                            }
>     }
> )
> ```
> 
> OTOH, if `full_experiment()` is used, then the `test_stats` structure passed to `calculate_overall_stats()` is a dictionary
> **Form 2**
> 
> ```
> {
>     'y': {'loss':..., 'accuracy':...., 'hits_at_k': ....., 'predictions': ....., 'probabilities': ....},
>     'combined': {'loss'....}
> }
> ```
> 
> Given current structure of `calculate_overall_stats()` it appears that **Form 1** is the desired structure. Is this correct?
> ...
> This means Form 2 is the correct structure for test_stat.

Got it. So we can change this the way we want. Before there was a single dictionary returned with both the predictions and the metrics because of the was TF1 worked, that was the easiest way to go. That's also the reason for all those `only_predict` flags adn logic you see arund. We can get rid of all of that right now and just have two functions, one that computes only predictions, `predict` and one that computes only stats, maybe called `eval` because `test` is too overloaded. this will simplify things everywhere, from model.py to postprocessing to the API. What I would suggest to do here anyway is, first fix the visualization tests to work with the Form 2, then later open a new PR for this code reorganization, how does that sound?

jimthompson5802:
Got it...
* Do not reimplement `COMBINED ACCURACY`
* Fix visualizations to work with FORM 2 predictions.

jimthompson5802:
Status update:  18 out of 22 visualization api tests pass, working on the failing 4 tests.

w4nderlust:
Wow great progress!

jimthompson5802:
Completed work on `test_visualization_api.py` unit tests.  All test pass.
```
pytest tests/integration_tests/test_visualization_api.py
platform linux -- Python 3.6.9, pytest-5.4.3, py-1.8.2, pluggy-0.13.1
rootdir: /opt/project
plugins: pycharm-0.6.0, typeguard-2.9.1
collected 22 items

tests/integration_tests/test_visualization_api.py ......................                                                             [100%]

============================================================= warnings summary =============================================================

==================================================== 22 passed, 3507 warnings in 56.27s ====================================================
```

This is ready to merge.

w4nderlust:
Looks great, thanks! Merging. I guess next step is to fix the other non API viz tests that were failing.

jimthompson5802:
Thank you.  I'll take a look at the other viz unit tests.

PR #762, opened 2020-07-12
Fixed wandb and comet for tf2, improved integration tests, enabled contrib with programmatic API

tgaddair:
- Removed TF2 incompatible code from comet and verified both comet and wandb end-to-end.
- Added integration test for comet and made wandb and comet tests use subprocess isolation to avoid test leakage.
- Added `ludwig.contrib.use_contrib` helper to allow contribs to be used from the programmatic API.

### DIFF
--- a/ludwig/contrib.py
+++ b/ludwig/contrib.py
@@ -23,6 +23,20 @@ import sys
 from .contribs import contrib_registry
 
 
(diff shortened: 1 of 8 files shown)


dsblank:
Excellent work!

w4nderlust:
Thanks for checking this Doug!

PR #766, opened 2020-07-12
Update Hyperopt for TF2 Porting

kaushikb11:
- Remove the Close Session logic for Tf2
- Fix Hyperopt Integration Tests
- Update logic for ParallelExecutor for hyperopt for the Tf2 porting.


w4nderlust:
This looks good to me so far. Just let's make sure it is doing the right thing with the pid / nvidia-smi.

kaushikb11:
@w4nderlust : This is for three GPUs and 6 workers (ParallelExecutor). Params `gpus` and `gpu_memory_limit` weren't being passed to the Model Class and added support for `gpus` of type `str` in the `initialize_tensorflow` [function](https://github.com/uber/ludwig/blob/tf2_porting/ludwig/utils/tf_utils.py#L48). 
mul_gpus

w4nderlust:
This looks good now, thank you!

w4nderlust:
@tgaddair just to confirm the tensorflow initialization
@calio for confirming this is fine using fiber too

ludwig/utils/tf_utils.py
```diff
@@ -69,7 +69,12 @@ def initialize_tensorflow(gpus=None,
 
     if horovod is not None and gpus is None:
         gpus = [horovod.local_rank()]
-    gpus = [gpus] if isinstance(gpus, int) else gpus
+
+    if isinstance(gpus, int):
+        gpus = [gpus]
+    else:
```
tgaddair, line 75: This will not work with line 71 above.  I would suggest:

```
elif isintance(gpus, str):
    ...
```

w4nderlust:
> Just one minor issue. Also, does this PR conflict with [3609963](https://github.com/uber/ludwig/commit/36099634bcc84eab4516b1584523aea4d922777d)? Can we do anything to make the merging less painful if so?

I think it shouldn't conflict because I first merged hyperopt in master and then merged master in tf2_porting, and this PR is on tf2_porting.

ludwig/utils/tf_utils.py
w4nderlust: good catch

calio:
This looks good to me and should not affect FiberExecutor. As long as it works with SerialExecutor, it should work with FiberExecutor.

PR #768, opened 2020-07-14
[WIP] Add bayesian strategy to hyperopt

dme65:
This PR adds a wrapper around [pySOT](https://github.com/dme65/pySOT), a surrogate optimization toolbox for hyperparameter tuning for Ludwig models.

There is currently only one test and more should be added before merging.


w4nderlust:
Thanks for the PR!
A couple things:

### 1)
it seems like the grid strategy was removed. Is that on purpose?

### 2)
could you please use the `tf2_porting` branch as the base branch?

### 3)
What is the meaning of the `n_init` parameter?
When I tried to set `num_samples` to 4 I got an assertion error:
```
assert max_evals > n_init and max_evals > batch_size
```
So should we add something like:
```
if n_init >= self.max_evals: n_init = max(0, max_evals)
```
?

### 4)
`Turbo1` is initialized with `batch_size = 1`, should probably make it work with batches too.

### 5)
history is used only for keeping track of the suggestions, maybe replace with a int counter, or do you plan to use history in some other way?

w4nderlust:
Thanks for refactoring to use PySot!
As it's a non fast forward pr, could you also merge uber:tf2_porting into it and resolve the conflicts? That way i can actually checkout it and try it on my machine. All the other points i listed before are non relevant anymore ;)

w4nderlust:
I actually did the merging and conflict resolution myself, no need to do it anymore.

PR #769, opened 2020-07-14
TF2 porting: Fix kfold_cv unit test (WIP)

jimthompson5802:
# Code Pull Requests

Unit test `tests/integration_tests/test_kfold_cv.py::test_kfold_cv_cli` fails on this call
https://github.com/uber/ludwig/blob/ff7159992eb59d5e8cb2bc3b4850d8a10a91f702/ludwig/experiment.py#L571-L572
because `kfold_cv_stats` structure contained `tf.Tensors` created by the output features `predict()` method.

Initially tried to resolve the issue by using output feature's `postprocess_results()` method, which converts the `tf.Tensor` object into numpy arrays.  Given current data structures at that point in processing, the structure for the `metadata` parameter required in `postprocess_results()` does not appear to be addressable.

As a work-around, the `tf.Tensor` objects, which represent predictions from the test data set are removed from `kfold_cv_stats`.  All other performance metrics are still available in `kfold_cv_stats` structure.

I'm still working on this to see what it will take to be able to use `post process_results()` method.  In the meantime, I wanted to provide some visibility in the work.

**Current view of unit tests with the above work-around**
```
platform linux -- Python 3.6.9, pytest-5.4.3, py-1.8.2, pluggy-0.13.1
rootdir: /opt/project
plugins: pycharm-0.6.0, typeguard-2.9.1
collected 460 items

../../../tests/integration_tests/test_api.py ..                                                                                      [  0%]
../../../tests/integration_tests/test_contrib_comet.py F                                                                             [  0%]
../../../tests/integration_tests/test_contrib_wandb.py .                                                                             [  0%]
../../../tests/integration_tests/test_encoders.py ......                                                                             [  2%]
../../../tests/integration_tests/test_experiment.py ....F........................................................................... [ 19%]
.................................................................................................................................... [ 48%]
..................................................................................................................................F. [ 76%]
.......                                                                                                                              [ 78%]
../../../tests/integration_tests/test_generic_features.py ..                                                                         [ 78%]
../../../tests/integration_tests/test_horovod.py Fs                                                                                  [ 79%]
../../../tests/integration_tests/test_hyperopt.py F                                                                                  [ 79%]
../../../tests/integration_tests/test_kfold_cv.py ...                                                                                [ 80%]
../../../tests/integration_tests/test_model_save_and_load.py .                                                                       [ 80%]
../../../tests/integration_tests/test_model_training_options.py ..FF.............                                                    [ 84%]
../../../tests/integration_tests/test_savedmodel.py F                                                                                [ 84%]
../../../tests/integration_tests/test_server.py F                                                                                    [ 84%]
../../../tests/integration_tests/test_simple_features.py ..........                                                                  [ 86%]
../../../tests/integration_tests/test_visualization.py ..........................                                                    [ 92%]
../../../tests/integration_tests/test_visualization_api.py ......................                                                    [ 97%]
../../../tests/ludwig/models/modules/test_encoder.py FFF.                                                                            [ 98%]
../../../tests/ludwig/utils/test_data_utils.py .                                                                                     [ 98%]
../../../tests/ludwig/utils/test_horovod_utils.py s                                                                                  [ 98%]
../../../tests/ludwig/utils/test_hyperopt_utils.py ....                                                                              [ 99%]
../../../tests/ludwig/utils/test_image_utils.py ..                                                                                   [ 99%]
../../../tests/ludwig/utils/test_normalization.py .                                                                                  [100%]

================================================================= FAILURES =================================================================

========================================================= short test summary info ==========================================================
FAILED ../../../tests/integration_tests/test_contrib_comet.py::test_contrib_experiment - assert 1 == 0
FAILED ../../../tests/integration_tests/test_experiment.py::test_experiment_multiple_seq_seq - tensorflow.python.framework.errors_impl.In...
FAILED ../../../tests/integration_tests/test_experiment.py::test_visual_question_answering - tensorflow.python.framework.errors_impl.Inva...
FAILED ../../../tests/integration_tests/test_horovod.py::test_horovod_implicit - assert 127 == 0
FAILED ../../../tests/integration_tests/test_hyperopt.py::test_hyperopt_executor - AttributeError: 'Model' object has no attribute 'close...
FAILED ../../../tests/integration_tests/test_model_training_options.py::test_model_progress_save[False-False] - AssertionError: assert 2 ...
FAILED ../../../tests/integration_tests/test_model_training_options.py::test_model_progress_save[False-True] - AssertionError: assert 2 == 3
FAILED ../../../tests/integration_tests/test_savedmodel.py::test_savedmodel - AttributeError: 'Model' object has no attribute '_session'
FAILED ../../../tests/integration_tests/test_server.py::test_server_integration - AssertionError: assert ['error'] == ['category_F0..._pr...
FAILED ../../../tests/ludwig/models/modules/test_encoder.py::test_image_encoders_resnet - AttributeError: 'ResNetEncoder' object has no a...
FAILED ../../../tests/ludwig/models/modules/test_encoder.py::test_image_encoders_stacked_2dcnn - assert (2, 2) == 2
FAILED ../../../tests/ludwig/models/modules/test_encoder.py::test_sequence_encoder_embed - AttributeError: 'EmbedSequence' object has no ...
=================================== 12 failed, 446 passed, 2 skipped, 4367 warnings in 901.37s (0:15:01) ===================================

```

### DIFF
--- a/ludwig/experiment.py
+++ b/ludwig/experiment.py
@@ -493,11 +493,11 @@ def kfold_cross_validate(
     # consolidate raw fold metrics across all folds
     raw_kfold_stats = {}
     for fold_name in kfold_cv_stats:
-        for category in kfold_cv_stats[fold_name]['fold_metric']:
-            if category not in raw_kfold_stats:
-                raw_kfold_stats[category] = {}
+        for output_name in kfold_cv_stats[fold_name]['fold_metric']:
+            if output_name not in raw_kfold_stats:
+                raw_kfold_stats[output_name] = {}
             fold_outputs = \
-                kfold_cv_stats[fold_name]['fold_metric'][category]
+                kfold_cv_stats[fold_name]['fold_metric'][output_name]
 
             # todo revisit to see if can use output feature.postprocess_results()
             # convert tensors created during predictions to numpy arrays to
@@ -517,20 +517,20 @@ def kfold_cross_validate(
                     'roc_curve',
                     'precision_recall_curve'
                 }:
-                    if metric not in raw_kfold_stats[category]:
-                        raw_kfold_stats[category][metric] = []
-                    raw_kfold_stats[category][metric] \
+                    if metric not in raw_kfold_stats[output_name]:
+                        raw_kfold_stats[output_name][metric] = []
+                    raw_kfold_stats[output_name][metric] \
                         .append(fold_outputs[metric])
 
     # calculate overall kfold statistics
     overall_kfold_stats = {}
-    for category in raw_kfold_stats:
-        overall_kfold_stats[category] = {}
-        for metric in raw_kfold_stats[category]:
-            mean = np.mean(raw_kfold_stats[category][metric])
-            std = np.std(raw_kfold_stats[category][metric])
-            overall_kfold_stats[category][metric + '_mean'] = mean
-            overall_kfold_stats[category][metric + '_std'] = std
+    for output_name in raw_kfold_stats:
+        overall_kfold_stats[output_name] = {}
+        for metric in raw_kfold_stats[output_name]:
+            mean = np.mean(raw_kfold_stats[output_name][metric])
+            std = np.std(raw_kfold_stats[output_name][metric])
+            overall_kfold_stats[output_name][metric + '_mean'] = mean
+            overall_kfold_stats[output_name][metric + '_std'] = std
 
     kfold_cv_stats['overall'] = overall_kfold_stats
 
--- a/ludwig/experiment.py
+++ b/ludwig/experiment.py
@@ -28,8 +28,6 @@ import numpy as np
 import pandas as pd
 import yaml
 
-import tensorflow as tf
-
 from ludwig.constants import TRAINING
 from ludwig.contrib import contrib_command
 from ludwig.data.postprocessing import postprocess
@@ -464,8 +462,8 @@ def kfold_cross_validate(
             logger.info("training on fold {:d}".format(fold_num))
             (
                 _,  # model
-                _,  # preprocessed_data
-                _,  # experiment_dir_name
+                preprocessed_data,  # preprocessed_data
+                experiment_dir_name,  # experiment_dir_name
                 train_stats,
                 model_definition,
                 test_results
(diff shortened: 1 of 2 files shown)


ludwig/experiment.py
```diff
@@ -496,6 +498,15 @@ def kfold_cross_validate(
                 raw_kfold_stats[category] = {}
             category_stats = \
                 kfold_cv_stats[fold_name]['fold_metric'][category]
+
+            # clear out predictions Tensors
+            # todo revisit to see if we can save predictions output
+            #    not enough structures to use output feature.postprocess_results()
+            keys = [key for key in category_stats
```
w4nderlust, line 505: if I understand correctly here category should be renamed metric, right?

ludwig/experiment.py
```diff
@@ -496,6 +498,15 @@ def kfold_cross_validate(
...
+                    if isinstance(category_stats[key], tf.Tensor)]
+            for key in keys:
+                del category_stats[key]
```
w4nderlust, line 508: Wouldn't

```
category_stats[key] = category_stats[key].numpy()
```

work?

jimthompson5802: I initially thought of doing that as well.

For numerical feature, this would be workable.  

Categorical or sequence features, the above will return the encoded value for the `predictions` key, e.g., {'predictions'; [1, 1, 3, 2]}`.  OTOH `postprocess_results()` returns `'predictions': ['a', 'a', 'c', 'b']`.

On reflection, this might be a better work-around because this provides predictions in some form as opposed to no predictions plus  the `probabilities` array will be provided.  If we do this, I just need to make a note in the documentation, that `predictions` in the `fold_metric` dictionary is the encoded representation of the output feature.

This issue probably shows the need to add a few more tests to make sure other output feature types do not present other problems.

I'll proceed along this approach.

ludwig/experiment.py
jimthompson5802: At this point, the dictionary `category_stats` has values like `loss`, `accuracy`, but also `predictions` and `probabilities`.

So it is a mixture of metrics and predicted output.  That is why I chose the a generic name `category_stats`.  Maybe a better name would be `fold_stats_predictions`.

w4nderlust: Got it. maybe let's call them `fold_outputs`.

w4nderlust: and `category` can be `output_name`.

ludwig/experiment.py
w4nderlust: My boubt is: the values in `fold_metric` are basically the values from the test stats for the different folds. Because of that, they should have already been postprocessed at this point, meaning all output tensors alrady converted etc. so this should not be an issue at this point, maybe we should go back in the steps to figure out why they are not postprocessed.

ludwig/experiment.py
jimthompson5802: Incorporated both suggestions.

ludwig/experiment.py
jimthompson5802: I understand the point.  To answer your specific question
> maybe we should go back in the steps to figure out why they are not postprocessed.

The reason kfold_cv does not see post-processed values is that kfold_cv uses the `experiment()` method to generate predictions.  This method does not include a call to `postprocess_results()`. 
https://github.com/uber/ludwig/blob/ff7159992eb59d5e8cb2bc3b4850d8a10a91f702/ludwig/experiment.py#L135-L161

OTOH, the `full_experiment()` method includes both the call to `experiment()` and `postprocess_results()`.  
https://github.com/uber/ludwig/blob/ff7159992eb59d5e8cb2bc3b4850d8a10a91f702/ludwig/experiment.py#L314-L381.

When I initially looked to run `postprocess_results()` after `experiment()` in the kfold_cv function, it appeared that I did not have addressability to the `metadata` parameter.  I'll take another look to see if it is possible to run the `postprocess_result()` method.

w4nderlust:
I actually ended up looking into it, and discovered that `experiment` does not dopostprovessing, while `full_experiment` does, so I introduced the postprocessing in `kfold_cross_validate`.
I don-t particularly like this solution because introduces redundancy, but it is necessary because of the current way we treat predict outputs,experiment etc. I also added some commented code, if we want to save the vsvs of predictions and probabilities we can uncomment it.

After we reorganize the outputs of `model.predict()` we can revise this entirely, but for now this works and passes the tests.

Let me know what do you think.

jimthompson5802:
The changes look good.  I think for the moment we can leave out saving the csv of predictions/probabilties.

I'm OK with merging.

PR #771, opened 2020-07-15
Add user guide for Fiber Executor

calio:
Add user guide for Fiber Executor


w4nderlust:
@calio I'm opening a new PR based on this on the ludwig-docs repo: https://github.com/ludwig-ai/ludwig-docs .
After I merge it, it would be great if you could also add information about using a shared drive for the data isntead of putting it inside the docker.

PR #774, opened 2020-07-17
TF2: hyperopt

w4nderlust:
This pull request merges improvements from @kaushikb11 and @dme65 and adds better testing and slight improvements in the GPU support.
Tests don't pass yet.


w4nderlust:
@kaushikb11 could you please help cleaning this up and solving the tests that don't currently work?

kaushikb11:
@w4nderlust : Is using `tuple` for the `range` param, the right way to go?

w4nderlust:
> @w4nderlust : Is using `tuple` for the `range` param, the right way to go?

What do you mean? We could go back to low and high and have some code for mapping it into a tuple if we are using the JointSpace.

dme65:
> @w4nderlust : Is using `tuple` for the `range` param, the right way to go?

The idea is that you either provide a range or a set of values, so I think it's cleaner to have the range as just one argument. 

As an example:
```
>>> JointSpace({"x": {"type": "real", "range": (0, 2), "space": "linear"}})
>>> JointSpace({"x": {"type": "real", "values": [0, 0.5, 1, 1.5, 2], "space": "linear"}})
```

w4nderlust:
Got it. We didn't have range for real values before.

I guess in conclusion my opinion is the following: it isn't a big deal to be hones, both having a low and high parameter and having a tuple work. Although, having a tuple may have the disadvantage that you have to check for the size of the tuple every single time, while with 2 seaprate values you don't need to, so I probably slightly prefer the 2 variables solution. Regarding values, that list can be of any length, so you don't have to check for the length (just non emptiness at most).

PR #785, opened 2020-07-22
Added comet_ml to test requirements

tgaddair:


dsblank:
Nice!

w4nderlust:
@dsblank with this we should have solved all issues related to comet in the TF2 branch. It would be great if you could run a test training using the `tf1_porting` branch and doublecheck if everything looks good for you too.
As I noticed you added some features in the mean time (the confusion matrix for instance, but also others) there could be also improvements that could be done to the integration (and we can think about them later), but just a quick check that everything works for you nw would be great!

dsblank:
Good ideas! Will do.

dsblank:
And congrats to your team! That was a lot of work!

PR #797, opened 2020-08-02
fix: train_set_metadata_json arg

ifokeev:
There is a bug of experiment `cli` arg `train_set_metadata_json` which fixed in this PR

### DIFF
--- a/ludwig/experiment.py
+++ b/ludwig/experiment.py
@@ -648,7 +648,7 @@ def cli(sys_argv):
     )
 
     parser.add_argument(
-        '--metadata_json',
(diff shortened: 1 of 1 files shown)


w4nderlust:
What is the problem this is solving?

ifokeev:
@w4nderlust wrong arg inside the cli. Can't call it without this parameter because `full_experiment` accepts only `train_set_metadata_json` https://github.com/uber/ludwig/blob/a00df9d228b31d5f079f369ff39eada7bac83dd5/ludwig/experiment.py#L179

w4nderlust:
Thanks!

PR #801, opened 2020-08-07
Refactoring trainer model

w4nderlust:


ludwig/data/preprocessing.py
```diff
@@ -295,247 +298,140 @@ def load_metadata(metadata_file_path):
...
+        training_set, test_set, validation_set = load_hdf5(
+            dataset,
+            model_definition['input_features'],
+            model_definition['output_features'],
+            shuffle_training=True
+        )
 
-    data_hdf5_fp = None
+    elif data_format in DICT_FORMATS:
+        if dataset:
```
tgaddair, line 408: Nice way to condense this part:

```
dataset = dataset or pd.DataFormat(dataset)
```

w4nderlust: Not sure, my logic here is: if dataset/training_set/etc. are None, keep them none, if the are not None, then convert them to dataframes. I guess that if I do `dataset = dataset or pd.DataFormat(dataset)`, if it's None, I'll get an empy dataframe, and when it's not None, I'll get the dictionary/

ludwig/data/preprocessing.py
```diff
@@ -82,21 +85,21 @@ def build_dataset_df(
...
-        train_set_metadata = build_metadata(
+    if metadata is None:
+        metadata = build_metadata(
             dataset_df,
             features,
             global_preprocessing_parameters
         )
 
-    data_val = build_data(
+    _data = build_data(
```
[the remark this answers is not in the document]
w4nderlust: In the past, we had problem by using the `data` name, now I don't remember the details but it was obscuring the data module name, that's why we used data_val. but maybe I can used a better name to begin with.

w4nderlust:
Pushed code to do data format inference. I kept the different values for format for now, but now the default is none and in that case it's automatically determined. Will still need the option to specify it manually, because for instance for csv if the user specifies a file without the extension, at the moment it is not recognized.

ludwig/data/preprocessing.py
```diff
@@ -295,247 +298,140 @@ def load_metadata(metadata_file_path):
...
+    # setup
+    features = (model_definition['input_features'] +
+                model_definition['output_features'])
+
+    # in case data_format is csv,
+    # check if there's a cached hdf5 file with hte same name,
+    # and in case move on with the hdf5 branch
+    if data_format in CSV_FORMATS:
+        if dataset:
+            if (file_exists_with_diff_extension(dataset, 'hdf5') and
```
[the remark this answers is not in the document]
w4nderlust: This is a great point, I was struggling in the past trying to explain to users that if they change the csv they have to delete the hdf5 and json to recreate them before training again.
Another dimension of variability is the model definition: if  the preprocessing section or the preprocessing section inside each input and output feature are different, the that hdf5 is not valid.
So I think we can have a section in the hdf5 called checksums and add a checksum of the csv and a checksum of the preprocessing parameters, if they match that is valid.

This said, I believe this is not suepr urgent, we can put this into the pile of things to do in v0.4 for improving preprocessing.

ludwig/models/ecd.py
```diff
@@ -215,6 +231,239 @@ def reset_metrics(self):
             of_obj.reset_metrics()
         self.eval_loss_metric.reset_states()
 
+    def batch_predict(
+            self,
+            dataset,
+            batch_size,
+            horovod=None,
```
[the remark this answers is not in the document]
w4nderlust: Introduces Predictor class

ludwig/models/ecd.py
```diff
@@ -215,6 +231,239 @@ def reset_metrics(self):
...
+
+        # consolidate predictions from each batch to a single tensor
+        for of_name, of_predictions in predictions.items():
+            for pred_name, pred_value_list in of_predictions.items():
+                predictions[of_name][pred_name] = tf.concat(pred_value_list,
+                                                            axis=0)
+
+        return predictions
+
+    def batch_evaluation(
```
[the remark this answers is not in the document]
w4nderlust: Introduces Predictor class

ludwig/models/new_ludwig_model.py
```diff
@@ -719,6 +721,109 @@ def evaluate(
...
+            #                          gpus=gpus,
+            #                          gpu_memory_limit=gpu_memory_limit,
+            #                          allow_parallel_threads=allow_parallel_threads)
+
+            # predict
+            test_results = self.evaluate(
+                test_set,
+                data_format=data_format,
+                batch_size=batch_size,
+                collect_predictions=True,
```
w4nderlust, line 811: I would create a parameter for collect_predictions that the user can set.

ludwig/models/new_ludwig_model.py
```diff
@@ -719,6 +721,109 @@ def evaluate(
...
+                collect_overall_stats=True,
+                debug=debug
+            )
+        else:
+            test_results = None
+
+        return (
```
w4nderlust, line 818: The caller of this function already has a reference to the NewLudwigModel object, so I believe that returning self, self.exp_dir_name and self.model_definition are superfluous.
I would return only:
```
test_results,
train_stats,
preprocessed_data
```

ludwig/models/predictor.py
```diff
@@ -257,14 +257,14 @@ def calculate_overall_stats(
         training_set_metadata
 ):
     overall_stats = {}
-    for output_feature in output_features:
-        of_name = output_feature.feature_name
+    for of_name, output_feature in output_features.items():
         if of_name not in overall_stats:
             overall_stats[of_name] = {}
         output_feature.calculate_overall_stats(
```
w4nderlust, line 263: I would modify this function, instead of passing `overall_stats[of_name]`, which is an empty dictionary, let's pass only predictions, targets and metadata. And then remove
```
if of_name not in overall_stats:
             overall_stats[of_name] = {}
```
And just do:
```
overall_stats[of_name] = output_feature.calculate_overall_stats(
  predictions[of_name],  # predictions
  dataset.get(of_name),  # target
   training_set_metadata[of_name]  #output feature metadata
)
```

This will require also changing the implementation in various output feature classes to return a dictionary instead of modifying a dictionary.

ludwig/models/new_ludwig_model.py
jimthompson5802: OK..I'll make it match in what happens in `evalutate.py`, i.e., `collect_predictions=not skip_collect_predictions`

ludwig/models/predictor.py
jimthompson5802: Glad you pointed this out.  Because I was going to ask the question that the current approach has each feature's `calculate_overall_stats()` method by modifying the dictionary that is passed as a parameter instead of returning a dictionary. 
I prefer returning a dictionary.  I understand that I'll have to modify all the features's `calculate_overall_stats`.

ludwig/models/new_ludwig_model.py
jimthompson5802: done.

ludwig/models/predictor.py
jimthompson5802: Resolved by PR #879.

Issue #803, opened 2020-08-07
bug: predict 'NoneType' object is not subscriptable


ifokeev:
master branch:

<img width="681" alt="Screenshot 2020-08-07 at 22 48 14" src="https://user-images.githubusercontent.com/2017148/89682917-31fb4680-d900-11ea-9048-9a41b20e9278.png">

should be the checker, I think.

This is a bug note. Still researching

w4nderlust:
#804 fixes it, we can close, thank you!

PR #804, opened 2020-08-07
fix: 'NoneType' object is not subscriptable

ifokeev:
#803


w4nderlust:
Thank you, good catch!

PR #829, opened 2020-08-16
Fix gpu_memory_limit to apply even when gpus param is not specified

### LINKED ISSUE #828: gpu_memory_limit only works if gpus parameter is specified
I'm trying to use the new gpu_memory_limit parameter introduced by TF2 as opposed as the gpu_fraction one.
One thing I noticed is that it only works if `--gpus` is also specified, as per
https://github.com/uber/ludwig/blob/b31ce876f6cd364471f7154740485656ab4ea6c9/ludwig/utils/tf_utils.py#L87-L107
Once you both specify `--gpus 0 --gpu_memory_limit 4000` (for example), allocated GPU memory is correctly set to 4GB. Without the `--gpus 0` bit, all available memory is allocated as if no `--gpu_memory_limit 4000` was specified.

One more thing is that Hyperopt still references the "old" `--gpu_fraction`
https://github.com/uber/ludwig/blob/b31ce876f6cd364471f7154740485656ab4ea6c9/ludwig/hyperopt_cli.py#L473-L479

tgaddair:
Also fixes hyperopt CLI to use `gpu_memory_limit` in place of removed `gpu_fraction`.

Fixes #828.

### DIFF
--- a/ludwig/hyperopt_cli.py
+++ b/ludwig/hyperopt_cli.py
@@ -471,11 +471,11 @@ def cli(sys_argv):
         help="list of gpus to use"
     )
     parser.add_argument(
-        "-gf",
-        "--gpu_fraction",
-        type=float,
-        default=1.0,
-        help="fraction of gpu memory to initialize the process with",
+        '-gml',
+        '--gpu_memory_limit',
+        type=int,
+        default=None,
+        help='maximum memory in MB to allocate per GPU device'
     )
     parser.add_argument(
         "-uh",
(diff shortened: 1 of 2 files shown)


w4nderlust:
Looks good to me, although there are a couple failed tests in test_experiment.py to check. Now sure those are connected to the code in this PR though.

tgaddair:
@w4nderlust looks like it was just flakiness.  Rerunning the test fixed it.

PR #831, opened 2020-08-17
HF Transformers Integration

ANarayan:
First pass at a complete integration of HF encoders. Supports integration of 16 encoders.

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -34,6 +34,7 @@ class BERTEncoder(Layer):
             self,
             pretrained_model_name_or_path='bert-base-uncased',
             reduce_output='cls_pooled',
+            trainable = False,
             **kwargs
     ):
         super(BERTEncoder, self).__init__()
@@ -41,7 +42,7 @@ class BERTEncoder(Layer):
             pretrained_model_name_or_path
         )
         self.reduce_output = reduce_output
-        self.transformer.trainable = False
+        self.transformer.trainable = trainable
 
     def call(self, inputs, training=None, mask=None):
         transformer_outputs = self.transformer(
@@ -64,12 +65,16 @@ class GPTEncoder(Layer):
     def __init__(
             self,
             reduce_output='sum',
+            pretrained_model_name_or_path='openai-gpt',
+            trainable=False,
             **kwargs
     ):
         super(GPTEncoder, self).__init__()
-        self.transformer = TFOpenAIGPTModel.from_pretrained('openai-gpt')
+        self.transformer = TFOpenAIGPTModel.from_pretrained(
+            pretrained_model_name_or_path
+        )
         self.reduce_output = reduce_output
-        self.transformer.trainable = False
+        self.transformer.trainable = trainable
 
 
     def call(self, inputs, training=None, mask=None):
@@ -90,6 +95,7 @@ class GPT2Encoder(Layer):
             self,
             pretrained_model_name_or_path='gpt2',
             reduce_output='sum',
+            trainable=False,
             **kwargs
     ):
         super(GPT2Encoder, self).__init__()
@@ -97,7 +103,7 @@ class GPT2Encoder(Layer):
             pretrained_model_name_or_path
         )
         self.reduce_output = reduce_output
-        self.transformer.trainable = False
+        self.transformer.trainable = trainable
 
 
     def call(self, inputs, training=None, mask=None):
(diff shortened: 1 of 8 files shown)


ludwig/encoders/text_encoders.py
```diff
@@ -40,15 +41,22 @@ def __init__(
             pretrained_model_name_or_path
         )
         self.reduce_output = reduce_output
+        self.transformer.trainable = False
```
w4nderlust, line 44: Here and in all other encoders add a `trainable=False` parameter to init, so that users have the option to finetune.

ludwig/encoders/text_encoders.py
```diff
@@ -82,30 +97,45 @@ def __init__(
...
+        self.transformer = TFTransfoXLModel.from_pretrained(
+            pretrained_model_name_or_path
+        )
         self.reduce_output = reduce_output
+        self.transformer.trainable = False
+
 
     def call(self, inputs, training=None, mask=None):
-        transformer_outputs = self.transformer(inputs, training=training)
+        transformer_outputs = self.transformer(
```
w4nderlust, line 132: this call looks different from the other encoders, specifically attention_mask and token_type_ids are missing. Is there a specific reason?

ludwig/encoders/text_encoders.py
```diff
@@ -163,15 +205,21 @@ def __init__(
...
+            inputs, 
+            training=training,
+            attention_mask=mask,
+            token_type_ids=tf.zeros_like(inputs)
+        )
         if self.reduce_output == 'cls_pooled':
             hidden = transformer_outputs[1]
         else:
-            hidden = transformer_outputs[0]
+            hidden = transformer_outputs[0][:,1:-1,:]
```
w4nderlust, line 220: Some encoders are skipping the first and last element of the sequence with `[:,1:-1,:]`, some others don't. i guess it's an encoder specific thing right? You doublechecked the size of all of them?

ludwig/encoders/text_encoders.py
```diff
@@ -353,7 +452,7 @@ def call(self, inputs, training=None, mask=None):
         else:
             hidden = transformer_outputs[0]
             hidden = reduce_sequence(hidden, self.reduce_output)
-        return hidden
+        return {'encoder_output': hidden}
 
 
 # todo tf2 check lack of TFDialoGPTModel
```
w4nderlust, line 458: this looks like is still in todo, right?

ludwig/features/sequence_feature.py
```diff
@@ -141,19 +139,24 @@ class SequenceInputFeature(SequenceFeatureMixin, InputFeature):
     def __init__(self, feature, encoder_obj=None):
         super().__init__(feature)
         self.overwrite_defaults(feature)
+        feature.update(feature['preprocessing'])
```
w4nderlust, line 142: Trying to figure out why this is needed. i guess to update pad_idx?

ludwig/encoders/text_encoders.py
ANarayan: Yup. This encoder doesn't have the attention_mask and token_type_ids as parameters.

ludwig/encoders/text_encoders.py
ANarayan: This is an encoder specific thing. Some of them append bos token or CLS token to the start of the tokenized input and append a eos or sep token to the end of the tokenized input. The indexing is to not consider the encoded output of those tokens.

ludwig/encoders/text_encoders.py
ANarayan: Yup. Didn't look into it extensively, but still looks like HF doesn't have a TF implementation of these encoders.

ludwig/features/sequence_feature.py
ANarayan: It is essentially used to make the pretrained_model_name_or_path parameter value directly accessible when we initialize the encoder. In the original feature dict, the pretrained_model_name_or_path paramenter is nested inside the key 'preprocessing'  (which is a dict) and is thus not passed to the encoders at initialization time. But, by updating the feature dict to directly contain a key for pretrained_model_name_or_path, the encoder now has access to that value.

w4nderlust: Got it. This means that if someone specifies it at the encoder level, it gets overridden. For now this is fine, but it's a design choice I have to think about. I may end up going the other way around (the one you specify in the encoder end up used in the preprocessing).

w4nderlust:
@ANarayan please sign the CLA so I can merge ;)

ludwig/features/sequence_feature.py
ANarayan: Ah gotcha. This makes sense!

Issue #840, opened 2020-08-21
Caption on training improvement misleading


carlogrisetti:
Hello :)
This piece of code
https://github.com/uber/ludwig/blob/62430e4a0dd7a4fda08d6dcd615fbdbbf53c5377/ludwig/models/trainer.py#L1316-L1318
although procedurally correct (I want to track whether there has been an improvement since the last plateau and learning rate reduction, to look for the next plateau) leads to a technically wrong statement in the CLI output.
Let's assume that the model is training, 5 epochs have passed and the learning rate gets automatically dropped as per the `reduce_learning_rate_on_plateau` parameters. The "last improvement epoch" gets reset to the current epoch (to keep track of the fact that we reduced the learning rate, and we thus need to wait another 5 epochs without improvement before reducing the learning rate again), but in the CLI the next epoch will show that it's only 1 epoch since the last improvement (which should have been 6), the next one will show 2 epochs (but it should be 7), and this only gets fixed as soon as we have an actual improvement.

It's trivial, but maybe it gets to your nerves as much as it does to me :)

PR #846, opened 2020-08-24
CLI output improvement

carlogrisetti:
I didn't want to "complain" on a fix\improvement I requested on CLI output messages... and moreover I knew how to write that, so...
here it is :)

It's my best imaginable solution to issue #840 I raised, showing both the last improvement and the last learning rate reduction and\or batch size increase

The only issue I found is that, having to modify ProgressTracker, models created prior to this PR are not resumable with a build after this one (since it tries to load the new values to track LR\BS modifications from the model training progress json, and they're missing).

If you deem it to be too much of a breaking change we could implement some sort of "set default value" here, to inject default values in the loaded array before passing it to the ProgressTracker constructor:
https://github.com/uber/ludwig/blob/13eb6bf434f384c45ba1d92ad1dcda04fc6f0517/ludwig/models/trainer.py#L1422-L1424

I hope you think this is an useful improvement.


w4nderlust:
I understand what you are proposing and it makes sense.
There i something to discuss about regarding the logic though.
In both `reduce_learning_rate()` and `increase_batch_size()` you don't reset `last_improvement`, and if that is not done it changes the logic.

The current logic is: whenever the learning rate is decreased or the batch size is increased, reset the count of epochs since improvement fr early stopping. The reason is that, if I say patience is 5 epochs and I reduce the learning rate after 5 epochs without improvement, I want to give the reduction/increase a chance to have an effect before stopping.

Your new logic doesn't do that, it only decreases or increases without resetting last_improvement, which means that if I provide early_stopping = 7 and lr_patience = 5, I could reduce the lr at epoch 6 say (because the first epoch is peak validation performance and then it decreases) and then not improving performance for one epoch and terminate due to early stopping.

We can discuss about which is the best logic, but it's important to point out that they are different before merging.

carlogrisetti:
I didn't think about the early stopping function combined with the BS/LR plateau.

If you ask me, the intended behavior is exactly how it reads: after N epochs with no improvement, you should stop (no matter what).
Had I encountered that situation, I would have pointed out that "I'm not stopping because I reduced LR / increased BS" as a bug.
I interpret the LR/BS feature as an optimization, to "squeeze out" a fine tuned model starting from a good hyperparameter definition (so the early stopping applies to the whole training in my vision, especially for the fact that I usually set it pretty high, and wouldn't have such of a dramatic effect with or without this PR merged)

Of course it is your decision in the end, and I understand the logic that is behind the current implementation. If that stands, you only need to fix the messages by removing "accuracy" (since it's variable depending on what measure I chose in the definition), and all the rest is spot-on.

Feel free to trash this if you think so, I know it's not personal :)

w4nderlust:
In the last commit you removed references in the prints to how epochs passed since last icnrease / decrease. What is the rationale for that? Why is it redundant?

carlogrisetti:
The same information is already printed in the reduce_/increase_ functions when either of them is active.
I also made it so that the last reduction/increase is only displayed if we are not having a model validation improvement, not to clutter the cli output every time without reason (I don't care how many epochs have passed since the last LR reduction, if the model is improving)

carlogrisetti:
Oh well... I have to check why it is failing the build...

While I was thinking about this issue, it came to my mind that something might be wrong in how we decide whether to reduce LR or increase BS... We're deciding about that based on the validation measure improvement... but we are training based on the training dataset. The condition where we would want to increase BS or reduce LR is when we can't have an improvement on the training dataset, since we can't directly control the validation one, and it's only a "check" of how well we are performing.
In case where we have validation data that doesn't correctly mimic the training data (ie: more complex and\or too few training data to correctly predict validation data) we have a training loss that constantly goes down (because the network is overfitting to the training data maybe), but this is not seen as an improvement, thus tripping the LR reduction \ BS increase mechanism, even if it is not needed at all (at that LR\BS the model is still improving. It's the validation part of it that it's not, and that is unrelated to LR\BS, but to the specific composition of the validation dataset).
Am I completely wrong or does this make sense? (it's of course a separate issue)

ifokeev:
@carlogrisetti as I understand it's the paradigm of the validation (see "Validation dataset" https://en.wikipedia.org/wiki/Training,_validation,_and_test_sets):

> An application of this process is in early stopping, where the candidate models are successive iterations of the same network, and training stops when the error on the validation set grows, choosing the previous model (the one with minimum error)

By the way, I think the idea about early stop by training dataset is might be. But it will be kind of research.
Good to have some config field which switch early stop dataset.

Another idea: it's better to have the `ludwig` which able to stop/start from different epochs with changed-by-user LR and BS. We need hot-replacement for hyperparameters during the training – that potentially might increase convergence.
I made 500+ experiments with ludwig and see that neural networks turn on about hyperparameters a lot.

carlogrisetti:
Well... yes and no... what you are talking about is preventing overfitting of the model by keeping the model with the best validation  result, and not saving the next one which (for example) has a higher error.
This is already implemented in Ludwig, and it's one of the best features not having to worry about iterations\overfitting, since Ludwig takes care of it all.
Early stop can be defined on the validation improvement because I expect that training dataset evaluation will *always* improve, whilist validation dataset evaluation will improve upon a certain point, where we will have the best model for that training run, and then decay due to overfitting. Early stop on that would mean that we give the training N chances to further improve and turn around this validation evaluation decay in performance, and get a better result than the currently best one. If we cannot do that within N epochs, we're doomed and we shall stop.

The purpose behind LR reduction \ BS increase, is to push further with a training that, keeping the current learning rate, would not be able to improve its precision. That happens primarily on the training dataset (the model can't converge to a solution at all because it's too "rough" to even understand the training data), thus we need to lower the training rate or increase the batch size (or both).
If the model il still improving on the training dataset, reducing the learning rate or improving the batch size is in fact counterproductive, because we're limiting the growth it currenly is having due to the fact that we are inducing a slower learning (most of the time).
The fact that the validation performance is not going down is totally irrelevant to the training process of the model, which is getting better and better as epoch passes... the issue is that it is getting better only on the training data because the model is either too complex (and it is overfitting on the training data) or validation data is not on the same complexity\difficulty level to predict as the training data. Lowering LR or increasing BS should not have any improvement over this issue.

I might still be totally wrong, but that should be the correct reasoning behind changing the training parameters on the go. Do you agree with me if put in this way or do you have a different perspective on this?

carlogrisetti:
(in the meantime I suppose build errors had nothing to do with this PR, let's wait for the gurus feedback on this :))

w4nderlust:
A couple things:
1. the failing test doesn't have much to do with the pr, will check it out
2. I see your point on training set vs. validation set used as a criterion for lr and batch size. Will think about it a bit and search around, but I agree with the motivation, so would want to doublecheck in order to be sure. Either way, that could be another PR.

For this one, would you be able to write a little test where you have a list of validation accuracies and then we simulate a few epohs of lr and batch changes? It would be useful to make sure the behavior we want is reflected.

carlogrisetti:
I wouldn't know how to write a test that checks that, but if you are worried about the actual effectiveness, I wrote this PR because I'm actively using this (since I found out that decay and decay_rate were not working anymore), and it works exactly as intended... well... as I intended and explained in this PR at least.
I only tested LR reduction, but the code for BS increase is exactly the same.

If you point me to an example of a test I could try to patch something up.
... and that's only if you're talking about an automated test to run during a build. If you're asking for a synthetic dataset to repro and test the validity of this single PR (not to check on every PR\commit done), it's a whole other request and I could manage to do that, since it's already working for the data I'm working on :)

carlogrisetti:
I hope this is what you asked for :)

dataset.csv and model_definition.yaml: [ludwig-test.zip](https://github.com/uber/ludwig/files/5132895/ludwig-test.zip)

As soon as it starts model improves, and the message is the same as before (no sense in showing that LR has not been reduced in N epochs, since we're improving and potentially never need LR reduction):
image

When model stops improving, along the usual "last improvement" message, we're also notified of the last LR reduction:
image

In the meantime we have a model improvement, this resets the "last improvement" count as usual, LR does not get reduced since it had an improvement (because patience is calculated both on the last improvement and on the last LR reduction). Subsequently model does not improve but the last reduction counter is not reset due to the model being improved in the meantime:
image

Model keeps not improving and reaches the patience threshold. No sense in writing how many epochs were since the last LR reduction, since we're doing that right now:
image

Model still does not improve and we're having a LR reduction count reset, whilist the model improvement count is still counting up, despite having just set a lower LR:
image

Then another plateau is reached after another 5 epochs, the LR gets reduced, and so on...
image

w4nderlust:
I see. I guess if we want to separate the logics, having last improvement in the ifs could be a problem... have to think about it a bit :) The prints are fine anyway, but i want to be sure I'm 100% convinced of the logic before commiting to change it.

carlogrisetti:
Sure, as I told you before it's your choice, this is only my opinion on this on how to improve it.

Actually I thought about my last digression and If you'll want to flip the logic towards a training evaluation, you'll just need to add a last_improvement_on_training (or something like that) in the ProgressTracker, or something like that and update it in the training evaluation, of course checking for that one in the already modified code for reducing LR or increasing BS.
Then, the message would be displayed if you had last_improvement_on_training > 0 and the LR/BS value > 0

I might even be able to do it myself if you want... Maybe :)
Just think about it and let me know

carlogrisetti:
Oh, just as an aftertought... if you want to switch the LR\BS governance towards the training evaluation results, I think that should be done prior to 0.3.0, due to the breaking changes in the ProgressTracker that would be needed (it's just a parameter to add, but it will nonetheless be missing in previous models).

In the meantime it happened to me several times to need to stay on the same learning rate as before (since training was improving very fast) but Ludwig reduced the LR as per N epochs since an improvement on the validation. This slowed the learning performance, since I either had to wait for an improvement with the much lower LR (which seldomly happens, since it's going back to actually improving much slower than before after the LR reduction), or either increase the number of epochs of patience before LR reduction (which works, but it's somewhat inefficient, since when needed it could wait way less than the epochs I am forced to set).

I could give it a try if you want, and make a PR on that, but I would like a go\no-go on that before spending time on this

PS: as per this PR itself, I still find it useful to have the separate info on vali improvement and LR reduction, even if it now happens on vali and not on training (I'm currently using my forked branch to work)

w4nderlust:
This will be addressed in v0.3 don't worry. I'm just ofcusing on other things at the moment, but I will get back to this. In particular I believe I will allow to specify the split to calculate the improvement on, and by default set it to be training, but that requires some reworking, that's why I'm postponing it.

w4nderlust:
@carlogrisetti worked on a proposal for improving the logic of both learning rate reduction and batch size increase, it intrduces parameters for specifying the metric and the split to use for maximum flexibility and uses training loss as default.
Let me know what you think please.

carlogrisetti:
There's an issue in that there's something wrong with the loop, like it is not checking for training improvement and only for LR\BS edit epochs passed.
I had patience set to 10, training improved up to epoch 10 included, but at epoch 11 LR was reduced as if it didn't take into consideration the improvement on the evaluation.
I could reproduce it so it's no fluke. Will have to look deeper into it.

As a side note, I changed the output message to include the actual split used, not to always say "validation". It might be useful to actually append the epochs passed since the evaluation metric improved on the evaluation split too, since it will be a different number from both the last loss improvement on validation (by default) and the last LR\BS edit. Do you think that would be useful or just clutter (and one should just trust the engine inner workings)?

Bye (and thanks for the super-quick work on this. I didn't expect this so soon)

carlogrisetti:
I would also rename `reduce_learning_rate_validation_metric` to `reduce_learning_rate_evaluation_metric` or something like that, just to also switch variable\parameter names in favor of a generic evaluation metric that could be based either on training or validation (or test...)

Anyway I found the issue: you were checking for the model to improve evaluation on the defined metric... but the LR reduction \ BS increase should happen if the model is NOT improving :)
Also, you were never updating the `progress_tracker.best_reduce_learning_rate_valid_metric` and `progress_tracker.best_increase_batch_size_valid_metric` values, so the training loop always tought to be improving when deciding whether to reduce LR \ increase BS.
I also added tracking for the last value that those metrics improved on the (training) dataset, and used those to decide whether to increase BS \ reduce LR

The only concern I have now is that the LR reduction \ BS increase message might be too long... I tried splitting on two rows, but if you have both LR and BS active, it can get confusing as you would see a duplicate "improvement of loss on training happened N epochs ago" since it could be set on different splits and different measures.
Like it is now It's within the default margins of a windows cmd console, so LGTM, but let me know.
image
image

Now, aesthetic issues apart on variable naming, it seems to be working fine.

w4nderlust:
Thanks @carlogrisetti ! I modified a little bit the logic, now nothing is computed and no message is printed after the last increase / decrease, which was needed to cleanup the prints.
I'm merging it, but please try it out and let me know if there other things you would modify in this PR.

carlogrisetti:
I see... I can live with the fact of having to check other data to know whether LRr\BSi had stopped. I just ran a training and it seems to be working as intended.

The only thing I'm not sure about is this print:
https://github.com/carlogrisetti/ludwig/blob/89f1bd5a3172bb4a40db1161b119e77fc8783828/ludwig/models/trainer.py#L1021-L1022

I understand you wanting to specify the split we're improving on (it will always be validation, I suppose), but it's in a slightly different syntax than the other prints: if we mantain the same print logic as in the LRr\BSi code, it should be `last improvement of {metric} on {split} happened...`, but here we have the feature too...
I see two options to remain consistent:
1. `last {feature} improvement of {metric} on {split} happened...` --> `last combined improvement of loss on training happened...`
2. `last improvement of {feature} {metric} on {split} happened...` --> `last improvement of combined loss on training happened...`

Personally I like option 2 way more, it's more "natural language"-like and best fits to the other prints.

"fixed" (or not) this, LGTM for merging

w4nderlust:
Modified everywhere to read like "{feature} {split} {metric}" -> "object_type training loss" or "class validation accuracy" which is concise and clean ;)

PR #855, opened 2020-08-28
Load LudwigModel in test_savedmodel before creating saved model

### LINKED ISSUE #844: Exporting a savedmodel from a loaded model
**EDIT: nevermind... it's already marked as a TODO TF2**
https://github.com/uber/ludwig/blob/f167981683c067b50be6a3656cbf553efbf192e9/ludwig/models/trainer.py#L1116-L1145

Do you want a separate issue for the actual issue?
In any case, here's the code I tried to use (several variations, but with the same outcome)
`from ludwig.api import LudwigModel`
`ludwig_model = LudwigModel.load('C:\\blabla\\model_run\\model')`
`ludwig_model.model.save_savedmodel('C:\\blabla\\savedmodel')`

Note that running a prediction on said model works fine:
`import pandas as pd`
`prediction = model.predict(data_df=pd.DataFrame({'ImageFile': ['C:\\somewhere\\over\\the\\rainbow.jpg']}))`

I even tried running the prediction before trying to save the savedmodel (as per the raiserror instruction, somewhat), but with no avail. I suspect it has something to do on how each model might be defined. Different model definitions will 

WARNING:tensorflow:Skipping full serialization of Keras layer <ludwig.models.ecd.ECD object at 0x000002062CAFE2E8>, because it is not built.
  File "C:/somewhere/scratch.py", line 3, in <module>
    model.model.save_savedmodel('C:\\somewhere\\savedmodel')
  File "C:\Program Files\Python36\lib\site-packages\ludwig\models\trainer.py", line 1139, in save_savedmodel
    self.model.save(save_path)
  File "C:\Program Files\Python36\lib\site-packages\tensorflow\python\keras\engine\network.py", line 1052, in save
    signatures, options)
  File "C:\Program Files\Python36\lib\site-packages\tensorflow\python\keras\saving\save.py", line 138, in save_model
    signatures, options)
  File "C:\Program Files\Python36\lib\site-packages\tensorflow\python\keras\saving\saved_model\save.py", line 65, in save
    saving_utils.raise_model_input_error(model)
  File "C:\Program Files\Python36\lib\site-packages\tensorflow\python\keras\saving\saving_utils.py", line 95, in raise_model_input_error
    'model._set_inputs(inputs).'.format(model))
ValueError: Model <ludwig.models.ecd.ECD object at 0x000002062CAFE2E8> cannot be saved because the input shapes have not been set. Usually, input shapes are automatically determined from calling .fit() or .predict(). To manually set the shapes, call model._set_inputs(inputs).

I tried doing that with the save_for_serving API call, and it breaks down in a different way:
`from ludwig.api import LudwigModel`
`model = LudwigModel.load('C:\\somewhere\\model_run\\model')`
`model.save_for_serving('C:\\somewhere\\savedmodel')`

  File "C:/somewhere/scratch.py", line 6, in <module>
    model.save_for_serving('C:\\somewhere\\savedmodel')
  File "C:\Program Files\Python36\lib\site-packages\ludwig\api.py", line 313, in save_for_serving
    if (self.model is None or self.model._session is None or
AttributeError: 'Trainer' object has no attribute '_session'

which refers to this
https://github.com/uber/ludwig/blob/f167981683c067b50be6a3656cbf553efbf192e9/ludwig/api.py#L313-L315

_Originally posted by @carlogrisetti in https://github.com/uber/ludwig/issues/841#issuecomment-678616694_

tgaddair:
Fixes #844.

### DIFF
--- a/tests/integration_tests/test_savedmodel.py
+++ b/tests/integration_tests/test_savedmodel.py
@@ -18,6 +18,7 @@ import shutil
 from copy import deepcopy
 
 import numpy as np
+import pytest
 import tensorflow as tf
 
 from ludwig.api import LudwigModel
@@ -29,7 +30,8 @@ from tests.integration_tests.utils import generate_data
 from tests.integration_tests.utils import sequence_feature
 
 
-def test_savedmodel(csv_filename):
+@pytest.mark.parametrize('should_load_model', [True, False])
+def test_savedmodel(csv_filename, should_load_model):
     #######
     # Setup
     #######
@@ -78,7 +80,8 @@ def test_savedmodel(csv_filename):
     ###################
     # load Ludwig model
     ###################
-    ludwig_model = LudwigModel.load(ludwigmodel_path)
+    if should_load_model:
+        ludwig_model = LudwigModel.load(ludwigmodel_path)
 
     #################
     # save savedmodel
--- a/tests/integration_tests/test_savedmodel.py
+++ b/tests/integration_tests/test_savedmodel.py
@@ -75,6 +75,11 @@ def test_savedmodel(csv_filename):
     shutil.rmtree(ludwigmodel_path, ignore_errors=True)
     ludwig_model.save(ludwigmodel_path)
 
+    ###################
+    # load Ludwig model
+    ###################
+    ludwig_model = LudwigModel.load(ludwigmodel_path)
+
     #################
     # save savedmodel
     #################


w4nderlust:
Tests are still running, but i suppose that they pass, which says that with the recent changes it works also with loading the model, right? Maybe for regression purposes we can have a boolean flag in the tests signature `should_load_model` so that we can test both cases, what do you think?

tgaddair:
Sure, let me throw that in there!

tgaddair:
@w4nderlust should be good to go now, please take another look when you get a chance.

PR #875, opened 2020-09-12
Fixed collect_weights and collect_activations

tgaddair:
- Save model definition inside `NewLudwigModel.train` as `Trainer` lacks awareness of model definition.
- Added helper methods to load model from model definition and check model initialization.

### DIFF
--- a/ludwig/models/new_ludwig_model.py
+++ b/ludwig/models/new_ludwig_model.py
@@ -858,7 +858,7 @@ class NewLudwigModel:
             dataset=dataset,
             data_format=data_format,
             training_set_metadata=self.training_set_metadata,
-            include_outputs=False,
+            include_outputs=True,
         )
 
         logger.debug('Predicting')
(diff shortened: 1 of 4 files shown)


ludwig/models/new_ludwig_model.py
```diff
@@ -858,7 +858,7 @@ def collect_activations(
             dataset=dataset,
             data_format=data_format,
             training_set_metadata=self.training_set_metadata,
-            include_outputs=False,
+            include_outputs=True,
```
w4nderlust, line 861: why do you need include outputs to be true when preprocessing for prediction when you collect the activations? That parameter is there for when you need to calculate eval losses, so I'm not sure of the need there.

tgaddair: This is needed because `batch_collect_activations` feeds that targets to the model to obtain the activations.  Without it, the targets cannot be constructed.  We could probably refactor it to not require the targets, but because this is the existing behavior on master as well (see: https://github.com/uber/ludwig/blob/master/ludwig/models/trainer.py#L939), I made the change that would have the least deviation from the current behavior.

w4nderlust: Makes sense. We should probably change that because I expect collect activations to work also in case we have only inputs (like collecting the last hidden representation of a test image from a pretrained image model). Will add a comment with a todo in the code to remind us of that.

PR #876, opened 2020-09-12
Fixed comet and wandb tests

tgaddair:
- Fixed `experiment_cli` to return experiment directory from `NewLudwigModel`.
- Fixed usage of `dataset` with image features.
- Fixed comparison to Pandas DF in `evaluate`.

### DIFF
--- a/ludwig/contribs/wandb.py
+++ b/ludwig/contribs/wandb.py
@@ -40,10 +40,10 @@ class Wandb():
                 "Ignored --wandb: Please install wandb; see https://docs.wandb.com")
             return None
 
-    def train_model(self, model, *args, **kwargs):
+    def train_model(self, model, model_definition, *args, **kwargs):
         import wandb
         logger.info("wandb.train_model() called...")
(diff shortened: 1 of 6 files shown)


ludwig/contribs/wandb.py
```diff
@@ -40,10 +40,10 @@ def import_call(*args, **kwargs):
                 "Ignored --wandb: Please install wandb; see https://docs.wandb.com")
             return None
 
-    def train_model(self, model, *args, **kwargs):
+    def train_model(self, model, model_definition, *args, **kwargs):
```
w4nderlust, line 43: model here is a NewLudwigModel instance right? then instead of having another parameter we can obtain the model definition with `model.model_definition` I believe

tgaddair: No, it's `ECD`.

w4nderlust: got it, then let's keep it the way it is for now. Maybe later we can consider switching model to an LudwigModel if it makes sense.

PR #884, opened 2020-09-14
TF2 fix experiment unit test

jimthompson5802:
# Code Pull Requests

Initial fix for `test_experiment.py`.  I'm still tracking down couple of more fixes.  This is just a check to confirm these work on `travis-ci`.


ludwig/api.py
```diff
@@ -655,25 +655,6 @@ def evaluate(
...
-            self.model_definition['input_features'],
-            True
-        )
-        if num_overrides > 0:
-            logger.warning(
-                'Using in_memory = False is not supported for Ludwig API.'
-            )
-
-        # preprocessing
-        dataset, training_set_metadata = preprocess_for_prediction(
```
tgaddair, line 670: Can you clarify why this part is removed?  Are we handling preprocessing somewhere else?

jimthompson5802: The were duplicated code in that section of `evaluate()` method.  You can see it here https://github.com/uber/ludwig/blob/463ce5f511a6fd85b8e5764e487220854b718c27/ludwig/api.py#L658-L696.  I removed the redundant code.

tgaddair: Ah gotcha.  Thanks!

PR #893, opened 2020-09-17
Fixed collect_activations to not include the outputs

tgaddair:


w4nderlust:
This is great, makes a lot of sense :)

PR #896, opened 2020-09-18
fix(strings_utils): bool 2 str

ifokeev:
Getting this error for `bool` types:

```
AttributeError: 'bool' object has no attribute 'lower’
```

This is the fix

### DIFF
--- a/ludwig/utils/strings_utils.py
+++ b/ludwig/utils/strings_utils.py
@@ -51,7 +51,7 @@ def strip_accents(s):
 
 
 def str2bool(v):
-    return v.lower() in ('yes', 'true', 't', '1')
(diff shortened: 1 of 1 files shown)


w4nderlust:
If it is a bool, then binary_features.py:65 should be false, that function should not be executed. Dan paste here an example dataframe that gives you tht error
Because this works fine:
```
>>> df = pd.DataFrame({'col': [True, False]})
>>> df
     col
0   True
1  False
>>> df.col
0     True
1    False
Name: col, dtype: bool
>>> df.col.dtype == object
False
```

ifokeev:
@w4nderlust I have no idea why this happening, but this fixed the issue. So I think that's not bad to do coercion anyway.
Thanks for the merge.

PR #904, opened 2020-09-19
Refactoring: single model_definition in LudwigModel init signature

w4nderlust:
Modified LudwigModel init signature to use model_definition for both dict and file

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -72,8 +72,7 @@ logger = logging.getLogger(__name__)
 
 class LudwigModel:
     def __init__(self,
-                 model_definition=None,
-                 model_definition_fp=None,
+                 model_definition,
                  logging_level=logging.ERROR,
                  use_horovod=None,
                  gpus=None,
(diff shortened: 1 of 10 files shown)


ludwig/api.py
```diff
@@ -72,8 +72,7 @@
 
 class LudwigModel:
     def __init__(self,
-                 model_definition=None,
-                 model_definition_fp=None,
+                 model_definition,
```
tgaddair, line 75: Nice!  I had thought about making this change, so glad you went ahead and did it!

ludwig/experiment.py
```diff
@@ -308,52 +312,31 @@ def cli(sys_argv):
     # Data parameters
     # ---------------
     parser.add_argument(
-        '--data_csv',
-        help='input data CSV file. If it has a split column, it will be used '
-             'for splitting (0: train, 1: validation, 2: test), otherwise the '
-             'dataset will be randomly split'
+        '--dataset',
```
tgaddair, line 315: We should probably add a test for this in `test_cli.py` at some point.

w4nderlust: Yes test_cli should test all of them, otherwise we will keep forgetting changing the cli parameters as they don't ger refactored automatically.

ludwig/utils/misc_utils.py
```diff
@@ -216,3 +216,20 @@ def get_file_names(output_directory):
...
+        raise ValueError(
+            'Either model_definition of model_definition_file have to be'
+            'not None to initialize a LudwigModel'
+        )
+    if model_definition is not None and model_definition_file is not None:
+        raise ValueError(
+            'Only one between model_definition and '
+            'model_definition_file can be provided'
+        )
+    if not model_definition:
```
tgaddair, line 233: If the user provided `model_definition` as a string on the command line, we need to load it as a dict, right?  Otherwise, how do we disambiguate the model filename from the model definition after this function is called?

w4nderlust: You are right but there's no need for it because the model_definition parameter in args has type `yaml.save_load` which parses the string as yaml automatically and the values of args.model_definition is the parsed dict

tgaddair: Ah, I see.  Makes sense!

tgaddair (approved):

PR #909, opened 2020-09-20
Added export.py

w4nderlust:
Added a script with two different export functions for SavedModel and Neuropod.
Still needs to be tested and tests need to be written for it.


tgaddair (approved):

tests/integration_tests/test_export.py
```diff
@@ -0,0 +1,82 @@
...
+from ludwig.export import export_savedmodel, export_neuropod
+from tests.integration_tests.utils import binary_feature
+from tests.integration_tests.utils import category_feature
+from tests.integration_tests.utils import generate_data
+from tests.integration_tests.utils import numerical_feature
+
+
+@pytest.mark.parametrize('export_type',
+                         ['savedmodel', 'neuropod'])
+def test_export(export_type, csv_filename, tmpdir):
```
w4nderlust, line 15: Can you refactor to have two separate tests? And place them in text_cli and use the _cli methods from export.py? Becuase we already know that the functions on their own work, testing the cli commands is want we wnat to make sure works fine (including the parsed args).

jimthompson5802:
Just pushed up unit tests for `export_savedmodel` and export_neuropod`.

Here is the run locally
```
=================================================== test session starts ===================================================
platform linux -- Python 3.6.9, pytest-6.0.2, py-1.9.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /opt/project
plugins: forked-1.3.0, pycharm-0.7.0, xdist-2.1.0, typeguard-2.9.1
collected 2 items

../../tests/integration_tests/test_export.py::test_export[savedmodel] PASSED                                        [ 50%]
../../tests/integration_tests/test_export.py::test_export[neuropod] PASSED                                          [100%]


```

tests/integration_tests/test_export.py
jimthompson5802: Do you still want the `test_export.py`?  Or should I remove it?

w4nderlust: We probably don't need it I believe.

PR #917, opened 2020-09-23
Added DistributedSampler to allow for deterministic shuffling by epoch

tgaddair:
This addresses failures in #916.

By making the shuffling occur over a fixed set of indices (and not in-place), we can deterministically shuffle by seed + epoch when resuming training from a previous saved state.

This PR incorporates a `DistributedSampler` to do this index shuffling, based on the implementation in PyTorch, which also removes the need for the separate `DistributedBatcher`.


jimthompson5802:
I can confirm that the MNIST example converges as expected with the changes in this PR.

tgaddair:
Thanks for verifying that, @jimthompson5802!

ludwig/data/sampler.py
```diff
@@ -0,0 +1,64 @@
...
+
+import math
+
+import numpy as np
+
+
+class DistributedSampler:
+    """Adapted from `torch.utils.data.distributed.DistributedSampler`."""
+    def __init__(self, dataset, shuffle=True, seed=0, horovod=None):
+        self.dataset = dataset
```
w4nderlust, line 26: It seems like we need the dataset just for using its length, maybe we can pass a dataset_length parameter in the signature instead?

w4nderlust (approved):

ludwig/data/sampler.py
tgaddair: Changed to `dataset_size`.

PR #925, opened 2020-09-28
Added batch prediction endpoint to FastAPI serving

tgaddair:
This change allows users to make multiple predictions in a single request when using FastAPI serving.

It introduces a new endpoint: `/batch_predict`.

Using this endpoint, requests are expected to be in a multi-part form.  Form elements are either files keyed by filename or a `dataset` field representing a JSON encoding of the dataset. 

The `dataset` is expected to be in the Pandas "split" format to reduce payload size.  This format divides the dataset into three parts:

```
{
    columns: List[str]
    index: List[Union[str, int]]
    data: List[List[object]]
}
```

### DIFF
--- a/ludwig/serve.py
+++ b/ludwig/serve.py
@@ -144,7 +144,7 @@ def convert_batch_input(form):
         if type(v) == UploadFile:
             file_index[v.filename] = _write_file(v, files)
 
-    data = json.loads(form['json'])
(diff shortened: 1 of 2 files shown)


w4nderlust:
@tgaddair can you also ad an example in the docs? There is already an example of theregular predict, just add another one for the batch please.

PR #930, opened 2020-09-29
[WIP\proposal] use hparam plugin in tensorboard

carlogrisetti:
I tried logging some of the hyperparameters used when training a model, along with the "usual" summaries that get written.
This enables a nice and cool feature in TensorBoard (yep, I'm always up for data analysis), that allows to have a quick comparison between models trained on the same dataset (hopefully) but with different hyperparameters.

This is only a proof-of-concept, and might not be so useful (well... it's not completely unuseful either, and it should only be polished prior to merging it), but it still has some potential. I'm currently using this to compare "big" changes in the overall parameters, but would like a lot to extend it to map also common parameters I tweak, such as resnet size, dropout value, fc_layer_sizes and so on.
I am currently fixed on the only measure of combined/epoch_loss, and I know that's suboptimal (there's a TODO exactly for that), and when present a combined/epoch_accuracy would be great too (but it hasn't been reimplemented yet).
Also, multiple measures can be added, cycling through all the various output features and tracking each epoch_loss as an output measure (along with epoch_accuracy when present, or mean_absolute_error, and so on... you get the reasoning).

This is the current result sample:
image
A detail shown when drilling down into the actual measures that resulted from that training
image
A graphical representation "cause-effect" like
image

Logging everything would clutter the interface (the menu on the left shows all the hyperparameters you have logged), and by design works by specifying the hyperparameter range (which is a think I don't want to do, that would force us in defining the hyperparameter range prior to training, as opposed to just logging and plotting the results).
I tought of everything I could in terms of having a way to automatically choose the best yaml parameters to track, but something that is *very* important to me might be absolutely of no use for someone else.

The only thing that could resolve this in my opinion is to have a yaml "flag" for each actual parameter that toggles hparam logging on\off, like below.
Do you have any ideas about this? Not being able to log\map\track the dropout of an input\output feature (for example) is a huge dealbreaker for me, since the parameters I "play" with the most during optimization are exactly those linked to the input features, and I also might want to just track some and not all of them.
Of course this would mean that you should track always the same parameter values, to be consistent in the output\analysis that would follow.

yaml sample:

input_features:
    -
        name: in
        type: numerical
        dropout: 0.2
                hparam_logging: true

output_features:
    -
        name: out
        type: numerical
        
preprocessing:
        force_split: true
        split_probabilities: [0.8, 0.2, 0.0]
        
training:
        batch_size: 5
                hparam_logging: true
        learning_rate: 0.01
                hparam_logging: true
        epochs: 1000
        early_stop: 1000
        reduce_learning_rate_on_plateau: 100
        reduce_learning_rate_on_plateau_patience: 5
                hparam_logging: true
        reduce_learning_rate_on_plateau_rate: 0.5
                hparam_logging: true
        learning_rate_warmup_epochs: 10

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -50,6 +50,8 @@ from ludwig.utils.math_utils import learning_rate_warmup, \
     learning_rate_warmup_distributed, exponential_decay
 from ludwig.utils.misc_utils import set_random_seed
 
+from tensorboard.plugins.hparams import api as hp
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


w4nderlust:
This is nice. You can achieve a similar visualization using hiplot in the visualization module, but having it integrated in the TensorBoard is cool too. let me thing of a way to make it more generic / compatible with the hyperopt we are doing now, but thanks for starting this!

carlogrisetti:
There are many native modules in tensorboard, and some way cooler graphical visualizations, but they often apply to a specific task/data type (like a running spectrogram visualization with epochs on the time (depth) axis and accuracy per category/domain on the x,y axis)

I hoped you would be interested in implementing this specific one :)

PR #939, opened 2020-10-03
Using isort Module to Optimize the Imports

lnxpy:
`isort` module is one of the most efficient ways to optimize importing libs, modules, files, and etc. It eventually sorts all those lines at the top of any script. I've chopped each directory in the several commits. In spite, there is no big change in the main algorithm, now, everything looks more pretty and clean.

[isort module official Github page](https://github.com/PyCQA/isort)

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -26,7 +26,7 @@ import logging
 import os
 import tempfile
 from pprint import pformat
-from typing import List, Union, Dict, Tuple, Optional
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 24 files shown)


w4nderlust:
Wasn't aware of this tool. I think the imports were in a decent shape because we use the pycharm import organizer, but this does a better job as sorting names and indenting stuff. My fear would be that in the future when new imports will be added we would need to remember to run tis (as the file watcher setup sounds pretty clunky and not all contributors would want to do it). Anyway, I'm merging and appreciate the PR

lnxpy:
Thanks. You can use it any time you want. Just add it to the requirements. Would you pleased add the "hacktoberfest" topic to this PR session? I really appreciate your attention.

Thanks again.

PR #941, opened 2020-10-04
Hyperopt Utils: replace strategy with sampler

mbzhu1:
- Seems like `hyperopt_config` isn't using "strategy" anymore, and is instead using "sampler"
- changed the `update_hyperopt_params_with_defaults(hyperopt_params)` function to reflect this


w4nderlust:
Thank you! can you also please replace the constant STRATEGY in constants.py with a SAMPLER constant and use it instead of the "strategy" string?

mbzhu1:
Yup, just changed it. Also `hyperopt.constants` module didn't have a sampler constant so I added one next to the hyperopt constants

PR #947, opened 2020-10-07
Add Ludwig Model Serve Example

jimthompson5802:
# Code Pull Requests

Add example for using Ludwig's model serve function
* Sample command to start Ludwig model server with a pre-trained model
* Python client program illustrates invoking the `/batch_predict` REST API.


examples/serve/client_program.py
```diff
@@ -0,0 +1,34 @@
...
+import pandas as pd
+
+# retrieve data to make predictions
+test_df = pd.read_csv('../titanic/data/test.csv')
+print('retrieved {:d} records for predictions'.format(test_df.shape[0]))
+
+# create json representation of dataset for REST API
+prediction_request_json = test_df.to_json(orient='split')
+
+# execute REST API /batch_predict
```
w4nderlust, line 11: What about showing also an example of calling /predict with the first row of the df before this one for /batch_predict? I think the single predict is the main one people will use, and that way this example would cover all use cases.

jimthompson5802: Added /predict example

PR #948, opened 2020-10-09
Export TFLite

carlogrisetti:
Adds export_tflite CLI command.
It currently does not work when exporting resnet models (I only tested those), but installing tf-nightly (2.4.0) solves the issue (they are actively working on improving model definition support when converting to TFLite).

I don't know if it's merge-worthy or not, since with the current TF version requirements most exports would fail. Also there could have been a way better file management when doing the intermediate savedmodel step.

Bye :)

### DIFF
--- a/ludwig/cli.py
+++ b/ludwig/cli.py
@@ -56,6 +56,7 @@ Available sub-commands:
    collect_weights       Collects tensors containing a pretrained model weights
    collect_activations   Collects tensors for each datapoint using a pretrained model
    export_savedmodel     Exports Ludwig models to SavedModel
+   export_tflite         Exports Ludwig models to TensorFlow Lite
    export_neuropod       Exports Ludwig models to Neuropod
    preprocess            Preprocess data and saves it into HDF5 and JSON format
    synthesize_dataset    Creates synthetic data for tesing purposes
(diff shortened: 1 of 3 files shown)


w4nderlust:
This is cool and definitely a feature we want to have, which I never got to implement, thank you a lot!

One thing I don't know is: what are the limitations? Are there models or layers that don't work at all with tflite?
Can you provide some pointers I can look at?

A side note, it would be great if you could have a test for this like the test for the regular savedmodel and the neuropod export ;)

carlogrisetti:
I wrote a long and descriptive comment... and then closed the browser :)
TL;DR:

this model_definit... hem... config works currently [config.txt](https://github.com/uber/ludwig/files/5353490/config.txt)

More complex models fail but I don't think due to missing support, might be bugs

Here's a reference to operator compatibility: https://www.tensorflow.org/lite/guide/ops_compatibility

Tests should be on simpler (like the one above) and complex (like a resnet one I found the issue with) models. There should however be a way to encapsulate errors and rethrow them in a warning for complex models, because it's just unfair to fail the build until the bugs are resolver for such a "niche" feature.
I'll try to conjure up a test for a simple in\out model export

carlogrisetti:
Added a pytest as copypasta of test_savedmodel... Of course no weights\results comparison can be made, and I also had to "simplify" the model switching back from a sequence input and category output to a numerical input and numerical output (I suppose due to the bugs\limitations I spoke about earlier).
It passes tests on my machine, let's see if it goes well on travis also.

It might be a good idea to add an *EXPERIMENTAL* tag in the CLI output, or something like that, just to note that it could (and will) fail sometimes :)

carlogrisetti:
This is not relevant anymore :)

PR #949, opened 2020-10-09
Initial commit of datasets API

skanjila:
# Code Pull Requests

Please provide the following:

- Added the datasets for reuters and ohsumed
- Added the base class that defines the keras like contract for download, process and load
- Added unit tests to test the above dataset functionality for reuters and ohsumed

# Documentation Pull Requests

This PR basically sets the stage for all future dataset contracts, it downloads the dataset and adds a version and stores the dataset to the .ludwig_cache directory and then converts the downloaded dataset into a simple python dictionary, the load function converts the dictionary into a pandas dataframe


ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,42 @@
...
+    where is is represented by the name.version of the dataset
+    args:
+        self (BaseDataset): A pointer to the current class
+        dataset_name: The name of the dataset we need to retrieve
+    """
+    @abc.abstractmethod
+    def download(self, dataset_name) -> None:
+        raise NotImplementedError("You will need to implement the download method to download the training data")
+
+    """Process the dataset to get it ready to be plugged into a dataframe
```
tgaddair, line 22: Instead of loading this into memory, with this function we want to map the raw dataset to the processed dataset in persistent storage.

So:

```
.ludwig_cache/id/raw -> .ludwig_cache/id/processed
```

For some datasets like CSV, this could be a no-op, but if the dataset is in a weird format, or needs to be sanitized in some way, this is where we would do it.

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+         self (BaseDataset): A handle to the current class
+     Returns:
+          A pandas DataFrame
+    """
+    def load(self) -> pd.DataFrame:
+       result = os.path.isfile(self.__dest_location)
+       if result != True:
+            self.download("ohsumed")
+       processed_result = self.process()
+       return pd.DataFrame(list(processed_result.items()), columns=['text', 'class'])
```
tgaddair, line 73: With a CSV, we can read directly into a Pandas DataFrame with [pd.read_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html).

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+
+    """Download the raw data to the ludwig cache in the format ~/.ludwig_cache/id
+    where is is represented by the name.version of the dataset
+    args:
+        self (BaseDataset): A pointer to the current class
+        dataset_name: The name of the dataset we need to retrieve
+    return:
+        true or false depending on whether the file exists in the new location
+    """
+    def download(self, dataset_name) -> bool:
```
tgaddair, line 34: The parent interface suggests this should not return anything.  I think having no return value makes more sense here.  If the dataset cannot be downloaded, we should raise an exception.  What do you think?

ludwig/datasets/text/versions.yaml
```diff
@@ -0,0 +1,3 @@
+text:
+  ohsumed: 1.0
```
tgaddair, line 2: We may want to store a checksum instead.  Then we we download, we can emit a warning if the dataset has changed.

tgaddair (commented):
Thanks for getting this put together so quickly!  I have thoughts around the preprocessing step, would be good to get your opinion @skanjila.

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,42 @@
...
+
+"""An abstract base class that defines a set of methods for download,
+preprocess and plug data into the ludwig training API"""
+
+
+class BaseDataset(metaclass=abc.ABCMeta):
+
+    """Download the raw data to the ludwig cache in the format ~/.ludwig_cache/id
+    where is is represented by the name.version of the dataset
+    args:
```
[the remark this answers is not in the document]
skanjila: Changed to follow that format

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+       in the manner needed by the ludwig training API
+       args:
+           self (BaseDataset): A handle to the current class
+           dict_reader (csv.DictReader): a pointer to a dictionary that can be read
+       :returns:
+           a dictionary containing KV pairs in the format of the training data
+    """
+    def process(self) -> Dict:
+        result = os.path.isfile(self.__dest_location)
+        if result != True:
```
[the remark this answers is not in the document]
skanjila: Simplified this logic

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __init__(self):
+        self.__initial_path = os.path.abspath(os.path.dirname(__file__))
+        self.__source_location =  os.path.join(self.__initial_path , "../text/ohsumed/ohsumed-allcats.csv")
+        self.__cache_location = os.path.join(self.__initial_path, "../../../.ludwig_cache/")
+        self.__config_file_path = os.path.join(self.__initial_path , "../text/versions.yaml")
+        with open(self.__config_file_path) as config_file:
+            self.__config_file_contents = yaml.load(config_file, Loader=yaml.FullLoader)
+        cur_version = self.__config_file_contents["text"]["ohsumed"]
+        self.__dest_location = self.__cache_location + "ohsumed_" + str(cur_version) + ".csv"
```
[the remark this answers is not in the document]
skanjila: Using path now

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+"""A class to process the ohsumed training data"""
+import yaml
+
+
+class OhsuMed(BaseDataset):
+
+    def __init__(self):
+        self.__initial_path = os.path.abspath(os.path.dirname(__file__))
```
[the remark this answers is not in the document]
skanjila: Changed all private variables to use one underscore to follow standard

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+        self.__source_location =  os.path.join(self.__initial_path , "../text/ohsumed/ohsumed-allcats.csv")
+        self.__cache_location = os.path.join(self.__initial_path, "../../../.ludwig_cache/")
```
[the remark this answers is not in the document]
skanjila: Added this

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Changed

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Changed all methods other than load to return nothing

ludwig/datasets/text/versions.yaml
skanjila: Lets get the basics working , then we can use the checksum instead of existence like the latest code

ludwig/datasets/base_dataset.py
skanjila: Fixed

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
+import csv
```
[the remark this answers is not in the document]
skanjila: Followed this format for all code, please check to make sure I hadnt forgotten places on this

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,74 @@
...
+
+"""A class to process the ohsumed training data"""
+import yaml
+
+
+class OhsuMed(BaseDataset):
+
+    def __init__(self):
+        self.__initial_path = os.path.abspath(os.path.dirname(__file__))
+        self.__source_location =  os.path.join(self.__initial_path , "../text/ohsumed/ohsumed-allcats.csv")
```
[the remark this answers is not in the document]
skanjila: No these datasets cannot, I did some research on both of these, for every dataset that can be downloaded using a URL, I'll write a simple utility to do this, stay tuned for a future PR on that once we get the basics in place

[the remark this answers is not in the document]
tgaddair: Can you explain the difference between the datasets included here and the ones here: https://www.mat.unical.it/OlexSuite/Datasets/SampleDataSets-download.htm.  If we can obtain them from this link, we should do that instead of including them in the repo.

If for whatever reason we cannot obtain them automatically via a script (e.g., we need the user to authenticate or something), then we should host them ourselves in something like S3 (which I can do using the Horovod S3 account).  If we go this route, we need to be extra certain that the datasets are in the public domain, and not subject to any copyrights preventing us from hosting them ourselves.

In general, I don't think we should host them in the repo itself, however, as datasets can get very large and add a lot of bloat to git pull / clone.

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,102 @@
...
+"""A class to download, process and return the ohsumed training data to be plugged into the Ludwig training API"""
+
+
+class OhsuMed(BaseDataset):
+
+    def __init__(self):
+        self._initial_path = os.path.abspath(os.path.dirname(__file__))
+        self._config_file_location = os.path.join(self._initial_path, "../text/versions.yaml")
+        self._source_location = os.path.join(self._initial_path, "../text/ohsumed/ohsumed-allcats.csv")
+        self._cache_dir = str(Path.home().joinpath('.ludwig_cache'))
```
tgaddair, line 34: This stuff we can move into `BaseDataset`.

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,102 @@
...
+from pathlib import Path
+import pandas as pd
+from ludwig.datasets.base_dataset import BaseDataset
+
+"""A class to download, process and return the ohsumed training data to be plugged into the Ludwig training API"""
+
+
+class OhsuMed(BaseDataset):
+
+    def __init__(self):
```
tgaddair, line 30: Cache location should be configurable.  If it's not set, then we can default to `~/.ludwig_cache`.

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,102 @@
...
+                for key, value in self._result_dict.items():
+                    writer.writerow([key, value])
+        except IOError:
+            print("I/O error")
+
+    """Now that the ohsumed data is processed load and return it as a pandas dataframe
+       Returns:
+          A pandas DataFrame
+    """
+    def load(self) -> pd.DataFrame:
```
tgaddair, line 93: Some of the logic of these methods we should move into the base class, specifically:

```
def preprocess(self):
    if not self.downloaded:
        self.download()

def load(self):
    if not self.preprocessed:
        self.preprocess()
```

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,102 @@
...
+        processed_data_path = Path.home().joinpath('.ludwig_cache') \
+            .joinpath("ohsumed_" + str(self._cur_version)).joinpath('processed.csv')
+        result = os.path.isfile(processed_data_path)
+        column_names = ["text","class"]
+        if result:
+            return pd.read_csv(processed_data_path, names=column_names)
+        else:
+            raise FileNotFoundError("The transformed data for ohsumed does not exist so cant return pandas dataframe")
+
```
tgaddair, line 102: I think it would be worth adding a convenience function in here so users don't need to construct this object if they just want to jump straight to the loaded data.  Something like:

```
def load(cache=DEFAULT_CACHE_DIR):
    dataset = OhsuMed(cache=cache)
    return dataset.load()
```

What do you think?

ludwig/datasets/ohsumed/ohsumed.py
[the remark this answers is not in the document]
skanjila: @tgaddair point taken let me explore this further, I didnt see this link, will keep you posted

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,86 @@
...
+import yaml
+import pandas as pd
+
+
+"""An abstract base class that defines a set of methods for download,
+preprocess and plug data into the ludwig training API"""
+
+
+class BaseDataset(metaclass=abc.ABCMeta):
+    def __init__(self, dataset_name):
```
tgaddair, line 29: Cache location should be configured by the user as opposed to the dataset.  Let's make it a keyword arg:

```
def __init__(self, dataset_name, cache=DEFAULT_CACHE_DIR)
```

Then we can define the default cache constant above the class definition in this file:

```
DEFAULT_CACHE_DIR = str(Path.home().joinpath('.ludwig_cache'))
```

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,93 @@
...
+            value_to_store = None
+            for row in dict_reader:
+                for key, value in row.items():
+                    if key == "class":
+                        value_to_store = value
+                    else:
+                        key_to_store = value
+                        self._result_dict[key_to_store] = value_to_store
+        else:
+            raise FileNotFoundError("The raw dataset for ohsumed was not found")
```
tgaddair, line 73: Instead of raising an exception, we can just call `self.download()` here.

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,93 @@
...
+    """Now that the ohsumed data is processed load and return it as a pandas dataframe
+       Returns:
+          A pandas DataFrame
+    """
+    def load(self) -> pd.DataFrame:
+        column_names = ["text", "class"]
+        if super().check_file_existence(self._cur_version, "ohsumed", "processed.csv"):
+            return pd.read_csv(self.processed_data_path, names=column_names)
+        else:
+            raise FileNotFoundError("The transformed data for ohsumed does not exist so cant return pandas dataframe")
```
tgaddair, line 91: Instead or raising an exception, we can just call `self.preprocess()` first.

ludwig/datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,93 @@
...
+        if not result:
+            raise FileNotFoundError("The raw data was not downloaded correctly")
+
+    """Process the ohsumed dataset to get it ready to be plugged into a dataframe
+       in the manner needed by the ludwig training API, to do this we create
+       a new dictionary that contains the KV pairs in the format that we need.
+       Returns:
+           None
+    """
+    def process(self) -> None:
```
tgaddair, line 61: To simplify code reuse and class hierarchy, I think we should separate the public interface from the dataset-specific implementation.  Specifically, we can have the public interface implement the "high level" actions and the dataset implement only dataset-internal details:

```
# in base class
def process(self):
    if not self.downloaded():
        self.download()
    self.process_downloaded_dataset()
```

Then in the child, we implement the "protected" `process_downloaded_dataset()` function.  What do you think?

ludwig/datasets/base_dataset.py
skanjila: Added

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Added

ludwig/datasets/ohsumed/ohsumed.py
skanjila: We redo the previous steps now

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Done, see the updates

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+class BaseDataset(metaclass=abc.ABCMeta):
+
+    def __init__(self, dataset_name, cache_location):
+        self._dataset_name = dataset_name
+        self._initial_path = os.path.abspath(os.path.dirname(__file__))
+        self._config_file_location = os.path.join(self._initial_path, "./config/dataset_config.yaml")
+        with open(self._config_file_location) as config_file:
+            self._config_file_contents = yaml.load(config_file, Loader=yaml.FullLoader)
+        if cache_location is not None:
```
tgaddair, line 43: Can simplify to:

```
self._cache_location = cache_location or DEFAULT_CACHE_LOCATION
```

ludwig/datasets/ohsumed/ohsumed.py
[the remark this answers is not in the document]
skanjila: The file download capability is added now.

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Moved

ludwig/datasets/ohsumed/ohsumed.py
skanjila: See the new logic

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Fixed

ludwig/datasets/ohsumed/ohsumed.py
skanjila: See the mixin logic

ludwig/datasets/ohsumed/ohsumed.py
skanjila: I will wait on this till the mixin functionality is put in

ludwig/datasets/ohsumed/ohsumed.py
skanjila: Again check the mixin logic

ludwig/datasets/transform_to_dataframe_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class TransformToDataframeMixin:
+
+    def __init__(self, dataset_name, cache_location):
+        super().__init__(dataset_name, cache_location)
+
+    def transform_processed_data_to_dataframe(self) -> pd.DataFrame:
+        column_names = ["text", "class"]
+        if self.check_file_existence(self._processed_file_name):
```
tgaddair, line 31: The `if/else` can be removed once we change the `load()` method to check `self.processed()`.

ludwig/datasets/transform_to_dataframe_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""A class whose responsibility it is to take in a csv file and convert it into
+any type of destination dataframe"""
+
+import pandas as pd
+
+
+class TransformToDataframeMixin:
+
+    def __init__(self, dataset_name, cache_location):
```
tgaddair, line 26: As with the other mixins, we don't need the initializer here.

ludwig/datasets/transform_to_dataframe_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License.
+# ==============================================================================
+
+"""A class whose responsibility it is to take in a csv file and convert it into
+any type of destination dataframe"""
+
+import pandas as pd
+
+
+class TransformToDataframeMixin:
```
tgaddair, line 24: Let's call this `PandasLoadMixin`.

ludwig/datasets/zip_download_workflow_mixin.py
```diff
@@ -0,0 +1,45 @@
...
+from zipfile import ZipFile
+
+
+"""A mixin to simulate downloading the zip file containing the training data
+and extracting the contents"""
+
+
+class ZipDownloadWorkflowMixin:
+
+    def __init__(self, dataset_name, cache_location):
```
tgaddair, line 29: As above, can remove this.

ludwig/datasets/zip_download_workflow_mixin.py
```diff
@@ -0,0 +1,45 @@
...
+from io import BytesIO
+from urllib.request import urlopen
+from zipfile import ZipFile
+
+
+"""A mixin to simulate downloading the zip file containing the training data
+and extracting the contents"""
+
+
+class ZipDownloadWorkflowMixin:
```
tgaddair, line 27: Suggest renaming to `ZipDownloadMixin`

tests/ludwig/datasets/test_ohsumed_dataset_workflow.py
```diff
@@ -0,0 +1,58 @@
...
+            .joinpath('processed.csv')
+
+    def test_download_success(self):
+        self._ohsu_med_handle_mock.download()
+        self._ohsu_med_handle_mock.download.assert_called_once()
+
+    def test_process_success(self):
+        self._ohsu_med_handle.process()
+        result = os.path.isfile(self._processed_data_path)
+        assert(result, True)
```
tgaddair, line 47: No need for parens on these asserts.

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+
+    """Download the file from config url that represents the raw unprocessed training data.
+       The workflow for this involves unzipping the file and renaming it to raw.csv, which means
+       keep trying to download the file till successful.
+    :arg:
+        None
+    :return
+        None"""
+    def download(self) -> None:
+        self.downloaded_raw_dataset()
```
tgaddair, line 64: `download_raw_dataset` should be an abstract method in this class.

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+           a new dictionary that contains the KV pairs in the format that we need.
+           If we fail we redownload the file
+           Returns:
+               None
+        """
+    def process(self) -> None:
+        if not self._is_downloaded():
+            self.download()
+        self.transform_downloaded_dataset()
+
```
tgaddair, line 85: `process_downloaded_dataset` should be an abstract method of this class.

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+    """Now that the ohsumed data is processed load and return it as a pandas dataframe
+       if we cant load the dataframe redo the whole workflow
+           :return
+              A pandas DataFrame
+        """
+    def load(self) -> pd.DataFrame:
+        return self.transform_processed_data_to_dataframe()
+
```
tgaddair, line 93: `load_processed_dataset` should be an abstract method of this class.

tgaddair (commented):
Looking good.  Several comments, but it's all very minor stuff.

One other thing I would like to change, is to avoid having a separate file for every mixin (necessary in Java, but not very common for Python).  Instead, we can have separate files for the load, process, and download mixins like:

```
datasets/download.py
datasets/process.py
datasets/load.py
```

What do you think?

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+        None"""
+    def download(self) -> None:
+        self.downloaded_raw_dataset()
+
+    """A helper method to verify the download
+    :arg
+        None
+    :return
+        True or false identifying whether the file has been downloaded"""
+    def _is_downloaded(self) -> bool:
```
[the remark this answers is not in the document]
skanjila: Made this public and changed all the files above

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+           in the manner needed by the ludwig training API, to do this we create
+           a new dictionary that contains the KV pairs in the format that we need.
+           If we fail we redownload the file
+           Returns:
+               None
+        """
+    def process(self) -> None:
+        if not self._is_downloaded():
+            self.download()
+        self.transform_downloaded_dataset()
```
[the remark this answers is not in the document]
skanjila: Changed

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+
+    """Now that the ohsumed data is processed load and return it as a pandas dataframe
+       if we cant load the dataframe redo the whole workflow
+           :return
+              A pandas DataFrame
+        """
+    def load(self) -> pd.DataFrame:
+        return self.transform_processed_data_to_dataframe()
```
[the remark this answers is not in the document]
skanjila: Changed

ludwig/datasets/csv_workflow_mixin.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+
+"""A mixin to convert a raw csv file into a processed dictionary which itself
+is also a csv file, the dictionary should be ready to be absorbed into a destination
+dataframe."""
+
+
+class CsvWorkflowMixin:
+
+    def __init__(self, dataset_name, cache_location):
```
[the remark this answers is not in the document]
skanjila: Removed from all mixins

ludwig/datasets/csv_workflow_mixin.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__(dataset_name, cache_location)
+
+    def transform_downloaded_dataset(self):
+        if self.check_file_existence(self._raw_file_name):
```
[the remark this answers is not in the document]
skanjila: Removed

ludwig/datasets/zip_download_workflow_mixin.py
skanjila: I actually made it even more generic by calling it DownloadMixin

ludwig/datasets/base_dataset.py
skanjila: Added

ludwig/datasets/base_dataset.py
skanjila: Added

ludwig/datasets/base_dataset.py
skanjila: Added

ludwig/datasets/download.py
```diff
@@ -0,0 +1,48 @@
...
+from io import BytesIO
+from urllib.request import urlopen
+from zipfile import ZipFile
+
+
+"""A mixin to simulate downloading the zip file containing the training data
+and extracting the contents"""
+
+
+class DownloadMixin:
```
tgaddair, line 27: Rename: `ZipDownloadMixin`

ludwig/datasets/download.py
```diff
@@ -0,0 +1,48 @@
...
+                zfile.extractall(self._download_dir)
+        # we downloaded the file, now check that this file exists
+        downloaded_file = self._download_dir.joinpath(self._dataset_file_name)
+
+        # rename the file to raw.csv
+        if os.path.isfile(downloaded_file):
+            os.rename(downloaded_file, self._raw_file_name)
+
+        # check for file existence and recursively and retry as needed
+        if not os.path.isfile(self._raw_file_name):
```
tgaddair, line 47: This part should not be necessary, process gets called as a separate step.

ludwig/datasets/load.py
```diff
@@ -0,0 +1,37 @@
...
+# limitations under the License.
+# ==============================================================================
+
+"""A class whose responsibility it is to take in a csv file and convert it into
+any type of destination dataframe"""
+
+import pandas as pd
+
+
+class LoadMixin:
```
tgaddair, line 24: Rename: `PandasLoadMixin`

ludwig/datasets/load.py
```diff
@@ -0,0 +1,37 @@
...
+
+    """This method converts a transformed data into a dataframe
+    args:
+    ret:
+        The pandas dataframe"""
+    def load_processed_dataset(self) -> pd.DataFrame:
+        column_names = ["text", "class"]
+        if self.check_file_existence(self._processed_file_name):
+            return pd.read_csv(self._processed_file_name, names=column_names)
+        else:
```
tgaddair, line 34: Can remove everything in the `else` portion along with the `if` check.

ludwig/datasets/process.py
```diff
@@ -0,0 +1,58 @@
...
+# ==============================================================================
+import csv
+import os
+
+"""A mixin to convert a raw csv file into a processed dictionary which itself
+is also a csv file, the dictionary should be ready to be absorbed into a destination
+dataframe."""
+
+
+class ProcessMixin:
```
tgaddair, line 25: Rename: `CSVProcessMixin`

ludwig/datasets/process.py
```diff
@@ -0,0 +1,58 @@
...
+                else:
+                    key_to_store = value
+                    self._result_dict[key_to_store] = value_to_store
+        try:
+            with open(self._processed_file_name, 'w') as csv_file:
+                writer = csv.writer(csv_file)
+                for key, value in self._result_dict.items():
+                    writer.writerow([key, value])
+        except IOError:
+            print("I/O error")
```
tgaddair, line 49: Don't swallow this exception.

ludwig/datasets/download.py
skanjila: Hey Travis,
The way I was thinking about this is we would potentially download other types of files and the naming for this file is to abstract the download pattern, this is why I called it DownloadMixin

skanjila: I went ahead and renamed

ludwig/datasets/download.py
skanjila: Removed

ludwig/datasets/load.py
skanjila: Renamed

ludwig/datasets/load.py
skanjila: Removed

ludwig/datasets/process.py
skanjila: Renamed

ludwig/datasets/process.py
skanjila: I reraised exception in latest PR

ludwig/datasets/process.py
```diff
@@ -0,0 +1,58 @@
...
+                    value_to_store = value
+                else:
+                    key_to_store = value
+                    self._result_dict[key_to_store] = value_to_store
+        try:
+            with open(self._processed_file_name, 'w') as csv_file:
+                writer = csv.writer(csv_file)
+                for key, value in self._result_dict.items():
+                    writer.writerow([key, value])
+        except IOError:
```
tgaddair, line 48: This approach will lose the stack trace.  Better to just remove the `try/except` entirely, so just keep:

```
with open(self._processed_file_name, 'w') as csv_file:
                writer = csv.writer(csv_file)
                for key, value in self._result_dict.items():
                    writer.writerow([key, value])
```

tests/ludwig/datasets/test_ohsumed_dataset_workflow.py
```diff
@@ -0,0 +1,58 @@
...
+            .joinpath('processed.csv')
+
+    def test_download_success(self):
+        self._ohsu_med_handle_mock.download()
+        self._ohsu_med_handle_mock.download.assert_called_once()
+
+    def test_process_success(self):
+        self._ohsu_med_handle.process()
+        result = os.path.isfile(self._processed_data_path)
+        assert result, True
```
tgaddair, line 47: Don't need the `, True` here, just: `assert result`

tests/ludwig/datasets/test_ohsumed_dataset_workflow.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_load_success(self):
+        self._ohsu_med_handle.process()
+        transformed_data = self._ohsu_med_handle.load()
+        # we test a random assortment of keys
+        first_key = "Laparoscopic treatment of perforated peptic ulcer. Mouret P  Francois Y  Vignal J  Barth X  Lombard-Platet R."
+        second_key = "Cuff size and blood pressure  letter  comment  Gollin S."
+        tmp = transformed_data['class'].where(transformed_data['text'] == first_key)
+        tmp1 = transformed_data['class'].where(transformed_data['text'] == second_key)
+        assert(tmp[0] == 'Neg-')
```
tgaddair, line 57: Remove parens.

tests/ludwig/datasets/test_reuters_dataset_workflow.py
```diff
@@ -0,0 +1,54 @@
...
+            .joinpath('processed.csv')
+
+    def test_download_success(self):
+        self._reuters_handle_mock.download()
+        self._reuters_handle_mock.download.assert_called_once()
+
+    def test_process_success(self):
+        self._reuters_handle.process()
+        result = os.path.isfile(self._processed_data_path)
+        assert result, True
```
tgaddair, line 47: `assert result`

tests/ludwig/datasets/test_reuters_dataset_workflow.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def test_load_success(self):
+        self._reuters_handle.process()
+        transformed_data = self._reuters_handle.load()
+        first_key = "2 NEW YORK BANK DISCOUNT WINDOW BORROWINGS 64 MLN DLRS IN FEB 25 WEEK Blah blah blah 3  "
+        tmp = transformed_data['class'].where(transformed_data['text'] == first_key)
+        assert (tmp[16] == 'Neg-')
```
tgaddair, line 54: Remove parens.

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,118 @@
...
+        with open(self._config_file_location) as config_file:
+            self._config_file_contents = yaml.load(config_file, Loader=yaml.FullLoader)
+        if cache_location is not None:
+            self._cache_location = cache_location
+        else:
+            self._cache_location = DEFAULT_CACHE_LOCATION
+        self._dataset_version = self._config_file_contents[dataset_name]["version"]
+        self._download_url = self._config_file_contents[dataset_name]["download_url"]
+        self._dataset_file_name = self._config_file_contents[dataset_name]["extracted_file_name"]
+        self._download_dir = Path.home().joinpath('.ludwig_cache').joinpath(self._dataset_name + "_"
```
tgaddair, line 47: Replace with `self._cache_location`:

```
self._download_dir = os.path.join(self._cache_location, f'{self._dataset_name}_{self._dataset_version}')
```

Same for other file names below.

Also, I would consider moving the `_raw_file_name` into the Download mixin, processed filename into Process mixin, etc.

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,118 @@
...
+                                                                         + str(self._dataset_version))
+        self._raw_file_name = Path.home().joinpath('.ludwig_cache').joinpath(self._dataset_name + "_"
+                                                                    + str(self._dataset_version)). \
+            joinpath('raw.csv')
+        self._processed_file_name = Path.home().joinpath('.ludwig_cache').joinpath(self._dataset_name + "_"
+                                                                         + str(self._dataset_version)). \
+            joinpath('processed.csv')
+        self._result_dict = {}
```
tgaddair, line 55: Can we remove this `self._result_dict`?

ludwig/datasets/download.py
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import os
+from io import BytesIO
+from urllib.request import urlopen
+from zipfile import ZipFile
+
+
+"""A mixin to simulate downloading the zip file containing the training data
```
tgaddair, line 23: Note on Python docstrings: doc strings in Python go below the class / function signature, like so:

```
class ZipDownloadMixin:
    """A mixin to ..."
    
    def download_raw_dataset(self):
        """Download the raw dataset and extract the contents ..."
```

ludwig/datasets/base_dataset.py
```diff
@@ -0,0 +1,118 @@
...
+           Returns:
+               None
+        """
+    def process(self) -> None:
+        if not self.is_downloaded():
+            self.download()
+        self.process_downloaded_dataset()
+
+    @abc.abstractmethod
+    def downloaded_raw_dataset(self):
```
tgaddair, line 97: Rename: `downloaded_raw_dataset` -> `download_raw_dataset`

ludwig/datasets/download.py
```diff
@@ -0,0 +1,45 @@
...
+and extracting the contents"""
+
+
+class ZipDownloadMixin:
+
+    """Download the raw dataset and extract the contents
+    of the zip file and store that in the cache location.
+    args:
```
tgaddair, line 31: No need to specify empty args and empty return value.

ludwig/datasets/load.py
```diff
@@ -0,0 +1,32 @@
...
+any type of destination dataframe"""
+
+import pandas as pd
+
+
+class PandasLoadMixin:
+
+    """This method converts a transformed data into a dataframe
+    args:
+    ret:
```
tgaddair, line 28: For return value, use this style:

```
:returns: a pandas DataFrame
```

ludwig/datasets/base_dataset.py
skanjila: Changed self._download_dir and moved mixin variables inside the mixins

ludwig/datasets/base_dataset.py
skanjila: removed

ludwig/datasets/download.py
skanjila: Fixed for all classes

ludwig/datasets/base_dataset.py
skanjila: changed

ludwig/datasets/download.py
skanjila: Fixed

ludwig/datasets/load.py
skanjila: Fixed

tests/ludwig/datasets/test_ohsumed_dataset_workflow.py
skanjila: removed

tests/ludwig/datasets/test_ohsumed_dataset_workflow.py
skanjila: fixed

ludwig/datasets/process.py
skanjila: fixed

PR #957, opened 2020-10-17
Adding feature ID logic

w4nderlust:


ludwig/data/preprocessing.py
```diff
@@ -940,7 +940,30 @@ def build_dataset(
 
 def build_metadata(dataset_df, features, global_preprocessing_parameters):
     metadata = {}
+    already_processed_features_names = set()
+    already_processed_features_preproc_params = {}
+
     for feature in features:
+
+        # if a feature has the same name of a previously processed one
```
tgaddair, line 948: An alternative way to think about this that might simplify the code would to construct a set of `(feature_name, preprocessing_params)` (to deduplicate), then only work with this set for the remainder of the preprocessing.

tests/integration_tests/utils.py
```diff
@@ -164,12 +172,14 @@ def set_feature(**kwargs):
...
     }
+    feature[ID] = feature[NAME]
+    feature[HASH] = compute_feature_hash(feature)
     feature.update(kwargs)
     return feature
 
 
 def sequence_feature(**kwargs):
-    seq_feature = {
+    feature = {
```
tgaddair, line 182: To clean this up, I think we should introduce a class called `FeatureDict` that derives from `dict`.  When you initialize it, it should pre-populate the correct ID and HASH values.  This also allows us to add more convenience methods to the `features` in the future.  It can also raise errors if certain fields like NAME are not set.

A better name might be `FeatureMetadata`.

ludwig/data/preprocessing.py
```diff
@@ -940,79 +941,92 @@ def build_dataset(
...
-        get_feature_meta = get_from_registry(
-            feature[TYPE],
-            base_type_registry
-        ).get_feature_meta
-
-        metadata[feature[NAME]] = get_feature_meta(
-            dataset_df[feature[NAME]].astype(str),
-            preprocessing_parameters
-        )
+            metadata[feature[HASH]] = get_feature_meta(
```
tgaddair, line 982: I see how this gets confusing.  One solution, at the cost of adding more keystrokes, would be to rename as follows:

```
NAME -> INPUT_COLUMN
HASH -> PROCESSED_COLUMN
```

I think that makes the relationship more clear.

ludwig/data/preprocessing.py
```diff
@@ -940,79 +941,92 @@ def build_dataset(
...
-        ).add_feature_data
-        add_feature_data(
-            feature,
-            dataset_df,
-            dataset,
-            training_set_metadata,
-            preprocessing_parameters
-        )
+
+        if HASH not in feature:
```
tgaddair, line 1007: See comment about `FeatureDict` below.

ludwig/features/base_feature.py
```diff
@@ -44,6 +45,17 @@ def __init__(self, feature, *args, **kwargs):
             raise ValueError('Missing feature name')
 
         self.feature_name = feature[NAME]
+
+        if ID not in feature:
```
tgaddair, line 49: See comment about `FeatureDict` below.

ludwig/utils/misc_utils.py
```diff
@@ -233,3 +236,12 @@ def check_which_config(config, config_file):
...
     return config
+
+
+def hash_dict(d: dict, max_length: int = 6) -> bytes:
+    s = json.dumps(d, sort_keys=True, ensure_ascii=True)
+    h = hashlib.md5(s.encode())
+    d = h.digest()
+    b = base64.b64encode(d)
+    return b[:max_length]
+    # return hash(json.dumps(d, sort_keys=True))
```
tgaddair, line 247: We should remove this line.

tgaddair (commented):
I think overall this structure is good, but what trips me up is just the naming.  My suggestion would be as follows:

1. We change `id` to `name`.  This way, feature name is the logical concept of the feature in Ludwig, which seems the most consistent with existing behavior.
2. The new optional `name` changes to `column` or `input_column` in the config. This way the user understands that this is an override so that multiple features with different "names" can point to the same column in the data table.
3. The `hash` changes to `processed_column`, so it's clear that this is referring specifically to a field in the processed dataset, and does not have an equivalent in the user config.

What do you think?

ludwig/data/preprocessing.py
```diff
@@ -940,79 +941,92 @@ def build_dataset(
...
-                feature[TYPE],
-                input_type_registry
-            ).encoder_registry
-            encoder_class = encoders_registry[feature['encoder']]
-            if hasattr(encoder_class, 'fixed_preprocessing_parameters'):
-                encoder_fpp = encoder_class.fixed_preprocessing_parameters
+        if HASH not in feature:
+            feature[HASH] = compute_feature_hash(feature)
+
+        if feature[HASH] not in metadata:
```
tgaddair, line 950: I'm wondering if it would simplify things if `metadata` only dealt with `feature[NAME]` instead of `feature[HASH]`.  We could have a separate map of `feature[HASH]: metadata` so we would not need to recompute.  Then the metadata lookups become much more straightforward (just use `feature[NAME]`).

tests/integration_tests/utils.py
w4nderlust: This is a good idea

ludwig/utils/misc_utils.py
w4nderlust: I realized too late that python hash is not stable :)

ludwig/data/preprocessing.py
w4nderlust: I f this mapping is inside metadata, then it's probably fine, otherwise bringing around a new dictionary will be kinda clunky.

tgaddair: No need to bring it along, it would just be for this loop.  So something like:

```
processed_feature_to_metadata = {}
for feature in features:
    if feature[PROC_COLUMN] not in processed_feature_to_metadata:
        processed_feature_to_metadata[feature[PROC_COLUMN]] = ...
    metadata[feature[NAME]] = processed_feature_to_metadata[feature[PROC_COLUMN]]          
```

Then for the rest of the code, the way we access `metadata[feature[NAME]]` doesn't need to change.

PR #959, opened 2020-10-18
Reduce run-time for Sequence Encoder/Generator unit test

jimthompson5802:
# Code Pull Requests

Summary of changes:
* Add Sequence encoder test that confirm correctness of generated output
* Add Generator decoder test that confirm correctness of generated output
* Reduce parameter combinations when generating test cases for Sequence Encoder/Sequence Generator decoder that execute `experiment_cli()`

Before and after results on testing this PR locally:

|Metrics For Entire Test Suite|Before PR|After PR|
|-----------|:-----------:|:------------:|
|Generated Tests|895|718|
|Elapsed Time (mm:ss)|33:15|27:38|
|Overall Test Coverage|81%|81%|

Note:

* Horvord test (`test_horovod.py`) fails because there is no local Horovod setup
* After the change one of the graph execution test fails (`test_graph_exection.py::test_sequence_generator`). The error occurs in the TFA `dynamic_decode` function.  The stack trace indicates the issue is around use of TFA custom ops.  Still investigating this failure.

Here is consolidate log file from testing the PR.
[_pytest_entire_tests_after_test_sequence_generator_change.txt](https://github.com/uber/ludwig/files/5398324/_pytest_entire_tests_after_test_sequence_generator_change.txt)

### DIFF
--- a/tests/integration_tests/test_sequence_features.py
+++ b/tests/integration_tests/test_sequence_features.py
@@ -13,6 +13,9 @@ from ludwig.utils.batcher import initialize_batcher
 from ludwig.data.dataset_synthesizer import build_synthetic_dataset
 from tests.integration_tests.utils import sequence_feature
 from tests.integration_tests.utils import ENCODERS
+from tests.integration_tests.utils import generate_data
+from tests.integration_tests.utils import run_experiment
+
 
 DEFAULT_HIDDEN_SIZE = 8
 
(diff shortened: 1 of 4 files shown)


jimthompson5802:
For the record, resolution for the tf-nightly issue was the tf-nightly version.  Error does not occur with version `tf-nightly==2.4.0.dev20201020`.

tests/integration_tests/test_sequence_features.py
```diff
@@ -0,0 +1,350 @@
...
+                assert encoder_out['encoder_output_state'].shape.as_list() \
+                       == [batch_size, TEST_HIDDEN_SIZE]
+
+        elif enc_encoder == 'stacked_cnn':
+            assert encoder_out['encoder_output'].shape.as_list() \
+                   == [batch_size, 1, TEST_HIDDEN_SIZE]
+
+        elif enc_encoder == 'transformer':
+            assert encoder_out['encoder_output'].shape.as_list() \
+                   == [batch_size, seq_size, 256]
```
w4nderlust, line 194: why here it's 256 instead of TEST_HIDDEN_SIZE ?

tests/integration_tests/test_sequence_features.py
```diff
@@ -0,0 +1,350 @@
...
+        encoder_out = encoder(tf.cast(inputs[input_feature_name], dtype=tf.int32))
+
+        # check encoder output for proper content, type and shape
+        batch_size = batch[input_feature_name].shape[0]
+        seq_size = input_features[0]['max_len']
+
+        assert 'encoder_output' in encoder_out
+        assert isinstance(encoder_out['encoder_output'], tf.Tensor)
+
+        if enc_encoder == 'parallel_cnn':
```
w4nderlust, line 140: one general consideration about the shapes of the outputs in all those if cases: the rank of the encoder_output  tensor depends on the reduce_output parameter, while encoder_output_state is not affected by it. not sure if we want to test for this too or not, but wnated to point it out if needed

tests/integration_tests/test_sequence_features.py
jimthompson5802: The reason for 256.  The `sequence_feature()` does not set a value for the `hidden_size`, which is the parameter used in Transformer encoder.  This means the encoder will take the default value of 256.   All the other encoders use the `fc_size`/`state_size` to set the hidden state size.  These are specified in `sequence_feature()` and set to 8.   So I use TEST_HIDDEN_SIZE for the other encoders.  

For clarity I can add a constant like `TEST_TRANSFORMER_HIDDEN_SIZE = 256`.

Did this make sense?

tests/integration_tests/test_sequence_features.py
jimthompson5802: I understand the point.   I know at this time the `reduce_output` and `reduce_input` are specified as `None`.   Over time, we should the add other parameters for testing the encoders/decoders and adjust the assertion test to account for those parameters.

I wanted to show at least one approach for setting up unit tests specifically for encoders and decoders.  Does this approach seem reasonable?

tests/integration_tests/test_sequence_features.py
w4nderlust: I think you could set the transformer hidden size and the embedding size both to be equal to TEST_HIDDEN_SIZE and that should address the iddue (and speed up the test involving the transformer ;) )

tests/integration_tests/test_sequence_features.py
w4nderlust: Yes the approach is definitely reasonable. I was pointing this out just so we know there are potentially other parameters that we are not testing for at the moment that influence the shape ;)

tests/integration_tests/test_sequence_features.py
jimthompson5802: updated `sequence_feature()` to set `hidden_size` to 8 and replaced hardcoded 256 with `TEST_HIDDEN_SIZE`

PR #970, opened 2020-10-25
Added Dask processor and Petastorm reader to train large datasets

tgaddair:
This pull request accomplishes the following goals:

1. Add support for Dask as a (mostly) drop-in replacement for Pandas during preprocessing
2. Add support for training on large out-of-memory datasets with Petastorm

Previous implementation of `Dataset` assumed all data was stored in an in-memory dictionary.  Previous implementation of `Batcher` assumed the dataset provided random access. Now we introduce separate datasets for in-memory datasets and Parquet-file datasets (processed with Petastorm), as well a separate batchers for random access and iterable datasets.


tgaddair:
Superceded by #1090.

PR #981, opened 2020-10-28
Fix visualization unit tests

jimthompson5802:
# Code Pull Requests

First try at fixing a few tests by changing reference to feature hash.


jimthompson5802:
Commits https://github.com/uber/ludwig/pull/981/commits/bbcf353b8ea7ac16df78f326288703e81e43a2fc and https://github.com/uber/ludwig/pull/981/commits/88efdf18a097433a44a589df069c306ed9d42cc4 supports using source data sets as ground truth and adds support for new `split_file` parameter on `ludwig visualize` command.  This parameter is the file path to a text file with the split values used in Ludwig's training and is used in creating visualizations.

Limitations with current state of the code:
* support reading only from csv files for ground truth.  Additional is planned to support full range of Ludwig data formats.
* refactor the visualization unit tests to support this new capability which should solve the errors that arose with the introduction of the feature id/hash capabilities.  
* add unit test for `split_file` support.

ludwig/data/preprocessing.py
```diff
@@ -1306,6 +1308,10 @@ def _preprocess_file_for_training(
             random_seed=random_seed
         )
 
+        # save split values for use by visualization routines
+        split_fp = os.path.splitext(dataset)[0] + '_split.txt'
```
w4nderlust, line 1312: maybe .csv?

ludwig/features/feature_utils.py
```diff
@@ -51,3 +55,32 @@ def set_str_to_idx(set_string, feature_dict, tokenizer_name):
...
+    # returns either the requested feature dictionary or None
+    for f in feature_list:
+        if f[NAME] == feature_name:
+            # return the requested feature dictionary
+            return f
+
+    # requested feature not found in the list
+    return None
+
+def retrieve_feature_hash(
```
w4nderlust, line 71: This should not be needed anymore right?

ludwig/visualize.py
```diff
@@ -24,18 +24,51 @@
...
+                                     read_excel, read_parquet,
+                                     read_pickle, read_fwf, read_feather,
+                                     read_html, read_orc, read_sas,
+                                     read_spss, read_stata,
+                                     figure_data_format_dataset)
+from ludwig.utils.misc_utils import get_from_registry
 
 logger = logging.getLogger(__name__)
 
+cacheable_data_format_reader_registry = {
```
w4nderlust, line 55: this should live in data utils probaly.

ludwig/visualize.py
```diff
@@ -147,6 +180,97 @@ def _validate_output_feature_name_from_test_stats(
...
+        np.array
+    """
+    return str2idx[raw]
+
+def _extract_ground_truth_values(
+        ground_truth: str,
+        output_feature_name: str,
+        ground_truth_split: int,
+        ground_truth_metadata: str,
+        split_file: Union[str, None]) -> np.ndarray:
```
w4nderlust, line 204: the return should not be a numpy array, but a pandas column i guess

tests/integration_tests/test_visualization.py
```diff
@@ -292,6 +293,10 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
                     ground_truth,
                     '--output_feature_name',
                     output_feature_name,
+                    '--split_file',
+                    split_file,
+                    '--ground_truth_metadata',
```
w4nderlust, line 298: would probably call it training_set_metadata or just metadata isntead of ground_truth_metadata

tests/integration_tests/test_visualization_api.py
```diff
@@ -92,14 +93,15 @@ def __init__(self, csv_filename):
         self.probability = predictions.iloc[:, 1:(num_probs + 2)].values
         self.ground_truth_metadata = self.model.training_set_metadata
         target_predictions = test_df[self.output_feature_name]
+        output_feature_hash = self.output_features[0][HASH]
         self.ground_truth = np.asarray([
-            self.ground_truth_metadata[self.output_feature_name]['str2idx'][
+            self.ground_truth_metadata[output_feature_hash]['str2idx'][
```
w4nderlust, line 98: this has been changed to be just NAME again ;)

ludwig/features/feature_utils.py
jimthompson5802: removed function

ludwig/visualize.py
jimthompson5802: I'd like to confirm the intent of the guidance.

I can make the change to return the pandas column.  This means that I'll have to move the encoding of string values to elsewhere.  I believe we will still need encoded numeric values.  Many of lower level visualization functions work on the encoded values, such as [here](
https://github.com/uber/ludwig/blob/0644620329aef02a4a9517386bc7839fd0192620/ludwig/visualize.py#L1079-L1103)
and [here](https://github.com/uber/ludwig/blob/0644620329aef02a4a9517386bc7839fd0192620/ludwig/visualize.py#L1165-L1198)

If the intent of the guidance is to move the translation of string to encode value outside of the `_extract_ground_truth_values()` function, I can make that change.  It just means duplicating a few lines of code that translates the string values to the encoded values wherever ground truth is needed.

OTOH, if the intent is to have the visualizations work only on the pandas raw string values, then this means more extensive changes are needed elsewhere in the visualization processing, which I'm not sure on the benefit.

w4nderlust: To make the single viz functions more reusable I'd probably make them accept pandas columns, and change their logic accordingly. I actually believe it would clean them and make tem simpler as it would not require mapping and unmapping values. And in the worst case scenario, one could just do the cast to numpy at the beginning of those functions. What about trying to do this for one function only, the compare performance from probs one for instance, and that would give us some indication if it's worth doing it for the other ones?

tests/integration_tests/test_visualization.py
jimthompson5802: Quick question...the `ludwig visualization` cli has a pre-existing parameter by the name of `ground_truth_metadata`
image

From your perspective, how does this `ground_truth_metadata` differ with `training_set_metadata`?

Since we are retrieving the prediction raw values from the predictions csv files, I need a way to translate the string values to the encoded numeric values since many of the visualization functions assume use of the encoded numeric representation.

ludwig/visualize.py
jimthompson5802: When I moved to the data_utils.py module, I noticed you already had a external reader registry built.  So I'm just reusing what you is already there instead of creating the `cacheable_data_format_reader_registry`.

ludwig/data/preprocessing.py
jimthompson5802: I can give it the `.csv` suffix.  Do you also want a header row ('split') or a headerless csv file is OK?

jimthompson5802: Actually, I think we should use a headerless csv file because this allows use of the `load_array()` function.

tests/integration_tests/test_visualization_api.py
jimthompson5802: reverted back to use NAME

ludwig/visualize.py
jimthompson5802: made modification.

jimthompson5802: I have a first working version of the modified viz functions that work with pandas columns directly.
Here is the modified [`compare_classifiers_performance_from_prob()`](https://github.com/jimthompson5802/ludwig/blob/fcba7649b476369796d7d3518e327603e4382617/ludwig/visualize.py#L1147-L1266).  This contains the fix for the inconsistency issues that I noted on the `master` branch.
This is all remains in [`compare_classifiers_performance_from_prob_cli()`](https://github.com/jimthompson5802/ludwig/blob/fcba7649b476369796d7d3518e327603e4382617/ludwig/visualize.py#L282-L324).  This is now just a minimal wrapper function now.  Maybe we can eliminate this.

Let me know what you think.  If this looks good, I'll do the same with the remaining viz functions.

tests/integration_tests/test_visualization.py
w4nderlust: I think it's just a name choice, we can stick to ground truth metadata for consistency.

ludwig/data/preprocessing.py
```diff
@@ -1312,6 +1314,10 @@ def _preprocess_file_for_training(
             random_seed=random_seed
         )
 
+        # save split values for use by visualization routines
+        split_fp = os.path.splitext(dataset)[0] + '_split.csv'
```
w4nderlust, line 1318: I would create a function in data utils, that given a dataset path return the split file path.
It will make it easy for other parts of the code ot obtain the correct path without knowing the actual logic.
Also, to be similar to what we do for metadata where se use `.meta.json` here we can use `.split.csv` as extension.

ludwig/features/feature_utils.py
```diff
@@ -14,14 +14,18 @@
...
 import numpy as np
+from typing import Union
 
 from ludwig.constants import SEQUENCE, PREPROCESSING, NAME
 from ludwig.constants import TEXT
 from ludwig.constants import TIMESERIES
 from ludwig.utils.misc_utils import hash_dict
 from ludwig.utils.strings_utils import UNKNOWN_SYMBOL
 from ludwig.utils.strings_utils import tokenizer_registry
+from ludwig.utils.data_utils import load_json
```
w4nderlust, line 28: these new imports are probably not needed, right? please run oranize imports to make sure ;)

ludwig/utils/data_utils.py
```diff
@@ -63,6 +63,7 @@
                                 STATA_FORMATS))
 
 
+
```
w4nderlust, line 66: also this return is probably not needed, please run reformat code to be sure.

tests/integration_tests/test_savedmodel.py
```diff
@@ -119,10 +119,12 @@ def test_savedmodel(csv_filename, should_load_model):
     restored_model = tf.saved_model.load(savedmodel_path)
 
     if_name = list(ludwig_model.model.input_features.keys())[0]
+    if_name_hash = ludwig_model.model.input_features[if_name].proc_column
```
w4nderlust, line 122: if_name_hash -> if_proc_column

ludwig/visualize.py
```diff
@@ -24,15 +24,19 @@
 import numpy as np
 import pandas as pd
 import sklearn
+import h5py
```
w4nderlust, line 27: this shouldn't be needed anymore

ludwig/visualize.py
```diff
@@ -147,6 +151,74 @@ def _validate_output_feature_name_from_test_stats(
...
+    :param split_file: (Union[str, None]) optional file path to split values.
+
+    # Return
+
+    :return pd.Series: ground truth values from source data set
+
+    """
+    # determine ground truth data format and get appropriate reader
+    data_format = figure_data_format_dataset(ground_truth)
+    if data_format not in CACHEABLE_FORMATS:
```
w4nderlust, line 192: I would say instead of returning an error, if the dataset is non cacheable (meaning it-s a DF) just do `gt_df = ground_truth`

ludwig/visualize.py
```diff
@@ -221,18 +296,43 @@ def compare_classifiers_performance_from_prob_cli(
...
+    ground_truth = _extract_ground_truth_values(
+        ground_truth,
+        output_feature_name,
+        ground_truth_split,
+        split_file
+    )
+
+    feature_metadata = metadata[output_feature_name]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
```
w4nderlust, line 325: These 2 lines should likely happen inside the viz function,, not outside. The viz functin should accept a pandas column ground truth. If this means that you also need to pass the metadata for it work, we can add the metadata parameter. in that case i would rather have both fature name and metadata as parameters.

ludwig/visualize.py
```diff
@@ -241,44 +341,72 @@ def compare_classifiers_performance_from_pred_cli(
...
+        output_feature_name,
+        ground_truth_split,
+        split_file
+    )
+
+    feature_metadata = metadata[output_feature_name]
+
+    # translate string to encoded numeric value
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
```
w4nderlust, line 386: Same as above

ludwig/visualize.py
```diff
@@ -291,26 +419,50 @@ def compare_classifiers_performance_subset_cli(
...
+    # retrieve ground truth from source data set
+    ground_truth = _extract_ground_truth_values(
+        ground_truth,
+        output_feature_name,
+        ground_truth_split,
+        split_file
+    )
+    feature_metadata = metadata[output_feature_name]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
```
w4nderlust, line 446: like above

ludwig/visualize.py
```diff
@@ -324,19 +476,38 @@ def compare_classifiers_performance_changing_k_cli(
...
```
w4nderlust, line 502: like above

ludwig/visualize.py
```diff
@@ -381,24 +554,47 @@ def compare_classifiers_predictions_cli(
...
```
w4nderlust, line 579: like above, although for this function I'm not sure you need any transformation into index at all.

ludwig/visualize.py
```diff
@@ -412,26 +608,47 @@ def compare_classifiers_predictions_distribution_cli(
...
+    ground_truth = _extract_ground_truth_values(
+        ground_truth,
+        output_feature_name,
+        ground_truth_split,
+        split_file
+    )
+
+    feature_metadata = metadata[output_feature_name]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
```
w4nderlust, line 634: like above

ludwig/visualize.py
```diff
@@ -445,26 +662,46 @@ def confidence_thresholding_cli(
...
+    # retrieve ground truth from source data set
+    ground_truth = _extract_ground_truth_values(
+        ground_truth,
+        output_feature_name,
+        ground_truth_split,
+        split_file
+    )
+    feature_metadata = metadata[output_feature_name]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
```
w4nderlust, line 687: like above

ludwig/visualize.py
```diff
@@ -479,26 +716,46 @@ def confidence_thresholding_data_vs_acc_cli(
...
```
w4nderlust, line 741: like above

ludwig/visualize.py
```diff
@@ -513,19 +770,37 @@ def confidence_thresholding_data_vs_acc_subset_cli(
...
```
w4nderlust, line 795: like above

ludwig/visualize.py
```diff
@@ -548,27 +824,43 @@ def confidence_thresholding_data_vs_acc_subset_per_class_cli(
...
```
w4nderlust, line 847: like above

ludwig/visualize.py
```diff
@@ -591,21 +886,36 @@ def confidence_thresholding_2thresholds_2d_cli(
...
         threshold_output_feature_names[0],
-        ground_truth_split
+        ground_truth_split,
+        split_file
     )
-    gt2 = load_from_file(
+    feature_metadata = metadata[threshold_output_feature_names[0]]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth0 = vfunc(ground_truth0, feature_metadata['str2idx'])
+
```
w4nderlust, line 902: llike above

ludwig/visualize.py
```diff
@@ -636,21 +951,36 @@ def confidence_thresholding_2thresholds_3d_cli(
...
         ground_truth,
         threshold_output_feature_names[0],
-        ground_truth_split
+        ground_truth_split,
+        split_file
     )
-    gt2 = load_from_file(
+    feature_metadata = metadata[threshold_output_feature_names[0]]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth0 = vfunc(ground_truth0, feature_metadata['str2idx'])
```
w4nderlust, line 966: like above

ludwig/visualize.py
```diff
@@ -672,26 +1004,45 @@ def binary_threshold_vs_metric_cli(
...
+    # retrieve ground truth from source data set
+    ground_truth = _extract_ground_truth_values(
+        ground_truth,
+        output_feature_name,
+        ground_truth_split,
+        split_file
+    )
+    feature_metadata = metadata[output_feature_name]
+    vfunc = np.vectorize(_encode_categorical_feature)
+    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
```
w4nderlust, line 1030: like above

ludwig/visualize.py
```diff
@@ -705,18 +1056,35 @@ def roc_curves_cli(
...
```
w4nderlust, line 1082: like above

ludwig/visualize.py
```diff
@@ -758,24 +1128,43 @@ def calibration_1_vs_all_cli(
...
```
w4nderlust, line 1154: liek above

ludwig/visualize.py
```diff
@@ -789,18 +1178,35 @@ def calibration_multiclass_cli(
...
```
w4nderlust, line 1204: like above

ludwig/visualize.py
```diff
@@ -1119,8 +1526,8 @@ def compare_classifiers_performance_from_prob(
 
 
 def compare_classifiers_performance_from_pred(
-        predictions_per_model: List[list],
-        ground_truth: np.array,
+        predictions_per_model: List[np.ndarray],
+        ground_truth: np.ndarray,
```
w4nderlust, line 1530: this is true everywhere:
both predictions and ground truth should be of the types that allow an API interface like this:
(pseudocode)
```
trainset, valset, testset, ... = moedl.train(dataset)
preds = model.predict(testset)
viz_function(ground_truth=testset[output_feature_name], predictions=[preds[output_feature_name][PREDICTIONS]])
```
or something that looks like this. So I'm not sure if ndarray is the correct one.

jimthompson5802: @w4nderlust   I want to confirm my understanding of the guidance.  Following is a mock up of how I would move code from _cli function to the related viz function.  Does this look right?

#### Currently a _cli and its related viz function looks like this
```
def compare_classifiers_performance_from_prob_cli(
        probabilities: Union[str, List[str]],
        ground_truth: str,
        ground_truth_split: int,
        split_file: str,
        ground_truth_metadata: str,
        output_feature_name: str,
        output_directory: str,
        **kwargs: dict
) -> None:

    # retrieve feature metadata to convert raw predictions to encoded value
    metadata = load_json(ground_truth_metadata)

    # translate string to encoded numeric value
    # retrieve ground truth from source data set
    ground_truth = _extract_ground_truth_values(
        ground_truth,
        output_feature_name,
        ground_truth_split,
        split_file
    )

    feature_metadata = metadata[output_feature_name]
    vfunc = np.vectorize(_encode_categorical_feature)
    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])

    probabilities_per_model = load_data_for_viz(
        'load_from_file', probabilities, dtype=float
    )

    compare_classifiers_performance_from_prob(
        probabilities_per_model,
        ground_truth,
        output_directory=output_directory,
        **kwargs
    )


def  compare_classifiers_performance_from_prob():
   #
   # code to create the visualization
   #
```

#### My interpretation of the guidance "both predictions and ground truth should be of the types that allow an API interface like this:" changes to the following:
```
def compare_classifiers_performance_from_prob_cli(
        probabilities: Union[str, List[str]],
        ground_truth: str,
        ground_truth_split: int,
        split_file: str,
        ground_truth_metadata: str,
        output_feature_name: str,
        output_directory: str,
        **kwargs: dict
) -> None:

    compare_classifiers_performance_from_prob(
        probabilities,
        ground_truth,
        ground_truth_split,
        split_file,
        ground_truth_metadata,
        output_feature_name,
        output_directory=output_directory,
        **kwargs
    )


def  compare_classifiers_performance_from_prob(
    probabilities: str,  # file path string
    ground_truth: pd.Series, # pandas Series for the ground truth
    ground_truth_split: int,
    split_file: str,  #file path to split file
    ground_truth_metadata: str,  #file path to training_meta_data.json 
    output_directory: str,  #directory to store the visualization
):

   # following functions moved from the _cli to the viz function
    # retrieve feature metadata to convert raw predictions to encoded value
    metadata = load_json(ground_truth_metadata)

    # translate string to encoded numeric value
    # retrieve ground truth from source data set
    ground_truth = _extract_ground_truth_values(
        ground_truth,
        output_feature_name,
        ground_truth_split,
        split_file
    )

    feature_metadata = metadata[output_feature_name]
    vfunc = np.vectorize(_encode_categorical_feature)
    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])

    probabilities_per_model = load_data_for_viz(
        'load_from_file', probabilities, dtype=float
    )
   # end of functions moved from _cli to viz function

   #
   # code to create the visualization
   #

```

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+        csv_filename
+):
+    """Ensure pdf and png figures from the experiments can be saved.
+
+    Probabilities are loaded from csv file.
+    :param csv_filename: csv fixture from tests.fixtures.filenames.csv_filename
+    :return: None
+    """
+    input_features = [
+        text_feature(vocab_size=10, min_len=1, representation='sparse'),
```
w4nderlust, line 343: remove the text feature, kep only the category one as it is much faster for test purposes, and reduce it's vocab to 2.

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+        category_feature(vocab_size=10)
+    ]
+    output_features = [category_feature(vocab_size=2, reduce_input='sum')]
+
+    # Generate test data
+    rel_path = generate_data(input_features, output_features, csv_filename)
+    input_features[0]['encoder'] = 'parallel_cnn'
```
w4nderlust, line 350: remove

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+
+    config = {
+        'input_features': input_features,
+        'output_features': output_features,
+        'combiner': {
```
w4nderlust, line 355: the combiner is superflous, remove

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+        _,
+        preprocessed_data,
+        exp_dir_name
+    ) = model.train(dataset=rel_path)
+
+    _, _, test_set, metadata = preprocessed_data
+
+    model.predict(
+        dataset=test_set,
+        skip_save_predictions=False,
```
w4nderlust, line 377: Set skip_save_predictions to True and remov exp_dir_name

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+
+    model = LudwigModel(
+        config,
+        logging_level=logging.WARN
+    )
+
+    (
+        _,
+        preprocessed_data,
+        exp_dir_name
```
w4nderlust, line 370: exp_dir_name -> _

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+    ) = model.train(dataset=rel_path)
+
+    _, _, test_set, metadata = preprocessed_data
+
+    model.predict(
```
w4nderlust, line 375: collect the outputs of predict

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+        dataset=test_set,
+        skip_save_predictions=False,
+        output_directory=exp_dir_name
+    )
+
+    output_feature_name = config['output_features'][0][NAME]
+    probability = os.path.join(exp_dir_name, '{}_probabilities.csv').format(
```
w4nderlust, line 382: obtain the probabilities from the tensors obtain from predict instead of reading them from disk

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+        output_feature_name)
+    experiment_source_data_name = csv_filename.split('.')[0]
+    split_file = get_split_path(csv_filename)
+
+    ground_truth = _extract_ground_truth_values(
+        rel_path,
+        output_feature_name,
+        ground_truth_split=2,
+        split_file=split_file
+    )
```
w4nderlust, line 392: not needed, you already have the test set from the output of training

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
...
+
+    probabilities_per_model = load_data_for_viz(
+        'load_from_file', [probability, probability], dtype=float
+    )
```
w4nderlust, line 396: no need to load from file, you already have them in memory

tests/integration_tests/test_visualization.py
```diff
@@ -326,6 +330,94 @@ def test_visualization_compare_classifiers_from_prob_csv_output_saved(
             print("Error: %s - %s." % (e.filename, e.strerror))
 
 
+def test_visualization_compare_classifiers_from_prob_csv_output_saved_api(
```
w4nderlust, line 333: run this using a temp file context

tests/integration_tests/test_visualization.py
jimthompson5802: default for skip_save_predictions is True, removed `output_directory=exp_dir_name`

tests/integration_tests/test_visualization.py
jimthompson5802: done

tests/integration_tests/test_visualization.py
jimthompson5802: @w4nderlust   I'd like to discuss this item.

Based on an earlier discussion, we talked about the viz function would be the place to translate the categories raw string values to the encoded numeric values.  See highlighted section.
image

This current method works with how the _cli function works.   

So to make the viz function support both the api and cli test functions, I'll have to modify the cli test function to do the translation of categorical raw string values to the encoded numeric value.

I just want to confirm that this is OK because, if I recall correctly, the location for this function has changed several times between  within the cli or within the viz functions.

w4nderlust: i see, you are probably right, before the data outputted by train was unprocessed, now it is processed. So, what i propose to do is the following: in the viz (as in non cli) function i will add a check like:
```
if not isinstance(ground_truth, Dataset):
  # do the mapping string to int
else:
  # it's already ints
```
So the type for `ground_truth` should be `Union[pd.Series, Dataset]` or something like that.
Does ths make sense? Sorry for the back and forth about this issue, but I'm also figuring out the best API as we go :) But I think with this change the viz function will become waaay more useable.

jimthompson5802: Implemented the recommended test

tests/integration_tests/test_visualization.py
jimthompson5802: done

tests/integration_tests/test_visualization.py
jimthompson5802: done

tests/integration_tests/test_visualization.py
jimthompson5802: done

tests/integration_tests/test_visualization.py
jimthompson5802: done

ludwig/visualize.py
jimthompson5802: removed

ludwig/visualize.py
jimthompson5802: With the last pushes, this should be taken care of.

tests/integration_tests/test_savedmodel.py
jimthompson5802: no longer using proc_column

ludwig/features/feature_utils.py
jimthompson5802: removed

jimthompson5802:
With the last set of commits, I've corrected the `test_visuazliation_api.py` test suite. 

Both the `test_visualization.py` and `test_visualization_api.py` complete successfully
```
root@da123aa80cd0:/opt/project/sandbox/pytest_area# pytest /opt/project/tests/integration_tests/test_visualization.py /opt/project/tests/integration_tests/test_visualization_api.py
platform linux -- Python 3.6.9, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir: /opt/project
plugins: typeguard-2.10.0, forked-1.3.0, pycharm-0.7.0, xdist-2.1.0
collected 50 items

../../tests/integration_tests/test_visualization.py ..........................                                                       [ 52%]
../../tests/integration_tests/test_visualization_api.py ........................                                                     [100%]

<<<<<<<<removed warning messages >>>>>>>>>>>
=============================================== 50 passed, 235 warnings in 202.49s (0:03:22) ===============================================

```

PR #985, opened 2020-11-01
Added SST2 datasets and made modifications to GoEmotions

ANarayan:
Added sst2 dataset. Fixed bug in GoEmotions dataset.


ludwig/datasets/goemotions/__init__.py
```diff
@@ -34,4 +34,17 @@ class GoEmotions(UncompressedFileDownloadMixin, MultifileJoinProcessMixin, CSVLo
...
+        if filetype == 'tsv':
+            file_df = pd.read_table(os.path.join(self.raw_dataset_path, filename), header=None)
+        return file_df
+
+    def process_downloaded_dataset(self):
+        super(GoEmotions, self).process_downloaded_dataset()
+        # format emotion ids to be a set of emotion ids vs. string
+        processed_df = pd.read_csv(os.path.join(self.processed_dataset_path, self.csv_filename))
+        processed_df.columns = ['text', 'emotion_ids', 'comment_id', 'split']
+        processed_df['emotion_ids'] = processed_df['emotion_ids'].str.split(",")
```
w4nderlust, line 49: this makes the value a list, but how does this list end up been written in the csv?
Usually set features assume space separated strings, so you may do a `" ".join(list)` to have such string if useful.

ludwig/datasets/goemotions/__init__.py
```diff
@@ -34,4 +34,17 @@ class GoEmotions(UncompressedFileDownloadMixin, MultifileJoinProcessMixin, CSVLo
...
 
     def __init__(self, cache_dir=DEFAULT_CACHE_LOCATION):
-        super().__init__(dataset_name="goemotions", cache_dir=cache_dir)
\ No newline at end of file
+        super().__init__(dataset_name="goemotions", cache_dir=cache_dir)
+
+    def read_file(self, filetype, filename):
+        if filetype == 'tsv':
+            file_df = pd.read_table(os.path.join(self.raw_dataset_path, filename), header=None)
+        return file_df
```
w4nderlust, line 42: if filetype is not tsv, file_df is not initialized.
If this function is local and used only in this class and you know the files are all tsv you probably don-t need the if, otherwise provide and else.

ludwig/datasets/goemotions/__init__.py
ANarayan: got it. made modifications in new commit

PR #990, opened 2020-11-03
Limit numpy\h5py max versions due to tensorflow 2.3.1 max supported versions

carlogrisetti:
Numpy 1.19.4 has an issue on windows which prevents it from running correctly (numpy package refers to this URL https://tinyurl.com/y3dm3h86).
I just tested with 1.19.3 and the issue is not present yet

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 Cython>=0.25
 h5py>=2.6,<2.11.0
-numpy>=1.15,<1.19.0
+numpy>=1.15 ; sys_platform != "Windows"
+numpy>=1.15,<1.19.0 ; sys_platform == "Windows"
 pandas>=0.19
 scipy>=0.18
 tabulate>=0.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 Cython>=0.25
-h5py>=2.6,<2.11.0
+h5py>=2.6 ; sys_platform != "Windows"
+h5py>=2.6,<2.11.0 ; sys_platform == "Windows"
 numpy>=1.15 ; sys_platform != "Windows"
 numpy>=1.15,<1.19.0 ; sys_platform == "Windows"
 pandas>=0.19
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 Cython>=0.25
-h5py>=2.6
-numpy>=1.15,<=1.19.3
+h5py>=2.6,<2.11.0
+numpy>=1.15,<1.19.0
 pandas>=0.19
 scipy>=0.18
 tabulate>=0.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 Cython>=0.25
 h5py>=2.6
-numpy>=1.15
+numpy>=1.15,<=1.19.3
 pandas>=0.19
 scipy>=0.18
 tabulate>=0.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,6 @@
 Cython>=0.25
-h5py>=2.6 ; sys_platform != "Windows"
-h5py>=2.6,<2.11.0 ; sys_platform == "Windows"
-numpy>=1.15 ; sys_platform != "Windows"
-numpy>=1.15,<1.19.0 ; sys_platform == "Windows"
+h5py>=2.6,<2.11.0
+numpy>=1.15,<1.19.0
 pandas>=0.19
 scipy>=0.18
 tabulate>=0.7
--- a/requirements_serve.txt
+++ b/requirements_serve.txt
@@ -2,4 +2,4 @@ uvicorn
 fastapi
 pydantic
 python-multipart
-neuropod
+neuropod ; sys_platform != "Windows"
--- a/requirements_serve.txt
+++ b/requirements_serve.txt
@@ -2,4 +2,4 @@ uvicorn
 fastapi
 pydantic
 python-multipart
-neuropod ; sys_platform != "Windows"
+neuropod ; platform_system != "Windows"


requirements.txt
```diff
@@ -1,6 +1,6 @@
 Cython>=0.25
-h5py>=2.6
-numpy>=1.15
+h5py>=2.6,<2.11.0
+numpy>=1.15,<1.19.0
```
tgaddair, line 3: Should this be platform-specific?

```
numpy>=1.15 ; sys_platform == "Windows"
numpy>=1.15,<1.19.0 ; sys_platform != "Windows"
```

carlogrisetti: Oh well, didn't even know it was possible!
That should be done for neuropod too, since it's not available on windows and fails installation of ludwig[full]

tgaddair: Yes, I agree, that's a good idea.

requirements.txt
```diff
@@ -1,6 +1,8 @@
 Cython>=0.25
-h5py>=2.6
-numpy>=1.15
+h5py>=2.6 ; sys_platform != "Windows"
```
tgaddair, line 2: Is the platform check needed for `h5py`?  Is this related to https://github.com/h5py/h5py/issues/1732?

w4nderlust: I actually added it because of the numpy dependency, but maybe it's overzealous of me (pip should resolve the h5py version that better suits the windows numpy constraint). Wasn't aware of that issue. I believe that shouldn't be an issue as we are saving preprocessed data and not model weights in h5, but the bug behind the reason why weights saving is broken may pop up somewhere in Ludwig too.

carlogrisetti: Actually that constraint should be not platform specific: tensorflow 2.3.1 requires that constraint independent of platform (and I don't know why pip doesn't know that from the beginning and only tells that in the end)

I found out about this with a `pip install ludwig --upgrade --upgrade-strategy eager`

Might be a bug / missing feature on pip, which should correctly map the support/requirements map of each package, but it upgrades both numpy and hdf5 beyond what's supported by tensorflow 2.3.1, giving a warning in the end.
I thought the quickest solution was to limit the version in requirements

carlogrisetti: The platform constraint should be set only for neuropod, since it's a package that's not available under windows, and makes a `pip install ludwig[full]` fail, also ludwig[serve] fails for the same issue

carlogrisetti:
Let's try to recap:
The focus switched from having a lower maximum version for numpy on windows because the latest one (1.19.4) is bugged on windows, to a lower again version (<1.19.0) due to a warning during tensorflow 2.3.1 installation, which tells that the supported numpy version is <1.19.0.
h5py did not have any windows related bug, I added the upper limit always to comply with tensorflow 2.3.1 maximum supported version.

The issue is not the windows bug anymore, since tensorflow 2.3.1 requires a much lower version of numpy, and we can use that limit to fix both issues.

In order to do that we should also modify the requirements as per my last commit.

Is there something i'm missing?

tgaddair:
@carlogrisetti, not sure about "win32", my suggestion was taken directly from the Python standard: https://www.python.org/dev/peps/pep-0508/#environment-markers

carlogrisetti:
platform_system is way better, since it generalizes based on architecture. Fixed that, thanks!

requirements.txt
w4nderlust: It's weird though, because without those constraints it looks like on unix and mac machines there is no issue, so i guess pip does it's job at figuring out the correct versions for those platforms.

carlogrisetti: It seems to be perfectly OK if I run a `pip install ludwig`, but it breaks without the constraints if I run a `pip install ludwig --upgrade --upgrade-strategy eager`

Limiting the max version might only be a workaround for a fix on the pip side (if this is even considered an issue and not a default behavior when trying to install direct dependencies to their most up to date version, not considering the cascade of constraints).

w4nderlust: I see, my understanding is that that's just the way the eager strategy works (see this example thread: https://github.com/pypa/pip/issues/5804 ). In your case probably the only_if_needed strategy would be better, despite not quite solving the issue, which I believe is "update dependencies to the latest compatible version" which is currently not done in pip (see this issue opened in 2013 https://github.com/pypa/pip/issues/988 ).

In conclusion I believe the best solution is:
```
h5py>=2.6
numpy>=1.15 ; platform_system != "Windows"
 numpy>=1.15,<1.19.0 ; platform_system == "Windows"
```
do you both agree @carlogrisetti and @tgaddair ?

carlogrisetti: Why do that on numpy and not h5py? They suffer from the same issue.
So from my understanding neither of those get updated outside TF boundaries by running `pip install ludwig --upgrade --upgrade-strategy eager` in linux and other OSs, and this is only a bug on windows?

If so, yes, this is the simplest fix for now (adding h5py)

Didn't want to spin up such a discussion on something this trivial :)

w4nderlust: No worries, it's good to have those discussions.
The reson why I suggest only numpy is:

- I believe the version of h5py will be determined by TF and numpy version anyway, so setting the numpy one to be platform dependent should be sufficient.
- I wouldn't want to optimize for things to work under the eager update strategy but just for installation from scratch, as the behavior of the eager strateg is not the desired one not because of wrong dependency constraints, but because it just ignores the mutual version dependencies of the various packages it's upgrading. When updating to a new ludwig version, I would actually suggest to create an ad hoc new virtualenv rather than upgrading an existing one, because of the mutual dependencies between ludwig dependencies.

Hope my reasoning is clear enough :)

carlogrisetti: It is clear. In that case you shouldn't need to edit anything (apart from neuropod), since dependencies are correctly computed during install (will have to double-check that, but I think they are ok), and you can let pip figure out the ceiling version.
The numpy bug on windows that initially throw me into this is past the maximum supported version in tensorflow (due to the eager update), so if we're not taking eager update into account we can just forget this PR and only fix the neuropod on windows requirement.

Let me check for a correct `pip install ludwig` behavior on requirements under windows, just to be sure

carlogrisetti: Ok nevermind, `pip install ludwig` installs the most up to date version of numpy, regardless of TF requirements, and results in this:
`tensorflow 2.3.1 requires h5py<2.11.0,>=2.10.0, but you'll have h5py 3.1.0 which is incompatible.`
`tensorflow 2.3.1 requires numpy<1.19.0,>=1.16.0, but you'll have numpy 1.19.4 which is incompatible. `

At this point, your proposal on constraining the version only on windows only works if this issue is only present on windows. If linux\other correctly expand the dependencies and figure out that one of the dependent packages (tensorflow) has a more strict requirement than the base package (ludwig), it's ok. If this issue is also present on linux\others, I stand with my current code edit, limiting the uppermost version directly in ludwig's own requirements.txt

Another way would be to remove the dependency on numpy and h5py alltogether, and rely on the chained dependency of tensorflow, making it choose the most appropriate version for itself.
For example, TF 2.3.1's numpy must be >=1.16.0, whilist Ludwig's requirement is >=1.15.0 (and we could potentially have the same issue in the lower end of the spectrum, having pip thinking it's all ok because numpy's version is 1.15.0, while tensorflow 2.3.1 requirement would be >= 1.16.0).

I know it's a little bit tangled of an explanation, but I hope it's clear :)

w4nderlust: Just tested on Mac:
```
Collecting h5py>=2.6
  Downloading h5py-3.1.0-cp38-cp38-macosx_10_9_x86_64.whl (2.9 MB)
     |████████████████████████████████| 2.9 MB 26.0 MB/s
Collecting numpy>=1.15
  Downloading numpy-1.19.4-cp38-cp38-macosx_10_9_x86_64.whl (15.3 MB)
     |████████████████████████████████| 15.3 MB 9.6 MB/s
tensorflow 2.3.1 requires h5py<2.11.0,>=2.10.0, but you'll have h5py 3.1.0 which is incompatible.
tensorflow 2.3.1 requires numpy<1.19.0,>=1.16.0, but you'll have numpy 1.19.4 which is incompatible.
```

With additional suggestion:

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default
```

I tried following the suggestion of using `--use-feature=2020-resolver` and I got:

```
Collecting h5py>=2.6
  Downloading h5py-2.10.0-cp38-cp38-macosx_10_9_x86_64.whl (3.0 MB)
     |████████████████████████████████| 3.0 MB 2.6 MB/s
Collecting numpy>=1.15
  Downloading numpy-1.18.5-cp38-cp38-macosx_10_9_x86_64.whl (15.1 MB)
     |████████████████████████████████| 15.1 MB 3.9 MB/s
```

Which is indeed the behavior one would expect from the dependency resolution. I can test on linux too, and on windows, but I believe it's not a platform issue.

Now te question is what do we want to do about this? My take is currently the following: for the time being I would be ok with putting the same constraints of TF 2.3.1 on both numpy and h5py, although this is not very sustainable moving forward, as we would have to keep track of them every single minor release of TF. So what I would do is to monitor when pip 20.3 is released and remove the constraints as soon as that happens. Should have been release in October, so I believeit won't be long.

What do you think @carlogrisetti and @tgaddair ?

carlogrisetti: That sounds good for someone that keeps its system maintained... The user that had an open issue recently on #1000 (heh, what a milestone), was using pip 8.x for example, and although I hope it will be a sporadic aituation, many current and slightly previous pip versions will be around for a lot of time.

The question is: do we want to ease the life of older pips, or reply to anyone having this issue to "enlarge your pip"?

Why did you dismiss the "remove numpy/h5py requirement and just let TF choose its own versions" option I proposed? (It's an actual genuine question) Do we have some narrower requirements for Ludwig itself? (I don't think so)

w4nderlust: That's a good point, people using older pips may keep on posting issues. On the other hand the reson for avoiding this is maintainability. But you are probably right, we can just follow whatever restrictions tensorflow has and work with that, but that would make being compatible with multiple version of tensorflow like we do now (2.2 2.3 and nightly if you look at the tests) more difficult.

PR #999, opened 2020-11-06
Fix typo on CLI output

carlogrisetti:
most likely a copy and paste error that resulted in the wrong variable being used in the info message


w4nderlust:
Thanks!

PR #1001, opened 2020-11-08
feat: integrate ray tune to hyperopt

kaushikb11:
## Description

The motivation for this Pull Request is to integrate Ray Tune with Ludwig's Hyperopt design. This will support the `samplers` provided by Ray Tune by using the `RayTuneExecutor`.

Further steps would be Resource Allocation, support for Search Algorithms & Trial Schedulers, tests & docs.

#### Example Config for RayTuneExecutor

```
hyperopt:
  sampler:
    type: ray
    search_alg:
      type: bayesopt
	num_samples: 2
  executor:
    type: ray
    cpu_resources_per_trial: 2
	gpu_resources_per_trial: 1
  parameters:
    training.learning_rate:
      space: uniform
      lower: 0.001
      upper: 0.1
    combiner.num_fc_layers:
      space: qrandint
      lower: 2
      upper: 6
      q: 2
    utterance.cell_type:
      space: grid_search
      values: [rnn, gru, lstm]
  goal: minimize
```


ludwig/hyperopt/ray.py
```diff
@@ -0,0 +1,133 @@
...
+                output_directory="results",
+                gpus=None,
+                gpu_memory_limit=None,
+                allow_parallel_threads=True,
+                use_horovod=None,
+                random_seed=default_random_seed,
+                debug=False,
+                **kwargs):
+
+        self.experiment_kwargs = dict(
```
kaushikb11, line 97: This will be replaced by [tune.with_parameters](https://docs.ray.io/en/master/tune/api_docs/execution.html#tune-with-parameters) functionality in ray v1.1.0

ludwig/hyperopt/execution.py
```diff
@@ -585,6 +588,103 @@ def execute(
...
+                output_directory="results",
+                gpus=None,
+                gpu_memory_limit=None,
+                allow_parallel_threads=True,
+                use_horovod=None,
+                random_seed=default_random_seed,
+                debug=False,
+                **kwargs):
+
+        self.experiment_kwargs = dict(
```
kaushikb11, line 649: This will be replaced by [tune.with_parameters](https://docs.ray.io/en/master/tune/api_docs/execution.html#tune-with-parameters) functionality in ray v1.1.0

tgaddair:
Sweet!

cc @richardliaw

richardliaw:
this is awesome! I'll take a look later.

ludwig/hyperopt/execution.py
```diff
@@ -585,6 +588,103 @@ def execute(
...
+        self.output_feature = output_feature
+        self.metric = metric
+        self.split = split
+        self.goal = goal
+        self.trial_id = 0
+        self.experiment_kwargs = {}
+
+    def _run_experiment(self, config):
+
+        self.trial_id += 1
```
richardliaw, line 606: This won't work since the function passed into `tune.run` will be duplicated and executed on multiple processes (therefore, `self` will not reference the same object)

kaushikb11: Ahh, @richardliaw is there a way to enumerate the processes and have it accessible inside the function?

ludwig/hyperopt/execution.py
```diff
@@ -585,6 +588,103 @@ def execute(
         return hyperopt_results
 
 
+class RayTuneExecutor(HyperoptExecutor):
+    def __init__(self, hyperopt_sampler, output_feature, metric, split, goal, **kwargs):
+        ray.init(ignore_reinit_error=True)
```
richardliaw, line 593: Tune already does this, I think you can skip this step?

ludwig/hyperopt/execution.py
richardliaw: Hmm, it might be difficult. Just curious, what are you trying to achieve by doing this?

ludwig/hyperopt/execution.py
richardliaw: Actually should be available in 1.0.1 :) which should be available now!

kaushikb11: Perfect! Code updated.

ludwig/hyperopt/execution.py
kaushikb11: The model and statistics are saved in a directory `[output_dir]/[experiment_name]_[model_name]_trial_id` where all variables are resolved to user-specified ones and `trial_id` is an increasing number starting from 0 used to differentiate among repeated runs.

richardliaw: Ah, I see. Does `trial_id` need to be numerical? In Tune, we have a unique identifier for trial_id but it is a hex value.

kaushikb11: Oh, that's the convention we have been following. @w4nderlust Would like your take on this?

w4nderlust: The hex may work fine. It's basically just a way to avoid saving to the same directory when running on the same filesystem or on a shared filesystem. Unique hexes would achieve the same goal ;)

ludwig/hyperopt/execution.py
```diff
@@ -585,6 +588,101 @@ def execute(
...
+        trial_id = tune.get_trial_id()
+        modified_config = substitute_parameters(
+            copy.deepcopy(hyperopt_dict["config"]), config)
+        hyperopt_dict["config"] = modified_config
+        hyperopt_dict["experiment_name"] = f'{hyperopt_dict["experiment_name"]}_{trial_id}'
+
+        train_stats, eval_stats = run_experiment(**hyperopt_dict)
+        metric_score = self.get_metric_score(eval_stats)
+
+        tune.report(parameters=str(config), metric_score=metric_score,
```
richardliaw, line 613: hmm Tune already automatically logs the config

ludwig/hyperopt/execution.py
richardliaw: Nice!

ludwig/hyperopt/execution.py
```diff
@@ -585,6 +588,101 @@ def execute(
...
+            output_directory=output_directory,
+            gpus=gpus,
+            gpu_memory_limit=gpu_memory_limit,
+            allow_parallel_threads=allow_parallel_threads,
+            use_horovod=use_horovod,
+            random_seed=random_seed,
+            debug=debug,
+        )
+
+        analysis = tune.run(tune.with_parameters(
```
richardliaw, line 677: are there parameters for configuring the number of evaluations (i.e., num_samples)? 
Also, it'd be nice to set the `resources_per_trial` of `tune.run`. This would allow Tune to allocate GPUs correctly. 

Finally, I'm interested in this `use_horovod` mechanism; what would it take to get that working?

tgaddair: We're going to be implementing the `use_horovod` support in a follow-up PR.  As part of the Dask / Ray integration, we're changing this API so that `use_horovod` gets replaced with a `Backend` interface.

richardliaw: Sounds good - yeah, i think this entire comment can be left to future work.

ludwig/hyperopt/execution.py
w4nderlust: So you are suggesting to remove `parameters=str(config)`?

ludwig/hyperopt/execution.py
w4nderlust: > are there parameters for configuring the number of evaluations (i.e., num_samples)?

They are actually in the sampler rather than in the exectuor. So far we tried to keep the two consepts separated.

ludwig/hyperopt/execution.py
kaushikb11: @richardliaw The intention here was to emulate the results of other `HyperoptExecutors` with the key `parameters` to have a similar interface for the further steps. 
@w4nderlust

richardliaw: @w4nderlust yeah, though I guess it doesn't hurt to keep it (as @kaushikb11 suggests that there's some later steps that require it)

ludwig/hyperopt/execution.py
kaushikb11: > are there parameters for configuring the number of evaluations (i.e., num_samples)?

Yes, I have added it to the sampler. The `sampler` will be returning the `search_space` and `num_samples`. This way, we could have the `sampler` and `executor` decoupled.

kaushikb11: @richardliaw I have a question regarding Resource Allocation for `resources_per_trial`. We get the parameters `gpus` and `gpu_memory_limit` from the User. `gpus` is GPUs to use (it uses the same syntax of CUDA_VISIBLE_DEVICES) and `gpu_memory_limit` is the maximum memory in MB to allocate per GPU device. 
Is there a way to allocate resources using `gpu_id` (CUDA_VISIBLE_DEVICES Syntax) for `tune.run`?Also, these params will be further passed to the experiment for the [Initialize Tensorflow](https://github.com/uber/ludwig/blob/master/ludwig/utils/tf_utils.py#L48) functionality.

ludwig/hyperopt/sampling.py
```diff
@@ -15,17 +15,19 @@
...
 # ==============================================================================
 import copy
+import inspect
 import itertools
 import logging
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Iterable, List, Tuple
 
 import numpy as np
+from ray import tune
```
w4nderlust, line 25: Can this import live inside the function that uses tune?

ludwig/hyperopt/execution.py
```diff
@@ -2,18 +2,21 @@
 import multiprocessing
 import signal
 from abc import ABC, abstractmethod
+from typing import Union
+
+import ray
+from ray import tune
```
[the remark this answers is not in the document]
tgaddair: +1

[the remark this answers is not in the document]
kaushikb11: Yup, thought about it. But if I import it in the `__init__` of the Ray Executor, it wouldn't be accessible to the class methods and the worker function.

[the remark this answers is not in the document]
richardliaw: ah, you could do things like:

```
try:
   import ray
except ImportError:
   ray = None
```
and then in the executor:
```
def __init__(self):
    if ray is None: raise ImportError
```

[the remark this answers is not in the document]
kaushikb11: Uff, thanks. Resolved

PR #1006, opened 2020-11-12
Cache checksum

w4nderlust:
Adds a checksum in the training set metadata to check if the hdf5 cache is still valid

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1192,7 +1192,8 @@ def preprocess_for_training(
                     logger.info(
                         "Found hdf5 and meta.json with the same filename "
                         "of the dataset, but checksum don't match, "
-                        "so they will be overridden"
+                        "if saving of processed input is not skipped "
+                        "they will be overridden"
                     )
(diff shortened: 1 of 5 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1348,8 +1357,13 @@ def _preprocess_file_for_training(
             logger.info('Writing preprocessed dataset cache')
             data_hdf5_fp = replace_file_extension(dataset, 'hdf5')
             data_utils.save_hdf5(data_hdf5_fp, data, training_set_metadata)
-            training_set_metadata[DATA_TRAIN_HDF5_FP] = data_hdf5_fp
+
             logger.info('Writing train set metadata')
+            training_set_metadata[DATA_TRAIN_HDF5_FP] = data_hdf5_fp
+            training_set_metadata[CHECKSUM] = calculate_checksum(
+                training_set,
```
jimthompson5802, line 1364: instead of specifying `training_set` should be `dataset` because this section of code is executed if `dataset=` parameter is specified on the train().

PR #1008, opened 2020-11-14
Refactored SST classes in 1 abstract one and 2 concrete ones for SST-2 and SST-5 + fixed in split in MNIST

w4nderlust:


ludwig/datasets/goemotions/__init__.py
```diff
@@ -37,13 +38,20 @@ def __init__(self, cache_dir=DEFAULT_CACHE_LOCATION):
...
     def process_downloaded_dataset(self):
         super(GoEmotions, self).process_downloaded_dataset()
         # format emotion ids to be a set of emotion ids vs. string
-        processed_df = pd.read_csv(os.path.join(self.processed_dataset_path, self.csv_filename))
+        processed_df = pd.read_csv(os.path.join(self.processed_dataset_path,
+                                                self.csv_filename))
         processed_df.columns = ['text', 'emotion_ids', 'comment_id', 'split']
-        processed_df['emotion_ids'] = processed_df['emotion_ids'].apply(lambda e_id: " ".join(e_id.split(",")))
-        processed_df.to_csv(os.path.join(self.processed_dataset_path, self.csv_filename), index=False)
+        processed_df['emotion_ids'] = processed_df['emotion_ids'].apply(
```
skanjila, line 51: be careful with lambda functions, they're difficult to debug in production mostly because you cant sometimes see whats happening inside them, for example in the scenario where e_id doesnt contain a comma, is that something that could happen?

ludwig/datasets/mixins/load.py
```diff
@@ -38,13 +38,20 @@ def load_processed_dataset(self, split) -> Union[pd.DataFrame,
...
-            val_set = data_df[data_df[SPLIT] == "1"]
-            test_set = data_df[data_df[SPLIT] == "2"]
-            return training_set, test_set, val_set
+            if SPLIT in data_df:
+                training_set = data_df[data_df[SPLIT] == 0]
+                val_set = data_df[data_df[SPLIT] == 1]
+                test_set = data_df[data_df[SPLIT] == 2]
+                return training_set, test_set, val_set
+            else:
+                raise ValueError("The dataset does not have splits, "
```
skanjila, line 53: if we're here whats the ask from the user, do they have to re-create the dataset?  I am wondering if we should have some explicit instructions on what they need to do?

ludwig/datasets/mnist/__init__.py
```diff
@@ -58,8 +57,11 @@ def process_downloaded_dataset(self):
         os.makedirs(self.processed_temp_path, exist_ok=True)
         for dataset in ["training", "testing"]:
             print(f'>>> create ludwig formatted {dataset} data')
-            labels, data = self.read_source_dataset(dataset, self.raw_dataset_path)
-            self.write_output_dataset(labels, data, os.path.join(self.processed_temp_path, dataset))
+            labels, data = self.read_source_dataset(dataset,
```
skanjila, line 60: do we need to do any checks with both labels and/or data before calling write_output_dataset?

ludwig/datasets/sst2/__init__.py
```diff
@@ -14,19 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-from ludwig.datasets.base_dataset import BaseDataset, DEFAULT_CACHE_LOCATION
-from ludwig.datasets.mixins.download import ZipDownloadMixin
-from ludwig.datasets.mixins.load import CSVLoadMixin
-from ludwig.datasets.mixins.process import *
+from ludwig.datasets.base_dataset import DEFAULT_CACHE_LOCATION
+from ludwig.datasets.sst2.sst_utils import SST
```
skanjila, line 18: what is SST, should we rename to something more descriptive :)

ludwig/datasets/sst2/__init__.py
```diff
@@ -50,91 +47,10 @@ class SST2(ZipDownloadMixin, MultifileJoinProcessMixin, CSVLoadMixin,
...
-            final_csv = pd.DataFrame(pairs)
-            final_csv.columns = ['sentence', 'label']
-            final_csv.to_csv(os.path.join(self.raw_dataset_path,
-                                          f'{split_name}.csv'),
-                             index=False)
-
-        super(SST2, self).process_downloaded_dataset()
+    def get_sentiment_label(self, id2sent, phrase_id):
+        sentiment = id2sent[phrase_id]
+        if sentiment <= 0.4:  # negative
```
skanjila, line 52: we should put 0.4 and 0.4 in a constants file or make them configs

ludwig/datasets/sst2/sst_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            return formatted_sent
+
+        def get_sentence_idxs(split):
+            return set(
+                datasplit_df[
+                    datasplit_df['splitset_label'] == split
+                    ]['sentence_index']
+            )
+
+        def get_sentences(sentences_idxs):
```
skanjila, line 96: could you add some contextual comments to this method mostly so that someone has a big picture on what this code is trying to accomplish before reading any code

ludwig/datasets/goemotions/__init__.py
tgaddair: Python handles these scenarios pretty gracefully.  If the input does not have a comma, it returns a list with one element containing the entire string, which is what we want.

ludwig/datasets/mixins/load.py
tgaddair: Good question.  In this case, they need to set `split=False` to return the entire dataset, then either manually split themselves or let Ludwig automatically assign splits (this will happen automatically if the user passes in `mode.train(dataset=dataset)`).

w4nderlust: Thought about it and decided for the ValueError for the following reasons:

- users will call `train, test, val = load(split=True)` so returning a single dataset would anyway raise an error, so better be explicit about it.
- the datasets that we are providing here either have official splits or they don't. In case they don't I'd rather have either the user come up with a split of having them trusting Ludwig with the splits, so that the dataset interface task is just to provide datasets as they are, without any surprise.

ludwig/datasets/sst2/__init__.py
w4nderlust: It's the acronym for Stanford Sentiment Treebank, which is a bit long... :)

ludwig/datasets/sst2/__init__.py
w4nderlust: they are a fixed convention of the construction of this specific variant of the SST dataset. We could make them constants, but they won't change so probably there's no need to make them configurable.

ludwig/datasets/sst2/sst_utils.py
w4nderlust: Good idea, @ANarayan maybe you could add come comments for all those internal functions to clarify what they are doing?

ludwig/datasets/mnist/__init__.py
w4nderlust: Probably not, @ANarayan could tell you more

ludwig/datasets/sst2/sst_utils.py
ANarayan: Yes @w4nderlust, I will add comments!

Issue #648, opened 2020-03-06
[Feature Request] Allow using custom languages/models for spaCy NLP


ZeroAurora:
**Is your feature request related to a problem? Please describe.**
Other related issues: #408 #251 
I trained a Chinese model for spaCy, linked it to `[spacy's package folder]/data/zh` (using `spacy link`) and want to use that for ludwig. However, when I tried to set the config for ludwig, I received an error, which tell me that there is no way to load the Chinese model.
```
ValueError: Key chinese_tokenizer not supported, available options: dict_keys(['characters', 'space', 'space_punct', 'underscore', 'comma', 'untokenized' (...) 'bert'])
```

**Describe the use case**
By allowing using custom languages for spacy, users using other language would be able to process their texts quicker and easier.

**Describe the solution you'd like**
Here's the current solution...
```yaml
input_features:
  -
    name: input
    type: text
    preprocessing:
      word_tokenizer: english_tokenize
```
...which I think could be changed to this...
```yaml
input_features:
  -
    name: input
    type: text
    preprocessing:
      word_tokenizer: spacy_tokenize
      spacy_model: zh #(or en, xx, etc.)
```

**Describe alternatives you've considered**
I've considered not to use spacy but to use a custom script to simply split sentences to words using some processors like "jieba". However, by using this method I would lose nearly all benefits from NLP.

**Additional context**
I think that's all :)
I don't know whether my advice could be accepted. But if it got solved I would be very thankful. 
BTW since I'm not a native English speaker, there may be some mistakes. Please don't mind it :p

w4nderlust:
We are thinking about changing the way you define the tokenizer to be more flexible, and that would allow you to do what you are looking for.

In the mean time, if you are using the API, you can do the following:
```
from ludwig.utils.nlp_utils import language_module_registry
from ludwig.utils.strings_utils import tokenizer_registry
from ludwig.utils.strings_utils import BaseTokenizer

language_module_registry['zh'] = 'your_model_name'  # for example 'en_core_web_sm'

class ChineseTokenizer(BaseTokenizer):
    def __call__(self, text):
        return process_text(text, load_nlp_pipeline('zh'))

tokenizer_registry['chinese_tokenizer'] = ChineseTokenizer
```

After you do this, you can refer to `chinese_tokenizer` in your model configuration within the same script where you run the code above.

w4nderlust:
[...]

PR #1012, opened 2020-11-18
feat: Add `pretrained_tokenizer_fast` configuration for HFTokenizer

brightsparc:
##  Description

Please provide the following:
- As per [discussion](https://github.com/uber/ludwig/issues/648#issuecomment-729239220) suggested adding support for the `use_fast` argument passed to the Hugging Face [AutoTokenizer](https://huggingface.co/transformers/model_doc/auto.html#transformers.AutoTokenizer)

```python
   self.tokenizer = AutoTokenizer.from_pretrained(
        pretrained_model_name_or_path, use_fast=True
    )
```

Validated that I was able to pass the following argument to `create_vocabulary`

```python
>>> from ludwig.utils.strings_utils import create_vocabulary
>>> create_vocabulary('abc',tokenizer_type='hf_tokenizer',
pretrained_model_name_or_path='/opt/ml/code/transformer',
pretrained_tokenizer_fast=True)
```
I tested with pulling my fork to pip install ludwig, and train a model with the following config.yml

```yaml
input_features:
    -   name: review_body
        type: text
        level: word
        encoder: bert
        preprocessing:
            pretrained_tokenizer_fast: true
        pretrained_model_name_or_path: /opt/ml/code/transformer
output_features:
    -   name: star_rating
        type: category
```

## Docs

I am not clear on how to update the `docs`.  The update would be to the following page:
https://ludwig-ai.github.io/ludwig-docs/user_guide/#text-features-preprocessing

  * pretrained_tokenizer_fast (default false): if the string has to be lowercased before being handled by the tokenizer.

### DIFF
--- a/ludwig/features/text_feature.py
+++ b/ludwig/features/text_feature.py
@@ -46,6 +46,7 @@ class TextFeatureMixin(object):
         'char_most_common': 70,
         'word_tokenizer': 'space_punct',
         'pretrained_model_name_or_path': None,
+        'pretrained_tokenizer_fast': False,
         'word_vocab_file': None,
         'word_sequence_length_limit': 256,
         'word_most_common': 20000,
@@ -94,7 +95,9 @@ class TextFeatureMixin(object):
             unknown_symbol=preprocessing_parameters['unknown_symbol'],
             padding_symbol=preprocessing_parameters['padding_symbol'],
             pretrained_model_name_or_path=preprocessing_parameters[
-                'pretrained_model_name_or_path']
+                'pretrained_model_name_or_path'],
+            pretrained_tokenizer_fast=preprocessing_parameters[
+                'pretrained_tokenizer_fast']
         )
         return (
             char_idx2str,
--- a/ludwig/utils/strings_utils.py
+++ b/ludwig/utils/strings_utils.py
@@ -96,7 +96,8 @@ def create_vocabulary(
         vocab_file=None,
         unknown_symbol=UNKNOWN_SYMBOL,
         padding_symbol=PADDING_SYMBOL,
-        pretrained_model_name_or_path=None
+        pretrained_model_name_or_path=None,
+        pretrained_tokenizer_fast=False,
 ):
     vocab = None
     max_line_length = 0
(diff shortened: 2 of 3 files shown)


brightsparc:
Note, I got a message in my logs - is the fork something that can be avoided, not sure if this is in my code?

```
  | 2020-11-18T15:12:02.388+11:00 | 2020-11-18 04:12:01.735590: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudnn.so.7
  | 2020-11-18T15:12:04.388+11:00 | huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
  | 2020-11-18T15:12:04.388+11:00 | To disable this warning, you can either:
  | 2020-11-18T15:12:04.388+11:00 | #011- Avoid using `tokenizers` before the fork if possible
  | 2020-11-18T15:12:04.388+11:00 | #011- Explicitly set the environment variable TOKENIZERS_PARALLELISM=(true \| false)
```

w4nderlust:
The tokenizer is used in two places, one is the `add_feature_meta` that you already addressed, the other one is `feature_data` and in turn `build_sequence_matrix`, so this is still missing.

Regarding forking, it depends what you are doing really. In theory image data preprocessing does some multiprocessing under some cirmumstances, and hyperopt may also do it, so if you are running text data preprocessing within an hyperopt call, this could be an explaination of why you are getting this. How are you actually running the tests?

brightsparc:
I notice a post about FastTokenizer by default coming out in the next release so maybe we don’t need this PR this still?
https://github.com/huggingface/transformers/releases/tag/v4.0.0-rc-1

w4nderlust:
> I notice a post about FastTokenizer by default coming out in the next release so maybe we don’t need this PR this still?
> https://github.com/huggingface/transformers/releases/tag/v4.0.0-rc-1

Right that is what i was referring to before. They are already in rc1, so I'd say let's wait a bit more for the release and then decide how to move forward accordingly, ok?

brightsparc:
Yep, sounds good.  I will close this PR and wait for v4 release.

PR #1014, opened 2020-11-18
Added Backend interface for abstracting DataFrame preprocessing steps

tgaddair:
The purpose of this PR is to introduce a generic `Backend` interface that allows for plugging in alternative DataFrame libraries to Pandas, including Dask, Modin, and Koalas.

This PR splits off the `Backend` abstraction from #970 into its own set of changes, allowing us to first prove out the refactor of the preprocessing steps before adding support for Dask.


ludwig/utils/time_utils.py
```diff
@@ -68,15 +68,15 @@ def proc(self):
         return time.clock() - self._proc
 
     def tic(self):
-        """Like Matlab tic/toc for wall time and processor time"""
+        """Like Matlab tic/toc for wall time and dataframe time"""
```
w4nderlust, line 71: this and the next replace are likely wrong :)

tgaddair: I blame PyCharm!

PR #1015, opened 2020-11-19
Add support for transforming numeric predictions that were normalized

jimthompson5802:
# Code Pull Requests

Proposed solution for the following situation:  If a numerical output feature is 'normalized', e.g., 'zscore' or 'minmax', predictions are returned in the 'normalized' value space instead of the output feature's value space.  This PR adds functions to do the inverse transformation on output features that were normalized.


jimthompson5802:
@w4nderlust   just marked this ready for review.  Summary of changes:
* added inverse transformations for `zscore` and `minmax` normalizations for output feature
* added new normalization `log1p` that performs $$ln(1 + x)$$ and its inverse $$exp(x) - 1$$.
* Wrapped the above in an object structure with `transform()` and `inverse_transform()` methods
* Created registry for the transformation objects.

Let me know what you think.  If this looks good, I'll add a unit test for the transformation/inverse transformations.

ludwig/features/numerical_feature.py
```diff
@@ -81,17 +84,46 @@ def add_feature_data(
...
+                    metadata[feature[NAME]]['min'],
+                    metadata[feature[NAME]]['max']
+                )
+            elif normalization_type == 'log1p':
+                numeric_transformer = NumericTransformer()
+            else:
+                raise ValueError(
+                    'Normalization "{}" not supported. Valid values are '
+                    '"minmax" or "zscore"'.format(normalization_type)
+                )
```
w4nderlust, line 109: Thinking about ways to remove this if.
If the `NumericTransformer.__init__` function has `**kwargs`, we can pass the entire `metadata[feature[NAME]]` without problems.
So we sould just need:
```
NumericTransformer = get_from_registry(
                 preprocessing_parameters['normalization'],
                 numeric_transformation_registry
             )(**metadata[feature[NAME]])
```
and entirely remove this if. What do you think?

ludwig/features/numerical_feature.py
```diff
@@ -267,7 +299,37 @@ def postprocess_predictions(
...
+                    numeric_transformer = NumericTransformer(
+                        metadata['min'],
+                        metadata['max']
+                    )
+                elif normalization_type == 'log1p':
+                    numeric_transformer = NumericTransformer()
+                else:
+                    raise ValueError(
+                        'Normalization "{}" not supported. Valid values are '
+                        '"minmax" or "zscore"'.format(normalization_type)
```
w4nderlust, line 323: Same thing as before regarding how to remove this if.

ludwig/features/numerical_feature.py
```diff
@@ -314,3 +376,49 @@ def populate_defaults(output_feature):
...
+        return x * self.range + self.min_value
+
+
+class Log1pTransformer:
+
+    def transform(self, x: np.ndarray) -> np.ndarray:
+        if np.any(x <= 0):
+            raise ValueError(
+                'One or more values are non-positive.  '
+                'log1p normalization only defined for positive values.'
```
w4nderlust, line 412: "is defined only for"?

ludwig/features/numerical_feature.py
jimthompson5802: done

ludwig/features/numerical_feature.py
jimthompson5802: done

ludwig/features/numerical_feature.py
jimthompson5802: done

ludwig/features/numerical_feature.py
```diff
@@ -81,17 +84,17 @@ def add_feature_data(
...
+            NumericTransformer = get_from_registry(
+                normalization_type,
+                numeric_transformation_registry
+            )
+
+            numeric_transformer = NumericTransformer(**metadata[feature[NAME]])
+
+            values = dataset[feature[PROC_COLUMN]]
+            dataset[feature[PROC_COLUMN]] = numeric_transformer.transform(
+                values)
```
w4nderlust, line 97: as we are not using `values` anywhere else, this can become a single line:
```
dataset[feature[PROC_COLUMN]] = numeric_transformer.transform(dataset[feature[PROC_COLUMN]])
```

jimthompson5802:
@w4nderlust your comments were perfect, code looks cleaner.  Once I add the unit test, this should finish off the PR.

ludwig/features/numerical_feature.py
```diff
@@ -267,7 +270,21 @@ def postprocess_predictions(
...
+                )
+
+                numeric_transformer = NumericTransformer(**metadata)
+                values_to_return = numeric_transformer.inverse_transform(
+                    predictions[PREDICTIONS].numpy()
+                )
+            else:
+                values_to_return = predictions[PREDICTIONS].numpy()
+
+            postprocessed[PREDICTIONS] = values_to_return
```
w4nderlust, line 287: We may further avoid this if (and the corresponding one in preprocessing) by having an `IdentityTransformer` in the registry associated with a `None` key, and obtaining the transformer by:
```
NumericTransformer = get_from_registry(
                      metadata['preprocessing'].get('normalization', None),
                      numeric_transformation_registry
                  )
```

ludwig/features/numerical_feature.py
```diff
@@ -60,6 +61,8 @@ def get_feature_meta(column, preprocessing_parameters):
                     'min': column.astype(np.float32).min(),
                     'max': column.astype(np.float32).max()
                 }
+            elif preprocessing_parameters['normalization'] == 'log1p':
+                return {}
```
w4nderlust, line 65: We could avoid this if by adding an additional function to the transformer interface like `fit_transform_params` like:
```
def fit_transform_params(column):
   ...
   return {calculated_values_like_min_and_max}
```

ludwig/features/numerical_feature.py
jimthompson5802: done

ludwig/features/numerical_feature.py
jimthompson5802: done

ludwig/features/numerical_feature.py
jimthompson5802: I understand the comment is to add a `fit_transform_params()` method to the numeric transformer classes.

I'd like to discuss comment.  I have question and concern.

**Question**: If I understand the intent, we replace this code in the `get_feature_metadata()`
```
        if preprocessing_parameters['normalization'] is not None:
            if preprocessing_parameters['normalization'] == 'zscore':
                return {
                    'mean': column.astype(np.float32).mean(),
                    'std': column.astype(np.float32).std()
                }
            elif preprocessing_parameters['normalization'] == 'minmax':
                return {
                    'min': column.astype(np.float32).min(),
                    'max': column.astype(np.float32).max()
                }
            elif preprocessing_parameters['normalization'] == 'log1p':
                return {}
            else:
                logger.info(
                    'Currently zscore and minmax are the only '
                    'normalization strategies available. No {}'.format(
                        preprocessing_parameters['normalization'])
                )
                return {}
        else:
            return {}
```

Does `get_feature_metadata()` become something like this?

```
def get_feature_metadata(column, preprocessing_parameters):

     # some code to instantiate  the numeric transformer object
        numeric_transformer = get_from_registry(
            preprocessing_parameters.get('normalization', None),
            numeric_transformation_registry
        )(**metadata[feature[NAME]])

      return numeric_transformer.fit_transform_params(column)
```

**Concern**:  If the above is directionally correct, then it is not clear to me how `# some code to instantiate the numeric transformer object` will work.  For two of the transformer classes, the constructor needs to know the pre-computed values for `mean`/`std` or `min`/`max`.  This seems like a chicken and egg kind of situation.

One possibility to get around this is to change the constructor `def __init__()` method to call `fit_transform_params()`.  This should work during pre-processing because we have values in the original value space.  However, I don't see how this work in post-processing because we only have values in the transformed space and we do not save the  numeric transformer object.  The numeric transformer object is instantiated in both pre-processing and post-processing.

Anyway, I think there are some nuances that need to be worked out with the `fit_transform_params()` method.

w4nderlust: I see your point. i believe it is solved by making the fit function a static/class function that doesn't require the class to be instantiated to be used.

jimthompson5802: I have a proposed solution for the `fit_transform_params()` method.  However to get it to work, even with a static `fit_transform_params()`, I had to handle two different situations when the the numeric transformer object is instantiated.  Pay close attention to comments in the `__init__()` and parameter validations performed in `transform()` and `Inverse_transform()` methods.

Does this look reasonable?

ludwig/features/numerical_feature.py
```diff
@@ -49,26 +50,36 @@ class NumericalFeatureMixin(object):
...
-                    'normalization strategies available. No {}'.format(
-                        preprocessing_parameters['normalization'])
-                )
-                return {}
-        else:
-            return {}
+        numeric_transformer = get_from_registry(
+            preprocessing_parameters.get('normalization', None),
+            numeric_transformation_registry
+        )(**preprocessing_parameters)
```
w4nderlust, line 56: As `fit_transform_params` is a static method, we don't need `(**preprocessing_parameters)`.

ludwig/features/numerical_feature.py
w4nderlust: I'm guessing you are refferring to `# When parameters are None we only need object to use
         # the fit_transform_params method, other methods should not be used`
I believe we don't need this additional consideration, swe don't need the parameters to be set to None by default, and also we don't need to check if they are not none. i think the doubt stems from line 56 where you initialize an obect before calling the static method, but you actually do not need to initialize it, you can call the `fit_transform_params` directly from the call reference you get from the registry. This should solve the outstanding doubts i believe.

w4nderlust:
@tgaddair we are almost done on this, but before merging I'd first merge your PRs, as this may need to be adapted acordingly. Do you agree?

tgaddair:
@w4nderlust yeah that sounds good.  It should be fine to merge #1014 followed by this PR.  The other should not conflict.

ludwig/features/numerical_feature.py
jimthompson5802: learning something new every day.  done.

ludwig/features/numerical_feature.py
jimthompson5802: done

ludwig/features/numerical_feature.py
```diff
@@ -357,15 +357,17 @@ class Log1pTransformer:
     def __init__(self, **kwargs: dict):
         pass
 
-    def transform(self, x: np.ndarray) -> np.ndarray:
+    @staticmethod
```
w4nderlust, line 360: I wouldn't make those methods static even if they don't need to use self, because we are calling them as `instance.transform(x)`, so having `self` as the first parameter of the function captures the `instance` variable.

jimthompson5802: done.

w4nderlust:
This looks good to me now. Let's hold a sec before merging as @tgaddair suggested. In the mean time we could add a test ;)

jimthompson5802:
Added unit test for numeric transformers.
While not related to numeric transformers, I adjusted internal variable names in the cached checksum test.

ifokeev:
Thanks for the feature. We use this code to postprocess numeric:

```
def postprocess_value(feature_name, value, train_set_metadata):
    # remove suffix
    norm_feature_name = feature_name.replace('_predictions', '')
    post_value = value

    if norm_feature_name in train_set_metadata:
        feature_data = train_set_metadata[norm_feature_name]

        idx2str = feature_data.get('idx2str')
        normalization = feature_data.get('preprocessing', {}).get('normalization')

        if idx2str:
            post_value = idx2str[value]
        elif normalization == 'zscore':
            mean = feature_data['mean']
            std = feature_data['std']
            post_value = float(post_value) * std + mean
        elif normalization == 'minmax':
            min_val = feature_data['min']
            max_val = feature_data['max']
            post_value = float(post_value) * (max_val - min_val) + min_val

    return post_value
```

and that's pretty awkward

jimthompson5802:
@ifokeev   Thank you.  Glad the feature is helpful.

jimthompson5802:
@ifokeev   I forgot to ask.  In your `postprocess_value()` function, I noticed this code fragment:
```

        idx2str = feature_data.get('idx2str')
        normalization = feature_data.get('preprocessing', {}).get('normalization')

        if idx2str:
            post_value = idx2str[value]
```
This code implies an output feature type that is non-numeric.  Would a similar capability be useful in that other feature type?  What is the use case around that other feature type?

ifokeev:
@jimthompson5802 

> Would a similar capability be useful in that other feature type?

Yeah. That's very useful.

> What is the use case around that other feature type?

As I understand there are still no support for `category` postprocessing. And maybe some others.
I had to postprocess myself all the time.

https://github.com/uber/ludwig/blob/ecbcf334f762484b22cd171a666bea75921dc202/tests/integration_tests/test_savedmodel.py#L170

w4nderlust:
The category postprocessing should alreadydo that:
https://github.com/uber/ludwig/blob/master/ludwig/features/category_feature.py#L407-L411

ludwig/features/numerical_feature.py
```diff
@@ -327,10 +326,14 @@ def inverse_transform(self, x: np.ndarray) -> np.ndarray:
         return x * self.sigma + self.mu
 
     @staticmethod
-    def fit_transform_params(column: np.ndarray) -> dict:
+    def fit_transform_params(
+            column: np.ndarray,
+            backend: 'LocalBackend'
```
tgaddair, line 331: The type here can be `Backend` (the base class).

jimthompson5802: done

PR #1017, opened 2020-11-19
Added custom encoder / decoder registration decorator

tgaddair:
The purpose of this PR is to make it easy for users to extend Ludwig by adding custom encoders and decoders:

```
@register(name='my_encoder')
class CustomNumericalEncoder(NumericalEncoder):
    def call(self, inputs, training=None, mask=None):
        ...
```

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -21,16 +21,13 @@ from abc import ABC
 import tensorflow as tf
 
 from ludwig.encoders import sequence_encoders
-from ludwig.encoders.base import Encoder, register
+from ludwig.encoders.base import Encoder, Registry, register
 from ludwig.modules.reduction_modules import SequenceReducer
 
 logger = logging.getLogger(__name__)
 
 
-ENCODER_REGISTRY = {
-    # TODO(travis): this will not work with custom sequence encoders added after this module is imported
-    **sequence_encoders.ENCODER_REGISTRY
-}
+ENCODER_REGISTRY = Registry(sequence_encoders.ENCODER_REGISTRY)
 
 
 class TextEncoder(Encoder, ABC):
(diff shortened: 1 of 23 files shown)


ludwig/encoders/text_encoders.py
```diff
@@ -16,16 +16,31 @@
...
+from ludwig.encoders import sequence_encoders
+from ludwig.encoders.base import Encoder, register
 from ludwig.modules.reduction_modules import SequenceReducer
 
 logger = logging.getLogger(__name__)
 
 
-class BERTEncoder(Layer):
+ENCODER_REGISTRY = {
+    # TODO(travis): this will not work with custom sequence encoders added after this module is imported
```
w4nderlust, line 31: I think this PR encapsulates perfectly what we discussed and it's great. i guess this todo is the only point to figure out.
On one hand, this may be fine: If a user wants to add an encoder on the fly to both sequence and text, they will have to do it manually, or maybe extending both SequenceEncoder and TextEncoder (i believe it should work, but I may be wrong).
Or maybe an alternative could be that the registr decorator has a registry parameter? In that case the user could use the decorator twise to add to both registries.

tgaddair: I like the idea of adding it to both through inheritance.  Let me try that and see how it goes.

tgaddair: Okay, I found what I feel is a good solution to this problem.  Added a new `Registry` type, that is just a normal dictionary, except that you can define parent-child relationships, so that if the parent (Sequence Encoder) is updated, so is the child (Text Encoder).  What do you think?

w4nderlust: That's interesting. There's a corner case, which is what if the children already set a value for a key and then we add something different with the same key to the parent, the previous values would be overridden. But actually, I don't believe this would be an issue to be honest.

tgaddair: The new design should account for this issue.

PR #1019, opened 2020-11-19
toxic comments dataset

flozi00:
This was my first try understanding the new datasets API, hope it's correct.
Training running without errors


ludwig/datasets/toxicComments/config.yaml
```diff
@@ -0,0 +1,4 @@
+version: 1.0
+download_urls:
+  - https://raw.githubusercontent.com/balajibalasubramanian/Kaggle-Toxic-Comments-Challenge/master/data/train.csv
```
tgaddair, line 3: Is this the same dataset as https://www.kaggle.com/c/jigsaw-toxic-comment-classification-challenge/data ?  If so, we may wish to use the Kaggle URL as the source of truth.  @skanjila is working on a PR that will allow us to download from Kaggle.  Once that's ready, we can adapt this PR to use that as well.  Does that sound good?

tgaddair (commented):
Hey @flozi00, thanks for the PR!  Really excited to see the datasets API being used.  As I mentioned in one of the comments, this could be a good candidate to integrate with our upcoming Kaggle utility.  Let me know if that sounds reasonable to you.

ludwig/datasets/toxicComments/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+from ludwig.datasets.mixins.load import CSVLoadMixin
+from ludwig.datasets.mixins.process import IdentityProcessMixin
+
+
+def load(cache_dir=DEFAULT_CACHE_LOCATION, split=False):
+    dataset = toxicComments(cache_dir=cache_dir)
+    return dataset.load(split=split)
+
+
+class toxicComments(UncompressedFileDownloadMixin, IdentityProcessMixin, CSVLoadMixin, BaseDataset):
```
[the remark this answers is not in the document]
flozi00: done

ludwig/datasets/toxicComments/config.yaml
flozi00: yeah, it's the jigsaw challenge
sounds good

I am thinking about writing converter for huggingface datasets API, what do you think ?
Would it make sense ?

w4nderlust: That may be nice to have! The reason why we did not use their dataset API directly is that 1) it's limited to nlp datasets, while in out case we want to cover also non nlp datasets and 2) their goal is slightly different, they want to have an in memory representation that is handy, while here we just want to encapsulate data preprocessing so that you can get a csv/dataframe in a shape that is useful for Ludwig. Bt this said, there's obviously an intersection between the two projects, so thinking about a bridge and coming up with pros and cons could be a good idea!

PR #1020, opened 2020-11-22
[WIP] Added schema validation to config files

tgaddair:
The purpose of this PR is to catch typos or other misconfigurations in the Ludwig `config.yaml` before training.

One common occurrence is that users will set a parameter in the config expecting it to do something, but because Ludwig does not validate the schema and will assume reasonable defaults, this can result in errors that go undetected.

This PR introduces schema validation based on `jsonschema` that is called at `LudwigModel` creation time.  It enforces that certain fields are set, and that certain fields (like `preprocessing`) do not have any unrecognized subfields.

Note that we do not apply strict subfield checking to every field in the config, in order to make it easier for users to switch between different encoders that take different parameters, for example.  We do, however, check that feature types, encoders, tokenizers, and other common parameters all have valid values and that preprocessing in particular does not have unrecognized fields.

### DIFF
--- a/ludwig/utils/schema.py
+++ b/ludwig/utils/schema.py
@@ -36,11 +36,14 @@ def get_schema():
                     'properties': {
                         'name': {'type': 'string'},
                         'type': {'type': 'string', 'enum': INPUT_FEATURE_TYPES},
+                        'column': {'type': 'string'},
+                        'proc_column': {'type': 'string'},
                         'preprocessing': {},
                         'encoder': {'type': 'string'}
                     },
                     'allOf': get_input_encoder_conds(),
-                    'required': ['name', 'type']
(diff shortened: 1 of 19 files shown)


ludwig/utils/schema.py
```diff
@@ -0,0 +1,96 @@
...
+                        'encoder': {'type': 'string'}
+                    },
+                    'allOf': get_input_encoder_conds() + get_input_preproc_conds(),
+                    'required': ['name', 'type'],
+                }
+            },
+            'combiner': {},
+            'output_features': {},
+            'training': {},
+            'preprocessing': {},
```
w4nderlust, line 47: missing hyperopt, but maybe it's worth its own separate schema

PR #1021, opened 2020-11-23
Added Kaggle API integration and Titanic dataset

skanjila:
- Introduced a mixin for kaggle workflow for download and authentication
- Added a unit test that tests the workflow using mocks


ludwig/datasets/titanic/config.yaml
```diff
@@ -0,0 +1,7 @@
+version: 1.0
+split_filenames:
+  train_file: titanic_train.csv
+  test_file: titanic_test.csv
+download_file_type: csv
+csv_filename: titanic.csv
+download_urls: ["http://fake.org", "http://fake1.org"]
```
w4nderlust, line 7: Maybe here we can just put `None`

tests/ludwig/datasets/titanic/test_titanic_workflow.py
```diff
@@ -0,0 +1,72 @@
...
+        'cabin': [2, 2, 1],
+        'embarked': ["Q", "Q", "C"],
+        'boat': [0, 0, 0],
+        'body': [0, 1, 0],
+        'home.dest': ["Sweden", "Slovenia", "Italy"],
+        'survived': [0, 1, 0]
+    })
+    titanic_train_filename = "titanic_train.csv"
+    titanic_test_filename = "titanic_test.csv"
+    with tempfile.TemporaryDirectory() as source_dir:
```
jimthompson5802, line 51: In discussions with @w4nderlust about the unit tests in general, we looking to standardize on [pytest's fixture `tmpdir`](https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture) for temporary directories.  We're in the process of updating the other tests to use this fixture.

Is it possible for this test to be adapted to use pytest's `tmpdir` fixture?

I see later on that this test creates a second temporary directory within the first temporary directory (`source_dir`).  This comment does not apply to the second temporary directory.  I think it can be your call on how you manage internal temporary directories.  About the only consideration is the potential naming conflict between your second temporary directory and pytest's `tmpdir` fixture.

skanjila: sure will change @jimthompson5802 , thanks for the input and moving forward please flag me if appropriate for some of those discussions

ludwig/datasets/titanic/config.yaml
skanjila: Actually I was thinking of just removing this altogether as there is not really a concept of a download url since the kaggle client abstracts that away from ys

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,38 @@
...
+class KaggleMixin:
+    """A mixin to abstract away the details of the kaggle API which includes
+    the ability to authenticate against the kaggle API, list the various datasets
+    and finally download the dataset"""
+    config: dict
+    raw_dataset_path: str
+    raw_temp_path: str
+    name: str
+
+    def list_downloads(self) -> list:
```
tgaddair, line 18: Is this method needed?

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def download_raw_dataset(self):
+        """
+        Download the raw dataset and extract the contents of the zip file and
+        store that in the cache location.
+        """
+        os.makedirs(self.raw_temp_path, exist_ok=True)
+        # Download all files for a competition
+        # Signature: competition_download_files(competition, path=None, force=False, quiet=True)
+        api.competition_download_files('titanic', path=self.raw_temp_path)
```
tgaddair, line 33: The competition name `'titanic'` should be provided in the config:

```
@property
def competition_name(self):
    return self.config["competition"]
```

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,38 @@
...
+        os.rename(self.raw_temp_path, self.raw_dataset_path)
+
+    @property
+    def download_urls(self):
```
tgaddair, line 37: It doesn't look like this is needed.

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,38 @@
+import os
+from pathlib import Path
+from kaggle.api.kaggle_api_extended import KaggleApi
+
+DEFAULT_CACHE_LOCATION = str(Path.home().joinpath('.ludwig_cache'))
+api = KaggleApi()
+api.authenticate()
+
+class KaggleMixin:
```
tgaddair, line 9: Let's call this `KaggleDownloadMixin` for consistency with the others.

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,38 @@
+import os
+from pathlib import Path
+from kaggle.api.kaggle_api_extended import KaggleApi
+
+DEFAULT_CACHE_LOCATION = str(Path.home().joinpath('.ludwig_cache'))
```
tgaddair, line 5: Doesn't look like `DEFAULT_CACHE_LOCATION` is needed here.

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,38 @@
...
+api = KaggleApi()
```
tgaddair, line 6: Let's call this within `download_raw_dataset` so we give the user a chance to fix the environment without needing to recreate the interpreter state.

ludwig/datasets/titanic/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    dataframe that can fit into Ludwig's training API.
+    """
+
+    config: dict
+    raw_temp_path: str
+    raw_dataset_path: str
+    processed_temp_path: str
+    processed_dataset_path: str
+
+    def __init__(self, dataset_name="titanic", cache_dir=DEFAULT_CACHE_LOCATION):
```
tgaddair, line 43: Doesn't look like we need this `__init__` function.

ludwig/datasets/titanic/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        super().__init__(dataset_name, cache_dir)
+
+    def process_downloaded_dataset(self):
+        """ The final method where we create a training and test file by iterating through
+        both of these files"""
+        train_df = pd.read_csv(os.path.join(self.processed_temp_path, "titanic_train.csv"))
```
tgaddair, line 49: We should pull these filenames from the config:

```
@propety
def train_file(self):
    return self.config["train_file"]

@propety
def test_file(self):
    return self.config["test_file"]
```

ludwig/datasets/titanic/config.yaml
tgaddair: Yeah, looks like we can remove it and replace it with:

```
competition: titanic
```

ludwig/datasets/titanic/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        test_df = pd.read_csv(os.path.join(self.processed_temp_path, "titanic_test.csv"))
+        train_df["split"] = 0
+        test_df["split"] = 2
+        final_df = pd.concat([train_df, train_df], axis=1)
+        final_df.to_csv(self.processed_dataset_path)
+
+    @property
+    def download_url(self):
```
tgaddair, line 57: This appears to be unneeded.

requirements.txt
```diff
@@ -10,6 +10,8 @@ tensorflow>=2.2
 tfa-nightly==0.12.0.dev20200820045606
 PyYAML>=3.12
 absl-py
+kaggle
+slugify
```
tgaddair, line 14: Do we need `slugify`?

tests/ludwig/datasets/titanic/test_titanic_workflow.py
```diff
@@ -0,0 +1,73 @@
...
+        'train_file': titanic_train_filename,
+        'test_file': titanic_test_filename
+        },
+        'csv_filename': 'fake_titanic.csv',
+    }
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with mock.patch('ludwig.datasets.base_dataset.read_config',
+                        return_value=config):
+            with mock.patch('ludwig.datasets.mixins.kaggle.KaggleMixin.api.authenticate', return_value=None):
```
tgaddair, line 66:
```
with mock.patch('ludwig.datasets.mixins.kaggle.KaggleApi') as mock_kaggle_cls:
    mock_kaggle_cls.return_value = Mock()
    ...
```

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class KaggleMixin:
+    """A mixin to abstract away the details of the kaggle API which includes
+    the ability to authenticate against the kaggle API, list the various datasets
+    and finally download the dataset"""
+    config: dict
+    raw_dataset_path: str
+    raw_temp_path: str
+    name: str
```
tgaddair, line 17: Here we can assume the presence of member variables for the username as key:

```
kaggle_username: str
kaggle_api_key: str
```

ludwig/datasets/titanic/__init__.py
tgaddair, line 43:
```
def __init__(kaggle_username=None, kaggle_api_key=None, ...):
    self.kaggle_username = kaggle_username
    self.kaggle_api_key = kaggle_api_key
```

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,40 @@
...
+            a list of competition objects associated with Titanic"""
+
+        return api.competitions_list(search="titanic")
+
+    def download_raw_dataset(self):
+        """
+        Download the raw dataset and extract the contents of the zip file and
+        store that in the cache location.
+        """
+        os.makedirs(self.raw_temp_path, exist_ok=True)
```
tgaddair, line 32:
```
with update_env(KAGGLE_USERNAME=self.kaggle_username, KAGGLE_API_KEY=self.kaggle_api_key):
    api = KaggleApi()
    api.authenticate()
```

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,40 @@
+import os
+from pathlib import Path
+from kaggle.api.kaggle_api_extended import KaggleApi
+
```
tgaddair, line 4: from contextlib import contextmanager

@contextmanager
def update_env(**kwargs):
    prev_env = {}
    for k, v in kwargs.items():
        if v is not None:
            prev_env = os.environ.get(k)
            os.environ[k] = v
    try:
         yield
    finally:
        for k, v in prev_env.items():
            os.environ[k] = v
```

Comment (tgaddair) [inline ludwig/datasets/mixins/kaggle.py:4]: Similar to this: https://github.com/horovod/horovod/blob/master/test/utils/common.py#L139

But we need to make one change, which is that we want to keep the original environment as well.

Comment (tgaddair) [reply ludwig/datasets/mixins/kaggle.py:4]:
```
@contextmanager
def update_env(**kwargs):
    override_env = {k: v for k, v in kwargs.items() if v is not None}
    old = os.environ
    try:
        os.environ = {os.environ, **override_env}
        yield
    finally:
        os.environ = old
```

File: requirements.txt (hunk 12 above)
Comment (skanjila) [reply requirements.txt:14]: Yes kaggle has a dependency on that

File: tests/ludwig/datasets/titanic/test_titanic_workflow.py (hunk 13 above)
Comment (skanjila) [reply tests/ludwig/datasets/titanic/test_titanic_workflow.py]: Fixed, note that assert_called_once doesnt have a parenthesis

File: ludwig/datasets/titanic/__init__.py (hunk 11 above)
Comment (skanjila) [reply ludwig/datasets/titanic/__init__.py]: Removed

File: ludwig/datasets/titanic/__init__.py (hunk 10 above)
Comment (skanjila) [reply ludwig/datasets/titanic/__init__.py:51]: We'll get to this once I get the example working

File: ludwig/datasets/mixins/kaggle.py (hunk 8 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py:4]: Fixed

File: ludwig/datasets/mixins/kaggle.py (hunk 7 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py]: Removed

File: ludwig/datasets/mixins/kaggle.py (hunk 6 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py]: Changed

File: ludwig/datasets/mixins/kaggle.py (hunk 5 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py]: Removed

File: ludwig/datasets/mixins/kaggle.py (hunk 4 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py:25]: Added

File: ludwig/datasets/mixins/kaggle.py (hunk 3 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py]: Removed

File: requirements.txt (hunk 12 above)
Comment (tgaddair) [reply requirements.txt:14]: Looks like they already pull in `python-slugify` in their requirements: https://github.com/Kaggle/kaggle-api/blob/master/setup.py#L39

Are you sure we need both?  That's a bug on their end if so.

Comment (skanjila) [reply requirements.txt:14]: I thought I saw one pulling in the other, I think you're right, will remove and rebuild

File: ludwig/datasets/mixins/kaggle.py (hunk 8 above)
Comment (tgaddair) [reply ludwig/datasets/mixins/kaggle.py:4]: Did you push your changes?  The code here still has the previous behavior.

Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py:4]: I did actually and looking at this locally shows this function

File: requirements.txt (hunk 12 above)
Comment (skanjila) [reply requirements.txt]: Removed

File: ludwig/datasets/titanic/__init__.py (hunk 10 above)
Comment (skanjila) [reply ludwig/datasets/titanic/__init__.py]: Fixed, see latest MR

File: ludwig/datasets/titanic/__init__.py (hunk 9 above)
Comment (skanjila) [reply ludwig/datasets/titanic/__init__.py]: Added

File: ludwig/datasets/mixins/kaggle.py (hunk 15 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py:27]: Added

File: ludwig/datasets/mixins/kaggle.py (hunk 14 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py:13]: Added

File: ludwig/datasets/mixins/kaggle.py (hunk 16 above)
Comment (skanjila) [reply ludwig/datasets/mixins/kaggle.py:4]: Added

File: ludwig/datasets/mixins/kaggle.py [hunk 17]
```diff
@@ -0,0 +1,46 @@
...
+        kaggle.json file we lookup the passed in username and the api key and
+        perform authentication.
+        """
+        with self.update_env(KAGGLE_USERNAME=self.kaggle_username, KAGGLE_API_KEY=self.kaggle_api_key):
+            api = KaggleApi()
+            api.authenticate()
+        os.makedirs(self.raw_temp_path, exist_ok=True)
+        # Download all files for a competition
+        # Signature: competition_download_files(competition, path=None, force=False, quiet=True)
+        api.competition_download_files('titanic', path=self.raw_temp_path)
```
tgaddair, line 30: Should be changed to `api.competition_download_files(self.competition_name, ...)`

ludwig/datasets/mixins/kaggle.py
```diff
@@ -0,0 +1,46 @@
...
+        os.rename(self.raw_temp_path, self.raw_dataset_path)
+
+    @contextmanager
+    def update_env(**kwargs):
```
tgaddair, line 34: Need to add `self` param if defining as a member function.

ludwig/datasets/titanic/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Titanic(CSVLoadMixin, KaggleDownloadMixin, BaseDataset):
+    """The Titanic dataset.
+
+    This pulls in an array of mixins for different types of functionality
+    which belongs in the workflow for ingesting and transforming training data into a destination
+    dataframe that can fit into Ludwig's training API.
+    """
+    config: dict
```
tgaddair, line 36: We don't need to define all these variables here since we can infer their existence from the parent classes.

tests/ludwig/datasets/titanic/test_titanic_workflow.py
```diff
@@ -0,0 +1,72 @@
...
+        },
+        'csv_filename': 'fake_titanic.csv',
+    }
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with mock.patch('ludwig.datasets.base_dataset.read_config',
+                        return_value=config):
+            with mock.patch('ludwig.datasets.mixins.kaggle.KaggleApi') as mock_kaggle_cls:
+                mock_kaggle_api = MagicMock()
+                mock_kaggle_cls.return_value = mock_kaggle_api
```
tgaddair, line 69: Probably need to call `titanic.download()` here.

ludwig/datasets/mixins/kaggle.py
skanjila: Fixed

ludwig/datasets/mixins/kaggle.py
skanjila: Added

ludwig/datasets/titanic/__init__.py
skanjila: Removed

tests/ludwig/datasets/titanic/test_titanic_workflow.py
skanjila: Added

ludwig/datasets/titanic/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+# ==============================================================================
+import os
+import pandas as pd
+from ludwig.datasets.base_dataset import BaseDataset, DEFAULT_CACHE_LOCATION
+from ludwig.datasets.mixins.kaggle import KaggleDownloadMixin
+from ludwig.datasets.mixins.load import CSVLoadMixin
+
+
+def load(cache_dir=DEFAULT_CACHE_LOCATION, split=False):
```
tgaddair, line 24: We should expose the Kaggle params here as well:

```
def load(cache_dir=DEFAULT_CACHE_LOCATION, split=False, kaggle_username=None, kaggle_api_key=None):
```

examples/titanic/simple_model_training.py
```diff
@@ -19,14 +19,15 @@
 # Define Ludwig model object that drive model training
 model = LudwigModel(config='./model1_config.yaml',
                     logging_level=logging.INFO)
+dataset = Titanic().load()
```
tgaddair, line 22: Let's use the convenience function:

```

dataset = titanic.load()
```

examples/titanic/simple_model_training.py
```diff
@@ -19,14 +19,15 @@
...
 
 # initiate model training
 (
     train_stats,  # dictionary containing training statistics
     preprocessed_data,  # tuple Ludwig Dataset objects of pre-processed training data
     output_directory # location of training results stored on disk
  ) = model.train(
-    dataset='./data/train.csv',
+    dataset=dataset,
```
tgaddair, line 30: We probably only want the train split here.

tests/ludwig/datasets/titanic/test_titanic_workflow.py
```diff
@@ -0,0 +1,75 @@
...
+        'competition': 'titanic'
+    }
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with mock.patch('ludwig.datasets.base_dataset.read_config',
+                        return_value=config):
+            with mock.patch('ludwig.datasets.mixins.kaggle.KaggleApi') as mock_kaggle_cls:
+                mock_kaggle_api = MagicMock()
+                mock_kaggle_cls.return_value = mock_kaggle_api
+                mock_kaggle_api.authenticate.assert_called_once
```
tgaddair, line 71: This assert needs to come after the download.

examples/titanic/simple_model_training.py
skanjila: added

tests/ludwig/datasets/titanic/test_titanic_workflow.py
skanjila: moved, however the test still is making the actual api call instead of mocking, would love some help here

examples/titanic/simple_model_training.py
skanjila: added with the split, see the latest code

tgaddair (approved):

w4nderlust:
Thank you so much @skanjila . It would be great if you could add a couple words about this in the user guide ;)

PR #1024, opened 2020-11-24
feat(progress_tracker): save current learning rate

ifokeev:
There is no option to track current learning rate. I tried to update `progress_tracker.learning_rate` but it destroyed the training.

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -583,6 +583,8 @@ class Trainer:
                         self.optimizer.variables(), root_rank=0)
 
                 progress_tracker.steps += 1
+                progress_tracker.learning_rate = current_learning_rate
+
                 if is_on_master():
                     progress_bar.update(1)
                 first_batch = False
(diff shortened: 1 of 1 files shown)


w4nderlust:
Yeah there's some minor confusion about LR at the moment, as there are some mechanism that always apply (decay, warmup) and other that apply only based on validation (plateau logic and batch size increase). the learning rate that is kept track of in the tracker is affected by the validation logic, but not by the always on mechanisms (they use the lr to actually compute the current learning rate to use).
If you look at the tesnorboard, the learning rate tracked there is the one actually used.
I will change the logic of the progress tracker. Maybe having two value `base_learning_rate` and `current_learning_rate` may work, but i have to think a little bit about what's the best solution here.

ifokeev:
Ok, then easiest solution for now is just to add `progress_tracker.current_learning_rate`, I think. I integrated it already and can view learning rate changes at least

PR #1027, opened 2020-11-28
Added cast_columns function to preprocessing and cast_column function to all feature mixin classes

w4nderlust:
Added cast_columns function to preprocessing and cast_column function to all feature mixin classes

The intent is to cast each df column to the correct type, so that `handle_missing_values` can actually do its job. In particular, for numerical values, if we don't do this, then a empty string value will make it so that the whole column would be read by pandas as `object` and empty string would not be interpreted as a missing value. With this PR empty string becomes NaN for numerical columns, so `handle_missing_values` now works fine.

To test before and after, try creating a dataset with 2 numerical columns and replace some entries with empty string and try to provide it to Ludwig and see the difference in behavior.

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -17,11 +17,11 @@
 import logging
 import os
 from abc import ABC, abstractmethod
-from copy import copy
 
-import ludwig
 import numpy as np
 import pandas as pd
+
+import ludwig
 from ludwig.backend import LOCAL_BACKEND
 from ludwig.constants import *
 from ludwig.constants import TEXT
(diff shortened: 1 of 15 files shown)


ludwig/features/bag_feature.py
```diff
@@ -41,6 +41,11 @@ class BagFeatureMixin(object):
         'fill_value': UNKNOWN_SYMBOL
     }
 
+    @staticmethod
+    def cast_column(feature, input_df, proc_df, backend):
+        proc_df[feature[COLUMN]] = input_df[feature[COLUMN]]
```
tgaddair, line 46: Can just make this a no-op.

ludwig/data/preprocessing.py
```diff
@@ -1028,7 +1036,29 @@ def build_dataset(
     return dataset, metadata
 
 
-def build_metadata(dataset_df, features, global_preprocessing_parameters, backend):
+def cast_columns(input_df, features, global_preprocessing_parameters, backend):
+    # todo figure out if global_preprocessing_parameters is needed
+    proc_df = backend.df_engine.empty_df_like(input_df)
```
tgaddair, line 1041: Similar to the way we do missing value replacement, I think we should apply these cases directly to the input_df. Otherwise, the `proc_df` will contain both the input columns like `text` as well as the processed columns like `text_HDSDP1393`.  We only want processed columns in the `proc_df`.

tgaddair (approved):
LGTM!

PR #1033, opened 2020-11-29
Added label-wise probability to binary feature predictions

tgaddair:
Prior to this PR, prediction outputs for binary features contain the column `feature_probabilities` which is the raw output of the model used to determine the label `{True, False}`.  In order to know the model's confidence in a prediction, the user needs to check the label, then mentally subtract this value from 1 if the label is False to get the model's confidence.  This PR simplifies things by adding a new `feature_probability` column that does just this.

### DIFF
--- a/ludwig/data/postprocessing.py
+++ b/ludwig/data/postprocessing.py
@@ -16,6 +16,7 @@
 # ==============================================================================
 import pandas as pd
 
+from ludwig.constants import BINARY
 from ludwig.features.feature_utils import SEQUENCE_TYPES
 from ludwig.utils.data_utils import DICT_FORMATS, DATAFRAME_FORMATS, \
     normalize_numpy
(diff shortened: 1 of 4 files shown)


w4nderlust:
Trying to figure out the logic.

Right now we have:
- predictions containing True and False,
- probabilities containing the probability of the positive class

After this PR we would have:
- predictions containing True and False,
- probabilities containing the probability of the positive class
- probability contaning the probability of the predicted class

I would say we can do 1 more step to make the output of binary more conformant with the output ofr category with something like:
- predictions containing True and False,
- probabilities containing 2 columns, the probability of the negative class and the probability of the positive class (prob of the negative is just 1- pos)
- probability contaning the probability of the predicted class

This would address also @carlogrisetti 's request

w4nderlust:
Pushed a commit (blindly, may be broken).
Also postprocessing.py::convert_to_df will lkely need to be adapted  to figrure outthe names of the column to be "probabpility_False" and "probability_True" for the PROBABILITIES

tgaddair:
Cool, I like this approach.

Issue #746, opened 2020-06-22
Image pre-trained model for transfer learning


StuckinPhD:
Hi,

I have a relatively small image dataset and would like to utilize transfer learning to get better results. I could not find much details related to transfer learning online, and the only mention of it is in the update post here related to BERT: https://eng.uber.com/ludwig-v0-2/.

I am still unsure about how to perform image transfer learning as is possible with using pretrained models on imageNet etc.

Is it possible? if so how would the model definition look for this? as the model starting weights would be different to my reduced classes at output?

Thanks

w4nderlust:
At the moment there areno pretrained image models importable in Ludwig, from version 0.3 on we will start to add them. I'll consider this a feature reqeust.

Also let me know if you would be interested in collaborating in adding this feature, it would be pretty easy.

StuckinPhD:
Id be happy to work on this. 
Ive never contributed to any repo before but let me know what I need to do and I'll do my best :)

w4nderlust:
> Id be happy to work on this.
> how can I help?

Great! Thank you for being willing to help.
We are now orking a lot on porting to TF2, so any contribution should start from the `tf2_porting` branch.
Ludwig has this concept of encoders for input features that makes it easy to add  new pretrained image encoders.
If you look at the `features/image_feature.py` you'll find a disctionary of encoder associated with the input feature.
Ideally you may want to add a pretrained model there, giving it a name and implementing the class. Classes for image encoders are in `model/modules/image_encoders.py`.
Basically one need to implement an object with an `__init__` function that contains the parameters (maybe in this case the path to the pretrained model) and a `call` function that takes as inputs tensors of chape `[batch, h, w, c]` and return either [batch, h', w', c']` or just a `[batch, hidden]` tensord, depending where you cut the original preprocessed model, if before or after flattening.
Let me know if these instructions are clear for you or if you have additional questions.

rajatsahay:
Hey there!
Is anybody still working on this? I would love to help.

w4nderlust:
@rajatsahay thank you for being willing to help! The instructions in my previous message, combined with the instructions in the developer guide should be a good starting point. Let me know if after reading them you still have doubts :)

PR #1035, opened 2020-11-30
Image Captioning Example

michaelbzhu:
I'm trying to implement the [image captioning example](https://ludwig-ai.github.io/ludwig-docs/examples/#image-captioning) from the [Ludwig](ludwig.ai) website using the Flickr 8K dataset; would it be cool if I added the finished code to the `examples/` folder?

I labelled it as work in progress right now because my current configuration seems to not perform super well; the model generates a lot of the same phrases for different images, so I'm still working on it to improve the performance
```
config = {
    "input_features": [
        {
            "name": "image_path",
            "type": "image",
            "encoder": "stacked_cnn",
            "preprocessing": {      # resize every image to be the same size
                "height": 300,
                "width": 300,
                "resize_method": "interpolate"
            }
        }
    ],
    "output_features": [
        {            
            "name": "caption",
            "type": "text",
            "level": "word",
            "decoder": "generator",
            "cell_type": "lstm"
        }
    ]
}
```


w4nderlust:
This is great, yes it would be a very nice example to have.
I suggest you change the size to something smaller to begin with (by resizing) and then adapt the stacked cnn to be of the right depth / width. Taking inspiration from some early image captioning paper regarding the parameters of the cnn could be a good idea to have a starting point.

examples/image_captioning/image_captioning.py
```diff
@@ -0,0 +1,115 @@
...
+import tensorflow as tf
+import os
+import pandas as pd
+import numpy as np
+from ludwig.api import LudwigModel
+import logging
+
+################### DOWNLOAD AND PREPROCESS DATA ###################
+# Download data
+image_folder = os.path.abspath('.') + '/Flicker8k_Dataset/'
```
tgaddair, line 11: Can you add this as a dataset to `ludwig.datasets`?  That way it can be reused even outside of this example.

michaelbzhu: @tgaddair just made a PR for this at #1053

michaelbzhu:
Sorry I kind of paused on this PR over winter break, but I updated the model implementation which seems to get much better results

Previous: Image encoder 
- this model pretty much underfit to output only "a a a a a a ... "
- I thought it was due to either small dataset or not enough model capacity
- I referenced some papers and tensorflow implementations and most used a pretrained image model with pretty good results

Now: Preprocess the images using a pretrained VGG16 image model and use a vector encoder on the extracted features
- most of the predicted captions follow a standard sentence structure
- seems to do pretty good at recognizing dogs and environments
- possible improvements in that it sometimes outputs weird sentences like "a football player in a football"

But overall I think its performs well enough to add as an example; the main idea is to show how to use image/vector encoder along with text decoder, and the update also shows how to use a pretrained image model with ludwig

examples/image_captioning/image_captioning.py
```diff
@@ -0,0 +1,158 @@
...
+
+"""
+Here we will define a function that will use the pretrained model to extract 
+features from the images in our dataset. Every image will be converted into 
+a size-4096 vector encoding.
+"""
+from keras.preprocessing.image import load_img, img_to_array
+from keras.applications.vgg16 import preprocess_input
+
+def extract_features(dataset, model):
```
tgaddair, line 51: This is cool, I like how this example shows how Ludwig can be extended / combined with other tools.  This makes me think about how we would accomplish a similar result through adding custom encoder / preprocessing.  Today I think it would be more effort to do this by extending Ludwig classes, but ideally we'd like to get to a point where this could be done as part of the LudwigModel itself (which would allow us to do things like distribute using Dask).  @w4nderlust what do you think?

michaelbzhu: Yea I was thinking about that too, and there seems to have been an older conversation on adding pretrained image encoders to ludwig at #746 

I can try to help with that; I'm currently looking at ludwigs docs and previous examples to see how that might be done

Also fyi, there is a list of available pretrained image models through keras at https://keras.io/api/applications/.

w4nderlust: @tgaddair I chatted with @michaelbzhu about it, i suggested to add an encoder similar to what we did for the HF text encoders, which also contains specific preprocessing instructions. That could be a nice way to add additional pretrained image encoders.

dionisioedu (approved):

PR #1038, opened 2020-12-02
Added bool2str postprocessing for binary features so prediction vocab matches inputs

tgaddair:
This PR makes it so that when the user has input data like: `{'binary_feature': ['Yes', 'Yes', 'No']}`, the prediction vocabulary will match the input. So instead of getting outputs like `True` and `False`, we will map them back to `Yes` and `No`.

### DIFF
--- a/ludwig/data/postprocessing.py
+++ b/ludwig/data/postprocessing.py
@@ -85,7 +85,12 @@ def convert_to_df(
                             class_name = training_set_metadata[of_name][
                                 'idx2str'][i]
                         elif output_feature.type == BINARY:
-                            class_name = 'True' if i == 1 else 'False'
+                            if (of_name in training_set_metadata and
+                                    'bool2str' in training_set_metadata[of_name]):
+                                class_name = training_set_metadata[of_name][
+                                    'bool2str'][i]
+                            else:
+                                class_name = 'True' if i == 1 else 'False'
                         else:
                             class_name = str(i)
                         data_for_df[
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1087,7 +1087,7 @@ def build_metadata(dataset_df, features, global_preprocessing_parameters, backen
             ).get_feature_meta
 
             metadata[feature[NAME]] = get_feature_meta(
-                dataset_df[feature[NAME]].astype(str),
+                dataset_df[feature[NAME]],
                 preprocessing_parameters,
                 backend
             )
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1086,8 +1086,12 @@ def build_metadata(dataset_df, features, global_preprocessing_parameters, backen
                 base_type_registry
             ).get_feature_meta
 
+            column = dataset_df[feature[NAME]]
+            if column.dtype == object:
+                column = column.astype(str)
+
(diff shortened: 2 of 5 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1087,7 +1087,7 @@ def build_metadata(dataset_df, features, global_preprocessing_parameters, backen
             ).get_feature_meta
 
             metadata[feature[NAME]] = get_feature_meta(
-                dataset_df[feature[NAME]].astype(str),
```
tgaddair, line 1090: @w4nderlust do you know which features rely on this conversion for metadata computation?  Can we move it into those features only?

tgaddair: Per our conversation: changing so that we only cast to string if the dtype is object.

w4nderlust: lgtm

ludwig/utils/strings_utils.py
```diff
@@ -36,6 +36,8 @@
 COMMA_REGEX = re.compile(r'\s*,\s*')
 UNDERSCORE_REGEX = re.compile(r'\s*_\s*')
 
+BOOL_TRUE_STRS = {'yes', 'y', 'true', 't', '1'}
```
w4nderlust, line 39: as we are not lowercasing anywhere (I believe) maybe we can add also "Yes", "YES", "Y", "TRUE", "True", "T"

w4nderlust:
Just a minor question: what are the names of the columns of the predicted DF when the the column is not object/string? if they are predictions, probability, probabilityes_0 and probabilities_1 like I expect, then i believe we are good :)

ludwig/utils/strings_utils.py
tgaddair: The lowercasing occurs in the `str2bool` function below.

tgaddair:
> Just a minor question: what are the names of the columns of the predicted DF when the the column is not object/string? if they are predictions, probability, probabilityes_0 and probabilities_1 like I expect, then i believe we are good :)

If the type is bool: `_probabilities_False` and `_probabilities_True`
If the type is string (yes/no): `_probabilities_no` and `_probabilities_yes`

Does that work?  `_probabilities_0` and `_probabilities_1` are also doable, but I felt True/False was more clear.

ludwig/data/postprocessing.py
```diff
@@ -85,7 +85,12 @@ def convert_to_df(
                             class_name = training_set_metadata[of_name][
                                 'idx2str'][i]
                         elif output_feature.type == BINARY:
-                            class_name = 'True' if i == 1 else 'False'
+                            if (of_name in training_set_metadata and
+                                    'bool2str' in training_set_metadata[of_name]):
+                                class_name = training_set_metadata[of_name][
+                                    'bool2str'][i]
+                            else:
+                                class_name = 'True' if i == 1 else 'False'
```
tgaddair, line 93: @w4nderlust here's where we assign the column name to `_probabilities_False` and `_probabilities_True`.

w4nderlust:
> If the type is bool: `_probabilities_False` and `_probabilities_True`
> If the type is string (yes/no): `_probabilities_no` and `_probabilities_yes`
> 
> Does that work? `_probabilities_0` and `_probabilities_1` are also doable, but I felt True/False was more clear.

that's actually even better.

Issue #1028, opened 2020-11-28
Loading pretrained model when features change


mangleddata:
I had previously opened https://github.com/uber/ludwig/issues/1009 (resolved now, thanks!) - but found the following behaviour.

Run 1: Have features_{0..100} and have a trained model.
Run 2: I now have a new feature feature_101 added to the same data set and have the new features added to the model file (all else being the same). Essentially added one more column. Now using the pretrained model in run 1 for run 2 still seems to run - but from what I could tell, it drops the newly added feature (I don't see the feature in the input_features array in description.json)

Is this expected ?

w4nderlust:
Yes, right now there is no mechanism in Ludwig to load a model trained with a different model definition (even if only different by one feature) as a starting point for another model.
What happens in your case is that the model definition of the model you are loading is what ends up being used instead.
So yes, this is expected behavior right now.

I guess we can turn this into a feature request.
If I understand correctly (please confimr) what you would like to happen here is that you would want to load the pieces of the previous model that match with the new model definition ( encoders of feature 0-100 and decoder of the output feature) and randomly initialize the parts that don't match(encoder of feature 101 and the combiner).
This is interesting and something we may consider in the future.

mangleddata:
What you wrote is exactly what I wanted to do! Thanks for considering the feature request.

jenishah:
We can use `model.get_weights()` and `model.set_weights()` to keep the previous weights for the new model. This can be used if the architecture does not change, but only the number of hidden units.  
@w4nderlust, if you think this can be used, please let me know.  I would like to work on this.

w4nderlust:
@jenishah thanks! Yes we can do that, the issue are some layers that may end up being different depending on the architecture, so there needs to be something like a check that figures out for ech encoder combiner and decoder if the name match and if the size matches and only in that case load the weights and communicate with modules have ben loaded and which have not been loaded.

jenishah:
@w4nderlust will include these factors while iterating. Thanks!

camaya7:
Hi @jenishah, thanks for posting your tip. This is a bit outdated but do you happen to remember where `model.get_weights()`and `model.set_weights()` can be modified? I'm having trouble finding them. 

@w4nderlust thank you for such amazing attention to your users! Any guidance on how to transfer trained model weights to a new model with a new dataset's model definition?

Thanks in advance.

w4nderlust:
[...]

PR #1048, opened 2020-12-09
Transfer weights from a pretrained model when the a new feature is added

jenishah:
- Feature for issue https://github.com/uber/ludwig/issues/1028
- There is a pretrained model with `n` number of features, and the current model has `n+1` number of features. Load weights from the pretrained model onto the current model.

@w4nderlust this PR is not meant to be the final PR, but I wanted to show if this is the way you would like to go forward. Please suggest changes in terms of API. Currently, I have modified the `train` method.  I will add the tests once the overall idea is approved.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -21,6 +21,44 @@
     Date last modified: 5/21/2019
     Python Version: 3+
 """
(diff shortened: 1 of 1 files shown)


jenishah:
@w4nderlust can you please share your comment on this method?

PR #1056, opened 2020-12-16
FIX: fail in reading xls/xlsx data sources

### LINKED ISSUE #1055: Use openpyxl instead of xlrd for reading newer Excel files
With the release of xlrd 2, support for newer Excel formats like xlsx has been dropped.  It is recommended to use openpyxl instead and to only use xlrd for legacy Excel formats.

See: https://stackoverflow.com/questions/65254535/xlrd-biffh-xlrderror-excel-xlsx-file-not-supported

jimthompson5802:
# Code Pull Requests

Fix #1055 

Dependent package `xlrd` changed the excel formats it supports.  This PR resolves the errors caused by that change.  Other changes in this PR
* update to unit test to test reading both `.xls` and `.xlsx` file formats
* update `requirements.txt` to remove pinning of `xlrd` package.

### DIFF
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -136,7 +136,12 @@ def read_jsonl(data_fp, df_lib):
 
 
 def read_excel(data_fp, df_lib):
-    return df_lib.read_excel(data_fp)
+    fp_split = os.path.splitext(data_fp)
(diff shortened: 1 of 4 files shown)


tgaddair (approved):
LGTM!

PR #1068, opened 2020-12-24
compare_performance documenting example usage

modernlearner:
Adds documentation of how to use the `compare_performance` visualization with two models.

*TODO: Generate the docs and add them to this PR using `mkdocs/code_docs_autogen.py`*

### DIFF
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -1354,11 +1354,11 @@ def compare_performance(
     # Example usage:
     
     ```python
-    model_a = LudwigModel(config,logging_level=logging.INFO)
-    model_a.train(dataset=dataset)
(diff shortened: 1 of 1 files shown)


w4nderlust:
It's our bad, we renamed function to "eval*" but in the visualization we still refer to their outputs as "test_statistics". We will replace with "eval_statistics". In the meantime, for this usafe, you need to use eval_statistics" for both a and b models, they can be obtained by either training and then running evaluate or by loading a pretrained model and running evaluate.
Could you please fix it?

modernlearner:
Ah okay, I updated the documentation to reflect what you suggested @w4nderlust let me know if further changes are needed. Thanks!

PR #1069, opened 2020-12-25
Refactor if statements and unused variables

milyiyo:
This pull request aims to refactor some `if` statements that can be joined and delete variables that were created but not used.


w4nderlust:
thanks for this cleanup :)

Issue #1070, opened 2020-12-28
TypeError: Failed to convert object of type <class 'list'> to Tensor.


Bhavesh-Joshi:
**Describe the bug**
An error occurred while saving model with method **save_savedmodel**

**TypeError: Failed to convert object of type <class 'list'> to Tensor. Contents: [None, -1]. Consider casting elements to a supported type.**

**To Reproduce**
1. Go to 'https://gist.github.com/Bhavesh-Joshi/4db376e4aa6a2ef808c4e54618acc987#file-vqa_train-py'
2. Run vqa_train.py
3. See an error

**Expected behavior**
The model should be saved.

**Environment :**
 - OS: Ubuntu
 - Version 20.04.1 LTS
- Python version 3.6
- Ludwig version 0.3

**Additional context**
Just want to add one scenario which I observed is the error only occurred when I train the model with the image dataset. With another kind of data like text, numbers it works fine.

w4nderlust:
Thanks @Bhavesh-Joshi will look into it.

jimthompson5802:
@Bhavesh-Joshi   I wanted to follow-up with you re: the error you encountered.  First thank for the information you provided.  Using that information  I was able to receate the error you are seeing.  Although I'm still doing research into root cause to figure out a fix, I wanted to provide you a quick status update.  

Lastly, thank you for taking the time to report the issue and helping us improve Ludwig.

jimthompson5802:
@Bhavesh-Joshi  I forgot to ask in the previous posting, if it is possible to get a copy of the image you used `Antenna.jpg`?

Bhavesh-Joshi:
![Antenna](https://user-images.githubusercontent.com/29869080/104283455-8ff71000-54d6-11eb-9cce-32868b9c637c.jpg)

Thanks for the update the @jimthompson5802. Please find the attached Antenna.jpg

PR #1074, opened 2021-01-13
Fix: failure in save_savedmodel() call with input image feature

jimthompson5802:
# Code Pull Requests

Fix Issue #1070 with reported error `TypeError: Failed to convert object of type <class 'list'> to Tensor. Contents: [None, -1]. Consider casting elements to a supported type.`

Ran advocates test to ensure restored model weights matched original trained model weights.

### DIFF
--- a/ludwig/decoders/sequence_decoders.py
+++ b/ludwig/decoders/sequence_decoders.py
@@ -144,7 +144,8 @@ class SequenceGeneratorDecoder(Layer):
             logger.debug('  {}'.format(self.decoder_rnncell))
 
     def build(self, input_shape):
-        print(">>>>>>>>>>>entered build()")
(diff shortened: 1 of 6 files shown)


jimthompson5802:
Pushed up latest commits to share where I am on this work.  PR is not ready for prime time yet, many of the unit tests will fail.

One of the issues relate the the `SequenceGeneratorDecoder`.  Right now it is not clear that the `build()` method for this class gets invoked.  I'm still tracking down this issue.   An illustration of this issue is with the `test_experiment.py::test_experiment_seq_seq` unit test.  One possible explanation is thatl run-time control to the `SequenceGeneratorDecoder` does not pass through the `call()` method, instead control flows through `_logits_training()`.

jimthompson5802:
Restarting a new PR.

PR #1080, opened 2021-01-17
Abstracted Horovod params into the Backend API

tgaddair:
This PR is a precursor to adding a the Ray and Dask backends.


ludwig/collect.py
```diff
@@ -321,11 +324,10 @@ def cli_collect_activations(sys_argv):
...
     parser.add_argument(
-        '-uh',
-        '--use_horovod',
-        action='store_true',
-        default=None,
-        help='uses horovod for distributed training'
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 330: I guess the default is None if we don't specify it, but maybe being explicit that the default is the local one may be easier for the users and for the help message?

ludwig/evaluate.py
```diff
@@ -242,11 +243,10 @@ def cli(sys_argv):
...
     parser.add_argument(
-        '-uh',
-        '--use_horovod',
-        action='store_true',
-        default=None,
-        help='uses horovod for distributed training'
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 249: same as above

ludwig/experiment.py
```diff
@@ -532,11 +533,10 @@ def cli(sys_argv):
...
     parser.add_argument(
-        '-uh',
-        '--use_horovod',
-        action='store_true',
-        default=None,
-        help='uses horovod for distributed training'
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 539: same as above

ludwig/features/category_feature.py
```diff
@@ -394,12 +393,7 @@ def postprocess_predictions(
         postprocessed = {}
         name = self.feature_name
 
-        npy_filename = None
-        if is_on_master():
```
w4nderlust, line 398: here the logic was that each worker was actually doing postprocessing so they would override each other's outputs if they were on the same machine (like one machine with multiple gpus). I'm not sure about how this issue is addressed now.

w4nderlust: Replicated for all output features

ludwig/hyperopt_cli.py
```diff
@@ -365,11 +366,10 @@ def cli(sys_argv):
...
     parser.add_argument(
-        "-uh",
-        "--use_horovod",
-        action="store_true",
-        default=False,
-        help="uses horovod for distributed training",
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 372: same as above

ludwig/models/ecd.py
```diff
@@ -271,6 +278,14 @@ def recurse_weights(model, prefix=None):
 
         return weights
 
+    def __setstate__(self, newstate):
```
w4nderlust, line 281: trying to get a better understanding of __setstate__ and __reduce__. Why are they needed? How are they used?

ludwig/models/predictor.py
```diff
@@ -10,17 +11,61 @@
...
+            dataset,
+            bucketing_field=None
+    ):
+        raise NotImplementedError()
+
+    # Remote implementations may override this
+    def shutdown(self):
+        pass
+
+    # Functions needed to treat Trainer as a context manager
```
w4nderlust, line 60: Trainer -> Predictor

ludwig/models/trainer.py
```diff
@@ -191,8 +236,8 @@ def __init__(
         self.batch_size = batch_size
         self.eval_batch_size = batch_size if eval_batch_size < 1 else eval_batch_size
         self.bucketing_field = bucketing_field
-        self.validation_field = validation_field
-        self.validation_metric = validation_metric
+        self._validation_field = validation_field
```
w4nderlust, line 239: why are _ added here?

ludwig/predict.py
```diff
@@ -216,11 +217,10 @@ def cli(sys_argv):
...
     parser.add_argument(
-        '-uh',
-        '--use_horovod',
-        action='store_true',
-        default=None,
-        help='uses horovod for distributed training'
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 223: same as above

ludwig/preprocess.py
```diff
@@ -238,6 +240,12 @@ def cli(sys_argv):
              'to a random number generator: data splitting, parameter '
              'initialization and training set shuffling'
     )
+    parser.add_argument(
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 247: same as above

ludwig/train.py
```diff
@@ -375,11 +376,10 @@ def cli(sys_argv):
...
     parser.add_argument(
-        '-uh',
-        '--use_horovod',
-        action='store_true',
-        default=None,
-        help='uses horovod for distributed training'
+        "-b",
+        "--backend",
+        help="specifies backend to use for parallel / distributed execution",
+        choices=ALL_BACKENDS,
```
w4nderlust, line 382: same as above

requirements.txt
```diff
@@ -12,6 +12,7 @@ PyYAML>=3.12
 absl-py
 kaggle
 requests
+tables
```
w4nderlust, line 15: why is this required?

ludwig/features/category_feature.py
tgaddair: This is handled in the `postprocess` function which calls into these output feature functions here: https://github.com/ludwig-ai/ludwig/blob/backend_refactor/ludwig/data/postprocessing.py#L35

This saves us from having to pass around the `backend` to all these features just for checking this one condition, and removes the duplicate code.

ludwig/collect.py
tgaddair: It's usually the local backend, but the tricky thing is that if the user runs with Horovod like: `horovodrun ludwig train ...` then the backend will be Horovod.  We can specify this in the help message if you think it would add clarity, but I think we should leave the value itself as `None` for this reason.

ludwig/models/trainer.py
tgaddair: These needed to be made private so that we could hide the public interface behind a property here:

https://github.com/ludwig-ai/ludwig/blob/backend_refactor/ludwig/models/trainer.py#L819

This is important for the upcoming Ray backend, which needs to override this behavior.

requirements.txt
tgaddair: This is actually needed for the Dask PR, if I remember correctly, so I'll remove this for now and add it in that PR.

ludwig/collect.py
w4nderlust: Oh you are right, forgot about that. Yes specifying in the message that horovodrun makes it autodetect horovod, and that this overrides the autodetection is a good idea for clarity.

requirements.txt
w4nderlust: not a huge deal, but moving it will likely keep things cleaner

tgaddair: Fixed.

PR #1082, opened 2021-01-19
updating hyperopt sampling functions to handle list values

ANarayan:
This PR updates the RandomSampler and PySOTSampler to handle categorical values of type list. Also updates the hyperopt tests to check this functionality

### DIFF
--- a/ludwig/hyperopt/sampling.py
+++ b/ludwig/hyperopt/sampling.py
@@ -295,7 +295,7 @@ class PySOTSampler(HyperoptSampler):
     def update(self, sampled_parameters: Dict[str, Any], metric_score: float):
         for key in sampled_parameters:
             if key in self.cat_params_values_types:
-                if type(sampled_parameters[key]) == list:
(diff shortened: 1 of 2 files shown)


ludwig/hyperopt/sampling.py
```diff
@@ -282,7 +295,11 @@ def sample(self) -> Dict[str, Any]:
     def update(self, sampled_parameters: Dict[str, Any], metric_score: float):
         for key in sampled_parameters:
             if key in self.cat_params_values_types:
-                sampled_parameters[key] = str(sampled_parameters[key])
+                if type(sampled_parameters[key]) == list:
```
w4nderlust, line 298: Maybe here to be more consistent we can say `is not in {bool, int, float, str}`? This will make it work also for dicts and other json-serializable types.

ANarayan: good point. will make the necessary edits.

PR #1088, opened 2021-02-01
Added scheduler support to Ray Tune hyperopt and fixed GPU usage

tgaddair:
Ray will set `CUDA_VISIBLE_DEVICES` for each trial automatically, so we do not need to manually restrict GPUs in Ludwig.

See the Ray Tune [documentation](https://docs.ray.io/en/latest/tune/user-guide.html#using-gpus) for details.

cc @ANarayan.


ANarayan:
I think we may want to encode the configs in tune.report [here](https://github.com/ludwig-ai/ludwig/blob/5ad60b81f5b552c504b5d0b280b3387b1925086c/ludwig/hyperopt/execution.py#L713) using json.dumps() rather than str(config). I am having a hard time parsing the returned training_stats, config, and parameters using json.loads() and other decoding methods.

PR #1090, opened 2021-02-04
Added Ray remote backend and Dask distributed preprocessing

tgaddair:


tgaddair:
cc @richardliaw @clarkzinzow would be great if one or both of you could also take a look (should be ready for review at this point).

richardliaw:
Nice! I'll add this to my TODOs and will get back to you Monday.

ludwig/data/dataframe/dask.py
```diff
@@ -0,0 +1,109 @@
...
+
+class DaskEngine(DataFrameEngine):
+    def __init__(self, parallelism=None):
+        self._parallelism = parallelism or multiprocessing.cpu_count()
+
+    def empty_df_like(self, df):
+        # Our goal is to preserve the index of the input dataframe but to drop
+        # all its columns. Because to_frame() creates a column from the index,
+        # we need to drop it immediately following creation.
+        return df.index.to_frame(name=TMP_COLUMN).drop(columns=[TMP_COLUMN])
```
tgaddair, line 46: @clarkzinzow let me know if there's a better way to do this

clarkzinzow: Would something like `pd.DataFrame(index=df.index)` work? Assuming that this is supposed to return a Pandas dataframe, if it's supposed to return a Dask dataframe then passing that to `dd.from_pandas()` would do the trick.

tgaddair: Both the input and output dataframes are Dask in this case.  It's an interesting situation because we have a "source" DF and a "sink" in our processing.  We take the columns from the source, transform them, and they become the columns of the sink (one-to-many, so one source column may become multiple columns in the sink DF).  That's why we create this empty DF, but in order to align the columns correctly, we need to preserve the index (hence this hacky looking tmp column).

clarkzinzow: Cool, got it. If you're hoping to avoid the grossness of creating and immediately dropping a temporary column, it seems as if `dd.from_pandas(pd.DataFrame(index=df.index))` would do the trick.

tgaddair: Oh interesting, thanks I'll give that a shot!

ludwig/backend/ray.py
```diff
@@ -0,0 +1,174 @@
...
+    # Our goal is to maximize the number of training resources we can
+    # form into a homogenous configuration. The priority is GPUs, but
+    # can fall back to CPUs if there are no GPUs available.
+    key = 'GPU' if use_gpu else 'CPU'
+
+    # Bucket the per node resources by the number of the target resource
+    # available on that host (equivalent to number of slots).
+    buckets = defaultdict(list)
+    for node_resources in resources:
+        buckets[int(node_resources.get(key, 0))].append(node_resources)
```
richardliaw, line 48: I think moving forward, we want to move away from the `ray.nodes()` API.

ludwig/backend/ray.py
```diff
@@ -0,0 +1,174 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def get_horovod_kwargs():
+    resources = [node['Resources'] for node in ray.state.nodes()]
+    use_gpu = int(ray.cluster_resources().get('GPU', 0)) > 0
+
+    # Our goal is to maximize the number of training resources we can
+    # form into a homogenous configuration. The priority is GPUs, but
```
richardliaw, line 40: is it possible to support non-homogenous configurations?

ludwig/backend/ray.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+    def shutdown(self):
+        self.executor.shutdown()
+
+
+class RayBackend(RemoteTrainingMixin, Backend):
+    def __init__(self, horovod_kwargs=None):
+        super().__init__()
+        self._df_engine = DaskEngine()
```
richardliaw, line 140: do we need to set the dask scheduler to be ray somewhere?

ludwig/backend/ray.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def load(self):
+        obj = self.cls(*self.args)
+        obj.__setstate__(ray.get(self.state))
+        return obj
+
+
+class RayTrainer(BaseTrainer):
+    def __init__(self, horovod_kwargs, trainer_kwargs):
+        setting = RayExecutor.create_settings(timeout_s=30)
```
richardliaw, line 76: should we expose more settings here?

ludwig/backend/ray.py
```diff
@@ -0,0 +1,174 @@
...
+        self._df_engine = DaskEngine()
+        self._horovod_kwargs = horovod_kwargs or {}
+        self._tensorflow_kwargs = {}
+
+    def initialize(self):
+        try:
+            ray.init('auto', ignore_reinit_error=True)
+        except ConnectionError:
+            logger.info('Initializing new Ray cluster...')
+            ray.init()
```
richardliaw, line 149: you could also `ray.init(ignore_reinit_error=True)` here too

ludwig/data/dataframe/dask.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def parallelize(self, data):
+        return data.repartition(self.parallelism)
+
+    def persist(self, data):
+        return data.persist()
```
clarkzinzow, line 52: I should point out that, contrary to Dask Distributed's `client.persist()` implementation which returns a Dask DataFrame containing references to futures, Dask-on-Ray currently uses plain Dask's default behavior of returning a Dask DataFrame containing the fully materialized data. We have an [open issue](https://github.com/ray-project/ray/issues/14227) for addressing this in the next sprint or two, but because of Dask's `df.persist()` implementation, the API might have to be slightly different (e.g. a `ray_persist()` function that we provide).

clarkzinzow: Good news here! Proper `dask.persist()` support for the Dask-on-Ray scheduler has been implemented, and should just work as expected: https://github.com/ray-project/ray/pull/14294

ludwig/backend/ray.py
tgaddair: From Horovod's perspective: definitely. I think we would just need to rework the RayExecutor interface a little. Namely, you could imagine the user just saying `num_gpus=N`, then we place however many processes per host we need to in order to meet this request (so no more num_hosts or num_slots params in this mode).

ludwig/backend/ray.py
tgaddair: Good catch! This must have gotten dropped during a merge at some point. Fixed.

ludwig/data/dataframe/dask.py
tgaddair: Thanks! Adding a reference to this issue. Looks like it's closed, what the resolution to fully support the expected semantics of `df.persist()`? Any timeline on when this will be released?

ludwig/backend/ray.py
tgaddair: Good point. I definitely want to make this more configurable via the Ludwig YAML or similar. I think we can do this in a follow-up to allow specifying the backend in a YAML file, so I will add a TODO for now. Does that seem reasonable to you?

ludwig/data/dataframe/dask.py
tgaddair: Looks like my comment came a little late, I will remove the link to the ticket then. This is great, thanks!

ludwig/backend/ray.py
richardliaw: yeah sounds good!

ludwig/backend/ray.py
richardliaw: I think if we get HorovodRay to just support `num_gpus=N`, then we wouldn't need to do the accounting. 

Generally, we're trying to move towards a more 'serverless' abstraction where programmers think about 'resources' rather than 'nodes'.

richardliaw: I think it's safe to keep this here for now (at least until we support num_gpus in horovod)

ludwig/backend/ray.py
richardliaw: yeah, that sounds good. Let me make an issue on horovod then!

ludwig/backend/ray.py
tgaddair: Sweet, I'll add a TODO referencing this issue.

ludwig/backend/ray.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_dask_kwargs():
+    # TODO ray: select this more intelligently,
+    #  must be greather than or equal to number of Horovod workers
+    return dict(
+        parallelism=int(ray.cluster_resources()['CPU'])
```
tgaddair, line 41: @clarkzinzow does this make sense as the default `repartition` value?  One partition per CPU? Not sure if there's a more reasonable heuristic for this. The one restriction we have is that for Petastorm, we must have at least one row group per Horovod worker, and the safest way to guarantee this at the moment is to repartition the dataframe.

clarkzinzow: That's the typical heuristic, yes, under the soft constraint of those chunks/partitions fitting nicely into each worker's memory.

ludwig/data/dataframe/dask.py
```diff
@@ -0,0 +1,112 @@
...
+        dataset_parquet_fp = os.path.join(cache_dir, f'{tag}.parquet')
+
+        # Workaround: https://issues.apache.org/jira/browse/ARROW-1614
+        features = get_combined_features(config)
+        for feature in features:
+            name = feature[NAME]
+            proc_column = feature[PROC_COLUMN]
+            reshape = training_set_metadata[name].get('reshape')
+            if reshape is not None:
+                dataset[proc_column] = self.map_objects(dataset[proc_column], lambda x: x.reshape(-1))
```
w4nderlust, line 81: Curious about this, what is it a work around for? (probably worth adding a comment about it too)

ludwig/data/dataset/parquet.py
```diff
@@ -0,0 +1,83 @@
...
+            feature[PROC_COLUMN]: list((-1, *training_set_metadata[feature[NAME]]['reshape']))
+            for feature in features
+            if 'reshape' in training_set_metadata[feature[NAME]]
+        }
+
+    def get(self, feature_name, sample):
+        t = getattr(sample, feature_name)
+        reshape_dim = self.reshape_features.get(feature_name)
+        if reshape_dim is not None:
+            t = tf.reshape(t, reshape_dim)
```
w4nderlust, line 45: I guess my previous question had to do with this, right? Tensors with multiple ranks get squashed and then put back in shape, right?

tests/integration_tests/test_ray.py
```diff
@@ -0,0 +1,134 @@
...
+@pytest.fixture
+def ray_start_4_cpus():
+    res = ray.init(num_cpus=4)
+    try:
+        yield res
+    finally:
+        ray.shutdown()
+
+
+def train_with_backend(backend, config, dataset=None, training_set=None, validation_set=None, test_set=None):
```
w4nderlust, line 52: if it's the same as the one in the dask test file, maybe we could put them in the util file?

w4nderlust (approved):
Absolutely amazing job with this! Added couple very minor comments

ludwig/data/dataframe/dask.py
tgaddair: Basically, PyArrow cannot serialize the data as a multi-dimensional tensors, so we need to store the shape so when we read it back, we can reshape it into its correct form. Will add a comment to this effect.

ludwig/data/dataset/parquet.py
tgaddair: Yes, exactly, I can add a comment here.

tests/integration_tests/test_ray.py
tgaddair: Good catch. Fixed.

PR #1111, opened 2021-03-05
Fix serve.py adding a try around the form parsing

w4nderlust:


ludwig/serve.py
```diff
@@ -71,8 +71,15 @@ def check_health():
 
     @app.post('/predict')
     async def predict(request: Request):
-        form = await request.form()
-        files, entry = convert_input(form)
+        try:
+            form = await request.form()
+            files, entry = convert_input(form)
+        except Exception as e:
+            logger.error("Failed to parse batch_predict form: {}".format(
```
tgaddair, line 78: `str(e)` drops the stack trace. An easy way to log the exception would be with [logger.exception](https://docs.python.org/3/library/logging.html#logging.exception):

```
logger.exception("Failed to parse batch_predict form")
```

Though since the current style also uses logger.error, no worries if you would rather stick to the file convention.

tgaddair (approved):

PR #1117, opened 2021-03-12
Adding register_trainable logic to RayTuneExecutor

ANarayan:
This PR registers the trainable function passed to `tune.run` using the `ray.tune` util function `register_trainable`. This fix is necessary in order ensure that the trainable function is accessible to all Ray processes in a cluster. Moreover, when running several parallel `ludwig.hyperopt` calls on a given ray cluster, it is necessary to have a unique trainable function for each experiment. Doing so prevents the underlying objects of different experiments (i.e. `self.decode_ctx`) from being shared. A unique trainable function name is created by generating a hash of the experiment config.

cc: @tgaddair


tgaddair (approved):
Nice! LGTM.

PR #1125, opened 2021-03-18
Update Titanic and serve examples so they work with Kaggle based dataset

mejackreed:
#1021 added dataset download from Kaggle. This pull request is a follow on so that the `serve` example can work out of the box for users.

There is also an additional documentation commit (f4446d9) that might be useful for users downloading data from Kaggle for the first time.


tgaddair (approved):
Nice! Thanks for the PR.

PR #1128, opened 2021-03-21
Perform inference with Dask when using the Ray backend

tgaddair:
This PR introduces a new `PartitionedDataset` constructed by the backend when calling `create_inference_dataset` within `preprocess_for_prediction`. Unlike other datasets, the `PartitionedDataset` runs the batch prediction in parallel on each partition of the distributed DataFrame.

A consequence of this rearchitecture is that instead of `Predictor.batch_predict` returning a `dict`, it now returns a DataFrame (as the dataset could be large enough to not fit into memory). Therefore, all the postprocessing now happens on the DataFrame with a flat structure, instead of the nested dict structure.

There a few API limitations to the current implementation:

1. No activation collection for Ray or Dask (usually you wouldn't need to run this in a massive dataset, anyway). We can add this in the future.
2. No batch evaluation when using Dask backend. The evaluation requires some tricky state management to collect the metrics across partitions, which we can do using Ray actors (though it is not ideal). We should rearchitect this in the future so that metrics collection can occur after we run the prediction.
3. No overall stats collection for partitioned datasets. It's not impossible, but will require more work, and is not a top priority at this time.

Another limitation we should address soon is that this implementation does not optimize for GPU inference. Because it runs a separate model replica on each partition, we have no way of preventing multiple model replicas from landing on the same GPU. With standalone Dask, this is possible when configuring your workers (see [here](https://docs.dask.org/en/latest/gpu.html#specifying-gpus-per-machine)). However, this approach doesn't work for Ray at the moment. One possibility is to create workers for each GPU and then use `ray.remote` to send partitions to GPU workers. If we go this route, we need be smart with how we route partitions to model replicas to ensure minimal network overhead. We can do this in a follow-up PR.

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -25,7 +25,7 @@ from ray.exceptions import RayActorError
 from ray.util.dask import ray_dask_get
 
 from ludwig.backend.base import Backend, RemoteTrainingMixin
-from ludwig.constants import NAME
+from ludwig.constants import NAME, PARQUET
 from ludwig.data.dataframe.dask import DaskEngine
 from ludwig.data.dataset.partitioned import PartitionedDataset
 from ludwig.models.predictor import BasePredictor, Predictor, get_output_columns
@@ -193,11 +193,16 @@ class RayPredictor(BasePredictor):
 
 
 class RayBackend(RemoteTrainingMixin, Backend):
-    def __init__(self, horovod_kwargs=None):
-        super().__init__()
+    def __init__(self, horovod_kwargs=None, data_format=PARQUET, **kwargs):
+        super().__init__(data_format=data_format, **kwargs)
         self._df_engine = DaskEngine()
         self._horovod_kwargs = horovod_kwargs or {}
         self._tensorflow_kwargs = {}
+        if data_format != PARQUET:
+            raise ValueError(
+                f'Data format {data_format} is not supported when using the Ray backend. '
+                f'Try setting to `parquet`.'
+            )
 
     def initialize(self):
         try:
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -21,14 +21,14 @@ from collections import defaultdict, OrderedDict
 import dask
 import ray
 from horovod.ray import RayExecutor
-from ray.exceptions import RayTaskError, RayActorError
+from ray.exceptions import RayActorError
 from ray.util.dask import ray_dask_get
 
 from ludwig.backend.base import Backend, RemoteTrainingMixin
 from ludwig.constants import NAME
 from ludwig.data.dataframe.dask import DaskEngine
 from ludwig.data.dataset.partitioned import PartitionedDataset
-from ludwig.models.predictor import BasePredictor, Predictor, EXCLUE_PRED_SET
+from ludwig.models.predictor import BasePredictor, Predictor, get_output_columns
 from ludwig.models.trainer import BaseTrainer, RemoteTrainer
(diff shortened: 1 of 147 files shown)


ifokeev:
so cool feature

ludwig/data/postprocessing.py
```diff
@@ -36,15 +39,35 @@ def postprocess(
...
+        predictions = output_feature.postprocess_predictions(
+            predictions,
             training_set_metadata[of_name],
             output_directory=output_directory,
-            skip_save_unprocessed_output=skip_save_unprocessed_output
+            backend=backend,
         )
-    return postprocessed
+
+    # Save any new columns but do not save the original columns again
```
w4nderlust, line 54: why do we need this two times? Can't we do it only once at the end?
Will likely also remove the need for `saved_keys`

ludwig/features/category_feature.py
```diff
@@ -388,61 +393,44 @@ def postprocess_predictions(
...
 
-            if not skip_save_unprocessed_output:
-                np.save(
-                    npy_filename.format(name, 'predictions_top_k'),
-                    preds_top_k
+        top_k_col = f'{self.feature_name}_predictions_top_k'
+        if top_k_col in predictions:
+            if 'idx2str' in metadata:
+                predictions[top_k_col] = backend.df_engine.map_objects(
+                    predictions[top_k_col],
```
w4nderlust, line 429: not 100% sure of how the content of the top k column looks like, but before each was a list of values, so in line 431 there was a loop for each value, while here it sems like there's only a loop over preds and not inside each pred

ludwig/backend/ray.py
```diff
@@ -141,35 +144,109 @@ def shutdown(self):
...
+@ray.remote
+class MetricCollector(object):
+    def __init__(self):
+        self.metrics = []
+
+    def add_metrics(self, metrics):
+        self.metrics.append(metrics)
+
+    def collect(self):
+        return sum_dicts(
```
w4nderlust, line 156: This still does not convince me 100%, as most of the metrics summed / averaged across partitions do no make much sense. maybe we can deactivate evaluation also in the ray case, think it through and then reactivate it.

ludwig/data/postprocessing.py
tgaddair: At least in the original implementation, existing columns were written to numpy before postprocessing, while new columns were written after postprocessing. This change is meant to preserve that behavior, but we can definitely remove the earlier lines if that change is acceptable.

ludwig/features/category_feature.py
tgaddair: Good catch, every row here should be a `pred_top_k`.  Fixed.

ludwig/backend/ray.py
tgaddair: I agree, though this implies it is also currently broken in the normal Horovod case as well:

https://github.com/ludwig-ai/ludwig/blob/51e357dca8b01cb3b9e687b13e1d270329c5bddf/ludwig/models/predictor.py#L289

I believe once we adopt the Metrics package for PyTorch that natively supports distributed aggregation this problem will be addressed. Until then, I agree that disabling makes sense.

w4nderlust: I think it depends on what we are aggregating, it it's accuracies for instance, this wouldn't be correct, it it's aggregation by summing the counts of correct predictions and all predictions and then later diving correct prediction by all predictions, this would be correct. But I believe we are doing hte formaer.

w4nderlust (approved):

PR #1129, opened 2021-03-22
Fixed bug in EthosBinary dataset class and model directory copying logic in RayTuneReportCallback

ANarayan:
This PR fixes two bugs:
1) processes the downloaded ethos binary dataset to be comma separated in oppose to semicolon separated.
2) ensures that the model directory copying process in the `RayTuneReportCallback ` is atomic: prior to this addition, issues arose where files within the model checkpoint directory were being accessed before being completely written.

cc: @tgaddair


ludwig/hyperopt/execution.py
```diff
@@ -722,7 +723,14 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                 if trainer.is_coordinator():
                     with tune.checkpoint_dir(step=progress_tracker.epoch) as checkpoint_dir:
                         checkpoint_model = os.path.join(checkpoint_dir, 'model')
-                        shutil.copytree(save_path, checkpoint_model)
+                        if not os.path.isdir(checkpoint_model):
```
w4nderlust, line 726: Maybe add a comment here about the fact that the previous implementation was copytree (you can leave that line commented) but that that's not atomic and so we have to do this temp + move thing, otherwise in the future we may forget about it and maybe replace it with copytree again :)

ANarayan: great feedback -- addressed in this commit:  here https://github.com/ludwig-ai/ludwig/pull/1129/commits/c924b9f1ceef634538c78648358c8c49d4d12238

tgaddair (approved):
LGTM! Atomic copy is awesome.

PR #1131, opened 2021-03-27
Initial time series implementation

nimz:
This PR contains the following (so far):
1. Enable loading of time series data from column-major as well as row-major sources (specified by a config option). If column-major, all input data must currently be time series (and this is checked), while the row-major case proceeds normally.
2. Enable undersampling time series signals, and doing direct k-step ahead prediction (rather than just predicting the next step). Currently, undersampling simply samples every n datapoints.
3. Refactor numeric feature transformations into `ludwig/features/feature_transform_utils.py`, for ease of use across classes (for now, numerical and time series features).
4. Add RMSE metric.

Example test case instructions:
- Download and unpack hourly weather data from https://www.kaggle.com/selfishgene/historical-hourly-weather-data
- Run `ludwig experiment --dataset temperature.csv --config_file config.yaml` with config.yaml below

config.yaml:
```
preprocessing:
        column_major: True
        splits_in_order: True

input_features:
    -
        name: Los Angeles in
        column: Los Angeles
        type: timeseries
        encoder: rnn
        state_size: 32
        preprocessing:
                normalization: zscore
                timeseries_length_limit: 20
                padding_value_strategy: fill_with_mean
                missing_value_strategy: fill_with_mean
                padding: left
    -
        name: Seattle
        type: timeseries
        encoder: rnn
        state_size: 32
        preprocessing:
                normalization: zscore
                timeseries_length_limit: 20
                padding_value_strategy: fill_with_mean
                missing_value_strategy: fill_with_mean
                padding: left

output_features:
    -
        name: Los Angeles out
        column: Los Angeles
        type: numerical
        preprocessing:
                normalization: zscore
                missing_value_strategy: fill_with_mean
```

@w4nderlust @tgaddair


ludwig/features/timeseries_feature.py
```diff
@@ -103,21 +120,93 @@ def pad(vector):
...
+        last_index = lambda i: max(0, i - n_steps_ahead)
+        ts_vals = timeseries.values.astype(np.float32)
+        # Iff zeros are appended on the right, put most recent values first
+        rev = 1 if padding == 'right' else -1
+        history = [np.zeros(0, dtype=np.float32) if i < n_steps_ahead else \
+            ts_vals[last_index(i)::-undersample_ratio][:length_limit][::rev] \
+            for i in range(len(timeseries))]
+        ts_vectors = backend.df_engine.df_lib.Series(history)
+
+        # Duplication of logic - not ideal. Also, need error handling.
```
w4nderlust, line 163: You should compute the mean or an other transform parameters in `get_feature_meta ` and then from `build_data` pass the relevant paramenters to this function, so at this point you should already know the mean. Also, this is conflating the padding value with the missing value strategy, maybe there's a better way to handle this and also maybe sending a warning to the user that we are overriding the padding value they specified with the mean is a good idea.

nimz: Good point. I've added an explicit "padding_value_strategy" option so that the behavior can be explicit, and updated per your suggestion. (Likely the user would usually want this to be the same as the missing value strategy, but right now they are specified independently.)

Issue #1132, opened 2021-03-28
Dask Scalar is not json serializable when generating dataset metadata on a dask dataframe


zhisbug:
**Describe the bug**
When calling the `precompute_fill_value` during generating the dataset metadata, if the `dataset_df` is a dask.dataframe (which is a requirement when running distributed ludwig on Ray), calling `df.mean()` for FILL_WITH_MEAN returns a symbolic value, which later cannot be serialized into JSON. 

**To Reproduce**
See my repo on running ludwig on ray with the petfinder dataset:
https://github.com/zhisbug/ludwig-petfinder/blob/dask-df/petfinder-adoption-prediction/train.py#L16

**Expected behavior**
Dask dataframe is lazy. We should check whether the dataframe is a dask.dataframe, and if it is, call `.compute()` to calculate the metadata value.

**Environment (please complete the following information):**
 - OS: Ubuntu 18
- Python version: 3.7
- Ludwig version: master

zhisbug:
#1133 should fix it.

tgaddair:
Fixed by #1133.

PR #1133, opened 2021-03-28
check dask dataframe at preprocessing

zhisbug:
# Code Pull Requests
This PR is my fix to issue #1132

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -20,6 +20,7 @@ from abc import ABC, abstractmethod
 
 import numpy as np
 import pandas as pd
+import dask.dataframe as dd
 
 import ludwig
 from ludwig.backend import LOCAL_BACKEND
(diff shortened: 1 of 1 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1173,8 +1174,12 @@ def precompute_fill_value(dataset_df, feature, preprocessing_parameters):
                 'Filling missing values with mean is supported '
                 'only for numerical types',
             )
-        return dataset_df[feature[COLUMN]].mean()
-
+        if isinstance(dataset_df, pd.DataFrame):
```
tgaddair, line 1177: Good catch! To make this more generic, what we can do is pass in the `backend` param from where this function is called in `build_metadata()`. So we can change the signature here to 

```
precompute_fill_value(dataset_df, feature, preprocessing_parameters, backend)
```

Then to get the mean, we just need to call:

```
return backend.df_engine.compute(dataset_df[feature[COLUMN]].mean())
```

Then we can remove the import of Dask at the top (since it is not a required package).

zhisbug: Made the changes! thanks!

tgaddair (approved):
LGTM! Can you rebase from master to fix tests?

PR #1140, opened 2021-03-30
update example-titanic config

ronaldyang:
# Update Multi Model Training in titanic example

- `fill_with_median` in preprocessing is no longer supported in latest ludwig, replace it with `fill_with_mean`
- `python multiple_model_training.py` will now run through with the fix


tgaddair (approved):
Thanks for the fix!

PR #1164, opened 2021-04-23
Added support for cloud object storage (S3, GCS, ADLS, etc.)

tgaddair:
Also consolidated HDF5 and Parquet caching into a single abstraction layer as part of the backend called `CacheManager`. Now, local backends can optionally cache in either HDF5 or Parquet, and the new `DatasetManager` abstraction will do the work of creating the appropriate dataset based on the cache format.

Depends on https://github.com/uber/petastorm/pull/665.

### DIFF
--- a/ludwig/utils/fs_utils.py
+++ b/ludwig/utils/fs_utils.py
@@ -25,9 +25,11 @@ import h5py
 from fsspec.core import split_protocol
 
 
-def get_fs_mode(mode):
+def get_binary_mode(mode):
     if mode == 'r':
         return 'rb'
+    elif mode == 'r+':
+        return 'rb+'
     elif mode == 'w':
         return 'wb'
     return mode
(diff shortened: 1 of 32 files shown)


ludwig/utils/fs_utils.py
```diff
@@ -0,0 +1,123 @@
...
+
+@contextlib.contextmanager
+def open_file(url, *args, **kwargs):
+    of = fsspec.open(url, *args, **kwargs)
+    with of as f:
+        yield f
+
+
+@contextlib.contextmanager
+def download_h5(url):
```
w4nderlust, line 104: why download_h5 and not open_h5?

w4nderlust: also does this work with remote urls? I guess if it's remote fsspec.open_local downloads it first?

w4nderlust (approved):

ludwig/utils/fs_utils.py
tgaddair: `download_h5` will first download the remote file to local, then open it for read-only.

`upload_h5` opens the file for write then uploads to the remote url.

The `open_local` just means to download the file locally before opening, yes, as opposed to streaming. This is primarily for the purpose of caching (if you intend to re-read repeatedly).

PR #1167, opened 2021-04-25
Add KDD Cup 2009 datasets

kanishk16:
Adds the following KDD datasets:
- [x] Appetency
- [x] Churn
- [x] Upselling


ludwig/datasets/kdd_appetency/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+            file_df = -pd.read_csv(
+                os.path.join(self.raw_dataset_path, filename), header=None)[0]
+        elif filetype == 'txt':
+            file_df = pd.read_csv(
+                os.path.join(self.raw_dataset_path, filename), header=None)
+        return file_df
+
+    def process_downloaded_dataset(self, header=0):
+        super().process_downloaded_dataset()
+        unprocessed_df = pd.read_csv(os.path.join(self.processed_dataset_path,
```
w4nderlust, line 58: this should likely be `raw_dataset_path`

ludwig/datasets/kdd_appetency/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+        processed_val_df['target'] = target.iloc[val_idx[0]]
+        processed_val_df['split'] = 1
+        
+        self.data_df = pd.concat([processed_train_df, processed_val_df])
+
+    def load_processed_dataset(self, split) -> Tuple[pd.DataFrame,
+                                                     pd.DataFrame,
+                                                     pd.DataFrame]:
+        if split:
+            if SPLIT in self.data_df:
```
w4nderlust, line 95: the pattern of setting `self.data_df` can be improved. We want the different steps (download, process and load) to be atomic and independent, because if the directories are already crated, they are not repeated, but this breaks that assumption.

ludwig/datasets/kdd_appetency/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+                test_set = self.test_df[self.test_df[SPLIT] == 2]
+            return training_set, test_set, val_set
+        else:
+            raise ValueError("The dataset does not have splits, "
+                             "load with `split=False`")
+
+
+def process_categorical_features(df, categorical_features):
+    for i in categorical_features:
+        df.iloc[:, i].fillna("?", inplace=True)
```
w4nderlust, line 108: you can use the empty string instead

ludwig/datasets/kdd_appetency/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+            return training_set, test_set, val_set
+        else:
+            raise ValueError("The dataset does not have splits, "
+                             "load with `split=False`")
+
+
+def process_categorical_features(df, categorical_features):
+    for i in categorical_features:
+        df.iloc[:, i].fillna("?", inplace=True) 
+        df.iloc[:, i].apply(lambda x: to_float_str(x))
```
w4nderlust, line 109: why do you need this?

ludwig/datasets/kdd_appetency/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+
+def process_numerical_features(df, categorical_features):
+    columns_to_impute = []
+    for i, column in enumerate(df.columns):
+        if i not in categorical_features and pd.isnull(df[column]).any():
+            columns_to_impute.append(column)
+            
+    for column_name in columns_to_impute:
+        df[column_name + "_imputed"] = pd.isnull(df[column_name]).astype(float)
+        df[column_name].fillna(0, inplace=True)
```
w4nderlust, line 128: Ludwig does this imputation at preprocessing time, no need to do it here, you can leave the empty string

ludwig/datasets/mixins/download.py
```diff
@@ -163,3 +163,35 @@ def download_raw_dataset(self):
     @property
     def download_url(self):
         return self.config["download_urls"]
+
+
+class ZipandUncompressedFileDownloadMixin:
```
w4nderlust, line 168: pandas can load a zipped file, so you can use the `UncompressedFileDownloadMixi` without the need for this new one.

ludwig/datasets/kdd_appetency/__init__.py
kanishk16: I believe `processed_dataset_path` works fine as `process_download_dataset` has been called to concatenate the train and test data into the csv file.

ludwig/datasets/kdd_appetency/__init__.py
kanishk16: I'm not sure about this. Perhaps, I'm missing something but this was part of the preprocessing they followed (c.f. - https://github.com/catboost/benchmarks/blob/master/quality_benchmarks/prepare_appetency_churn_upselling/prepare_appetency_churn_upselling.ipynb).

w4nderlust (approved):

PR #1176, opened 2021-05-07
Fix build_mateadata to use COLUMN instead of NAME + CacheManager file names

w4nderlust:
Fix build_mateadata to use COLUMN instead of NAME and by populating proc_feature_to_metadata.
Also fixes the way the cache manager calculates cache file names.


w4nderlust:
Superseeded by https://github.com/ludwig-ai/ludwig/pull/1189

PR #1186, opened 2021-05-26
Added schema validation to config files

tgaddair:
The purpose of this PR is to catch typos or other misconfigurations in the Ludwig config.yaml before training.

One common occurrence is that users will set a parameter in the config expecting it to do something, but because Ludwig does not validate the schema and will assume reasonable defaults, this can result in errors that go undetected.

This PR introduces schema validation based on jsonschema that is called at LudwigModel creation time. It enforces that certain fields are set, and that certain fields (like preprocessing) do not have any unrecognized subfields.

Note that we do not apply strict subfield checking to every field in the config, in order to make it easier for users to switch between different encoders that take different parameters, for example. We do, however, check that feature types, encoders, tokenizers, and other common parameters all have valid values and that preprocessing in particular does not have unrecognized fields.


ludwig/features/binary_feature.py
```diff
@@ -47,6 +47,12 @@ class BinaryFeatureMixin:
         'fill_value': 0
     }
 
+    preprocessing_schema = {
+        'missing_value_strategy': {'type': 'string', 'enum': MISSING_VALUE_STRATEGY_OPTIONS},
+        'fill_value': {'type': 'integer'},
```
w4nderlust, line 52: this may be also True, False and a string in case of binry features with a vocab like Yes/No

ludwig/features/vector_feature.py
```diff
@@ -43,7 +43,14 @@ class VectorFeatureMixin:
...
     preprocessing_defaults = {
         'missing_value_strategy': FILL_WITH_CONST,
-        'fill_value': ""
+        'fill_value': 0,
+    }
+
+    preprocessing_schema = {
+        'vector_size': {'type': 'integer', 'minimum': 0},
+        'missing_value_strategy': {'type': 'string', 'enum': MISSING_VALUE_STRATEGY_OPTIONS},
+        'fill_value': {'type': 'number'},
```
w4nderlust, line 52: this should likely be a list of numbers, is there a way to enforce it?
Otherwise we can leave `number` but we probably need more logic for making sure we set the fill to a vector of the right size with that number, like `fill_missing(column, vector_size, fill_value): fill_vec = [fill_value] * vector_size; colum.fill_na(fill_vec)`

ludwig/features/binary_feature.py
tgaddair: Added all string possibilities.

ludwig/features/vector_feature.py
tgaddair: Added space-separated list regex.

PR #1187, opened 2021-05-29
Refactored all contrib classes to use the Callback interface

tgaddair:
This change replaces all global contrib handlers with callbacks, removing the handler instances from global state and aligning the API with hyperopt callbacks.


w4nderlust:
@dsblank would you mind giving feedback on this? It changes the architecture a bit wrt how we implemented the contrib back then, but it shouldn't change the functionality.

w4nderlust (approved):

w4nderlust:
@dsblank we are moving forward and merging, let me know if you have opinions, suggestions or concerns though

dsblank:
Thanks! We'll be actively looking at this over the few days. Much appreciated!

PR #1192, opened 2021-06-03
Added export to MLflow pyfunc model format

tgaddair:
This PR adds a new `mlflow.pyfunc` model variant called `LudwigMlflowModel` that allows users to load Ludwig models and use them for prediction alongside other Mlflow model types. When using experiment tracking, the `model` directory will now be uploaded to the run as an Mlflow model at the end of each epoch, instead of simply a directory.

This PR also adds a new `export_mlflow` command to take an existing Ludwig model and convert it to Mlflow or (when `model_name` is provided) register it in the Mlflow model registry.


ludwig/api.py
```diff
@@ -484,64 +484,65 @@ def train(
...
+                                'Best validation model {0} on test set {1}: {2}'.format(
+                                    validation_metric,
+                                    validation_field,
+                                    best_vali_metric_epoch_test_metric
+                                )
+                            )
+                        logger.info(
+                            '\nFinished: {0}_{1}'.format(experiment_name, model_name))
+                        logger.info('Saved to: {0}'.format(output_directory))
+                finally:
```
tgaddair, line 543: Note: this whole block of changes is just adding a `try ... finally` to ensure `on_train_end` is also called in the event of an exception.

Issue #965, opened 2020-10-21
Assertion Error while running hyperopt notebook


Juggernaut1997:
Hi, From last couple of weeks I am exploring Ludwig. While exploring I came across the Hyper-parameter optimization visualization done in Ludwig. So, I forked the repo and start looking into it. While doing that I got an error related to multiprocessing pool which I am not sure what it is. So I am mentioning the error below, please look into it and guide me what's causing this error and how I can resolve it. The notebook I am using is [**hyperopt notebook**](https://github.com/uber/ludwig/blob/master/examples/hyperopt/model_hyperopt_example.ipynb).
> ---------------------------------------------------------------------------
> RemoteTraceback                           Traceback (most recent call last)
> RemoteTraceback: 
> """
> Traceback (most recent call last):
>   File "/usr/lib/python3.6/multiprocessing/pool.py", line 119, in worker
>     result = (True, func(*args, **kwds))
>   File "/usr/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar
>     return list(map(*args))
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/hyperopt/execution.py", line 202, in _run_experiment
>     train_stats, eval_stats = run_experiment(**hyperopt_dict)
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/hyperopt/execution.py", line 703, in run_experiment
>     debug=debug,
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/api.py", line 1035, in experiment
>     debug=debug,
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/api.py", line 397, in train
>     random_seed=random_seed
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/data/preprocessing.py", line 1208, in preprocess_for_training
>     random_seed=random_seed
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/data/preprocessing.py", line 169, in preprocess_for_training
>     random_seed=random_seed
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/data/preprocessing.py", line 1423, in _preprocess_df_for_training
>     random_seed=random_seed
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/data/preprocessing.py", line 925, in build_dataset
>     metadata
>   File "/usr/local/lib/python3.6/dist-packages/ludwig/data/preprocessing.py", line 1014, in build_data
>     preprocessing_parameters
[...]

PR #1207, opened 2021-06-12
Allow hyperopt parallel workers to create subprocess for image pre-processing

jimthompson5802:
# Code Pull Requests

Fix Issue #965.  `AssertionError: daemonic processes are not allowed to have children` is raised when hyperopt parallel workers have an image input feature in the training data.  image preprocessing will have each parallel worker create subprocesses to prepare images for training.

This PR modifies how hyerpopt starts it parallel workers to allow them to create subprocesses themselves.  With this PR in place the above `AssertionError` no longer occurs.

However, when running unit test `test_hyperopt.py` the unit test hangs in this situation: `test_hyperopt.py::test_hyperopt_executor`  The unit test hangs under this condition:
* `serial` unit test runs to completion
* `parallel` unit test hangs on the first test case.

If the `serial` unit test is run **after** the `parallel` unit test, **no hang occur**.  Also **no hang occurs on the main branch**.  

Although the PR addresses the original issue #965 the method by which this is accomplished appear to have some unintended side-effects.  

Submitting PR to allow a broader review of the changes to help identify the root cause for the hang.

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -7,6 +7,7 @@ import signal
 import shutil
 from abc import ABC, abstractmethod
 from typing import Union
+from concurrent.futures import ProcessPoolExecutor
 
 from ludwig.api import LudwigModel
 from ludwig.callbacks import Callback
@@ -30,26 +31,6 @@ try:
 except ImportError:
     ray = None
 
-
-######
-# todo clean up once non-daemonic processes work
-# import multiprocessing.pool
-
-class NoDaemonProcess(multiprocessing.Process):
-    @property
-    def daemon(self):
-        return False
-
-    @daemon.setter
-    def daemon(self, value):
-        pass
-
-
-class NoDaemonContext(type(multiprocessing.get_context('spawn'))):
-    Process = NoDaemonProcess
-
-
-#####
 
 class HyperoptExecutor(ABC):
     def __init__(self, hyperopt_sampler: Union[dict, HyperoptSampler],
@@ -385,9 +366,8 @@ class ParallelExecutor(HyperoptExecutor):
             debug=False,
             **kwargs
     ) -> HyperoptResults:
-        # todo clean up code
-        # ctx = multiprocessing.get_context('spawn')
-        ctx = NoDaemonContext()
+
+        ctx = multiprocessing.get_context('spawn')
 
         if gpus is None:
             gpus = get_available_gpus_cuda_string()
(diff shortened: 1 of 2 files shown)


jimthompson5802:
@w4nderlust  @tgaddair 


I may have a fix for Issue #965, which generates an `AssertionError: daemonic process are not allowed to have children.` during a hyperopt run when the training data contains an image feature.  This last change also solves the hang in the `test_hyperopt.py` unit test described in the initial posting.   

This change takes a different approach from the original one of providing a Ludwig custom class for `NonDaemonicProcessPool`.   I could not get to the bottom of the hang issue with that approach.  So decide to rethink the solution.

However, the solution implemented in the last set of commits impacts the Python version used with Ludwig.  The fix makes use of the  `concurrent.futures` library that comes with py3.7 or later. 

### Summary of change:
Replaced `multiprocessing.Pool` with `concurrent.futures.PoolProcessExecutor`.  Py3.7 or later is required because the version of `PoolProcessExecutor` in py3.7 is the first version that supports specification of work process initializer and specification of the multiprocessing context (aka 'spawn').  Prior to Py3.7, the `PoolProcessExcecutor` did not have support of these options.

You can see this issue in the `py3.6, tf2.3.1, distrubuted` test run.  Here are screenshots illustrating this point.  This screenshot shows failures in the three hyperopt Parallel unit tests.
image


This is the error from one of those tests, showing the multiprocessing context parameter is not supported.
image


Here is the run for `py3.7, tf2.5.0, distributed`.  The test that failed with `py3.6` do not fail with `py3.7`
image


The other failures in the unit test are related to the missing `torch` module.

@tgaddair   Note:  I replaced the original `try-finally` construct with a `try-except`  because the `PoolProcessExecutor` class does not support `join()` and `close()` methods.  Right now I'm just capturing any exception and printing it out.  I'll replace the `print()` with a `logger` call.  Just looking for any feedback on how best to handle exceptions.

Let me know what you guys think, especially re: the py3.7 requirement.

jimthompson5802:
Couple other comments:
* From manual observation of the reproducible example provided by the user in #965, I can see multiple hyperopt parallel workers active and when each work is doing image pre-processing, I can see the specified number of parallel processes (`num_processes`).
* From the hyperopt results, the training (MNIST) is converging, here is example log from one of the parallel workers:
image

jimthompson5802:
After reviewing [source code for the `concurrent.futures.PoolProcessExecutor`](https://github.com/python/cpython/blob/main/Lib/concurrent/futures/process.py) decided to remove the `try-finally` structure because the parallel worker clean-up is performed by `PoolProcessExecutor`.   This commit https://github.com/ludwig-ai/ludwig/pull/1207/commits/65ed7be943ad12cf6bbca2aae7daabe0a4657df4 makes this change.

jimthompson5802:
@w4nderlust  @tgaddair    I noticed the issue with distributed unit test was fixed.  I merged the fix into this PR.

Now all the test pass except for the one with py3.6. 

From my perspective, the only thing holding back this PR is the minimum Python version required for Ludwig.

jimthompson5802:
superesed by pytorch conversion implementation.

PR #1212, opened 2021-06-22
Update Readme to activate directly the virtualenv

vreyespue:
Minor change in order to help the unexperienced reader to correctly activate the virtual environment before installation. In my opinion this minor fix would make the suggestion of using a virtualenv more robust. Many thanks and best regards.


w4nderlust:
Thank you!

PR #1214, opened 2021-06-24
Add metric and loss modules for RMSE, RMSPE, and AUC

ANarayan:
This PR adds support for RMSE, RMSPE, and AUC.

Specifically it adds a metric and loss module for both RMSE and RMSPE and a metric module for AUC.


ludwig/modules/loss_modules.py
```diff
@@ -447,43 +462,63 @@ def weighted_sigmoid_cross_entropy(
...
+            range_max=num_classes,
         )
     else:
-        raise ValueError('Unsupported sampler {}'.format(sampler))
+        raise ValueError("Unsupported sampler {}".format(sampler))
     return sampled_values
+
+
+def rmspe_loss(targets, predictions):
+    if type(predictions) == dict and PREDICTIONS in predictions.keys():
```
w4nderlust, line 515: I would remove this, and just have the loss passing the right tensors and the metric passing the tensor associated with predictions key. This way this function is more atomic and reusable.

ludwig/features/numerical_feature.py
```diff
@@ -265,74 +273,74 @@ def __init__(self, feature):
...
+        if self.loss[TYPE] == "mean_squared_error":
+            self.metric_functions[LOSS] = MSEMetric(name="eval_loss")
+        elif self.loss[TYPE] == "mean_absolute_error":
+            self.metric_functions[LOSS] = MAEMetric(name="eval_loss")
+        elif self.loss[TYPE] == "root_mean_squared_error":
+            self.metric_functions[LOSS] = RMSEMetric(name="eval_loss")
+        elif self.loss[TYPE] == "root_mean_squared_percentage_error":
+            self.metric_functions[LOSS] = RMSPEMetric(name="eval_loss")
+
+        self.metric_functions[ERROR] = ErrorScore(name="metric_error")
```
w4nderlust, line 327: i think we can get rid of error, it's just not informative. can also get rid of ErrorScore.

ANarayan: @w4nderlust see [commit](c446be85bdc9755816985bd56b0a0a23ea14e867)

ludwig/modules/loss_modules.py
ANarayan: @w4nderlust  see [commit](https://github.com/ludwig-ai/ludwig/pull/1214/commits/d93696b53d409bd297c7ae67c6cc94bccadec41b)

PR #1219, opened 2021-06-29
Initial implementation of the end-to-end autotrain module

ANarayan:
This PR:

- adds automl module
- adds support for specifying ray tune time budget
- adds support for auto tuning batch size
- add support auto selecting learning rate


ludwig/automl/utils.py
```diff
@@ -0,0 +1,39 @@
...
+def get_avg_words(field: Series) -> int:
+    # sample a subset if dataframe is large
+    if len(field) > 5000:
+        field = field.sample(n=5000, random_state=40)
+    unique_entries = field.unique()
+    avg_words = Series(unique_entries).str.split().str.len().mean()
+    return avg_words
+
+
+def get_available_resources():
```
tgaddair, line 29: For Ray, we can use the resource info provided by Ray to obtain this:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L49

ludwig/models/trainer.py
```diff
@@ -331,6 +331,60 @@ def write_step_summary(
...
+
+            except tf.errors.ResourceExhaustedError as e:
+                gc.collect()
+                high = prev_batch_size
+                midval = (high + low) // 2
+                self.batch_size = midval
+                if high - low <= 1:
+                    break
+
+        # Restore original parameters to defaults
```
tgaddair, line 381: We can put this in a `try-finally` block to ensure we restore the params even on exception.

ludwig/models/trainer.py
```diff
@@ -331,6 +331,60 @@ def write_step_summary(
...
+        model,
+        training_set,
+        max_trials: int = 10,
+    ):
+        original_epochs = self.epochs
+        skip_save_model = self.skip_save_model
+        skip_save_progress = self.skip_save_progress
+        skip_save_log = self.skip_save_log
+        # Set temporary values
+        self.epochs = 3
```
tgaddair, line 345: Is it necessary to train for 3 full epochs or just 3 batches?

tgaddair (commented):
Awesome progress! Left a few comments.

ludwig/automl/base_config.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+        - 1
+    )
+
+    # TODO (ASN): clarify logic below
+
+    # Second pass to exclude fields that are too expensive given the constraints
+    for meta in metadata:
+        if input_count > 2 and meta["config"]["type"] == "text":
+            # By default, exclude text inputs when there are other candidate inputs
```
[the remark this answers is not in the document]
ANarayan: agreed -- this makes sense.

ludwig/models/trainer.py
ANarayan: Not sure if we need a finally clause here since we restore the original parameters outside of the while loop.

ludwig/models/trainer.py
```diff
@@ -26,29 +27,32 @@
...
-from tabulate import tabulate
-from tqdm import tqdm
-
-from ludwig.constants import LOSS, COMBINED, TRAINING, VALIDATION, TEST, TYPE
-from ludwig.globals import MODEL_HYPERPARAMETERS_FILE_NAME
-from ludwig.globals import MODEL_WEIGHTS_FILE_NAME
-from ludwig.globals import TRAINING_CHECKPOINTS_DIR_PATH
-from ludwig.globals import TRAINING_PROGRESS_TRACKER_FILE_NAME
-from ludwig.globals import is_progressbar_disabled
+#from ludwig.api import LudwigModel
```
tgaddair, line 35: Nit: remove commented line

ludwig/models/trainer.py
```diff
@@ -331,6 +335,230 @@ def write_step_summary(
...
+        skip_save_model = self.skip_save_model
+        skip_save_progress = self.skip_save_progress
+        skip_save_log = self.skip_save_log
+        # Set temporary values
+        self.skip_save_model = True
+        self.skip_save_progress = True
+        self.skip_save_log = True
+
+        # Turn eager mode on
+        tf.config.experimental_run_functions_eagerly(True)
```
tgaddair, line 504: We should probably wrap this in a try-finally just in case an exception is raised and handled by the user. So something like:

```
tf.config.experimental_run_functions_eagerly(True)
try:
    ...
finally:
    tf.config.experimental_run_functions_eagerly(False)
```

ludwig/automl/automl.py
```diff
@@ -0,0 +1,91 @@
...
+    default_configs = create_default_config(dataset, target, time_limit_s)
+    model_config, model_name = model_select(default_configs)
+    hyperopt_results = _train(model_config, dataset,
+                              output_dir, model_name=model_name)
+    experiment_analysis = hyperopt_results.experiment_analysis
+
+    # catch edge case where metric_score is nan
+    # TODO (ASN): Decide how we want to proceed if at least one trial has
+    # completed
+    for trial in hyperopt_results.ordered_trials:
```
tgaddair, line 64: Was this already fixed in the RayExecutor?

ludwig/automl/automl.py
```diff
@@ -0,0 +1,91 @@
...
+    :param dataset: (str) filepath to dataset.
+    :param target_name: (str) name of target feature
+    :param time_limit_s: (int, float) total time allocated to auto_train. acts
+                        as the stopping parameter
+
+    # Returns
+    :return: (str) path to best trained model
+    """
+
+    default_configs = create_default_config(dataset, target, time_limit_s)
```
tgaddair, line 55: I think it could be useful to separate out the `create_default_config` in case the user wants to do something like: create default config, modify, then auto-train.

So perhaps we could do this:

```
def auto_train(..., default=None):
    if config is None:
        config = create_auto_config(dataset, target, time_limit_s)
```

This `create_auto_config` would then combine the default config creation with model selection. Maybe instead of returning the model name we can obtain this from the combiner name? That would also simplify the API for end users.

tgaddair (commented):
Awesome stuff! Only a handful of minor comments.

ludwig/automl/automl.py
ANarayan: @tgaddair yeah so logic here is about notifying the user that there was an error in the experiment (namely 1 or more trials did not have the time to start due to the time budget). maybe we should just raise a warning and proceed (rather than throwing an exception)?

ludwig/api.py
```diff
@@ -481,16 +481,25 @@ def train(
...
                     tuned_batch_size = trainer.tune_batch_size(
-                        temp_model,
+                        self.config,
                         training_set,
+                        random_seed=random_seed
                     )
                     self.config[TRAINING]['batch_size'] = tuned_batch_size
 
+                # auto tune learning rate
+                if self.config[TRAINING]["learning_rate"] == "auto":
```
w4nderlust, line 494: AUTO, LEARNING_RATE and BATCH_SIZE can become constants

ludwig/automl/automl.py
```diff
@@ -8,8 +8,11 @@
 (2) Tunes config based on resource constraints
 (3) Runs hyperparameter optimization experiment
 """
+from logging import raiseExceptions
 from typing import Dict, Union
 
+import dask.dataframe as dd
```
w4nderlust, line 14: dask may not be installed, it is installed only in the fistributed package

ludwig/automl/automl.py
```diff
@@ -48,16 +56,36 @@ def auto_train(dataset: str, target: str, time_limit_s: Union[int, float], outpu
...
+
+    # catch edge case where metric_score is nan
+    # TODO (ASN): Decide how we want to proceed if at least one trial has
+    # completed
+    for trial in hyperopt_results.ordered_trials:
+        if np.isnan(trial.metric_score):
+            raise ValueError(
+                "There was an error running the experiment. "
+                "A trial failed to start. "
+                "Consider increasing the time budget for experiment. "
```
w4nderlust, line 69: Are we sure failing to start is the only possible reason for a NaN?

ludwig/automl/base_config.py
```diff
@@ -121,19 +127,20 @@ def get_field_info(dataset: str):
...
 
 
-def get_input_and_output_features(
+def get_features_config(
     fields: List[FieldInfo],
     row_count: int,
+    resources: Dict,
     target_name: str = None,
-) -> dict:
+) -> Dict:
```
w4nderlust, line 143: Here and in all other places we are using `typing.Dict` we should use `dict` if we don't specify the types of keys and values, and `Dict[keytype, valuetype]` otherwise. Just for consistency.

More info: https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict

ludwig/automl/base_config.py
```diff
@@ -220,18 +227,17 @@ def get_field_metadata(
...
-
-    # Second pass to exclude fields that are too expensive given the constraints
-    for meta in metadata:
-        if input_count > 2 and meta["config"]["type"] == "text":
-            # By default, exclude text inputs when there are other candidate inputs
-            meta["excluded"] = True
+    # Exclude text fields if no GPUs are available
+    if resources['gpu'] == 0:
+        for meta in metadata:
+            if input_count > 2 and meta["config"]["type"] == "text":
```
w4nderlust, line 233: CONFIG, TYPE and TEXT can all be constants

ludwig/models/trainer.py
```diff
@@ -364,34 +368,152 @@ def train_for_tuning(
...
-    def tune_batch_size(
+    def tune_learning_rate(
         self,
+        config,
         model,
         training_set,
+        random_seed: int = default_random_seed,
+        min_lr: float = 1e-8,
+        max_lr: float = 1.0,
+        total_training_steps: int = 100,
```
w4nderlust, line 381: this seems like a big number of steps

ANarayan: @w4nderlust totally agree. I pulled this default value from the pytorch lightning LR tuner. We can modify our implementation is you think something smaller is more appropriate.

ludwig/automl/automl.py
ANarayan: @tgaddair this make sense. Just to clarify, we would expect that if the user wants to modify the config themselves, they would call `create_default_config` first before calling `auto_train` right?

tgaddair: Yes, exactly.

ludwig/automl/automl.py
ANarayan: @w4nderlust  Not sure - let me investigate

ANarayan: Another way around it is to just check if the `training_stats` and `eval_stats` are empty dicts.

ludwig/automl/defaults/tabnet_config.yaml
```diff
@@ -0,0 +1,73 @@
...
+  optimizer:
+    type: adam
+  tune_batch_size:
+    type: bin_search
+    substitute_with_max: True
+
+hyperopt:
+  # goal: maximize
+  parameters:
+    training.learning_rate:
```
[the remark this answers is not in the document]
ANarayan: @tgaddair Reasoning behind the auto keyword is as follows. We want to tune the learning_rate and scale the batch_size only when those parameters are not part of the hyperparameter search space. The challenge with tuning the parameters during a hyperparameter optimization experiment is that the tuning algorithm has no knowledge that the value of the parameter it sample has changed and thus learns an incorrect parameter search space. This being said, I think the default behavior should be to do a hyperparameter search over learning_rate and batch_size (because we learn a relationship between batch_size and lr) vs tuning them independently. We can of course run experiments comparing the efficacy of tuning vs. search space optimization and change the default behavior.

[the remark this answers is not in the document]
tgaddair: I see, so in the current implementation the `auto` keyword will be overridden by the values selected from the search space.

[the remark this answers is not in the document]
ANarayan: yup!

ludwig/automl/automl.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+
+    autotrain_results = {
+        'path_to_best_model': experiment_analysis.best_checkpoint,
+        'trial_id': "_".join(experiment_analysis.best_logdir.split("/")[-1].split("_")[1:])
+    }
+    return autotrain_results
+
+
+def _create_auto_config(dataset, target, time_limit_s) -> dict:
```
tgaddair, line 93: Let's make this public by removing the underscore. But `create_default_config` and `model_select` it may make sense to be private.

ludwig/automl/base_config.py
```diff
@@ -0,0 +1,314 @@
...
+    experiment_resources = {
+        'cpu_resources_per_trial': 1
+    }
+    gpu_count, cpu_count = resources['gpu'], resources['cpu']
+    if gpu_count > 0:
+        experiment_resources.update({
+            'gpu_resources_per_trial': 1
+        })
+        if cpu_count > 1:
+            cpus_per_trial = int(cpu_count/gpu_count)
```
tgaddair, line 74: Maybe `max(int(cpu_count / gpu_count), 1)`

ludwig/automl/automl.py
```diff
@@ -0,0 +1,111 @@
...
+try:
+    import dask.dataframe as dd
+    import ray
+except ImportError:
+    logger.error(
+        ' ray is not installed. '
+        'In order to use auto_train please run '
+        'pip install ludwig[ray]'
+    )
+    sys.exit(-1)
```
tgaddair, line 32: I know we do this in a few other places in Ludwig, but for programatic usage, we should probably avoid calling `sys.exit` in case the user doesn't want their notebook to crash. Maybe raise an exception?

ludwig/models/trainer.py
w4nderlust: We are spanning through 8 orders of magnitude with the default parameters, probably less than 10 for each order of magnitude may be sufficient. Although let's keep it like this for now and if we figure out that it's too slow, 50 would probably be sufficient too.

ludwig/automl/automl.py
w4nderlust: Good point

ludwig/automl/automl.py
```diff
@@ -51,13 +66,12 @@ def auto_train(
     # Returns
     :return: (str) path to best trained model
     """
-
-    default_configs = create_default_config(dataset, target, time_limit_s)
-    model_config, model_name = model_select(default_configs)
-    hyperopt_results = _train(model_config, dataset,
+    if config is None:
+        config = _create_auto_config(dataset, target, time_limit_s)
+    model_name = config['combiner']['type']
```
w4nderlust, line 71: COMBINER and TYPE should be constants

ludwig/automl/automl.py
ANarayan: @tgaddair Totally agree with making `create_default_config` and `model_select` private. Whats the reasoning for making `create_auto_config` public?

tgaddair: The idea would be if the user wants to inspect the auto config and modify it before training, e.g.:

```
config = create_auto_config()
config['training']['learning_rate'] = 1
auto_train(..., config=config)
```

Does that seem reasonable to you?

ANarayan: Right right. this make total sense!

ludwig/automl/automl.py
tgaddair: I agree, I think a warning would make more sense.

tgaddair (approved):
LGTM! Can't wait to try it out.

PR #1220, opened 2021-06-29
[WIP] Added support for end-to-end TensorFlow graph inference

tgaddair:
Serving a Ludwig model consists of three stages:

1. Preprocessing
2. Model inference
3. Postprocessing

Prior to this PR, steps (1) and (3) took place outside of the TensorFlow graph, meaning that they needed to be run in a Python interpreter during serving time. This Python dependency comes with a number of shortcomings:

1. Requires creating a complete Python environment for serving the model. This is usually in the form of a Docker image or Conda environment, which adds a lot of bloat at serving time (multiple models cannot easily served in the same server).
2. Prevents fusing operations together into a single execution environment. One advantage of having the entire model in the TF graph is being able to place it directly onto the GPU to accelerate inference. However, this existing Python steps cannot easily be placed on GPU with the rest of the model, creating a bottleneck that greatly increases latency.
3. Backwards compatibility. TensorFlow graphs are generally backward compatible for many versions, allowing us to run inference on old graphs in newer serving stacks. The same cannot be said for Python Ludwig models, where the APi is subject to change between versions.

Currently, we support input/output features of the following types:

- Numerical
- Category

The plan is to expand this list to include all Ludwig types in the near future.


ludwig/utils/strings_utils.py
```diff
@@ -282,31 +283,49 @@ def __init__(self, **kwargs):
...
     def __call__(self, text):
         return text
 
-
 class SpaceStringToListTokenizer(BaseTokenizer):
     def __call__(self, text):
         return SPLIT_REGEX.split(text.strip())
 
+    @staticmethod
+    def call(t):
```
tgaddair, line 295: Could make this more specific: `def call_inference(t: tf.Tensor) -> tf.Tensor:`

ludwig/utils/strings_utils.py
```diff
@@ -282,31 +283,49 @@ def __init__(self, **kwargs):
...
+        # TODO(ksbrar)
+        splitter = tf_text.RegexSplitter(split_regex=r'\s+')
```
tgaddair, line 297: We should reuse the constant variables above to avoid duplication. We need probably convert them to back to strings.

If there's no way to go from regex to string, then we can separate them out:

```
SPLIT_REGEX_PATTERN = r'\s+'
SPLIT_REGEX = re.compile(SPLIT_REGEX_PATTERN)
```

ludwig/utils/strings_utils.py
```diff
@@ -282,31 +283,49 @@ def __init__(self, **kwargs):
...
 class CharactersToListTokenizer(BaseTokenizer):
     def __call__(self, text):
         return text
 
-
 class SpaceStringToListTokenizer(BaseTokenizer):
     def __call__(self, text):
         return SPLIT_REGEX.split(text.strip())
 
+    @staticmethod
```
tgaddair, line 294: We should handle the case where the `call_inference` method is not implemented. One way to do this is within `BaseTokenizer`:

```
def call_inference(t: tf.Tensor) -> tf.Tensor:
    raise RuntimeError(f'Graph inference not supported for tokenizer: {self.__class__.__name__)')
```

tests/integration_tests/test_pure_tf_model.py
```diff
@@ -88,8 +88,10 @@ def test_pure_tf_model(csv_filename, tmpdir):
         skip_save_processed_input=True,
     )
 
+    # TODO(ksbrar) 
+    # print(input_features[0])
     ludwig_tf_model = ludwig_model.create_inference_graph()
-    print(ludwig_tf_model)
+    # print(ludwig_tf_model)
```
tgaddair, line 94: Let's expand this test. We need to at least cover the following:

- Test different tokenizers
- Test handling invalid tokenizer
- Test equivalence of predictions from Python and TF models

tgaddair:
Looks like tests are failing due to import errors:

```
>         raise e.ag_error_metadata.to_exception(e)
E         AttributeError: in user code:
E         
E             /home/runner/work/ludwig/ludwig/ludwig/utils/tf_utils.py:212 call  *
E                 return self.tokenizer.call(t)
E             /home/runner/work/ludwig/ludwig/ludwig/utils/strings_utils.py:307 call  *
E                 splitter = tf_text.RegexSplitter(split_regex=r'\w+|[^\w\s]')
E         
E             AttributeError: module 'tensorflow_text' has no attribute 'RegexSplitter'
```

tgaddair:
Outstanding issues:

- Fix (or determine root cause of) py3.6 failures
- Add tests comparing model and graph outputs
- Make sure set features are working

ksbrar:
Update: The single python 3.6 failure is expected because it is using tf 2.3.1, and [tf_text.RegexSplitter was introduced in 2.4.1](https://pythonrepo.com/repo/tensorflow-text-python-natural-language-processing#:~:text=Dec%2017%2C%202020)-,Release%202.4.1,-Major%20Features%20and)

PR #1224, opened 2021-07-10
Non-absolute image path support

hungcs:
# Code Pull Requests

Please provide the following:

- Provides support for non-absolute image paths and adds naive image inference (if specified via preprocessing arguments).
- Images are inferred using the formula `min(avg([width | height]), 256)` sampling the first 100 images. Checking the sizes of all images seems [rather complicated](https://stackoverflow.com/questions/15800704/get-image-size-without-loading-image-into-memory/19035508) and involves format-specific logic that may break if header data is malformed.
- `is_http` check passes the url directly to `imread `because `fsspec` can error [trying to retrieve image info](https://github.com/dask/dask/issues/5222), which may be related to the jpeg/header issue above.
- Sample dataset with http image paths: [insuranceLite.csv](https://github.com/ludwig-ai/ludwig/files/6795923/insuranceLite.csv)


tgaddair (approved):
LGTM!

ludwig/features/image_feature.py
```diff
@@ -222,11 +226,23 @@ def _finalize_preprocessing_parameters(
...
-            height = first_img_height
-            width = first_img_width
+            # So infer height and width from a sample.
+            if INFER_IMAGE_DIMENSIONS in preprocessing_parameters and preprocessing_parameters[INFER_IMAGE_DIMENSIONS]:
+                should_resize = True
+                sample_size = min(len(input_feature_col), ImageFeatureMixin.IMAGE_SAMPLE_SIZE)
+                sample_images = [ImageFeatureMixin.read_image(
+                    ImageFeatureMixin.get_image_from_path(src_path, img)) for img in input_feature_col[:sample_size]]
+
+                height_avg = min(
```
w4nderlust, line 236: These operations should be max not min, as we want to make sure the inferred image size is not above a certain amount.

Also, having a default `ImageFeatureMixin.MAX_IMAGE_INFER_DIM = 256` is ok, but maybe that could be a preprocessing parameter too.

w4nderlust:
On more thing that would be great is to have a test for remote images (you could use the images in the Ludwig readme for this purpose)

PR #1235, opened 2021-07-17
[automl] refactored to provide extended public API

tgaddair:
This PR allows the user to specify different dataset input formats and to call AutoML functions in piecemeal fashion.

### DIFF
--- a/ludwig/automl/base_config.py
+++ b/ludwig/automl/base_config.py
@@ -180,7 +180,7 @@ def get_features_config(
     target_name: str = None,
 ) -> dict:
     """
-    Constructs FeildInfo objects for each feature in dataset. These objects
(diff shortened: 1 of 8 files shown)


ludwig/automl/base_config.py
```diff
@@ -15,13 +15,15 @@
...
 
 import os
+from dataclasses import dataclass
 from typing import List, Union
 
 import pandas as pd
+from ludwig.automl.data_source import DataSource, DataframeSource
 from ludwig.automl.utils import (FieldInfo, avg_num_tokens,
-                                 get_available_resources)
+                                 get_available_resources, _ray_init)
```
w4nderlust, line 24: as we are importing and using `_ray_init` in a few places, wouldn't it be better making it properly available without the `_` in the name?

Issue #1234, opened 2021-07-16
Predictions not getting saved


w4nderlust:
Hi there, I'm super new to ML and Ludwig, so I'm almost certainly missing something. I'm trying to complete an example in the getting started guide, but I'm stuck when it comes to getting predictions to work.

I'm using the `epinions-1.csv` (CSV file with a `class` and a `text` column) as suggested in the getting started guide and I'm running Ludwig using the official Docker container. The following is my configuration.

```
input_features:
- name: text
  type: text
  level: word
  encoder: parallel_cnn
output_features:
- name: class
  type: category
```

First off I start training using the `experiment` command. That seems to work and outputs a model. However when I try to use the following `predict` command on **the same data** with the `class` column removed (same CSV, just a single column), I don't get any CSVs with predictions as the docs seem to indicate I should. `prediction.shapes.json` contains an empty object and the NPY files also seem empty when loaded with `numpy.load`.

```
ludwig predict --dataset /data/epinions-1-text.csv --data_format csv --model_path /data/results/experiment_run/model --output_directory /data/predictions
```

Any idea what I'm doing wrong? Thanks! 🙏

_Originally posted by @ianmuscat in https://github.com/ludwig-ai/ludwig/discussions/1229_

PR #1236, opened 2021-07-18
Fix: reinstitute saving predictions as csv for LocalBackend

jimthompson5802:
# Code Pull Requests

Fix Issue #1234.  For LocalBackend save predictions as csv files.  This matches behavior pre-v0.4.  For other backends, predictions saved in parquet file.

Current Status:
* Successfully tested with numeric and categorical output feature.
* Confirmed that this fixes the issue described in #1234 

Remaining work:
* Test with other output features: binary and sequence
* Update unit tests for this change


jimthompson5802:
@w4nderlust   @tgaddair 
There appears to be one flaw in my original approach of
```
if instance(backend, LocalBackend):
   # generate csv prediction files
else:
   # generate parquet prediction file
```
I just realized with PR #1128 many of the visualization functions were modified to look for probabilities from the `parquet` predictions only.  As a results many of the visualization tests are now failing because these test run with `LocalBackend` and the `parquet` file no longer exists.

From my perspective, I see two options to address this problem:

**Option 1**: Generate the `parquet` predictions all the time and create the `csv` predictions file only for `LocalBackend`.  
```
if instance(backend, LocalBackend):
   # generate csv prediction files

# generate parquet prediction file
```
Pro:  For any other Ludwig functions (e.g., visualization) that were changed to use the `parquet` file, those functions will not have to change.  Less code changes.

Con: three copies of prediction values are stored:  `*.npy`, `predictions.parquet` and `*.csv` files  for `LocalBackend`.   However, for other backends, there won't be the three copies.


**Option 2**: Keep current approach
```
if instance(backend, LocalBackend):
   # generate csv prediction files
else:
   # generate parquet prediction file
```
 and modify `ludwig.visualize._get_cols_from_predictions()` to retrieve predictions from either the `parquet` file or from `*.csv` files.

Pro:  Only two copies of prediction values are stored for any given run: `*.npy` and `predictions.parquet` or `*.npy` and `*.csv`.  More efficient on space usage.

Con:  Added complexity to the `_get_cols_from_predictions()` function.  May be entering the "rabbit hole" of having to update other Ludwig functions that were similarly modified with #1128 that I have not encountered yet.

**Recommendation**:  I'm leaning toward **Option 1** because it is the simpler of the two options.  I figure with the cost of the three copies is probably reasonable because the `LocalBackend` is most likely with local testing using "smaller" data set sizes.

Let me know what you think.  I'm in a holding pattern for now.

ludwig/backend/base.py
```diff
@@ -129,3 +134,57 @@ def __init__(self, **kwargs):
...
+        # LocalBackend save predictions in csv format
+        # for each column in the postprocessed_output dataframe save as a csv
+
+        # setup to parse column names to form csv files
+        # column names are one of the following forms
+        #   <feature_name>_predictions
+        #   <feature_name>_probabilities
+        #   <feature_name>_probability
+        #   <feature_name>_probabilities_<token_text>
+        pattern = re.compile(
```
tgaddair, line 152: Would this function work for this? It should result in a dict with the correct format:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/postprocessing.py#L87

justinxzhao:
Closing as the original issue has been fixed.

PR #1243, opened 2021-07-22
Add image dimension inference to automl

hungcs:
Sets the `infer_image_dimensions` flag to `True` (default is False) for autoML


hungcs:
Will also have to add image inference to `infer_type`, doing that now

tgaddair (approved):
LGTM! @ANarayan can you also take a look as this affects automl features?

ANarayan:
@tgaddair @hungcs Looks great to me!!

PR #1244, opened 2021-07-23
[datasets] Add remote filesystem support to datasets module

ANarayan:
This PR adds support for reading and writing to remote filesystems (e.g. S3 buckets) in the `Datasets` module. 

cc: @w4nderlust , @tgaddair


ludwig/datasets/mixins/download.py
```diff
@@ -57,17 +65,27 @@ def download_raw_dataset(self):
...
+                    with ZipFile(BytesIO(zipresp.read())) as zfile:
+                        zfile.extractall(tmpdir)
+            protocol, _ = split_protocol(self.raw_dataset_path)
+            if protocol is not None:
+                fs = fsspec.filesystem(protocol)
+                fs.put(tmpdir, self.raw_dataset_path, recursive=True)
+            else:
+                shutil.copytree(tmpdir, self.raw_dataset_path)
+
+        #os.rename(self.raw_temp_path, self.raw_dataset_path)
```
tgaddair, line 82: Can remove this line.

ludwig/datasets/mixins/download.py
```diff
@@ -57,17 +65,27 @@ def download_raw_dataset(self):
...
         """
-        os.makedirs(self.raw_temp_path, exist_ok=True)
-        for url in self.download_urls:
-            with urlopen(url) as zipresp:
-                with ZipFile(BytesIO(zipresp.read())) as zfile:
-                    zfile.extractall(self.raw_temp_path)
-        os.rename(self.raw_temp_path, self.raw_dataset_path)
+        #os.makedirs(self.raw_temp_path, exist_ok=True)
+
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 70: Can we simplify to something like this?

```
with fs_utils.upload_output_file(self.raw_dataset_path) as tmpdir:
    ...
```

If that doesn't work, we should still make something like this in `fs_utils` as we do the same thing below.

ludwig/datasets/mixins/download.py
```diff
@@ -57,17 +65,27 @@ def download_raw_dataset(self):
         Download the raw dataset and extract the contents of the zip file and
         store that in the cache location.
         """
-        os.makedirs(self.raw_temp_path, exist_ok=True)
-        for url in self.download_urls:
-            with urlopen(url) as zipresp:
-                with ZipFile(BytesIO(zipresp.read())) as zfile:
-                    zfile.extractall(self.raw_temp_path)
-        os.rename(self.raw_temp_path, self.raw_dataset_path)
+        #os.makedirs(self.raw_temp_path, exist_ok=True)
```
tgaddair, line 68: Can remove this.

ludwig/datasets/mixins/download.py
```diff
@@ -80,32 +98,41 @@ def download_raw_dataset(self):
...
-            file_path = os.path.join(self.raw_temp_path, filename)
-            with tarfile.open(file_path) as tar_file:
-                tar_file.extractall(path=self.raw_temp_path)
-
-            for f in os.scandir(os.path.join(self.raw_temp_path, 
-                                             download_folder_name)):
-                shutil.copyfile(f, os.path.join(self.raw_temp_path, f.name))
-
-        os.rename(self.raw_temp_path, self.raw_dataset_path)
+        #os.makedirs(self.raw_temp_path, exist_ok=True)
```
tgaddair, line 101: Can remove this.

ludwig/datasets/mixins/download.py
```diff
@@ -80,32 +98,41 @@ def download_raw_dataset(self):
...
+
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 103: Same as above regarding using `fs_utils`.

ludwig/datasets/mixins/download.py
```diff
@@ -118,23 +145,30 @@ def download_raw_dataset(self):
...
-                )
-            gzip_content_file = '.'.join(filename.split('.')[:-1])
-            with gzip.open(
-                    os.path.join(self.raw_temp_path, filename)) as gzfile:
-                with open(os.path.join(self.raw_temp_path, gzip_content_file),
-                          'wb') as output:
-                    shutil.copyfileobj(gzfile, output)
-        os.rename(self.raw_temp_path, self.raw_dataset_path)
+        #os.makedirs(self.raw_temp_path, exist_ok=True)
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 149: Same as above.

ludwig/datasets/mixins/download.py
```diff
@@ -153,12 +187,18 @@ def download_raw_dataset(self):
         Download the raw dataset files and store in the cache location.
         """
         os.makedirs(self.raw_temp_path, exist_ok=True)
-        for url in self.download_url:
-            filename = url.split('/')[-1]
-            urllib.request.urlretrieve(url, os.path.join(self.raw_temp_path,
-                                                         filename))
-
-        os.rename(self.raw_temp_path, self.raw_dataset_path)
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 190: Same as above.

ludwig/datasets/mixins/process.py
```diff
@@ -26,7 +30,13 @@ class IdentityProcessMixin:
     processed_dataset_path: str
 
     def process_downloaded_dataset(self):
-        os.rename(self.raw_dataset_path, self.processed_dataset_path)
+        protocol, _ = split_protocol(self.processed_dataset_path)
```
tgaddair, line 33: I would move this into `fs_utils` as a utility function, like `fs_utils.rename`.

ludwig/datasets/mixins/process.py
```diff
@@ -26,7 +30,13 @@ class IdentityProcessMixin:
...
+        if protocol is not None:
+            fs = fsspec.filesystem(protocol)
+            fs.copy(self.raw_dataset_path,
```
tgaddair, line 36: Maybe we can use `fsspec.mv` or `fsspec.rename` here. Otherwise the behavior is different between local and remote versions.

ludwig/datasets/mixins/process.py
```diff
@@ -74,7 +84,11 @@ def process_downloaded_dataset(self, header=0):
             all_files.append(file_df)
 
         concat_df = pd.concat(all_files, ignore_index=True)
-        os.makedirs(self.processed_dataset_path, exist_ok=True)
+        #os.makedirs(self.processed_dataset_path, exist_ok=True)
+        #makedirs(self.process_downloaded_dataset, exist_ok=True)
+        with fsspec.open(self.processed_dataset_path, mode="wb") as f:
```
tgaddair, line 89: Is it necessary to write this file? Does it work without this?

ludwig/datasets/ieee_fraud/__init__.py
```diff
@@ -65,7 +66,8 @@ def process_downloaded_dataset(self):
         final_train = pd.merge(
             train_dfs['train_transaction'], train_dfs['train_identity'], on='TransactionID', how='left')
 
-        os.makedirs(self.processed_dataset_path, exist_ok=True)
+        with fsspec.open(self.processed_dataset_path, mode="wb") as f:
```
tgaddair, line 69: Do we need to create this file?

tgaddair (commented):
Nice! Few comments about moving some of this into `fs_utils`.

w4nderlust:
The new changes look good to me, are all places where fsspec was used covered? Asking just to be sure.
Other than that, le'ts address the failures, bt the structure of the code looks fine tome. Thanks!

ludwig/utils/fs_utils.py
```diff
@@ -51,9 +51,23 @@ def path_exists(url):
...
+        fs = fsspec.filesystem(protocol)
+        fs.mv(src, tgt, recursive=True)
+    else:
+        os.rename(src, tgt)
+
+
 def makedirs(url, exist_ok=False):
     fs, path = get_fs_and_path(url)
-    return fs.makedirs(path, exist_ok=exist_ok)
+    if fs == "s3":
```
tgaddair, line 65: Not sure why this is needed for s3, since directories don't exist. Is this so checks like `exists(dirname)` will pass?

This may also be true for other object stores. Maybe something more generic we could do here is:

```
fs.makedirs(path, exist_ok=exist_ok)
if not exists(paths):
    # Directory was not created -> no directory concept in filesystem
    with fsspec.open(url, mode="wb") as f:
            pass
```

What do you think?

ludwig/datasets/mixins/kaggle.py
```diff
@@ -33,19 +39,27 @@ def download_raw_dataset(self):
...
             api.authenticate()
-        os.makedirs(self.raw_temp_path, exist_ok=True)
 
-        if self.is_kaggle_competition:
-            download_func = api.competition_download_files
-        else:
-            download_func = api.dataset_download_files
-        # Download all files for a competition/dataset
-        download_func(self.competition_name, path=self.raw_temp_path)
+        #os.makedirs(self.raw_temp_path, exist_ok=True)
```
tgaddair, line 43: Can remove this line.

tgaddair (commented):
Nice! Just a couple things we can refactor.

tgaddair (approved):
LGTM!

PR #1252, opened 2021-08-07
Fixed TabNet conversion to TF graph with unknown batch size

tgaddair:


ANarayan:
Looks good to me! I am not longer seeing the original issue with the `NoneType` TypeError when calling `get_connected_model`.

PR #1257, opened 2021-08-16
[automl] Memory Aware Config Tuning

ANarayan:
This PR adds a new capability to the `automl` module which tightens the hyperparameter search space ranges in a Ludwig config file in order to avoid OOM issues. These OOM issues may be caused by implausible parameter assignments (given memory constraints) that can be sampled in the hyperparameter optimization experiment with the provided parameter search space ranges.

### DIFF
--- a/ludwig/automl/auto_tune_config.py
+++ b/ludwig/automl/auto_tune_config.py
@@ -1,0 +1,164 @@
+import copy
+from collections import OrderedDict
+
+import psutil
+import ray
+
+try:
+    import GPUtil
+except ImportError:
+    raise ImportError(
+        ' GPUtil is not installed. '
+        'In order to use this module please run '
+        'pip install GPUtil'
+    )
+
+from ludwig.api import LudwigModel
+from ludwig.automl.utils import get_available_resources
+from ludwig.data.preprocessing import preprocess_for_training
+from ludwig.features.feature_registries import update_config_with_metadata
+from ludwig.utils.defaults import merge_with_defaults
+from ludwig.constants import COMBINER, HYPEROPT, BATCH_SIZE, TRAINING, TYPE, PREPROCESSING, SPACE
+
+# maps variable search space that can be modified to minimum permissible value for the range
+RANKED_MODIFIABLE_PARAM_LIST = {
+    'tabnet': OrderedDict({
+        'training.batch_size': 32,
+        'combiner.size': 8,
+        'combiner.output_size': 8,
+    }),
+    'concat': OrderedDict({
+        'training.batch_size': 32,
+        'combiner.fc_size': 64,
+        'combiner.num_fc_layers': 1,
+
+    }),
+    'tabtransformer': OrderedDict({
+        'training.batch_size': 32,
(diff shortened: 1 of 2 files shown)


ludwig/automl/auto_tune_config.py
```diff
@@ -0,0 +1,164 @@
+import copy
+from collections import OrderedDict
+
+import psutil
+import ray
+
+try:
+    import GPUtil
```
tgaddair, line 8: Can we add this to the `requirements_ray.txt` instead of asking the user to install this by hand?

ludwig/automl/auto_tune_config.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def get_machine_memory():
+
+    if ray.is_initialized():  # using ray cluster
+        @ray.remote(num_gpus=1)
+        def get_remote_gpu():
+            gpus = GPUtil.getGPUs()
+            total_mem = gpus[0].memory_total
+            return total_mem * BYTES_TO_MB
```
tgaddair, line 64: This calculation is a little bit confusing, maybe use `total_mem_mb * BYTES_PER_MB` so it's clear we're converting from MB to bytes.

ludwig/automl/auto_tune_config.py
```diff
@@ -0,0 +1,164 @@
...
+        'training.batch_size': 32,
+        'combiner.num_heads:': 4,
+        'combiner.output_size': 8,
+        'combiner.num_layers': 4,
+        'combiner.num_fc_layers': 1,
+    }),
+}
+
+
+BYTES_TO_MB = 1e6
```
tgaddair, line 46: `nvidia-smi` (which is what GPUtil is using under the hood) reports in "MiB" (which is actually the correct way to measure MB, but I digress). So the correct conversion is:

```
BYTES_TO_MB = 1024 * 1024
```

ludwig/automl/auto_tune_config.py
```diff
@@ -0,0 +1,164 @@
...
+def get_new_params(current_param_values, hyperparam_search_space, params_to_modify):
+    for param, _ in params_to_modify.items():
+        if hyperparam_search_space[param][SPACE] == "choice":
+            current_param_values[param] = hyperparam_search_space[param]['categories'][-1]
+        else:
+            current_param_values[param] = hyperparam_search_space[param]['upper']
+    return current_param_values
+
+
+def memory_tune_config(config, dataset):
```
tgaddair, line 117: If would suggest running this with `ray.remote` if ray is available, just because TensorFlow has a tendency to have some issues with state corruption when creating models and potentially allocating GPU memory.

tgaddair (commented):
This is amazing! Just a few suggestions.

ludwig/automl/auto_tune_config.py
tgaddair: We should probably move the `import ray` in here, too, given the error message.

tgaddair (approved):
LGTM!

PR #1260, opened 2021-08-17
Ray data migration

ShreyaR:
# Code Pull Requests

This PR adds support for Ray Datasets for distributed prediction.

# Documentation Pull Requests

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -874,7 +874,6 @@ class LudwigModel:
             # calculate the overall metrics
             if collect_overall_stats:
                 # TODO ray: support calculating stats on partitioned datasets
-                # TODO(shreya): Confirm what's needed to enable this.
                 if isinstance(dataset, RayDataset):
                     raise ValueError(
                         'Cannot calculate overall stats on a ray dataset at this time. '
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -874,6 +874,7 @@ class LudwigModel:
             # calculate the overall metrics
             if collect_overall_stats:
                 # TODO ray: support calculating stats on partitioned datasets
+                # TODO(shreya): Confirm what's needed to enable this.
                 if isinstance(dataset, PartitionedDataset):
                     raise ValueError(
                         'Cannot calculate overall stats on a partitioned dataset at this time. '
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -721,7 +721,7 @@ class LudwigModel:
 
         # preprocessing
         logger.debug('Preprocessing')
-        dataset, training_set_metadata = preprocess_for_prediction(
+        dataset, _ = preprocess_for_prediction(
             self.config,
             dataset=dataset,
             training_set_metadata=self.training_set_metadata,
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -31,7 +31,7 @@ from collections import OrderedDict
 from pprint import pformat
 from typing import Dict, List, Optional, Tuple, Union
 
-from ludwig.data.dataset.partitioned import PartitionedDataset
+from ludwig.data.dataset.partitioned import RayDataset
 from ludwig.utils.fs_utils import upload_output_directory, path_exists, makedirs
 
 import numpy as np
@@ -875,9 +875,9 @@ class LudwigModel:
             if collect_overall_stats:
                 # TODO ray: support calculating stats on partitioned datasets
                 # TODO(shreya): Confirm what's needed to enable this.
-                if isinstance(dataset, PartitionedDataset):
+                if isinstance(dataset, RayDataset):
                     raise ValueError(
-                        'Cannot calculate overall stats on a partitioned dataset at this time. '
+                        'Cannot calculate overall stats on a ray dataset at this time. '
                         'Set `calculate_overall_stats=False`.'
                     )
 
--- a/ludwig/backend/dask.py
+++ b/ludwig/backend/dask.py
@@ -18,7 +18,6 @@
 from ludwig.backend.base import Backend, LocalTrainingMixin
 from ludwig.constants import NAME, PARQUET, PREPROCESSING, TFRECORD
 from ludwig.data.dataframe.dask import DaskEngine
-from ludwig.data.dataset.partitioned import RayDataset
 from ludwig.models.predictor import BasePredictor
 
 
@@ -56,11 +55,9 @@ class DaskPredictor(BasePredictor):
         )
 
     def _check_dataset(self, dataset):
-        if not isinstance(dataset, RayDataset):
-            raise RuntimeError(
-                f'Dask backend requires RayDataset for inference, '
-                f'found: {type(dataset)}'
-            )
+        raise NotImplementedError(
+            'Dask backend does not suppoer batch inference at this time. Try Ray.'
+        )
 
     def shutdown(self):
         pass
--- a/ludwig/backend/dask.py
+++ b/ludwig/backend/dask.py
@@ -18,7 +18,7 @@
 from ludwig.backend.base import Backend, LocalTrainingMixin
 from ludwig.constants import NAME, PARQUET, PREPROCESSING, TFRECORD
 from ludwig.data.dataframe.dask import DaskEngine
-from ludwig.data.dataset.partitioned import PartitionedDataset
+from ludwig.data.dataset.partitioned import RayDataset
 from ludwig.models.predictor import BasePredictor, Predictor, get_output_columns
 
 
@@ -70,9 +70,9 @@ class DaskPredictor(BasePredictor):
         )
 
     def _check_dataset(self, dataset):
-        if not isinstance(dataset, PartitionedDataset):
+        if not isinstance(dataset, RayDataset):
             raise RuntimeError(
-                f'Dask backend requires PartitionedDataset for inference, '
+                f'Dask backend requires RayDataset for inference, '
                 f'found: {type(dataset)}'
             )
 
--- a/ludwig/backend/dask.py
+++ b/ludwig/backend/dask.py
@@ -99,13 +99,17 @@ class DaskPredictor(BasePredictor):
             self.output_columns = output_columns
             self.features = features
             self.data_hdf5_fp = data_hdf5_fp
-            self.batch_predict = partial(self.predictor.batch_predict, *args,
-                                         **kwargs)
+            self.batch_predict = partial(
+                self.predictor.batch_predict, *args, **kwargs)
 
         def __call__(self, df: pd.DataFrame) -> pd.DataFrame:
             pd_ds = PandasDataset(df, self.features, self.data_hdf5_fp)
             predictions = self.predictor.batch_predict(
-                model=self.model, dataset=pd_ds)
+                model=self.model,
+                dataset=pd_ds,
+            )
+            for output_feature in self.model.output_features.values():
+                predictions = output_feature.flatten(predictions)
             ordered_predictions = predictions[self.output_columns]
             return ordered_predictions
 
--- a/ludwig/backend/dask.py
+++ b/ludwig/backend/dask.py
@@ -18,9 +18,8 @@
 from ludwig.backend.base import Backend, LocalTrainingMixin
 from ludwig.constants import NAME, PARQUET, PREPROCESSING, TFRECORD
 from ludwig.data.dataframe.dask import DaskEngine
-from ludwig.data.dataset.pandas import PandasDataset
 from ludwig.data.dataset.partitioned import RayDataset
-from ludwig.models.predictor import BasePredictor, Predictor, get_output_columns
+from ludwig.models.predictor import BasePredictor
 
 
 class DaskRemoteModel:
--- a/ludwig/backend/dask.py
+++ b/ludwig/backend/dask.py
@@ -59,7 +59,7 @@ class DaskPredictor(BasePredictor):
         num_gpus = int(ray.cluster_resources().get('GPU', 0) > 0)
         return dataset.ds.map_batches(
             batch_predictor,
-            compute='actores',
+            compute='actors',
             batch_format='pandas',
             num_gpus=num_gpus
         ).to_dask()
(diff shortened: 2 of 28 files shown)


ludwig/backend/ray.py
```diff
@@ -176,16 +177,20 @@ def batch_predict(self, model, dataset, *args, **kwargs):
         predictor_kwargs = self.predictor_kwargs
         output_columns = get_output_columns(model.output_features)
 
-        def batch_predict_partition(dataset):
+        def batch_predict_partition(df):
```
tgaddair, line 180: This will be more efficient if we turn it into a (stateful) actor:

```
class BatchInferModel:
    def __init__(self):
        self.model = remote_model.load()
        self.predictor = Predictor(**predictor_kwargs)
    
    def __call__(self, df: pd.DataFrame) -> pd.DataFrame:
        pd_ds = PandasDataset(df, dataset.features, dataset.data_hdf5_fp)
        predictions = self.predictor.batch_predict(self.model, pd_ds, *args, **kwargs)
        ordered_predictions = predictions[output_columns]
        return ordered_predictions

return dataset.ds.map_batchest(BatchInferModel, ...)
```

This allows us to cache the model and predictor in memory between calls.

ludwig/data/dataframe/dask.py
```diff
@@ -23,6 +23,7 @@
 import dask.dataframe as dd
 from dask.diagnostics import ProgressBar
 
+# TODO(shreya): Remove unused imports here? Esp. partitioned datasets
```
tgaddair, line 26: Yes, we should be able to remove these.

ludwig/backend/ray.py
```diff
@@ -176,16 +177,20 @@ def batch_predict(self, model, dataset, *args, **kwargs):
...
+            predictions = predictor.batch_predict(model, pd_ds, *args, **kwargs)
             ordered_predictions = predictions[output_columns]
             return ordered_predictions
-
-        return dataset.map_dataset_partitions(
-            batch_predict_partition,
-            meta=[(c, 'object') for c in output_columns]
+        
+        num_gpus = int(ray.cluster_resources().get('GPU', 0) > 0)
+        return dataset.ds.map_batches(
```
tgaddair, line 189: We may need to do a conversion to Dask DataFrame here:

```
dataset.ds.map_batches(...).to_dask()
```

The other option is to carry through the RayDataset and then only call the methods to save to a particular format in the end. This might be a better longterm solution, but might be too involved for this PR, since the code already assumes Dask DF will be used downstream. What do you think?

ShreyaR: Yeah I agree, I really like the idea of carrying Ray Datasets throughout the prediction and I can do that in a new PR.

ludwig/data/dataframe/dask.py
ShreyaR: Done

ludwig/backend/ray.py
ShreyaR: Thanks for the suggestion, done!

ludwig/backend/ray.py
```diff
@@ -163,6 +166,22 @@ def shutdown(self):
         self.executor.shutdown()
 
 
+class BatchInferModel:
+    def __init__(self, remote_model, predictor_kwargs, output_columns, features, data_hdf5_fp, *args, **kwargs):
+        self.model = remote_model.load()
```
tgaddair, line 171: We need to define this class in-line because we don't want to load the remote model until we're running in the worker context.

ludwig/data/dataset/tfrecord.py
```diff
@@ -190,6 +190,7 @@ def create(self, dataset, config, training_set_metadata):
         )
 
     def create_inference_dataset(self, dataset, tag, config, training_set_metadata):
+        # TODO(shreya): Confirm if this needs to be updated to RayDatasets too.
```
tgaddair, line 193: Probably we want to remove PartitionedDataset here and just raise an exception like:

```
raise ValueError('Batch inference not supported with TFRecord format at this time')
```

ludwig/api.py
```diff
@@ -872,6 +872,7 @@ def evaluate(
             # calculate the overall metrics
             if collect_overall_stats:
                 # TODO ray: support calculating stats on partitioned datasets
+                # TODO(shreya): Confirm what's needed to enable this.
```
tgaddair, line 875: Change `isinstance` check to `RayDataset`

ludwig/features/base_feature.py
```diff
@@ -490,3 +492,12 @@ def prepare_decoder_inputs(
         )
 
         return feature_hidden
+
+    def flatten(self, df: pd.DataFrame) -> pd.DataFrame:
+        """ Converts the output of batch_predict to a 1D array. """
+        return df
+
+    def unflatten(self, df: dd.DataFrame) -> dd.DataFrame:
```
tgaddair, line 500: We'll need to make Dask a core dependency for this type hint to work, as currently it's optional and this is on the critical path. I don't think that's a bad idea, though, as it's pretty lightweight.

ludwig/models/predictor.py
```diff
@@ -130,7 +130,13 @@ def batch_predict(
         for key, pred_value_list in predictions.items():
             predictions[key] = tf.concat(pred_value_list, axis=0).numpy()
 
-        return from_numpy_dataset(predictions)
+        predictions = from_numpy_dataset(predictions)
+
+        for of_name, output_feature in model.output_features.items():
```
tgaddair, line 135: We should probably pull this into `ray.py` as otherwise we will flatten but never unflatten for other backends.

ludwig/features/base_feature.py
tgaddair: Another option we may also want to consider using the DataFrameEngine abstraction we use elsewhere to make this agnostic to different DataFrame implementations we may want to use in the future (like Modin): 

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/base.py#L21

Here's an example of how we use it:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/strings_utils.py#L114

tgaddair (commented):
Nice!

ludwig/features/base_feature.py
ShreyaR: I like the idea of a DataFrame abstraction, but one concern I have is that some methods of the specific DF implementation might not be represented in `DataFrameEngine`. E.g., in the case of a Dask DataFrame, the `apply` method isn't defined for `DataFrameEngine`. I'm not sure how much of an issue you think this is?

One mitigation is to also keep adding methods to `DataFrameEngine` as needed.

ludwig/models/predictor.py
ShreyaR: Good catch! Fixed in the latest commit.

ludwig/backend/ray.py
```diff
@@ -210,6 +218,36 @@ def shutdown(self):
...
+        def __init__(
+                self,
+                remote_model: RayRemoteModel,
+                predictor_kwargs: Dict[str, Any],
+                output_columns: List[str],
+                features: List[Dict],
+                data_hdf5_fp: str,
+                *args, **kwargs
+        ):
+            self.model = remote_model.load()
```
tgaddair, line 231: Still not sure about how we're calling `remote_model.load()`, I feel it should cause some issues, but apparently it's not in the tests. My suggestion would be to wrap this in a function that returns a class instead of an instance, so something like this:

```
def get_batch_infer_mode(remote_model, ...):
    class BatchInferModel:
        def __init__(self):
            self.model = remote_mode.load()
            ...
    return BatchInferModel
```

Let me know if this makes sense. Basically, just want to avoid loading the model before we send it to the remote workers.

ludwig/backend/ray.py
```diff
@@ -210,6 +218,36 @@ def shutdown(self):
...
+            self.predictor = Predictor(**predictor_kwargs)
+            self.output_columns = output_columns
+            self.features = features
+            self.data_hdf5_fp = data_hdf5_fp
+            self.batch_predict = partial(
```
tgaddair, line 236: Is this being called? Looks like `self.predictor` is being used below.

ludwig/features/base_feature.py
tgaddair: That's true, but one way we work around this is to have a dummy implementation for Pandas that just does the simple thing:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/pandas.py#L52

But I think this is fine for now, as it does make it more clear that the context is different (full DF vs shard).

tgaddair (commented):
Nice work figuring out the flatten stuff. Just a couple small things, then we should be good to land.

ludwig/backend/ray.py
ShreyaR: Hey Travis, a couple of questions here:
- If the model is only loaded when sent to remote workers, we won't be able to cache the loaded model in between multiple runs? Following your comment from https://github.com/ludwig-ai/ludwig/pull/1260#discussion_r690647028
- With this implementation of using the function wrapper, does the instantiation of the `BatchInferModel` class happen inside `map_batches`?
- We could also potentially move `remote_model.load()` inside the `__call__` method?

ShreyaR: Also, I do think it's likely that this was causing issues. E.g., one stacktrace i saw was:

```
  File "/opt/hostedtoolcache/Python/3.6.14/x64/lib/python3.6/site-packages/ray/_private/function_manager.py", line 503, in _load_actor_class_from_gcs
    actor_class = pickle.loads(pickled_class)
  File "/home/runner/work/ludwig/ludwig/ludwig/models/ecd.py", line 294, in __setstate__
    self.set_weights(newstate['weights'])
  File "/opt/hostedtoolcache/Python/3.6.14/x64/lib/python3.6/site-packages/tensorflow/python/keras/engine/base_layer.py", line 1810, in set_weights
    (self.name, len(weights), expected_num_weights, str(weights)[:50]))
ValueError: You called `set_weights(weights)` on layer "ecd" with a weight list of length 5, but the layer was expecting 3 weights. Provided weights: [array([[ 0.19552892, -0.07009441, -0.17480165, -0...
```

ludwig/backend/ray.py
ShreyaR: Good catch -- fixed now!

ludwig/backend/ray.py
ShreyaR: Updated!

ludwig/backend/dask.py
```diff
@@ -18,7 +18,8 @@
 from ludwig.backend.base import Backend, LocalTrainingMixin
 from ludwig.constants import NAME, PARQUET, PREPROCESSING, TFRECORD
 from ludwig.data.dataframe.dask import DaskEngine
-from ludwig.data.dataset.partitioned import PartitionedDataset
+from ludwig.data.dataset.pandas import PandasDataset
```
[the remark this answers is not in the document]
ShreyaR: Thanks, removed in the latest commit.

ludwig/backend/dask.py
```diff
@@ -70,9 +57,9 @@ def batch_collect_activations(self, model, *args, **kwargs):
         )
 
     def _check_dataset(self, dataset):
-        if not isinstance(dataset, PartitionedDataset):
+        if not isinstance(dataset, RayDataset):
```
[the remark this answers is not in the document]
ShreyaR: Done!

ludwig/features/base_feature.py
ShreyaR: Great!

ludwig/features/base_feature.py
```diff
@@ -17,6 +17,8 @@
 from abc import ABC, abstractmethod
 from typing import Dict
 
+import dask.dataframe as dd
```
tgaddair, line 20: Because Dask is optional and this module is always imported, we need to guard this import similar to how we do it [here](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/data_utils.py#L37):

```
try:
    import dask.dataframe as dd
except ImportError:
    pass
```

For the purposes of type checking, this should be fine.

ludwig/features/sequence_feature.py
```diff
@@ -16,7 +16,9 @@
 # ==============================================================================
 import os
 
+import dask.dataframe as dd
```
tgaddair, line 19: Same thing as above regarding the import guard.

ludwig/features/text_feature.py
```diff
@@ -18,7 +18,9 @@
 from collections.abc import Iterable
 
 import numpy as np
+import pandas as pd
 import tensorflow as tf
+import dask.dataframe as dd
```
tgaddair, line 23: Import guard here too.

ludwig/data/dataset/partitioned.py
```diff
@@ -14,21 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-from ludwig.data.dataset.pandas import PandasDataset
+from typing import Dict, List
 
+from dask.dataframe import DataFrame
```
tgaddair, line 19: Both of these (dask and ray) need to be import guarded, because we import this module in `api.py`.

tgaddair (commented):
Last thing we need to double check is the optional dependency imports. I added a few comments about how we can workaround that. Let me know if it makes sense.

ludwig/features/base_feature.py
ShreyaR: Thanks for pointing this out Travis! Just fixed this. I went with the first approach you listed.

ludwig/features/sequence_feature.py
ShreyaR: Fixed.

ludwig/features/text_feature.py
ShreyaR: Fixed.

ludwig/data/dataset/partitioned.py
ShreyaR: Fixed.

tgaddair (approved):
LGTM! Awesome work.

PR #1267, opened 2021-08-23
Adding ViT image encoder and insurance lite dataset

ShreyaR:
- Adds insurance lite dataset
- Adds an encoder for VIT
- Adds a config for training on insurance_lite

# Documentation Pull Requests


ludwig/datasets/mixins/download.py
```diff
@@ -165,3 +212,33 @@ def download_raw_dataset(self):
...
+
+
+class S3FileDownloadMixin:
+    """Downloads an uncompressed file from S3."""
+
+    config: dict
+    raw_dataset_path: str
+    raw_temp_path: str
+
+    def download_raw_dataset(self):
```
tgaddair, line 224: Can we use `fsspec` for this?

ludwig/encoders/image_encoders.py
```diff
@@ -280,4 +282,71 @@ def __init__(
...
+    def __init__(
+            self,
+            image_size,
+            patch_size: int = 16,
+            num_layers: int = 8,
+            d_model: int = 64,
+            num_heads: int = 4,
+            mlp_dim: int = 128,
+            channels: int = 3,
+            dropout: float = 0.1,
```
tgaddair, line 299: To implement loading the weights, something you could do is have a param here called `use_pretrained_weights` or similar that, when true, will load weights into the layers. The weights can be either store in the repo if small, or we can consider serving them in s3 or similar if they're large.

tgaddair (commented):
Looks good. Just a couple of comments.

PR #1269, opened 2021-08-24
[WIP] Add Ray backend to Ray hyperopt

Yard1:
# Code Pull Requests
WIP
https://github.com/ludwig-ai/ludwig/issues/1262


Yard1:
@tgaddair @w4nderlust This is ready for review, though merging should be held off until both PRs referenced in description are released in their respective packages. Working on a ludwig docs PR atm.

ludwig/api.py
```diff
@@ -537,7 +538,7 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
 
                     # save training statistics
                     if self.backend.is_coordinator():
-                        if not skip_save_training_statistics:
+                        if not skip_save_training_statistics and os.path.exists(os.path.dirname(training_stats_fn)):
```
tgaddair, line 541: Can this be a remote directory? If so, we should use `fs_utils.exists`.

ludwig/callbacks.py
```diff
@@ -50,6 +50,10 @@ def on_train_start(
     def on_train_end(self, output_directory):
         pass
 
+    def on_trainer_train_setup(self, trainer, save_path):
+        """Called in EVERY trainer before training starts."""
```
tgaddair, line 54: Every trainer means on every rank?

ludwig/data/dataframe/dask_df_utils.py
```diff
@@ -48,3 +54,33 @@ def dask_to_tfrecords(
     out = Delayed(name, graph)
     out = out.compute()
     return out
+
+
+class file_lock(AbstractContextManager):
+    def __init__(self, path, timeout=None, remove_file=False) -> None:
+        self.path = Path(path) if "://" not in path else None
```
tgaddair, line 61: More reliable way might be to do:

```
protocol, _ = split_protocol(path)
self.path = Path(path) if protocol is None else None
```

ludwig/hyperopt/execution.py
```diff
@@ -750,12 +775,34 @@ def __init__(
...
+    def _gpu_resources_per_trial_non_none(self):
+        return self.gpu_resources_per_trial or 0
+
+    def _get_sync_client_and_remote_checkpoint_dir(self, trial_dir: Path):
+        """Get the Ray sync client and path to remote checkpoint directory."""
+        remote_checkpoint_dir = os.path.join(
+            self.sync_config.upload_dir, *_get_relative_checkpoints_dir_parts(trial_dir))
+        return get_cloud_sync_client(remote_checkpoint_dir), remote_checkpoint_dir
+
+    def _validate_remote_fs_for_ray_backend(self, backend, dataset, output_directory):
```
tgaddair, line 795: As discussed, we should probably remove this. Maybe we can leave it in as a warning. Also, we can use the `split_protocol` approach to determine if it's remote.

ludwig/hyperopt/execution.py
```diff
@@ -936,6 +1094,27 @@ def run_experiment_trial(config, checkpoint_dir=None):
...
             )
 
+        if _is_ray_backend(backend):
+            # we can't set Trial actor's CPUs to 0 so we just go very low
+            resources_per_trial = PlacementGroupFactory(
+                [{"CPU": 0.001}] + ([{"CPU": 1, "GPU": 1}] * self._gpu_resources_per_trial_non_none) if self._gpu_resources_per_trial_non_none else (
+                    [{"CPU": 0.001}] + [{"CPU": 1}] * self._cpu_resources_per_trial_non_none)
+            )
+
+        if "://" in output_directory:
```
tgaddair, line 1104: Same thing here about `split_protocol`.

requirements_ray.txt
```diff
@@ -2,4 +2,4 @@ ray[default,tune]>=1.6.0
 pickle5
 tensorboardX<2.3
 GPUtil
-
+awscli
```
tgaddair, line 5: Is this needed?

ludwig/hyperopt/execution.py
```diff
@@ -27,11 +29,33 @@
...
+    get_horovod_kwargs = None
+
+    class RayBackend:
+        pass
+
+    class RayRemoteTrainer:
+        pass
+
+
+def _is_ray_backend(backend) -> bool:
```
tgaddair, line 51: In general, it would be nice if we could avoid calling `_is_ray_backend()` in so many places. This could be a big refactor, so feel free to just add a TODO. But basically my preference would be to encapsulate this behavior behind an interface, like a `HyperoptBackend` or similar.

tgaddair (commented):
Looks great! Only things I would change initially would be to have a standard way of dealing with remote vs local paths (i.e., `fs_utils`) and to abstract away some of the Ray backend specific logic into an interface. As mentioned in the comments, we can save some of this for a follow-up PR as there is a lot here already.

ludwig/callbacks.py
Yard1: Correct, will make it more explicit

requirements_ray.txt
Yard1: It's needed for Ray to sync from s3. We can leave it out, in which case Ray will throw an exception asking for it to be installed

tgaddair: Definitely fine to leave it in if needed. I wonder if this is also an issue for GCS and Azure Storage?

ludwig/hyperopt/execution.py
Yard1: Added a TODO note

tgaddair (approved):
LGTM! I will try and prepare a release for Horovod so we can land this.

PR #1272, opened 2021-08-28
Update Ludwig dependency specifications for pytorch support

jimthompson5802:
To support the TF to Torch port.

# Code Pull Requests

This adds PyTorch dependencies to Ludwig dependencies requirement*.txt file


tgaddair (commented):
Nice! Any particular reason for pinning the package versions? Ideally we would want to support as many as possible, at least >= the current versions.

jimthompson5802:
No particular reason for pinning.  I can alter the version to be more general >=

tgaddair (approved):

PR #1274, opened 2021-08-29
Update categorical features to torch

tgaddair:


ludwig/modules/initializer_modules.py
```diff
@@ -25,7 +26,7 @@
     # 'ones': tf.initializers.ones,
     # 'orthogonal': tf.initializers.orthogonal,
     # 'normal': tf.random_normal_initializer,
-    # 'uniform': tf.random_uniform_initializer,
+    'uniform': torch.nn.init.uniform_,
     # 'truncated_normal': tf.keras.initializers.TruncatedNormal,
     # 'variance_scaling': tf.keras.initializers.VarianceScaling,
     # 'glorot_normal': tf.initializers.glorot_normal,
```
jimthompson5802, line 32: For my purpose, I could use xavier_uniform and xavier_normal initializers.

tgaddair: These should now be available :)

PR #1277, opened 2021-08-30
Do not initialize Ray if already initalized

Yard1:
This PR ensures that Ray is not initialized if it's already initialized. The previous way of handling that case by `ignore_reinit_error` causes an exception to be thrown with Ray Client.


tgaddair (approved):
LGTM!

PR #1282, opened 2021-09-03
Torch image features

ShreyaR:
WIP PR for migrating image features from TF to PyTorch

### DIFF
--- a/ludwig/encoders/image_encoders.py
+++ b/ludwig/encoders/image_encoders.py
@@ -252,8 +252,8 @@ class ResNetEncoder(ImageEncoder):
 
         return {'encoder_output': hidden}
 
-    def get_output_shape(self, input_shape):
-        # TODO(shreya): Confirm that this is implemented
+    @property
+    def output_shape(self) -> torch.Size:
         return self.fc_stack.output_shape
 
 
--- a/ludwig/encoders/image_encoders.py
+++ b/ludwig/encoders/image_encoders.py
@@ -96,7 +96,7 @@ class Stacked2DCNN(ImageEncoder):
         self.conv_stack_2d = Conv2DStack(
             layers=conv_layers,
             num_layers=num_conv_layers,
-            default_in_channels=in_channels,
+            first_in_channels=in_channels,
             default_out_channels=out_channels,
             kernel_size=kernel_size,
             default_stride=stride,
--- a/ludwig/encoders/image_encoders.py
+++ b/ludwig/encoders/image_encoders.py
@@ -17,15 +17,13 @@
 import logging
 from abc import ABC
 
-# import tensorflow as tf
-# from tensorflow.keras.layers import Flatten
+import torch
 
 from ludwig.encoders.base import Encoder
 from ludwig.modules.mlp_mixer_modules import MLPMixer
 from ludwig.utils.registry import Registry, register, register_default
-from ludwig.modules.convolutional_modules import Conv2DStack, \
+from ludwig.modules.convolutional_modules import Conv2DStack, ResNet,\
     get_resnet_block_sizes
-from ludwig.modules.convolutional_modules import ResNet
 from ludwig.modules.fully_connected_modules import FCStack
 
 logger = logging.getLogger(__name__)
@@ -45,19 +43,23 @@ class Stacked2DCNN(ImageEncoder):
 
     def __init__(
             self,
+            img_height,
+            img_width,
             conv_layers=None,
             num_conv_layers=None,
-            filter_size=3,
-            num_filters=32,
-            strides=(1, 1),
+            in_channels=None,
+            out_channels=32,
+            kernel_size=3,
+            stride=1,
             padding='valid',
-            dilation_rate=(1, 1),
-            conv_use_bias=True,
-            conv_weights_initializer='glorot_uniform',
-            conv_bias_initializer='zeros',
-            conv_weights_regularizer=None,
-            conv_bias_regularizer=None,
-            conv_activity_regularizer=None,
+            dilation=1,
+            conv_bias=True,
+            padding_mode='zeros',
+            # conv_weights_initializer='glorot_uniform',
+            # conv_bias_initializer='zeros',
+            # conv_weights_regularizer=None,
+            # conv_bias_regularizer=None,
+            # conv_activity_regularizer=None,
             # conv_weights_constraint=None,
             # conv_bias_constraint=None,
             conv_norm=None,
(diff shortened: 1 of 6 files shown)


ludwig/utils/image_utils.py
```diff
@@ -18,8 +18,11 @@
 import os
 import sys
 from math import ceil, floor
+from typing import Tuple, Union
 
 import numpy as np
+# TODO(shreya): Import guard?
```
tgaddair, line 24: This should be okay, since torch is going to be a required dependency going forward.

ludwig/utils/image_utils.py
```diff
@@ -154,4 +159,51 @@ def num_channels_in_image(img):
...
+    """
+    if type(prop) == int:
+        return np.ones(2) * prop
+    elif type(prop) == tuple and len(prop) == 2:
+        return np.array(list(prop))
+    else:
+        raise TypeError(f'kernel_size must be int or tuple of length 2.')
+
+
+def get_img_output_shape(
```
tgaddair, line 179: I'm wondering if we can do something similar to what `pytorch-summary` provides here for more complicated modules such as CNNs:

https://github.com/sksq96/pytorch-summary

I will post about this on the Ludwig channel, please share your thoughts as well!

ludwig/features/image_feature.py
```diff
@@ -16,32 +16,35 @@
...
+# TODO(shreya): Confirm if it's ok to do per channel normalization
+# TODO(shreya): Also confirm if this is being used anywhere
+# TODO(shreya): Confirm if ok to use imagenet means and std devs
 image_scaling_registry = {
     'pixel_normalization': lambda x: x * 1.0 / 255,
-    'pixel_standardization': lambda x: tf.map_fn(
-        lambda f: tf.image.per_image_standardization(f), x)
+    'pixel_standardization': partial(
+        torchvision.transforms.functional.normalize,
+        mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]
```
tgaddair, line 46: Is this equivalent to what `tf.image` was doing?

ludwig/encoders/image_encoders.py
```diff
@@ -127,61 +131,64 @@ def __init__(
...
             default_activation=fc_activation,
             default_dropout=fc_dropout,
         )
 
-    def call(self, inputs, training=None, mask=None):
+    def call(self, inputs):
         """
             :param inputs: The inputs fed into the encoder.
-                    Shape: [batch x height x width x channels], type tf.uint8
+                    Shape: [batch x height x width x channels], type torch.uint8
```
tgaddair, line 143: Has this been changed to NCHW here, or is that still to be.done?

tgaddair (commented):
Nice work! Would be good to sync with Piero as well on the remaining TODOs.

ludwig/encoders/image_encoders.py
```diff
@@ -40,84 +39,89 @@ def register(cls, name):
...
+            fc_norm_params: Optional[Dict[str, Any]] = None,
+            fc_activation: str = 'relu',
+            fc_dropout: float = 0,
             **kwargs
     ):
         super().__init__()
 
         logger.debug(' {}'.format(self.name))
 
+        assert first_in_channels is not None
```
tgaddair, line 89: Generally prefer not to use assertions because:

- They cannot be recovered from using try-except
- They can be disabled at runtime based on user config

If this is something we want to check, and not just a sanity check for us as developers, then we should make this a `ValueError`.

ludwig/encoders/image_encoders.py
```diff
@@ -127,61 +131,64 @@ def __init__(
...
-            # default_weights_constraint=fc_weights_constraint,
-            # default_bias_constraint=fc_bias_constraint,
             default_norm=fc_norm,
             default_norm_params=fc_norm_params,
             default_activation=fc_activation,
             default_dropout=fc_dropout,
         )
 
-    def call(self, inputs, training=None, mask=None):
+    def call(self, inputs):
```
tgaddair, line 140: Should this be `forward`?

ludwig/encoders/image_encoders.py
```diff
@@ -127,61 +131,64 @@ def __init__(
...
         )
-        hidden = self.flatten(hidden, training=training)
+        hidden = self.flatten(hidden)
 
         # ================ Fully Connected ================
         outputs = self.fc_stack(hidden)
 
         return {'encoder_output': outputs}
 
+    def get_output_shape(self, input_shape):
```
tgaddair, line 157: I think this was changed to a property `output_shape`.

ludwig/encoders/image_encoders.py
```diff
@@ -223,52 +237,52 @@ def __init__(
...
-            # default_weights_constraint=fc_weights_constraint,
-            # default_bias_constraint=fc_bias_constraint,
             default_norm=norm,
             default_norm_params=norm_params,
             default_activation=activation,
             default_dropout=dropout,
         )
 
-    def call(self, inputs, training=None, mask=None):
+    def call(self, inputs):
```
tgaddair, line 246: Same as above regarding `forward` and `output_shape`.

tgaddair (approved):
LGTM! Just a few things we should fix either now or in a follow-up.

ludwig/features/image_feature.py
ShreyaR: Not really -- I couldn't find the equivalent of this in torch. I've standardized images using the ImageNet means and std devs right now, which is usually a reasonable standardization.

Another alternative -- I could change this to implement my own per image standardization. Lmk what you think!

ludwig/encoders/image_encoders.py
ShreyaR: This should be done.

ShreyaR: I need to test everything out, but I've updated image utils to use torchvision.

ludwig/encoders/image_encoders.py
ShreyaR: Done!

ludwig/utils/image_utils.py
ShreyaR: Sounds good!

ludwig/utils/image_utils.py
ShreyaR: Solved on slack.

ludwig/encoders/image_encoders.py
ShreyaR: Sounds good -- fixed in the latest commit.

ludwig/encoders/image_encoders.py
ShreyaR: Thanks, fixed!

ludwig/encoders/image_encoders.py
ShreyaR: Fixed!

PR #1283, opened 2021-09-03
Torch output shape support

ShreyaR:
This PR adds support for output and input shapes for all Ludwig components.

- Adds a `LudwigComponent` class that supports getting the `output_shape` via running `forward` on a dummy tensor of `input_shape`
- Adds `LudwigComponent` as the parent class of all modules.
- Adds `LudwigComponent` to encoders.
- Adds `LudwigComponent` to combiners.
- Update features to use `output_shape`, `input_shape` properties instead of methods `get_output_shape`, `get_input_shape`.

### DIFF
--- a/ludwig/features/binary_feature.py
+++ b/ludwig/features/binary_feature.py
@@ -145,8 +145,8 @@ class BinaryInputFeature(BinaryFeatureMixin, InputFeature):
     def get_input_dtype(cls):
         return tf.bool
 
-    def get_input_shape(self):
-        return ()
+    # def get_input_shape(self):
+    #     return ()
 
     @staticmethod
     def update_config_with_metadata(
@@ -232,8 +232,9 @@ class BinaryOutputFeature(BinaryFeatureMixin, OutputFeature):
     def get_output_dtype(cls):
         return tf.bool
 
-    def get_output_shape(self):
-        return ()
+    @property
+    def output_shape(self) -> torch.Size:
+        return super().output_shape
 
     @staticmethod
     def update_config_with_metadata(
--- a/ludwig/features/binary_feature.py
+++ b/ludwig/features/binary_feature.py
@@ -145,8 +145,9 @@ class BinaryInputFeature(BinaryFeatureMixin, InputFeature):
     def get_input_dtype(cls):
         return tf.bool
 
-    # def get_input_shape(self):
-    #     return ()
+    @property
+    def input_shape(self) -> torch.Size:
+        return torch.Size([1])
 
     @staticmethod
     def update_config_with_metadata(
@@ -234,7 +235,7 @@ class BinaryOutputFeature(BinaryFeatureMixin, OutputFeature):
 
     @property
     def output_shape(self) -> torch.Size:
-        return super().output_shape
+        return torch.Size([1])
 
     @staticmethod
     def update_config_with_metadata(
--- a/ludwig/features/h3_feature.py
+++ b/ludwig/features/h3_feature.py
@@ -118,12 +118,9 @@ class H3InputFeature(H3FeatureMixin, InputFeature):
     def get_input_dtype(cls):
         return tf.uint8
 
-    # def get_input_shape(self):
-    #     return H3_VECTOR_LENGTH,
-
     @property
     def input_shape(self) -> torch.Size:
-        return torch.Size((H3_VECTOR_LENGTH, ))
+        return torch.Size([H3_VECTOR_LENGTH, ])
 
     @staticmethod
     def update_config_with_metadata(
--- a/ludwig/features/h3_feature.py
+++ b/ludwig/features/h3_feature.py
@@ -17,6 +17,7 @@
 import logging
 
 import numpy as np
+import torch
 # import tensorflow as tf
 
 from ludwig.constants import *
@@ -117,8 +118,12 @@ class H3InputFeature(H3FeatureMixin, InputFeature):
     def get_input_dtype(cls):
         return tf.uint8
 
-    def get_input_shape(self):
-        return H3_VECTOR_LENGTH,
+    # def get_input_shape(self):
+    #     return H3_VECTOR_LENGTH,
+
+    @property
+    def input_shape(self) -> torch.Size:
+        return torch.Size((H3_VECTOR_LENGTH, ))
 
     @staticmethod
     def update_config_with_metadata(
--- a/ludwig/features/h3_feature.py
+++ b/ludwig/features/h3_feature.py
@@ -120,7 +120,7 @@ class H3InputFeature(H3FeatureMixin, InputFeature):
 
     @property
     def input_shape(self) -> torch.Size:
-        return torch.Size([H3_VECTOR_LENGTH, ])
+        return torch.Size([H3_VECTOR_LENGTH])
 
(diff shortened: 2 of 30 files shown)


ludwig/utils/torch_utils.py
```diff
@@ -93,6 +93,23 @@ def losses(self):
...
+
+    @property
+    def input_shape(self) -> torch.Size:
+        """ Returns the size of the input tensor without the batch dimension. """
+        raise NotImplementedError('Abstract class.')
+
+    @property
+    def output_shape(self) -> torch.Size:
+        """ Returns the size of the output tensor without the batch dimension."""
+        output_tensor = self.forward(torch.rand(*self.input_shape))
```
tgaddair, line 109: General advice is to avoid calling `forward` directly. Not sure it applies here, but probably good practice to follow:

https://discuss.pytorch.org/t/any-different-between-model-input-and-model-forward-input/3690

ludwig/utils/torch_utils.py
```diff
@@ -93,6 +93,23 @@ def losses(self):
...
+        super().__init__()
+
+    @property
+    def input_shape(self) -> torch.Size:
+        """ Returns the size of the input tensor without the batch dimension. """
+        raise NotImplementedError('Abstract class.')
+
+    @property
+    def output_shape(self) -> torch.Size:
+        """ Returns the size of the output tensor without the batch dimension."""
```
tgaddair, line 108: The input shape will not include batch size, so we should account for that. PyTorch Summary uses batch size 2 for batch norm. Here's an example:

https://github.com/sksq96/pytorch-summary/blob/master/torchsummary/torchsummary.py#L59

ludwig/utils/torch_utils.py
```diff
@@ -93,6 +93,23 @@ def losses(self):
             collected_losses.extend(layer.losses)
         return collected_losses
 
+
+class LudwigComponent(Module):
```
tgaddair, line 97: We probably want to inherit from `LudwigModule` so we can get the loss tracking features.

ludwig/features/binary_feature.py
```diff
@@ -232,8 +232,9 @@ def get_prediction_set(self):
     def get_output_dtype(cls):
         return tf.bool
 
-    def get_output_shape(self):
-        return ()
+    @property
+    def output_shape(self) -> torch.Size:
```
tgaddair, line 236: This should be `torch.Size([1])`

ludwig/features/binary_feature.py
```diff
@@ -145,8 +145,8 @@ def call(self, inputs, training=None, mask=None):
     def get_input_dtype(cls):
         return tf.bool
 
-    def get_input_shape(self):
-        return ()
+    # def get_input_shape(self):
+    #     return ()
```
tgaddair, line 149: This should be `torch.Size([1])`

ludwig/features/h3_feature.py
```diff
@@ -117,8 +118,12 @@ def call(self, inputs, training=None, mask=None):
...
         return tf.uint8
 
-    def get_input_shape(self):
-        return H3_VECTOR_LENGTH,
+    # def get_input_shape(self):
+    #     return H3_VECTOR_LENGTH,
+
+    @property
+    def input_shape(self) -> torch.Size:
+        return torch.Size((H3_VECTOR_LENGTH, ))
```
tgaddair, line 126: For consistency, I would use a list instead of a tuple here.

ludwig/utils/torch_utils.py
```diff
@@ -93,6 +93,23 @@ def losses(self):
...
+    def __init__(self):
+        super().__init__()
+
+    @property
+    def input_shape(self) -> torch.Size:
```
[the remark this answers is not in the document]
ShreyaR: Good catch, done!

ludwig/features/sequence_feature.py
```diff
@@ -301,8 +302,9 @@ def get_prediction_set(self):
     def get_output_dtype(cls):
         return tf.int32
 
-    def get_output_shape(self):
-        return self.max_sequence_length,
+    @property
+    def output_shape(self) -> torch.Size:
+        return torch.Size((self.max_sequence_length, ))
```
tgaddair, line 307: As above, we should stick to either list or tuple.

ludwig/features/set_feature.py
```diff
@@ -136,8 +137,12 @@ def call(self, inputs, training=None, mask=None):
...
         return tf.bool
 
-    def get_input_shape(self):
-        return len(self.vocab),
+    # def get_input_shape(self):
+    #     return len(self.vocab),
+
+    @property
+    def input_shape(self) -> torch.Size:
+        return torch.Size((len(self.vocab), ))
```
tgaddair, line 145: Same as above.

ludwig/features/set_feature.py
```diff
@@ -231,8 +236,12 @@ def get_prediction_set(self):
...
         return tf.bool
 
-    def get_output_shape(self):
-        return self.num_classes,
+    # def get_output_shape(self):
+    #     return self.num_classes,
+
+    @property
+    def output_shape(self) -> torch.Size:
+        return torch.Size((self.num_classes, ))
```
tgaddair, line 244: Same as above.

ludwig/utils/torch_utils.py
ShreyaR: I think the advantage of using `__call__` instead of `forward` is to call the forward and backward hooks. In our case it's probably more efficient to not call the hooks since we only want to use forward to compute the shapes, if in case the forward and backward hooks end up having some expensive computation.

I'm happy to switch to the `__call__` method if you prefer though.

ludwig/utils/torch_utils.py
ShreyaR: Done!

tgaddair (commented):
In general, we should remove the commented out code and align on using either list or tuple as the param type to Size. Other than that, looks great!

ludwig/utils/torch_utils.py
ShreyaR: Good catch, I've added a dummy batch dimension of 2 right now!

ludwig/utils/torch_utils.py
tgaddair: Since in most cases the hooks are just side effects, we probably don't want to call them here, I agree. So for now I think we can stick with forward.

ludwig/features/binary_feature.py
ShreyaR: Done

ludwig/features/binary_feature.py
ShreyaR: Done

ludwig/features/h3_feature.py
ShreyaR: Done

ludwig/features/sequence_feature.py
ShreyaR: Done

ludwig/features/set_feature.py
ShreyaR: Done

tgaddair (approved):

PR #1295, opened 2021-09-09
Cast `max_length` to int in `build_sequence_matrix::pad`

Yard1:
Running `tests/integration_tests/test_ray.py ` using Ray Client caused an exception due to `max_length` in  `build_sequence_matrix::pad` being of `numpy.float64` dtype. Casting it to int fixed the exception.


Yard1:
@tgaddair would love if this could get merged ASAP, thanks!

tgaddair (approved):

tgaddair:
Nice! Will merge once tests pass.

PR #1297, opened 2021-09-10
Torch images tests

ShreyaR:
Added tests for image modules.


ludwig/modules/convolutional_modules.py
```diff
@@ -101,7 +98,7 @@ def __init__(
             ))
 
         for layer in self.layers:
-            logger.debug('   {}'.format(layer.name))
+            logger.debug('   {}'.format(layer._get_name()))
```
tgaddair, line 101: Can we implement this `_get_name` as a property called `name` so we don't need to call private functions?

ludwig/modules/convolutional_modules.py
```diff
@@ -841,75 +855,71 @@ def __init__(
...
 
         self.conv1 = Conv2DLayerFixedPadding(
-            img_height=img_height,
-            img_width=img_width,
-            in_channels=first_in_channels,
-            out_channels=out_channels,
-            kernel_size=3,
-            stride=stride,
-        )
+            img_height, img_width, first_in_channels, out_channels,
```
tgaddair, line 872: Recommend using keyword args here, as using args alone can sometimes result in difficult to debug errors when interfaces change.

tgaddair (approved):
This looks great! Tests are amazing. I suggest @w4nderlust also taking a look at the conv modules changes.

ludwig/modules/convolutional_modules.py
ShreyaR: I could add a property in the LudwigModule, but the issue is that if some layers are pytorch layers directly then this would cause an issue. E.g. directly used Activation, Pooling layers that aren't wrapped in a LudwigModule would raise an error.

w4nderlust: makes sense, we don't want to wrap everything really

tests/ludwig/encoders/test_image_encoders.py
```diff
@@ -0,0 +1,40 @@
...
+    outputs = stacked_2d_cnn(inputs)
+    assert outputs['encoder_output'].shape[1:] == stacked_2d_cnn.output_shape
+
+
+@pytest.mark.parametrize('img_height,img_width', [(224, 224)])
+def test_resnet_encoder(img_height: int, img_width: int):
+    resnet = ResNetEncoder(img_height, img_width)
+    inputs = torch.rand(2, 3, img_height, img_width)
+    outputs = resnet(inputs)
+    assert outputs['encoder_output'].shape[1:] == resnet.output_shape
```
jimthompson5802, line 32: Just an observation...`reset.output_shape` is set by the fully connected layer `output_shape`.  The `output_shape` for fully connected is calculated by using the default `LudwigModule.output_shape` property, which sends a "fake" tensor through the `forward()` method.  

It seems that the `assert` test is comparing the outputs that are derived from the same mechanism.  So if an error occurred in constructing the output tensor from the fully connected layer, then the assertion would not catch the discrepancy.   If this was a math proof, isn't this like assuming the consequence. :-)

From what I can tell, this comment applies to the other tests for image encoders as well.

OTOH,  in the assert test for shapes in the 2D convolutional tests one side is derived directly from the parameters of the constructors, which is independent of the processing in the `forward()` methods.

ludwig/modules/activation_modules.py
```diff
@@ -1,24 +1,22 @@
-# import tensorflow as tf
+import torch
+import torch.nn as nn
 
 
 # taken from https://github.com/google-research/google-research/blob/master/tabnet/tabnet_model.py
-def glu(x, n_units=None):
+def glu(x: torch.Tensor, dim: int = -1):
```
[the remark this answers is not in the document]
ShreyaR: I'd updated the tabnet module to be consistent with the new function signature. The change is available in the file ludwig/modules/tabnet_modules.py in this patch below.

tests/ludwig/encoders/test_image_encoders.py
ShreyaR: @jimthompson5802 good catch, this set of tests won't catch the errors from the FC forward! IMO, the encoder output tests the the full forward pass of the encoder, while the resnet output property is the independently computed output shape of the module. For the FC layer, the output shape is computed by calling forward only on the FC layer, so errors that originate in piping the Resnet module to the FC layer will be picked up. However, I'm open to suggestions on how to catch errors in FC output tensors.

tests/ludwig/modules/test_utils.py
```diff
@@ -0,0 +1,19 @@
...
+        module: LudwigModule,
+        input_shape: Tuple[int]
+):
+    """ Runs a unit test to confirm that the out shape matches expected output.
+
+    module: Module to be tested.
+    input_shape: List of integers of the expected input shape (w/o batch dim).
+    """
+
+    inputs = torch.rand(2, *input_shape)
```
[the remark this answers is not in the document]
ShreyaR: Good catch, I've updated in the latest patch!

ludwig/modules/convolutional_modules.py
```diff
@@ -841,75 +855,71 @@ def __init__(
...
 
         self.conv2 = Conv2DLayerFixedPadding(
-            img_height=img_height,
-            img_width=img_width,
-            in_channels=out_channels,
-            out_channels=out_channels,
-            kernel_size=3,
-            stride=stride,
-        )
+            img_height, img_width, out_channels, out_channels, kernel_size=3,
```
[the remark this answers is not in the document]
ShreyaR: Done!

ludwig/modules/convolutional_modules.py
```diff
@@ -546,45 +543,35 @@ def __init__(
...
-            out_channels=out_channels,
-            kernel_size=kernel_size,
-            stride=stride,
-            padding=padding,
-            dilation=dilation,
-            groups=groups,
-            bias=bias,
-            padding_mode=padding_mode
-        ))
+            in_channels, out_channels, kernel_size, stride, padding, dilation,
```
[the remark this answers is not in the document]
ShreyaR: Done!

ludwig/utils/image_utils.py
```diff
@@ -180,18 +180,21 @@ def num_channels_in_image(img: torch.Tensor):
...
-    elif type(prop) == tuple and len(prop) == 2:
-        return np.array(list(prop))
+        return np.ones(2).astype(int) * prop
+    elif isinstance(prop, Iterable) and len(prop) == 2:
+        return np.array(list(prop)).astype(int)
+    elif type(prop) == np.ndarray and prop.size == 2:
+        return prop.astype(int)
     else:
-        raise TypeError(f'kernel_size must be int or tuple of length 2.')
+        raise TypeError(f'prop must be int or tuple of length 2, but is {prop}.')
```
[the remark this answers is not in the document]
ShreyaR: Good point -- updated!

ludwig/modules/convolutional_modules.py
ShreyaR: Done!

tests/ludwig/encoders/test_image_encoders.py
jimthompson5802: That's reasonable, i.e., if FC has the independent verification, then anything using it can assume proper operation and make use of its results

For the sequence encoders, the approach I took was to use the relevant values of the encoder parameters for the output shape assertion test.  [Here is an example.](https://github.com/jimthompson5802/ludwig/blob/f09358e4530cdd3871932c97db0e1c1b922c159c/tests/integration_tests/test_sequence_encoders.py#L90-L126)

I think either approach should meet the requirement.

ludwig/utils/image_utils.py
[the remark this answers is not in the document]
w4nderlust: I suggest tuple -> Iterable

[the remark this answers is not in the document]
w4nderlust: Ah sorry, GitHub is not showing the change but the original issue :)

tests/ludwig/encoders/test_image_encoders.py
w4nderlust: FYI I'm mergin the PR, if we want to do modifications tothe tests we could do them in a follow up PR for both seq and image tests

PR #1300, opened 2021-09-15
Refactor: Tf to torch implement input_dtype property in LudwigModule

jimthompson5802:
# Code Pull Requests

Implement the following refactoring:

- Move get_input_dtype() from InputFeature to LudwigModule
- change get_input_dtype() from method to property input_dtype
- Have the default behavior be return torch.float32 and allow subclasses to override

Successfully tested this change with the sequence encoders.


ShreyaR (commented):
lgtm -- thanks for putting together the PR Jim!

PR #1305, opened 2021-09-17
Tf to torch unit tests For Conv1D related Layers

jimthompson5802:
# Code Pull Requests

Provides more granular unit tests for PyTorch Conv1D related layers


tests/ludwig/modules/test_convolutional_modules.py
```diff
@@ -1,11 +1,109 @@
...
+BATCH_SIZE = 2
+SEQ_SIZE = 100
+HIDDEN_SIZE = 64
+
+
+###
+# Helper function to compute expected output shape
+# for Conv1D related layers
+###
+def expected_seq_size(
```
ShreyaR, line 20: You could use the [get_img_output_shape](https://github.com/ludwig-ai/ludwig/blob/81f86c1623646064cbabde716ee45b5f29f98663/ludwig/utils/image_utils.py#L179) function for this within image_utils.py. That function operates on convs and pools independently, so you could call the function twice if you wanted to perform both a conv followed by a pool.

tests/ludwig/modules/test_convolutional_modules.py
```diff
@@ -1,11 +1,109 @@
...
+        pool_padding: str,
+        pool_stride: int,
+        pool_function: str
+) -> None:
+    input = torch.randn([BATCH_SIZE, SEQ_SIZE, HIDDEN_SIZE],
+                        dtype=torch.float32)
+
+    conv1_layer = Conv1DLayer(
+        in_channels=HIDDEN_SIZE,
+        out_channels=256,
```
ShreyaR, line 81: (optional) you could take `out_channels` a pytest parameter

ShreyaR (commented):
looks great! added a few comments

tests/ludwig/modules/test_convolutional_modules.py
w4nderlust: and make it smaller potentially so speed up test :)

tests/ludwig/modules/test_convolutional_modules.py
jimthompson5802: As I understand the implementation of `get_img_output_shape`, it is geared for a 2D structure.  To adapt it for the 1D structures, do I set, for example, `img_height` to zero?

tests/ludwig/modules/test_convolutional_modules.py
jimthompson5802: done.

tests/ludwig/modules/test_convolutional_modules.py
jimthompson5802: On a lark, I tried setting `img_height` to zero and it works.  Here is the code structure:
```
    output_seq_size = get_img_output_shape(
        0,  # img_height set to zero for 1D structure
        SEQ_SIZE,  # img_width equates to sequence size
        kernel_size, # kernel_size for conv layer
        strides,  # stride for conv layer
        padding,  # padding for conv layer
        dilation  # dilation for conv layer
    )
    if pool_size is not None:
        # adjust expected output shape for pooling layer
        output_seq_size = get_img_output_shape(
            0,  # img_height set to zero for 1D structure
            output_seq_size[1],  # img_width equates to sequence size
            pool_size,  # kernel size for pooling layer
            pool_stride,  # stride for pooling layer
            pool_padding,  # padding for polling layer
            1  # pooling layer only support unit dilation
        )
    assert out_tensor.size() == (BATCH_SIZE, output_seq_size[1], NUM_FILTERS)

```

if this is what you had in mind, I can retire the `expected_seq_size()` or actually change the implementation in that function to use the `get_img_output_shape()` function instead of the hand coded formulas.

ShreyaR: Nice catch with setting `img_height=0`!

Yep, that's what I had in mind. Up to you -- I would be ok with either choice. Could you also remove the comments in the arguments, and use keyword args instead?

jimthompson5802: done.

tests/ludwig/modules/test_convolutional_modules.py
```diff
@@ -1,13 +1,291 @@
...
+from ludwig.utils.image_utils import get_img_output_shape
+from ludwig.modules.convolutional_modules import Conv2DLayer, Conv2DStack, \
+    Conv2DLayerFixedPadding, ResNetBlock, ResNetBottleneckBlock, \
+    ResNetBlockLayer, ResNet, Conv1DLayer, Conv1DStack, ParallelConv1DStack, \
+    ParallelConv1D
+
+BATCH_SIZE = 2
+SEQ_SIZE = 50
+HIDDEN_SIZE = 64
+NUM_FILTERS = 32
```
w4nderlust, line 15: Let's make them a bit smaller, for instance:
  SEQ_SIZE = 10
  HIDDEN_SIZE = 8
  NUM_FILTERS = 4
to speed up tests

jimthompson5802: I can make the change but then I'll have to reduce the combinations tested.  For example, with seq_size=10, will fail with the larger values in this set of parameters:
```
@pytest.mark.parametrize('pool_size, pool_padding, pool_stride',
                         [(None, None, None), (3, 'same', 1), (5, 'same', 1),
                          (3, 'valid', 2), (5, 'valid', 2)])
```
Any recommendations on the range of values to test for?

jimthompson5802: Here are some of the errors that occur with the smaller sizes:
```
>       return torch.max_pool1d(input, kernel_size, stride, padding, dilation, ceil_mode)
E       RuntimeError: Given input size: (4x1x4). Calculated output size: (4x1x0). Output size is too small
E       RuntimeError: Calculated padded input size per channel: (10). Kernel size: (13). Kernel size can't be greater than actual input size
E       RuntimeError: Given input size: (4x1x2). Calculated output size: (4x1x-1). Output size is too small
```

jimthompson5802: In terms for speed the `test_conv1d_layer` is the one most affected by the change in parameters.  With the sizing that is currently inlace, before your recommendations, 360 tests are generated and it takes about 2 seconds on my laptop.

jimthompson5802: @w4nderlust   made the adjustments per comments except for seq_size.  The minimum value required to allow testing some of the combinations of parameter is 17.  I did cut back on some of combinations.  So the number of generated test cases is about half.  From what I can tell, this gives the minimum combinations that will exercise all the code paths.

PR #1306, opened 2021-09-18
Vit image encoder v2

ShreyaR:
This PR
- Adds insurance lite dataset
- Adds an encoder for VIT along with unit tests
- Adds a config for training on insurance_lite


ludwig/data/preprocessing.py
```diff
@@ -1191,7 +1193,11 @@ def cast_columns(dataset_df, features, global_preprocessing_parameters,
 
 
 def build_metadata(
-        metadata, dataset_cols, features, global_preprocessing_parameters, backend
+        metadata: Dict[str, Any],
+        dataset_cols: Dict[str, dd.Series],
```
tgaddair, line 1197: This can also be a Pandas Series. One thing we may want to do is create our own typing module for Ludwig that has these generic DataFrame abstractions. One thing we need to incorporate here is the possibility that Dask is not installed (since it's optional), so that should be included in the type.

ludwig/data/preprocessing.py
```diff
@@ -16,11 +16,13 @@
 # ==============================================================================
 import logging
 from abc import ABC, abstractmethod
+from typing import Any, Dict, List
 
+import dask.dataframe as dd
```
tgaddair, line 21: See comment below, but either we will ned to import guard this, or (better) create a new type for a Series that can be either Pandas or Dask (iff Dask is installed).

ludwig/datasets/insurance_lite/config.yaml
```diff
@@ -0,0 +1,3 @@
+version: 1.0
+s3_download_urls: s3://datasets.us-west-2.predibase.com/insurance.csv
```
tgaddair, line 2: Is this public? Maybe we can use the Kaggle API to pull the dataset.

ludwig/datasets/mixins/download.py
```diff
@@ -165,3 +212,33 @@ def download_raw_dataset(self):
     @property
     def download_url(self):
         return self.config["download_urls"]
+
+
+class S3FileDownloadMixin:
```
tgaddair, line 217: No need to use S3 specific stuff, we can just use `fsspec`, which abstracts away these S3-like filesystems. See `fs_utils` in Ludwig for examples.

ludwig/encoders/image_encoders.py
```diff
@@ -296,3 +296,88 @@ def input_shape(self) -> torch.Size:
...
+        super().__init__()
+        try:
+            from transformers import ViTConfig, ViTModel
+        except ModuleNotFoundError:
+            logger.error(
+                ' transformers is not installed. '
+                'In order to install all image feature dependencies run '
+                'pip install ludwig[image]'
+            )
+            sys.exit(-1)
```
tgaddair, line 339: We should try to do away with these `sys.exit` calls, since they don't work for programmatic usage (there's no way to recover from them). Instead, let's just raise a `RuntimeError` with this message.

ludwig/encoders/image_encoders.py
```diff
@@ -296,3 +296,88 @@ def input_shape(self) -> torch.Size:
...
+
+        img_width = img_width or img_height
+        if not img_width == img_height:
```
tgaddair, line 342: More idiomatic would be: `img_weight != image_height`.

ludwig/encoders/image_encoders.py
```diff
@@ -296,3 +296,88 @@ def input_shape(self) -> torch.Size:
...
+            raise ValueError(f'img_height and img_width should be identical.')
+        self._input_shape = (in_channels, img_height, img_width)
+
+        if use_pretrained:
+            self.model = ViTModel.from_pretrained(pretrained_model)
+            self.model.train()
```
tgaddair, line 348: The BertEncoder (see `text_encoders.py`) has a `trainable=True` param the user can set, so that they can disable fine tuning. I think we may as well add something here, too.

tgaddair (commented):
Nice! Just a few suggestions.

ludwig/data/preprocessing.py
ShreyaR: Done!

ludwig/data/preprocessing.py
ShreyaR: Done!

ludwig/datasets/mixins/download.py
ShreyaR: Done!

ludwig/encoders/image_encoders.py
ShreyaR: Done!

ludwig/encoders/image_encoders.py
ShreyaR: Done!

ludwig/encoders/image_encoders.py
ShreyaR: Done!

ludwig/datasets/insurance_lite/config.yaml
ShreyaR: Fixed in the latest patch!

ludwig/data/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from typing import Union
+
+try:
+    import dask.dataframe as dd
+except ImportError:
+    pass
+import pandas as pd
+
+
+if 'dask.dataframe' in sys.modules:
```
tgaddair, line 11: Another way we could do this would be:

```
except ImportError:
    dd = None

...

if dd is not None:
```

But this works, too.

ludwig/datasets/insurance_lite/__init__.py
```diff
@@ -47,26 +44,29 @@ def __init__(self, cache_dir=DEFAULT_CACHE_LOCATION):
...
-        # options for using it once.
-        df = pd.read_csv(
-            file_name, header=0,
-            names=['id', 'image_path', 'insurance_company',
-                    'cost_of_vehicle', 'min_coverage', 'expiry_date',
-                    'max_coverage', 'condition', 'amount'],
-            usecols=list(range(1, 9))
+
+        for url in self.config['kaggle_dataset_files']:
+            file_name = os.path.join(self.raw_dataset_path, url.split('/')[-1])
```
tgaddair, line 49: Could use `os.path.basename(url)` instead.

ludwig/datasets/insurance_lite/__init__.py
```diff
@@ -47,26 +44,29 @@ def __init__(self, cache_dir=DEFAULT_CACHE_LOCATION):
...
+                names=['id', 'image_path', 'insurance_company',
+                        'cost_of_vehicle', 'min_coverage', 'expiry_date',
+                        'max_coverage', 'condition', 'amount'],
+                usecols=list(range(1, 9))
+                )
+            img_urls = df['image_path']
+            self.download_images(img_urls)
+            df['image_path'] = df['image_path'].apply(
+                lambda x: os.path.join(
+                    self.raw_dataset_path, 'images', x.split('/')[-1]
```
tgaddair, line 63: Same here, cleaner to use `os.path.basename(x)`

tgaddair (approved):
LGTM! Added a few minor suggestions. Not super critical, but feel free to apply them if you agree with the changes.

ludwig/data/__init__.py
ShreyaR: Thanks for the suggestion. I prefer this as well, updated in the latest patch.

ludwig/datasets/insurance_lite/__init__.py
ShreyaR: Done

ludwig/datasets/insurance_lite/__init__.py
ShreyaR: Done.

PR #1307, opened 2021-09-20
Return None for unreadable images and try to infer num channels

hungcs:


ludwig/features/image_feature.py
```diff
@@ -68,6 +69,7 @@ class ImageFeatureMixin:
         'height': {'type': 'integer', 'minimum': 0},
         'width': {'type': 'integer', 'minimum': 0},
         'num_channels': {'type': 'integer', 'minimum': 0},
+        'infer_image_num_channels': {'type': 'boolean'},
```
tgaddair, line 72: I'm trying to think if there's a scenario where the user wants to infer the height and width but not the channels. Right now I'm leaning towards `infer_image_dimensions` just handling all 3 to simplify things. Do you have some thoughts on this?

tgaddair (commented):
Nice fix! Only question is about need to new parameter.

ludwig/utils/image_utils.py
```diff
@@ -30,16 +31,15 @@
 logger = logging.getLogger(__name__)
 
 
+@functools.lru_cache(maxsize=32)
```
tgaddair, line 34: Is the purpose this cache to avoid downloading the image twice when we infer the image dimensions (1) and then process it (2)?

ludwig/features/image_feature.py
```diff
@@ -206,36 +218,39 @@ def _finalize_preprocessing_parameters(
...
+                height_avg = min(
+                    sum(x.shape[0] for x in inferred_sample) / len(inferred_sample),
+                    preprocessing_parameters[INFER_IMAGE_MAX_HEIGHT])
+                width_avg = min(
+                    sum(x.shape[1] for x in inferred_sample) / len(inferred_sample),
+                    preprocessing_parameters[INFER_IMAGE_MAX_WIDTH])
+
+                height, width = round(height_avg), round(width_avg)
+                logger.debug("Inferring height: {0} and width: {1}".format(height, width))
+            elif first_image is not None:
```
tgaddair, line 234: It looks like we can generalize this `first_image` code path as a special case of the `infer_image_dimensions` case. Specifically, what if we removed the `infer_image_dimensions` param and just used `infer_image_sample_size`. Then if the user sets this value to `1`, we get the same effect as setting `infew_image_dimensions=False`, right?

ludwig/utils/image_utils.py
hungcs: Yeah, or to avoid downloading images with the same url in each step

ludwig/features/image_feature.py
hungcs: sounds good!

hungcs: actually I just realized there is a slight difference -`infer_image_dimensions` also sets `should_resize = True`, which will prevent Ludwig from throwing an error on the first mismatch.

tgaddair: I see. In this case, I would be in favor of making `should_resize=True` the default (because the default is now to infer image dimensions). Then if the user wants the old behavior, they would need to set `should_resize=False` manually. An additional thing we can do in this case, when `should_resize=False` and `infer_image_sample_size > 1` is raise the error if not all the elements of the sample match. Would that solve the issue?

hungcs: By user setting the older behavior, do you mean exposing `should_resize` and `user_specified_num_channels` through preprocessing parameters? Or meaning they would just edit `should_resize = False` in the code.

If we expose `should_resize` and `user_specified_num_channels` through preprocessing parameters, then those would just be replacing `infer_image_dimensions`, so I don't know if that would decrease complexity. There also could be conflicts (i.e. user specifies explicit width and height but then has `should_resize` set to False

tgaddair: Hmm, yeah I see the issue here. I think given how much we would need to overhaul the config params to make it work, it's probably fine to leave it as-is for now.

tgaddair (approved):

PR #1308, opened 2021-09-22
Torch vector features

ShreyaR:
Migrates vector features to PyTorch


ludwig/data/preprocessing.py
```diff
@@ -52,7 +52,7 @@
 from ludwig.utils.misc_utils import (get_from_registry, merge_dict,
                                      resolve_pointers, set_random_seed,
                                      get_proc_features_from_lists)
-from ludwig.utils.type_utils import Column
+# from ludwig.utils.type_utils import Column
```
ShreyaR, line 55: I was running into an issue with importing Column. Will uncomment in a subsequent patch

ludwig/features/vector_feature.py
```diff
@@ -102,26 +97,26 @@ def add_feature_data(
             raise
 
         # Determine vector size
-        vector_size = backend.df_engine.compute(
+        output_size = backend.df_engine.compute(
```
w4nderlust, line 100: I'm not 100% sold on this choice of renaming. The way I see it `vector_size` as a name works well for bot input and output features, while `output_size` doesn't really work for the input features. Maybe I'm missing something though.

ludwig/modules/metric_modules.py
```diff
@@ -166,7 +167,7 @@ def __init__(self, name='r2_score'):
         self.add_state("sum_y_y_hat", default=torch.tensor(0, dtype=torch.float32))
         self.add_state("N", default=torch.tensor(0, dtype=torch.float32))
 
-    def update(self, y_hat, y):
+    def update(self, y_hat: Dict[str, torch.Tensor], y: torch.Tensor):
```
w4nderlust, line 170: this change would likely affect all the feature using this metric, namely the numerical ones. Should they be modified to pass the dictionary too?

ludwig/features/base_feature.py
```diff
@@ -220,7 +215,7 @@ def update_metrics(self, targets, predictions):
                 metric_fn.update(predictions, targets)
             else:
                 #metric_fn.update_state(targets, predictions[PREDICTIONS])
-                metric_fn.update(predictions[PREDICTIONS], targets)
+                metric_fn.update(predictions, targets)
```
w4nderlust, line 218: This is in the base feature, so should work across the baord for all features and all metrics

ludwig/features/vector_feature.py
ShreyaR: Hmm I see what you mean. The issue on the output vector features that requires renaming is that these parameters are passed directly to the decoder object (Projector class in this case) also used by other features. I'll think about how to fix this and upload another patch.

ludwig/modules/metric_modules.py
ShreyaR: Currently the metric computation seems to be very inconsistent. A lot of metrics assume a dictionary of tensors as an argument while others assume a tensor. The update metrics function is also called from the `base_features` so consistency in argument types across metrics would be important

ShreyaR:
@w4nderlust fixed the issue with `vector_size`. Per our discussion, I've kept the changes to the metric modules. Once we have consensus on next steps for unifying metric modules I can make changes to all metrics in Ludwig to be consistent.

tests/integration_tests/test_simple_features.py
```diff
@@ -162,18 +162,18 @@ def test_feature(input_test_feature, output_test_feature,
...
+        ([category_feature()],
+         [numerical_feature(), numerical_feature()]),
+        ([category_feature()],
+         [sequence_feature(vocab_size=5), sequence_feature(vocab_size=7)]),
+        ([category_feature()],
+         [set_feature(vocab_size=5), set_feature(vocab_size=7)]),
+        ([category_feature()],
+         [text_feature(vocab_size=5), text_feature(vocab_size=7)]),
         ([category_feature()],
          [vector_feature(), vector_feature()]),
```
w4nderlust: this looks great for testing output vector features, although it doesn't seem to me we are testing for input vector features, or ar we?

ludwig/data/preprocessing.py
ShreyaR: Fixed in the latest patch

ludwig/features/vector_feature.py
ShreyaR: Resolved offline

ludwig/modules/metric_modules.py
ShreyaR: Resolved offline

ludwig/features/base_feature.py
ShreyaR: Resolved offline

PR #1310, opened 2021-09-23
Tf to torch audio feature

jimthompson5802:
# Code Pull Requests

- Port Audio Input Feature to Torch
- Add unit test for audio input feature

### DIFF
--- a/ludwig/encoders/sequence_encoders.py
+++ b/ludwig/encoders/sequence_encoders.py
@@ -501,7 +501,8 @@ class ParallelCNN(SequenceEncoder):
         if self.reduce_output is not None:
             logger.debug('  FCStack')
             self.fc_stack = FCStack(
-                num_filters * self.num_conv_layers,
+                self.parallel_conv1d.output_shape[-1],
+                # num parallel * conv_layer output size
                 layers=fc_layers,
                 num_layers=num_fc_layers,
                 default_fc_size=fc_size,
@@ -1522,7 +1523,7 @@ class StackedRNN(SequenceEncoder):
         if self.reduce_output is not None:
             logger.debug('  FCStack')
             self.fc_stack = FCStack(
-                state_size,
+                self.recurrent_stack.output_shape[-1],  # state_size,
                 layers=fc_layers,
                 num_layers=num_fc_layers,
                 default_fc_size=fc_size,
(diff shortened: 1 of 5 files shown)


ludwig/encoders/sequence_encoders.py
```diff
@@ -1609,7 +1610,7 @@ def __init__(
             pretrained_embeddings=None,
             embeddings_on_cpu=False,
             conv_layers=None,
-            num_conv_layers=1,
+            num_conv_layers=None,
```
w4nderlust, line 1613: Not sure why this change is needed

jimthompson5802: The change was to address this error that came up in the unit test
```
 File "/opt/project/ludwig/encoders/sequence_encoders.py", line 1765, in __init__
    'Invalid layer parametrization, use either conv_layers or '
ValueError: Invalid layer parametrization, use either conv_layers or num_conv_layers
```

For the test, I use `tests.integration_tests.utils.audio_feature()` to create the audio input feature config definition.  This creates the config parameter to the audio input feature
```
       'conv_layers': [
            {
                'filter_size': 400,
                'pool_size': 16,
                'num_filters': 32,
                'regularize': 'false'
            },
            {
                'filter_size': 40,
                'pool_size': 10,
                'num_filters': 64,
                'regularize': 'false'
            }
        ]
```

this then runs afoul of this validation check in `StackedCNNRNN` because both `con_layers` and `num_conv_layers` are NOT None.  One or the other or both has to be `None`.
```
        if conv_layers is not None and num_conv_layers is None:
            # use custom-defined layers
            self.conv_layers = conv_layers
            self.num_conv_layers = len(conv_layers)
        elif conv_layers is None and num_conv_layers is not None:
            # generate num_conv_layers with default parameters
            self.conv_layers = None
            self.num_conv_layers = num_conv_layers
        elif conv_layers is None and num_conv_layers is None:
            # use default layers with varying filter sizes
            self.conv_layers = [
                {'pool_size': 3},
                {'pool_size': None}
            ]
            self.num_conv_layers = 2
        else:
            raise ValueError(
                'Invalid layer parametrization, use either conv_layers or '
                'num_conv_layers'
            )
```

If you think it is better to have the default value for `num_conv_layers=1`, then I can change the `tests.integration_tests.utils.audio_feature()` to explicitly set `num_conv_layers` to `None`. 

Let me know if this answered the question and if I need to make any changes.

w4nderlust: Makes sense, thanks for the clarification. Given this piece of code:
```
elif conv_layers is None and num_conv_layers is None:
            # use default layers with varying filter sizes
            self.conv_layers = [
                {'pool_size': 3},
                {'pool_size': None}
            ]
            self.num_conv_layers = 2
```
I'd say setting to None sounds good.

PR #1312, opened 2021-09-24
Add gray image/avg image fallbacks for unreachable images

hungcs:
If an image cannot be read (often due to an expired image url), return a uniform gray image of the same dimensions.

### DIFF
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -375,17 +375,17 @@ class ImageFeatureMixin:
                 # helper function for handling single image
                 def _get_processed_image(img_store):
                     if isinstance(img_store, str):
-                        return read_image_and_resize(
+                        res_single = read_image_and_resize(
                             get_abs_path(src_path, img_store)
                         )
                     else:
-                        return read_image_and_resize(img_store)
+                        res_single = read_image_and_resize(img_store)
+                    return res_single if res_single is not None else default_image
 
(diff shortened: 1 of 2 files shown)


ludwig/utils/image_utils.py
```diff
@@ -31,6 +31,14 @@
 logger = logging.getLogger(__name__)
 
 
+def get_gray_default_image(height, width, num_channels):
+    return np.full((height, width, num_channels), 128)
```
tgaddair, line 35: I believe we should make this `, dtype=np.uint8)`

ludwig/features/image_feature.py
```diff
@@ -377,10 +381,11 @@ def _get_processed_image(img_store):
                     else:
                         return read_image_and_resize(img_store)
 
-                proc_df[feature[PROC_COLUMN]] = backend.df_engine.map_objects(
+                res = backend.df_engine.map_objects(
                     input_df[feature[COLUMN]],
                     _get_processed_image
                 )
+                proc_df[feature[PROC_COLUMN]] = res if res is not None else default_image
```
tgaddair, line 388: I think `res` needs to be a series here. Can you move this default logic into `_get_processed_image` above?

tgaddair:
Test failures due to sklearn release.

Fix here: https://github.com/ludwig-ai/ludwig/pull/1325

tgaddair (approved):

PR #1313, opened 2021-09-24
Updated set encoders and embedding modules

ShreyaR:
This PR migrates set features to PyTorch.


ludwig/modules/embedding_modules.py
```diff
@@ -137,20 +119,6 @@ def embedding_matrix_on_device(
...
-        embeddings, embedding_size = embedding_matrix(
-            vocab,
-            embedding_size,
-            representation=representation,
-            embeddings_trainable=embeddings_trainable,
-            pretrained_embeddings=pretrained_embeddings,
-            force_embedding_size=force_embedding_size,
-            embedding_initializer=embedding_initializer
-        )
-    '''
```
w4nderlust, line 151: I think we need to force them being on cpu when `embeddings_on_cpu==True` though, otherwise if the model is on gpu also the embeddigns will be on gpu I believe.

ludwig/modules/embedding_modules.py
```diff
@@ -81,33 +81,16 @@ def embedding_matrix(
...
         embedding_size = vocab_size
-        '''
-        embeddings = tf.Variable(
-            get_initializer('identity')([vocab_size, embedding_size]),
-            trainable=False,
-            name='embeddings'
-        )
-        '''
         embeddings = torch.tensor(
             get_initializer('identity')([vocab_size, embedding_size]),
```
w4nderlust, line 89: there's an error here that probably was there since porting to TF2. the tesnor should be `[vocab_size, vocab_size]`, otherwise every sparse representation for tokens with id > vocab_size will be the same (all 0s)

ludwig/encoders/set_encoders.py
```diff
@@ -91,22 +91,28 @@ def __init__(
...
         """
-        hidden = self.embed_sparse(inputs, training=training, mask=mask)
+        hidden = self.embed(inputs, training=training, mask=mask)
         hidden = self.fc_stack(hidden, training=training, mask=mask)
 
         return hidden
+
+    @property
+    def input_shape(self) -> torch.Size:
+        return torch.Size([1])
```
w4nderlust, line 114: Is 1 correct here?

ludwig/modules/embedding_modules.py
```diff
@@ -210,20 +174,26 @@ def forward(self, inputs, training=None, mask=None):
 
         return embedded
 
+    @property
+    def input_shape(self) -> torch.Size:
+        return torch.Size([1])
```
w4nderlust, line 179: this is an interesting case, I believe the same ambed is used also for seuqence/text features where the input size is not 1.

ludwig/modules/loss_modules.py
```diff
@@ -509,3 +509,15 @@ def rmspe_loss(targets, predictions):
         )
     )
     return loss
+
+
+class SigmoidCrossEntropyLoss(nn.Module):
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.loss_fn = nn.BCEWithLogitsLoss()
```
w4nderlust, line 517: does ths work on vectors of outputs like in the set case?

ludwig/modules/embedding_modules.py
ShreyaR: @w4nderlust in line 94 `embedding_size` is set to be equal to `vocab_size`, so the equivalent sizes should be the same.
I can make the change regardless tho if you'd prefer

ludwig/modules/embedding_modules.py
ShreyaR: @w4nderlust i looked at the imports of `sequence_encoders.py`, it looks like they use `EmbedSequence` instead of `Embed`.

However, it does look like `Embed` is used in category encoders, date encoders and h3 encoders. Do you know if the shape would change for any of them?

ludwig/modules/loss_modules.py
ShreyaR: I tested this out on a set output feature and it worked

ludwig/encoders/set_encoders.py
ShreyaR: I can double check

ludwig/modules/embedding_modules.py
ShreyaR: Done

ludwig/modules/embedding_modules.py
w4nderlust: My bad, missed it.

ludwig/encoders/set_encoders.py
ShreyaR: I believe so -- set encoders pass the input directly to the `Embed` module, so they should have the same input shape

ludwig/modules/embedding_modules.py
w4nderlust: Oh ok, it sounds good then. In date and H3 it is used to embed the single components, so it makes sense to have have size 1.

ludwig/modules/loss_modules.py
w4nderlust: And what was the aggregation function? Were all the independent losses for each dimension of the vector summed or averaged? I believe they should be summed, as it would be equivalent to have N binary output features.

ShreyaR: Discussed offline

tests/ludwig/modules/test_embedding_modules.py
```diff
@@ -19,6 +19,24 @@ def test_embed(
...
         representation=representation,
     )
+    inputs = torch.randint(0, 2, size=(2, 1)).bool()
+    outputs = embed(inputs)
+    assert outputs.shape[1:] == embed.output_shape
+
+
+@pytest.mark.parametrize('vocab', [['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']])
+@pytest.mark.parametrize('embedding_size', [5, 10])
+@pytest.mark.parametrize('representation', ['dense', 'sparse'])
```
w4nderlust, line 29: maybe reduce these ranges like for the set ones?

ludwig/modules/embedding_modules.py
```diff
@@ -156,8 +157,73 @@ def __init__(
...
+                    inputs[b, i] indicates that token i is present in sample b.
+        """
+        # Convert multi-hot input to input of indices
+        inputs = inputs.int() * torch.arange(self.vocab_size)
+        embedded = self.embeddings(inputs.long())
+        # Mask out the 0th embedding
+        mask = torch.unsqueeze(inputs, -1)
+        embedded = embedded * mask
+        # Sum over all positive tokens
         embedded = torch.sum(embedded, dim=1)
```
w4nderlust, line 229: maybe allow for both sum and avg as aggregation function?

tests/ludwig/modules/test_embedding_modules.py
ShreyaR: Which range? The embedding size?

w4nderlust: vocab too

ShreyaR:
@w4nderlust ready for review!

PR #1338, opened 2021-10-01
Port date_feature and date_encoders to PyTorch

justinxzhao:
- Added unit tests for DateEmbed and DateWave in test_date_encoders.
- Added unit tests in test_date_features.

### DIFF
--- a/ludwig/encoders/date_encoders.py
+++ b/ludwig/encoders/date_encoders.py
@@ -263,14 +263,14 @@ class DateEmbed(DateEncoder):
         input_vector = inputs.type(torch.IntTensor)
 
         scaled_year = self.year_fc(input_vector[:, 0:1].type(torch.FloatTensor))
-        embedded_month = self.embed_month(input_vector[:, 1].unsqueeze(1) - 1)
-        embedded_day = self.embed_day(input_vector[:, 2].unsqueeze(1) - 1)
-        embedded_weekday = self.embed_weekday(input_vector[:, 3].unsqueeze(1))
+        embedded_month = self.embed_month(input_vector[:, 1:2] - 1)
+        embedded_day = self.embed_day(input_vector[:, 2:3] - 1)
+        embedded_weekday = self.embed_weekday(input_vector[:, 3:4])
         embedded_yearday = self.embed_yearday(
-            input_vector[:, 4].unsqueeze(1) - 1)
-        embedded_hour = self.embed_hour(input_vector[:, 5].unsqueeze(1))
-        embedded_minute = self.embed_minute(input_vector[:, 6].unsqueeze(1))
-        embedded_second = self.embed_second(input_vector[:, 7].unsqueeze(1))
+            input_vector[:, 4:5] - 1)
+        embedded_hour = self.embed_hour(input_vector[:, 5:6])
+        embedded_minute = self.embed_minute(input_vector[:, 6:7])
+        embedded_second = self.embed_second(input_vector[:, 7:8])
         periodic_second_of_day = torch_utils.periodic(
             input_vector[:, 8:9], 86400)
 
--- a/ludwig/encoders/date_encoders.py
+++ b/ludwig/encoders/date_encoders.py
@@ -283,42 +283,14 @@ class DateEmbed(DateEncoder):
             training=training,
             mask=mask
         )
-        embedded_month = self.embed_month(
-            input_vector[:, 1] - 1,
-            training=training,
-            mask=mask
-        )
-        embedded_day = self.embed_day(
-            input_vector[:, 2] - 1,
-            training=training,
-            mask=mask
-        )
-        embedded_weekday = self.embed_weekday(
-            input_vector[:, 3],
-            training=training,
-            mask=mask
-        )
+        embedded_month = self.embed_month(input_vector[:, 1].unsqueeze(1) - 1)
+        embedded_day = self.embed_day(input_vector[:, 2].unsqueeze(1) - 1)
+        embedded_weekday = self.embed_weekday(input_vector[:, 3].unsqueeze(1))
         embedded_yearday = self.embed_yearday(
-            input_vector[:, 4] - 1,
-            training=training,
-            mask=mask
-        )
-        embedded_hour = self.embed_hour(
-            input_vector[:, 5],
-            training=training,
-            mask=mask
-        )
-        embedded_minute = self.embed_minute(
-            input_vector[:, 6],
-            training=training,
-            mask=mask
-        )
-        embedded_second = self.embed_second(
-            input_vector[:, 7],
-            training=training,
-            mask=mask
-        )
-
+            input_vector[:, 4].unsqueeze(1) - 1)
+        embedded_hour = self.embed_hour(input_vector[:, 5].unsqueeze(1))
+        embedded_minute = self.embed_minute(input_vector[:, 6].unsqueeze(1))
+        embedded_second = self.embed_second(input_vector[:, 7].unsqueeze(1))
         periodic_second_of_day = torch.sin(
             input_vector[:, 8:9].type(torch.float32)
             * (2 * math.pi / 86400)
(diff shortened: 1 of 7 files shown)


ludwig/modules/embedding_modules.py
```diff
@@ -168,7 +168,7 @@ def forward(self, inputs: torch.Tensor) -> torch.Tensor:
 
     @property
     def input_shape(self) -> torch.Size:
-        return torch.Size([1])
+        return torch.Size([self.vocab])
```
w4nderlust, line 171: this should be 1

ludwig/utils/torch_utils.py
```diff
@@ -106,13 +106,14 @@ def input_dtype(self):
...
+        """ Returns size of the input tensor without the batch dimension."""
         raise NotImplementedError('Abstract class.')
 
     @property
+    @abstractmethod
     def output_shape(self) -> torch.Size:
-        """ Returns the size of the output tensor without the batch dimension."""
-        return self._compute_output_shape()
+        """ Returns size of the output tensor without the batch dimension."""
+        raise NotImplementedError('Abstract class.')
```
w4nderlust, line 116: here we want to keep that _compute_output_shape as default behavior and submodules can override it if they want

ludwig/encoders/date_encoders.py
```diff
@@ -232,34 +253,33 @@ def __init__(
...
 
-    def call(
+    def forward(
             self,
-            inputs,
-            training=None,
-            mask=None
-    ):
+            inputs: torch.Tensor,
+            training: Optional[bool] = None,
```
ShreyaR, line 265: You can remove the `training` and `mask` arguments here as well while calling the forward method of each encoder (e.g. `year_fc`). Pytorch doesn't need the training `flag` for backprop, and I believe we aren't using the `mask` arg anymore.

ludwig/encoders/date_encoders.py
```diff
@@ -232,34 +253,33 @@ def __init__(
...
+            :param mask: bool specifying masked values.
             :type mask: bool
          """
         # ================ Embeddings ================
-        input_vector = tf.cast(inputs, tf.int32)
+        input_vector = inputs.type(torch.int32)
 
         scaled_year = self.year_fc(
-            tf.cast(input_vector[:, 0:1], tf.float32),
+            input_vector[:, 0:1].type(torch.float32),
```
ShreyaR, line 282: Question about this -- from the `input_shape` property and the `DATE_INPUT_SIZE` definition, my understanding was that the year would be the first 4 indices of the input feature, where each index is a number between 0-9. So to get the input for `self.year_fc`, you'd need `input_vector[:, 0:4]` followed by converting the list of ints to a single integer. Similarly for other inputs to the month, day, etc. encoders.

I'm not sure if I'm missing some context tho or if my understanding of the input representation is incorrect?

ludwig/encoders/date_encoders.py
```diff
@@ -232,34 +253,33 @@ def __init__(
...
+            :param training: bool specifying if in training mode (important for
+                dropout).
             :type training: bool
-            :param mask: bool specifying masked values
+            :param mask: bool specifying masked values.
             :type mask: bool
          """
         # ================ Embeddings ================
-        input_vector = tf.cast(inputs, tf.int32)
+        input_vector = inputs.type(torch.int32)
```
ShreyaR, line 279: Looks like you cast every input to `torch.float32` before passing it to each embedding module, so you could potentially cast all inputs to `torch.int32` here and remove the casting in line 282 e.g.

ludwig/modules/fully_connected_modules.py
```diff
@@ -19,15 +19,22 @@
...
+from ludwig.utils.torch_utils import LudwigModule, initializer_registry, \
     activations, reg_loss
 
 logger = logging.getLogger(__name__)
 
 
-
 class FCLayer(LudwigModule):
 
+    @property
```
ShreyaR, line 30: Thanks for adding these!

ShreyaR (commented):
Looks great! Thanks for adding the feature unit tests. 

Added a few comments.

ludwig/modules/embedding_modules.py
justinxzhao: Done.

ludwig/utils/torch_utils.py
justinxzhao: Done.

ludwig/encoders/date_encoders.py
justinxzhao: If the size 19 inputs are broken down like: (YYYY)(MM)(DD)(W)(YDYD)(HH)(MM)(SS), then we should certainly pass in 0:4 for year, 4:6 for month, etc.

I've pinged the ludwig channel to confirm, since the previous embedding code seems to be working off of input of size 9.

ludwig/encoders/date_encoders.py
justinxzhao: Done.

ludwig/modules/fully_connected_modules.py
justinxzhao: Acknowledged. Thanks!

ludwig/encoders/date_encoders.py
justinxzhao: Should I/we also clean up, probably in a separate PR, the `training` and `mask` parameters in Encoder base class? I'm getting a warning about it.

ludwig/encoders/date_encoders.py
justinxzhao: Changed to 9. Based on offline discussion, 19 is likely copied from the H3 encoder code. date_features' returns are size 9 for (year, month, day, weekday, yearday, hours, minutes, seconds, seconds of day). I've also updated the encoder unit tests as well.

ludwig/encoders/date_encoders.py
justinxzhao: I believe this may be resolved with the latest push. Please reopen if not.

ludwig/modules/fully_connected_modules.py
```diff
@@ -48,9 +55,12 @@ def __init__(
     ):
         super().__init__()
 
-        #self.layers = []
+        # self.layers = []
```
ShreyaR, line 58: nit: Could you remove the commented out code from this file? Also in lines 106, 121, etc.

ludwig/encoders/date_encoders.py
```diff
@@ -15,21 +15,25 @@
...
 from ludwig.modules.embedding_modules import Embed
 from ludwig.modules.fully_connected_modules import FCStack
 
 logger = logging.getLogger(__name__)
 
-
 ENCODER_REGISTRY = Registry()
 
+# Year, month, day, weekday, yearday, hour, minute, seconds, second_of_day.
+# TODO: Share this constant with date_feature.DATE_VECTOR_SIZE.
```
ShreyaR, line 34: nit: You could import `DATE_INPUT_SIZE` in `date_feature.py` and remove the code that sets `DATE_VECTOR_LENGTH`.

So the `input_shape` property of `DateInputFeature` could return `torch.Size([DATE_INPUT_SIZE])`.

ShreyaR (commented):
lgtm, thanks for adding changes!

Added minor comments, feel free to ignore unless you're planning on adding another patch for a different reason based on @w4nderlust's review.

ludwig/encoders/date_encoders.py
justinxzhao: That would work. Another idea would be to have a constants.py file and put all of the constants there. I'll think about this more, and send a PR for this if it seems good.

ludwig/modules/fully_connected_modules.py
justinxzhao: Done.

w4nderlust (approved):

PR #1345, opened 2021-10-03
Syntactical port to PyTorch of TextInputFeature and BERTEncoder

w4nderlust:


ludwig/encoders/text_encoders.py
```diff
@@ -67,7 +68,7 @@ def __init__(
             )
             sys.exit(-1)
 
-        self.transformer = TFBertModel.from_pretrained(
+        self.transformer = BertModel.from_pretrained(
```
tgaddair, line 71: Something for a follow-up, but it could be worthwhile to do something similar for text transformers as we did for ViT, where the user can opt not to use the pretrained weights:

https://github.com/ludwig-ai/ludwig/pull/1306/files#diff-4b05fc9dd09199c0bfe796c2fb4a8814a8e13127a4eeba9177dc035765b52c07R350

PR #1347, opened 2021-10-04
Add combiner schema validation

ksbrar:


ludwig/combiners/combiners.py
```diff
@@ -39,6 +39,12 @@
 logger = logging.getLogger(__name__)
 
 
+# TODO: Where should these go? Are they complete?
+temp_weights_initializer_registry = ['glorot_uniform']
```
tgaddair, line 43: Docs on combiner params here: https://ludwig-ai.github.io/ludwig-docs/user_guide/configuration/#concat-combiner

ksbrar:
Just did another once-over of the params to make sure no defaults got changed - all good from what I can tell.

PR #1350, opened 2021-10-05
Port h3 features/encoder to Pytorch

justinxzhao:


ludwig/encoders/h3_encoders.py
```diff
@@ -180,7 +163,9 @@ def __init__(
         )
 
         logger.debug('  FCStack')
+        self.fc_size = fc_size
```
ShreyaR, line 166: Is this used anywhere?

ludwig/encoders/h3_encoders.py
```diff
@@ -368,105 +307,95 @@ def __init__(
...
-            activity_regularizer=None,
-            dropout=0.0,
-            recurrent_dropout=0.0,
-            reduce_output='last',
-            **kwargs
+            embedding_size: int = 10,
+            embeddings_on_cpu: bool = False,
+            num_layers: int = 1,
+            state_size: int = 10,
+            cell_type: int = 'rnn',
```
ShreyaR, line 359: Seems like this should be a `str` type

ludwig/modules/recurrent_modules.py
```diff
@@ -35,33 +35,16 @@
...
-            weights_regularizer=None,
-            recurrent_regularizer=None,
-            bias_regularizer=None,
-            activity_regularizer=None,
-            # kernel_constraint=kernel_constraint,
-            # recurrent_constraint=recurrent_constraint,
-            # bias_constraint=bias_constraint,
-            dropout=0.0,
-            recurrent_dropout=0.0,
+            input_size: int = None,
```
ShreyaR, line 38: Could you add a unit test for `RecurrentStack` in `tests/ludwig/modules/test_recurrent_modules.py`?

ludwig/modules/embedding_modules.py
```diff
@@ -409,9 +414,8 @@ def __init__(
         )
 
         if embedding_regularizer:
-            embedding_regularizer_obj = tf.keras.regularizers.get(
-                embedding_regularizer)
-            self.add_loss(lambda: embedding_regularizer_obj(self.embeddings))
+            self.add_loss(lambda: reg_loss(
+                self.embeddings, embedding_regularizer))
```
ShreyaR, line 418: Could you also add a test for `EmbedSequence` under `tests/ludwig/modules/test_embedding_modules.py`?

ShreyaR (commented):
Looks great! Added some comments for tests, and some minor comments

ludwig/encoders/h3_encoders.py
justinxzhao: Good catch! Done.

ludwig/modules/embedding_modules.py
justinxzhao: Done.

ludwig/modules/recurrent_modules.py
justinxzhao: Done.

tests/ludwig/modules/test_recurrent_modules.py
```diff
@@ -0,0 +1,36 @@
+import logging
+import torch
+
+from ludwig.modules import recurrent_modules
+
+logger = logging.getLogger(__name__)
+
+
+def test_recurrent_stack():
```
ShreyaR, line 9: You could potentially consolidate `test_recurrent_stack` and `test_recurrent_stack_NoSequenceSize` into a single test by using `pytest.mark.parameterize`.  E.g.
```
@pytest.mark.parameterize('sequence_size', [19, None])
```

This is probably good for now, but might be useful for future tests.

ShreyaR (approved):
Looks great! Feel free to merge once @w4nderlust approves

ludwig/modules/embedding_modules.py
```diff
@@ -79,9 +79,8 @@ def embedding_matrix(
 
     elif representation == 'sparse':
         embedding_size = vocab_size
-        embeddings = torch.tensor(
-            get_initializer('identity')([vocab_size, embedding_size]),
-            requires_grad=False)
+        embeddings = get_initializer('identity')(
+            [vocab_size, embedding_size]).clone().detach().requires_grad_(False)
```
w4nderlust, line 83: curious here, why are clone and detach needed and why this different way of setting requires grad to false?

ludwig/encoders/h3_encoders.py
```diff
@@ -42,70 +48,49 @@ class H3Embed(H3Encoder):
...
-                   is greater than 0).
-            :type regularize: Boolean
         """
         super().__init__()
         logger.debug(' {}'.format(self.name))
 
         self.embedding_size = embedding_size
         self.reduce_output = reduce_output
-        self.reduce_sequence = SequenceReducer(reduce_mode=reduce_output)
+        self.sum_sequence_reducer = SequenceReducer(reduce_mode=reduce_output)
```
w4nderlust, line 93: I prefer the old name, as which reduce operation to use is user specified, it could be something different than sum.

ludwig/modules/embedding_modules.py
justinxzhao: This was a warning/recommendation from pytorch:

```
UserWarning: To copy construct from a tensor, it is recommended to use source
Tensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True),
 rather than torch.tensor(sourceTensor)
```

It looks like it's slightly more efficient, though also less readable:
https://stackoverflow.com/questions/55266154/pytorch-preferred-way-to-copy-a-tensor

ludwig/encoders/h3_encoders.py
justinxzhao: Done.

justinxzhao: I kept the other sequence reducer in this file as sum_sequence_reducer since that one seems to be configured strictly with reduce_mode=sum.

ludwig/modules/embedding_modules.py
justinxzhao: I like the second option. Yeah it's not clear why in the original code we had to copy the tensor returned by get_initializer(...) into another tensor.

PR #1352, opened 2021-10-06
Added tests and optional pretraining for text features

ShreyaR:
For BERTEncoder:
- Add unit tests
- Allow using an untrained model.

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -17,6 +17,7 @@
 import logging
 import sys
 from abc import ABC
+from typing import Callable, Dict, Optional, Union
 
 import torch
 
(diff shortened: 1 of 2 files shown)


ludwig/encoders/text_encoders.py
```diff
@@ -50,16 +51,34 @@ class BERTEncoder(TextEncoder):
...
-            trainable=True,
-            num_tokens=None,
-            max_sequence_length=None,
+            use_pretrained: bool = True,
+            pretrained_model_name_or_path: str = 'bert-base-uncased',
+            trainable: bool = True,
+            reduce_output: str ='cls_pooled',
+            num_tokens: Optional[int] = None,
+            max_sequence_length: Optional[int] = None,
+            vocab_size: Optional[int] = 30522,
```
w4nderlust, line 60: Can we consolidate `num_tokens` and `vocab_size` into one singel parameter? I beleive they have similar goals.

ludwig/encoders/text_encoders.py
```diff
@@ -68,17 +87,47 @@ def __init__(
...
+                gradient_checkpointing=gradient_checkpointing,
+                position_embedding_type=position_embedding_type,
+                use_cache=use_cache,
+                classifier_dropout=classifier_dropout
+            )
+            self.transformer = BertModel(config)
+
         self.reduce_output = reduce_output
         if not self.reduce_output == 'cls_pooled':
             self.reduce_sequence = SequenceReducer(reduce_mode=reduce_output)
```
w4nderlust, line 120: As reduce_output can be None (on 'none' or 'None'), we want to make sure that the output_shape takes that into consideration, and contains an if like:
```
if self.reduce_output is None:
  return torch.Size([self.max_sequence_length, self.transformer.embedding_size])
```

ShreyaR: Thanks for the input on slack -- updated with the fix!

PR #1362, opened 2021-10-08
[automl] add experiment heuristics to automl module

ANarayan:
This PR adds heuristics from the tabular dataset experiments to the automl module. The key changes are to the values in the default configs and the model selection logic.

This PR also adds a feature that enables users to specify config parameters that can overwrite the automl generated params.

### DIFF
--- a/ludwig/automl/automl.py
+++ b/ludwig/automl/automl.py
@@ -15,6 +15,7 @@ from typing import Dict, Union
 
 import numpy as np
 import pandas as pd
+import pdb
 
 from ludwig.api import LudwigModel
 from ludwig.automl.base_config import (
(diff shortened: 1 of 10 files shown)


ludwig/automl/automl.py
```diff
@@ -171,28 +187,91 @@ def train_with_config(
...
+            total_numerical_feats += 1
+
+    percent_numerical_feats = total_numerical_feats / len(fields)
+
+    base_config = default_configs["base_config"]
+
+    # tabular dataset heuristics
+    if len(fields) > 3:
+        if percent_numerical_feats > 0.9:
+            base_config = (
```
amholler, line 221: This line seems to be missing a merge_dict, so when concat is selected, bad json config is created.  Saw this when running a version of this change set on tf-legacy on the higgs dataset.

amholler:
Hi, @ANarayan The variant of this PR, which is https://github.com/ludwig-ai/ludwig/pull/1507,
has been merged on tf-legacy and that variant PR should be landable on master as well.

PR #1385, opened 2021-10-13
AutoTransformerEncoder, ELECTRAEncoder, FlauBERTEncoder, and T5Encoder to Torch

justinxzhao:


justinxzhao:
> LGTM
> 
> As a meta-point, there's a lot of repeated / similar code in both these files. The tests, in particular, look like they could be parameterized by encoder class as well. Something for post-migration cleanup, perhaps.

I like this idea. We could split up the tests into encoders that have cl-pooled output and ones that don't. I'll think about this in a future change.

PR #1447, opened 2021-10-30
added the setting of an environment var and made directory writable

skanjila:
# Code Pull Requests

Please provide the following:

- set an environment variable identifying location of WANDB_DIR
- made the above directory writeable
- all local tests passing

### DIFF
--- a/tests/integration_tests/test_contrib_wandb.py
+++ b/tests/integration_tests/test_contrib_wandb.py
@@ -15,6 +15,8 @@ TEST_SCRIPT = os.path.join(os.path.dirname(__file__), 'scripts',
 
 @pytest.mark.distributed
 def test_contrib_experiment(csv_filename):
+    os.environ["WANDB_DIR"] = os.path.join(os.getcwd(), 'results')
+    subprocess.call(['chmod', '-R', '+w', os.environ["WANDB_DIR"]])
     cmdline = [
         sys.executable, TEST_SCRIPT,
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_contrib_wandb.py
```diff
@@ -15,6 +15,8 @@
 
 @pytest.mark.distributed
 def test_contrib_experiment(csv_filename):
+    os.environ["WANDB_DIR"] = os.path.join(os.getcwd(), 'results')
```
tgaddair, line 18: Should use a tmpdir for this:

```
def test_contrib_experiment(csv_filename, tmpdir):
    wandb_dir = os.path.join(os.getcwd(), 'results')
    os.makedirs(wandb_dir)
    subprocess.call(['chmod', '-R', '+w', wandb_dir])
    os.environ["WANDB_DIR"] = wandb_dir
```

skanjila: Updated to use tmp

tgaddair: Hey @skanjila, the updated code has the same issue. Please see the updated reference code in my comment above (there was a typo before). The important thing to note is use of the `tmpdir` fixture.

skanjila: Whoops sorry @tgaddair I changed the code to use tmpdir, let me know if anything else is missing

PR #1449, opened 2021-10-31
fix: incompatibility error in numeric input feature with  use of dense encoder

jimthompson5802:
# Code Pull Requests

Resolve error that occurs with the `dense` encoder with `NumericalInputFeature`.
```
 def linear(input: Tensor, weight: Tensor, bias: Optional[Tensor] = None) -> Tensor:
        r"""
        Applies a linear transformation to the incoming data: :math:`y = xA^T + b`.
    
        This operator supports :ref:`TensorFloat32<tf32_on_ampere>`.
    
        Shape:
    
            - Input: :math:`(N, *, in\_features)` N is the batch size, `*` means any number of
              additional dimensions
            - Weight: :math:`(out\_features, in\_features)`
            - Bias: :math:`(out\_features)`
            - Output: :math:`(N, *, out\_features)`
        """
        if has_torch_function_variadic(input, weight, bias):
            return handle_torch_function(linear, (input, weight, bias), input, weight, bias=bias)
>       return torch._C._nn.linear(input, weight, bias)
E       RuntimeError: expected scalar type Float but found Double
```

### DIFF
--- a/ludwig/features/numerical_feature.py
+++ b/ludwig/features/numerical_feature.py
@@ -207,11 +207,14 @@ class NumericalInputFeature(NumericalFeatureMixin, InputFeature):
     def forward(self, inputs):
         assert isinstance(inputs, torch.Tensor)
         assert inputs.dtype == torch.float32 or inputs.dtype == torch.float64
-        assert len(inputs.shape) == 1 or (len(inputs.shape) == 2 and inputs.shape[1] == 1)
+        assert len(inputs.shape) == 1 or (
+                    len(inputs.shape) == 2 and inputs.shape[1] == 1)
 
         if len(inputs.shape) == 1:
             inputs = inputs[:, None]
-        inputs_encoded = self.encoder_obj(inputs)
+
+        # ensure inputs are torch.float32 to avoid problems with use of nn.Linear
+        inputs_encoded = self.encoder_obj(inputs.type(torch.float32))
 
         return inputs_encoded
 


ludwig/features/numerical_feature.py
```diff
@@ -207,11 +207,14 @@ def __init__(self, feature, encoder_obj=None):
...
-        assert len(inputs.shape) == 1 or (len(inputs.shape) == 2 and inputs.shape[1] == 1)
+        assert len(inputs.shape) == 1 or (
+                    len(inputs.shape) == 2 and inputs.shape[1] == 1)
 
         if len(inputs.shape) == 1:
             inputs = inputs[:, None]
-        inputs_encoded = self.encoder_obj(inputs)
+
+        # ensure inputs are torch.float32 to avoid problems with use of nn.Linear
+        inputs_encoded = self.encoder_obj(inputs.type(torch.float32))
```
w4nderlust, line 217: wonder if it's better to silently cast or to remove float64 from the assert, which would for the user to do the cst outside. @tgaddair what do you think?

ludwig/features/numerical_feature.py
```diff
@@ -207,11 +207,14 @@ def __init__(self, feature, encoder_obj=None):
     def forward(self, inputs):
         assert isinstance(inputs, torch.Tensor)
         assert inputs.dtype == torch.float32 or inputs.dtype == torch.float64
-        assert len(inputs.shape) == 1 or (len(inputs.shape) == 2 and inputs.shape[1] == 1)
+        assert len(inputs.shape) == 1 or (
+                    len(inputs.shape) == 2 and inputs.shape[1] == 1)
```
w4nderlust, line 211: Here wondering if it would be better to be more prescriptive and support only one or the other clauses of the or. Allthough there's a worls where in the future we do more feature augmentation and `inputs.shape[1]` could be > 1, so maybe keeping the flexibility is ok. @tgaddair opinions?

ShreyaR: @w4nderlust I believe removing the assert for `len(inputs.shape) == 2 and inputs.shape[1] == 1` (i.e. supporting for a tensor of shape `batch` without support shape `batch x 1`) breaks some tests in `test_model_save_and_load.py`. Can't remember which ones right now, but I'd added the second assert since I figured the two shapes were functionally the same, and supporting `batch x 1` seemed more consistent with the other feature types.

PR #1450, opened 2021-11-01
Fix test_regularizers.py

ShreyaR:


ludwig/models/ecd.py
```diff
@@ -98,13 +99,27 @@ def save_savedmodel(self, save_path):
...
-        # parameter inputs is a dict feature_name -> tensor / ndarray
-        # or
-        # parameter (inputs, targets) where
-        #   inputs is a dict feature_name -> tensor/ndarray
-        #   targets is dict feature_name -> tensor/ndarray
-
+    def forward(
+            self,
+            inputs: Union[
+                Dict[str, torch.Tensor],
```
justinxzhao, line 105: `inputs` can also be a Dict[str, numpy.ndarray]

ludwig/models/ecd.py
```diff
@@ -98,13 +99,27 @@ def save_savedmodel(self, save_path):
...
+                Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]
+            ],
+            mask=None
+    ) -> Dict[str, torch.Tensor]:
+        """
+        Forward pass of the model.
```
justinxzhao, line 111: Thanks for adding this documentation!

tests/integration_tests/test_model_training_options.py
```diff
@@ -301,74 +301,75 @@ def test_optimizers(optimizer_type, generated_data_for_optimizer, tmp_path):
...
-        # retrieve training losses for all epochs
-        train_losses = np.array(train_stats['training']['combined']['loss'])
-        regularization_losses.append(train_losses[0])
-
-    # create a set of losses
-    regularization_losses_set = set(regularization_losses)
-
-    # ensure all losses obtained with the different methods are different
-    assert len(regularization_losses) == len(regularization_losses_set)
+# TODO(shreya): Figure out what to do about this.
```
justinxzhao, line 304: Instead of comment out the test, could we annotate it as skipped with pytest?

`@pytest.mark.skip(reason='  ... ')`

ludwig/models/ecd.py
ShreyaR: Thanks, updated!

ludwig/models/ecd.py
ShreyaR: Thanks!

tests/integration_tests/test_model_training_options.py
ShreyaR: Yeah! I actually ended up updating the test so that it works with the latest method of regularization.

justinxzhao (approved):

ludwig/decoders/sequence_decoders.py
```diff
@@ -111,29 +108,21 @@ def __init__(
         #     state_size,
```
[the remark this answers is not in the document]
ShreyaR: I think it's useful to keep this code around since sequence decoders still need to be ported, and having the commented tf code can act as a guide for that.

ludwig/models/ecd.py
```diff
@@ -208,14 +226,16 @@ def train_loss(self, targets, predictions, regularization_lambda=0.0):
...
-                # regularization losses and should be added
-                # to the total loss as they are
-                train_loss += loss
-            else:
-                # this assumes all unnamed losses are regularization losses
-                train_loss += regularization_lambda * loss
+            train_loss += loss
+
+        # Add regularization loss
+        if regularization_type is not None:
```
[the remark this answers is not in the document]
ShreyaR: In case of `None`, we don't add any regularization.

skanjila (approved):

PR #1452, opened 2021-11-01
Skip tests related to sequence output generation, sampled softmax, neuropod, and savedmodel

justinxzhao:


ludwig/features/binary_feature.py
```diff
@@ -116,15 +116,13 @@ class BinaryInputFeature(BinaryFeatureMixin, InputFeature):
     def __init__(self, feature, encoder_obj=None):
         super().__init__(feature)
         self.overwrite_defaults(feature)
-        if encoder_obj:
-            self.encoder_obj = encoder_obj
-        else:
-            self.encoder_obj = self.initialize_encoder(feature)
+        self.encoder_obj = self.initialize_encoder(feature)
```
ShreyaR, line 119: This would remove support for initialized encoders that are passed through.

ShreyaR (commented):
This is awesome! Added one comment.

ludwig/features/binary_feature.py
justinxzhao: This shouldn't be deleted -- added it back.

What's a use case for passing in an initialized encoder instead of initializing one from the feature definition?

ShreyaR (approved):
Looks great!

PR #1465, opened 2021-11-03
Drop legacy distributed data loaders in favor of RayDatasets

tgaddair:


ShreyaR (approved):
This is awesome! No comments/requests for changes.

PR #1467, opened 2021-11-04
Migrate savedmodel to tracing-based torchscript.

justinxzhao:
## Background

There are two ways to produce torchscript: scripting and tracing.

Useful references:
- https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html
- https://www.youtube.com/watch?v=2awmrMRf0dA
- https://www.youtube.com/watch?v=St3gdHJzic0

Scripting is the simplest but also strictest method. Tracing is more flexible, but also cannot express input-dependent control flow.

## Non-dict inputs (not strict)

Tracing seems to prefer fixed, non-dict inputs:

```
RuntimeError: Encountering a dict at the output of the tracer might cause the trace to be incorrect, this is only valid if the container structure does not change based on the module's inputs. Consider using a constant container instead (e.g. for `list`, use a `tuple` instead. for `dict`, use a `NamedTuple` instead). If you absolutely need this and know the side effects, pass strict=False to trace() to allow this behavior.
```

However, this can be bypassed with `strict=False`. I'm reasonably confident that our "container structure does not change based on the module's inputs". This will be worth checking during benchmarking.

## Refactoring to using a flat dictionary of output

Even with `strict=False`, only certain output tensor structures are permitted: 

```
traced region cannot be understood by the tracer, only outputs matchingdict[Union[str, Tensor], Union[Tensor, Tuple[Tensor, ...]]] can be a dictionary output of a traced function.
```

The structure returned by `ecd::forward()` is a dictionary of dictionary of tensors: `map(feature_name -> map(tensor_name -> tensor))`. This is not permitted by torchscript.

We work around this by refactoring the code to return a flat `map(str -> tensor)`, using concatenated keys i.e. `"{feature_name}::{tensor_name}" -> tensor`. A new utils library, `output_feature_utils.py` manages the concatenation / splitting.

## Defining `input_shape`s everywhere

Tracing seems to require `input_shape` and `output_shape` implementations for every class that inherits from `LudwigModule` that's instantiated, even if those properties are never directly referenced in the modeling codepath. My best guess is that torchscript may have some special treatment for class properties. A few of things to try to improve this in subsequent PRs:
- Define a new `LudwigTransform`, which doesn't have shape properties, and use them for ludwig modules that take arbitrary input (i.e. reduction_modules)
- Extend constructors of ludwig modules that take arbitrary input (i.e. reduction_modules) to be initialized with the explicit size of what those modules will be handling.
- Remove `@property` annotations, and replace references to output_shape and input_shape properties with explicit function calls.

## Tracing warnings

This [blog post](https://paulbridger.com/posts/mastering-torchscript/#use-tracing-if-you-must) calls out that torchscript tracing warnings are important to pay attention to, since they might highlight input-dependent behavior that will not be traced.

Here's a paste of our current tracer warnings.

```
tests/integration_tests/test_torchscript.py::test_torchscript[True]
  /Users/justin/ludwig/venv/lib/python3.8/site-packages/torch/nn/modules/conv.py:294: UserWarning: Using padding='same' with even kernel lengths and odd dilation may require a zero-padded copy of the input be created (Triggered internally at  ../aten/src/ATen/native/Convolution.cpp:660.)
    return F.conv1d(input, weight, bias, self.stride,

tests/integration_tests/test_torchscript.py::test_torchscript[True]
tests/integration_tests/test_torchscript.py::test_torchscript[False]
  /Users/justin/ludwig/ludwig/features/category_feature.py:125: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    assert len(inputs.shape) == 1 or (

tests/integration_tests/test_torchscript.py::test_torchscript[True]
tests/integration_tests/test_torchscript.py::test_torchscript[False]
  /Users/justin/ludwig/ludwig/modules/embedding_modules.py:157: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if inputs.ndim != 2 or inputs.shape[1] != 1:

tests/integration_tests/test_torchscript.py::test_torchscript[True]
tests/integration_tests/test_torchscript.py::test_torchscript[False]
  /Users/justin/ludwig/ludwig/modules/convolutional_modules.py:417: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if hidden.shape[1] == 0:

tests/integration_tests/test_torchscript.py::test_torchscript[True]
tests/integration_tests/test_torchscript.py::test_torchscript[False]
  /Users/justin/ludwig/venv/lib/python3.8/site-packages/torch/nn/functional.py:2246: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if size_prods == 1:

tests/integration_tests/test_torchscript.py::test_torchscript[True]
tests/integration_tests/test_torchscript.py::test_torchscript[False]
  /Users/justin/ludwig/ludwig/modules/convolutional_modules.py:283: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
    if hidden.shape[1] == 0:
```

It looks like most of these are related to control-flow and not necessarily traces that would break from different inputs. During benchmarking, if we may find that certain inputs fail, this could mean that we have a bad trace from `get_model_inputs()`. At that point, we would need to either refactor the code to make the tracing more input-independent, or change `get_model_inputs()` (which we use to do tracing) to better represent prediction data.

### DIFF
--- a/ludwig/features/date_feature.py
+++ b/ludwig/features/date_feature.py
@@ -162,6 +162,9 @@ class DateInputFeature(DateFeatureMixin, InputFeature):
     ):
         pass
 
+    def create_input(self):
+        return torch.Tensor([[2013, 2, 26, 1, 57, 0, 0, 0, 0], [2015, 2, 26, 1, 57, 0, 0, 0, 0]]).type(torch.int32)
+
     @staticmethod
     def populate_defaults(input_feature):
         set_default_value(input_feature, TIED, None)
--- a/ludwig/features/date_feature.py
+++ b/ludwig/features/date_feature.py
@@ -42,7 +42,7 @@ class DateFeatureMixin:
 
     preprocessing_schema = {
         'missing_value_strategy': {'type': 'string', 'enum':
-            MISSING_VALUE_STRATEGY_OPTIONS},
+                                   MISSING_VALUE_STRATEGY_OPTIONS},
         'fill_value': {'type': 'string'},
         'computed_fill_value': {'type': 'string'},
         'datetime_format': {'type': ['string', 'null']},
@@ -74,7 +74,7 @@ class DateFeatureMixin:
                 'The preprocessing fill in value will be used.'
                 'For more details: '
                 'https://ludwig.ai/user_guide/#date-features-preprocessing'
-                    .format(date_str)
+                .format(date_str)
             )
             fill_value = preprocessing_parameters['fill_value']
             if fill_value != '':
@@ -83,8 +83,8 @@ class DateFeatureMixin:
                 datetime_obj = datetime.now()
 
         yearday = (
-                datetime_obj.toordinal() -
-                date(datetime_obj.year, 1, 1).toordinal() + 1
+            datetime_obj.toordinal() -
+            date(datetime_obj.year, 1, 1).toordinal() + 1
         )
 
         midnight = datetime_obj.replace(
@@ -137,10 +137,8 @@ class DateInputFeature(DateFeatureMixin, InputFeature):
 
     def forward(self, inputs):
         assert isinstance(inputs, torch.Tensor)
-        assert inputs.dtype in [torch.int16, torch.int64]
-
+        assert inputs.dtype in [torch.int16, torch.int32, torch.int64]
         inputs_encoded = self.encoder_obj(inputs)
-
         return inputs_encoded
 
     @property
--- a/ludwig/features/date_feature.py
+++ b/ludwig/features/date_feature.py
@@ -163,7 +163,9 @@ class DateInputFeature(DateFeatureMixin, InputFeature):
         pass
 
     def create_input(self):
-        return torch.Tensor([[2013, 2, 26, 1, 57, 0, 0, 0, 0], [2015, 2, 26, 1, 57, 0, 0, 0, 0]]).type(torch.int32)
+        return torch.Tensor(
+            [[2013, 2, 26, 1, 57, 0, 0, 0, 0],
+             [2015, 2, 26, 1, 57, 0, 0, 0, 0]]).type(torch.int32)
 
     @staticmethod
     def populate_defaults(input_feature):
--- a/ludwig/features/date_feature.py
+++ b/ludwig/features/date_feature.py
@@ -162,7 +162,7 @@ class DateInputFeature(DateFeatureMixin, InputFeature):
     ):
         pass
 
-    def create_input(self):
+    def create_sample_input(self):
         return torch.Tensor(
             [[2013, 2, 26, 1, 57, 0, 0, 0, 0],
              [2015, 2, 26, 1, 57, 0, 0, 0, 0]]).type(torch.int32)
--- a/ludwig/features/numerical_feature.py
+++ b/ludwig/features/numerical_feature.py
@@ -205,8 +205,6 @@ class NumericalInputFeature(NumericalFeatureMixin, InputFeature):
             self.encoder_obj = encoder_obj
         else:
             self.encoder_obj = self.initialize_encoder(feature)
-        print(f'feature: {feature}')
-        print(f'self.encoder_obj: {self.encoder_obj}')
 
     def forward(self, inputs):
         assert isinstance(inputs, torch.Tensor)
@@ -249,6 +247,9 @@ class NumericalInputFeature(NumericalFeatureMixin, InputFeature):
         "None": PassthroughEncoder,
         None: PassthroughEncoder,
     }
+
+    def create_input(self):
+        return torch.Tensor([20, 23])
 
 
 class NumericalOutputFeature(NumericalFeatureMixin, OutputFeature):
--- a/ludwig/features/numerical_feature.py
+++ b/ludwig/features/numerical_feature.py
@@ -332,7 +332,7 @@ class NumericalOutputFeature(NumericalFeatureMixin, OutputFeature):
     def get_prediction_set(self):
         return {PREDICTIONS, LOGITS}
 
-    @ property
+    @property
     def input_shape(self) -> torch.Size:
         return torch.Size([self.input_size])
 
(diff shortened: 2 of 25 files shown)


jimthompson5802:
@justinxzhao To confirm my understanding of this....
>Right now, each decoder’s outputs are represented as a map(string -> tensor) (permitted structure). However, since 
> 
>  there can be multiple output features, the representation we are currently using over all decoders is to key each decoder’s output by the output feature name map(string -> map(string -> tensor)), which is not permitted.
> 
> Tracing works (for at least a subset of features) after refactoring output features to return NamedTuple instead of dictionaries.
> 
> It's worth noting that using NamedTuples imposes that all output features abide by the same attributes. This has implications for mixed-modality output features (i.e. binary+sequence, category+image).

Here is an example of our current way:
```
{
    "category_ouptut_feature": {
        "logits: torch.Tensor,
       "predictions":  torch.Tensor,
       "probabilities":  torch.Tensor  
    },
    "numeric_output_feature": {
        "logits": torch.Tensor,
        "predictions":  Torch.Tensor
    }
}
```

Which is not allowed by torchscript tracing.  Is this a correct understanding of the issue?

justinxzhao:
> @justinxzhao To confirm my understanding of this....
> 
> > Right now, each decoder’s outputs are represented as a map(string -> tensor) (permitted structure). However, since
> > there can be multiple output features, the representation we are currently using over all decoders is to key each decoder’s output by the output feature name map(string -> map(string -> tensor)), which is not permitted.
> > Tracing works (for at least a subset of features) after refactoring output features to return NamedTuple instead of dictionaries.
> > It's worth noting that using NamedTuples imposes that all output features abide by the same attributes. This has implications for mixed-modality output features (i.e. binary+sequence, category+image).
> 
> Here is an example of our current way:
> 
> ```
> {
>     "category_ouptut_feature": {
>         "logits: torch.Tensor,
>        "predictions":  torch.Tensor,
>        "probabilities":  torch.Tensor  
>     },
>     "numeric_output_feature": {
>         "logits": torch.Tensor,
>         "predictions":  Torch.Tensor
>     }
> }
> ```
> 
> Which is not allowed by torchscript tracing. Is this a correct understanding of the issue?

You're correct that this is an example of a structure that's prohibited by torchscript, but your example actually looks more like the map we return from `predictions()`. `predictions()` is actually fine since that's called outside of `ECD.forward()` (and thus excluded from torchscript tracing).

The issue is that `ECD.forward()` is not allowed return a map of a map. `ECD.forward()` returns something like the following, which is not allowed:

```
{
     "category_ouptut_feature": {
         "logits: torch.Tensor,
        "last_hidden":  torch.Tensor,
        "projection_input":  torch.Tensor  
     },
     "numeric_output_feature": {
         "logits": torch.Tensor,
         "last_hidden":  Torch.Tensor
     }
 }
```

In this PR, I standardize all outputs against a NamedTuple (logits, last_hidden) [here](https://github.com/ludwig-ai/ludwig/pull/1467/files#diff-6ee778ac6bfdda4199f94d4561eb176de95165863e689af86837e61e328656b3R290). This change also drops the additional "projection_input" output that's also returned by category_feature.

ludwig/features/date_feature.py
```diff
@@ -164,6 +162,11 @@ def update_config_with_metadata(
     ):
         pass
 
+    def create_input(self):
```
tgaddair, line 165: Maybe we can rename this `create_sample_input` to clarify that this is used for Torchscript to trace the graph.

ludwig/features/numerical_feature.py
```diff
@@ -17,6 +17,8 @@
 import logging
 
 import numpy as np
+import random
```
tgaddair, line 20: Nit: native Python imports go in the section of imports at the top:

```
import logging
import random
from typing import Dict

import numpy as np
import torch
```

ludwig/features/numerical_feature.py
```diff
@@ -320,7 +327,8 @@ def _setup_metrics(self):
         self.metric_functions[MEAN_SQUARED_ERROR] = MSEMetric()
         self.metric_functions[MEAN_ABSOLUTE_ERROR] = MAEMetric()
         self.metric_functions[ROOT_MEAN_SQUARED_ERROR] = RMSEMetric()
-        self.metric_functions[ROOT_MEAN_SQUARED_PERCENTAGE_ERROR] = RMSPEMetric()
+        self.metric_functions[ROOT_MEAN_SQUARED_PERCENTAGE_ERROR] = RMSPEMetric(
```
tgaddair, line 330: Ew, are you using an 80 char limit? We should bump this up to at least 100, I would say.

ludwig/modules/reduction_modules.py
```diff
@@ -40,8 +40,12 @@ def __init__(self, reduce_mode=None, **kwargs):
     def forward(self, inputs, mask=None):
         return self._reduce_obj(inputs, mask=mask)
 
+    # @property
```
tgaddair, line 43: Can remove commented code.

ludwig/utils/output_feature_utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def get_feature_name_from_concat_name(concat_name: str) -> str:
+    return '::'.join(concat_name.split('::')[:-1])
+
+
+def get_single_output_feature_tensors(
+        output_feature_dict: Dict[str, torch.Tensor],
+        feature_name: str) -> Dict[str, torch.Tensor]:
+    '''Returns a map of tensors related to the given feature_name.'''
```
tgaddair, line 22: Function docstrings use double-quotes:

```
"""Returns ..."""
```

tgaddair (commented):
Looks very good, nice work! Just a few minor comments.

tgaddair (approved):
LGTM! Failing test is flaky from a different issue, so good to land.

PR #1473, opened 2021-11-09
Classify binary data type only if the values are bool-like.

justinxzhao:


ludwig/automl/utils.py
```diff
@@ -26,7 +27,8 @@ class FieldInfo:
     name: str
     dtype: str
     key: str = None
-    distinct_values: int = 0
+    distinct_values: List = field(default_factory=lambda: [])
```
tgaddair, line 30: We can probably make this `None`.

tgaddair (commented):
Functionally, this looks good. The one thing we need to think some more about is how this works for large datasets in Dask and external sources like databases.

Two things spring to mind here:

1. We shouldn't call `distinct_values` and `num_distinct_values` as entirely separate calls.
2. In the worst case, the number of distinct values will be equal to the number of rows, which for large datasets can be prohibitive.

These goals are somewhat at odds with one another, so here's the best compromise I can think of at the moment:

1. We compute the number of distinct values first.
2. We have a second function on `DataSource` called `is_bool_type` similar to `is_string_type` that allows us to flexibly decide based on the data source whether the column is bool compatible or not. This means for big datasets we can optimize this call to avoid redoing the distinct values check (for example, if num_distinct_values > 2, we know is bool is false, and can skip the computation entirely).

Let me know if this makes sense or if you'd like to discuss this further.

amholler:
Hi, @justinxzhao Thanks for this nice functionality!  Your functionality has been landed on tf-legacy as part of another PR request and @tgaddair will also land on master, so you close this PR at your convenience.

PR #1481, opened 2021-11-12
Support BytesIO in read_image()

justinxzhao:
On the twitterbots dataset, training is failing when image features are enabled with this error AttributeError: '_io.BytesIO' object has no attribute 'shape'`.

This is caused by an issue with how `read_image()` doesn’t handle `BytesIO` objects returned by `get_image_from_http_bytes` when image features are URLs.

This PR fixes this, with minor cleanup to `read_image()`.


ludwig/utils/image_utils.py
```diff
@@ -107,14 +108,35 @@ def is_image_score(src_path, img_entry):
...
+            image_tensor = decode_image(
+                torch.frombuffer(buffer_view, dtype=torch.uint8))
+            del(buffer_view)
+            return image_tensor
+    if isinstance(img, BytesIO):
+        buffer_view = img.getbuffer()
+        try:
+            image_tensor = decode_image(
+                torch.frombuffer(buffer_view, dtype=torch.uint8))
+            del(buffer_view)
```
skanjila, line 133: is this basically memory cleanup?

ludwig/utils/image_utils.py
```diff
@@ -107,14 +108,35 @@ def is_image_score(src_path, img_entry):
...
+    Returns None if the image could not be read.
     """
+    if isinstance(img, torch.Tensor):
+        return img
     if isinstance(img, str):
         return read_image_from_str(img, num_channels)
-    return img
+    if isinstance(img, bytes):
+        with BytesIO(img) as buffer:
+            buffer_view = buffer.getbuffer()
```
skanjila, line 123: what happens if the img is too big to fit in the buffer?

ludwig/utils/image_utils.py
```diff
@@ -107,14 +108,35 @@ def is_image_score(src_path, img_entry):
...
-    If num_channels is not provided, he image is read in unchanged format.
+    If num_channels is not provided, the image is read in unchanged format.
+    Returns None if the image could not be read.
     """
+    if isinstance(img, torch.Tensor):
+        return img
     if isinstance(img, str):
         return read_image_from_str(img, num_channels)
-    return img
+    if isinstance(img, bytes):
```
ShreyaR, line 121: I believe @jimthompson5802's PR also adds some of these changes

ShreyaR (approved):
lgtm @justinxzhao! Jim's PR also adds a few of these changes in addition to tests for some image_utils functions, and it hasn't been merged yet. Feel free to merge if you can confirm that the tests pass!

ludwig/utils/image_utils.py
justinxzhao: Great observation!

I actually added my changes on top of Jim's MNIST fix. I'll wait for his PR to be merged first, and then rebase (and re-test) before merging.

ludwig/utils/image_utils.py
justinxzhao: I think so. I'm not sure if it's completely necessary to manually delete the buffer here. @jimthompson5802 what was your reasoning for calling `del` manually?

ludwig/utils/image_utils.py
justinxzhao: I imagine that the entire program would terminate with OOM. Hopefully no one's reading in images that are THAT big.

ludwig/utils/image_utils.py
skanjila: yes I was under the impression that garbage collector would kick in and do this for us, is that not the case?

PR #1487, opened 2021-11-14
Added whylogs integration and connected to metadata preprocessing

anmshkmr:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/callbacks.py
+++ b/ludwig/callbacks.py
@@ -116,16 +116,16 @@ class Callback(ABC):
     def on_test_end(self, trainer, progress_tracker, save_path):
         pass
 
-    def on_build_metadata_start(self, df):
+    def on_build_metadata_start(self, df, dataset_name):
         pass
 
-    def on_build_metadata_end(self, df):
+    def on_build_metadata_end(self, df, dataset_name):
         pass
 
-    def on_build_data_start(self, df):
+    def on_build_data_start(self, df, dataset_name):
         pass
 
-    def on_build_data_end(self, df):
+    def on_build_data_end(self, df, dataset_name):
         pass
 
     def on_visualize_figure(self, fig):
--- a/ludwig/callbacks.py
+++ b/ludwig/callbacks.py
@@ -116,6 +116,18 @@ class Callback(ABC):
     def on_test_end(self, trainer, progress_tracker, save_path):
         pass
 
+    def on_build_metadata_start(self, df):
+        pass
+
+    def on_build_metadata_end(self, df):
+        pass
+
+    def on_build_data_start(self, df):
+        pass
+
+    def on_build_data_end(self, df):
+        pass
+
     def on_visualize_figure(self, fig):
         pass
 
--- a/ludwig/callbacks.py
+++ b/ludwig/callbacks.py
@@ -116,16 +116,16 @@ class Callback(ABC):
     def on_test_end(self, trainer, progress_tracker, save_path):
         pass
 
-    def on_build_metadata_start(self, df, dataset_name):
+    def on_build_metadata_start(self, df, mode):
         pass
 
-    def on_build_metadata_end(self, df, dataset_name):
+    def on_build_metadata_end(self, df, mode):
         pass
 
-    def on_build_data_start(self, df, dataset_name):
+    def on_build_data_start(self, df, mode):
         pass
 
-    def on_build_data_end(self, df, dataset_name):
+    def on_build_data_end(self, df, mode):
         pass
 
     def on_visualize_figure(self, fig):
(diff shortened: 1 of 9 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1383,7 +1454,8 @@ def preprocess_for_training(
         skip_save_processed_input=False,
         preprocessing_params=default_preprocessing_parameters,
         backend=LOCAL_BACKEND,
-        random_seed=default_random_seed
+        random_seed=default_random_seed,
+        callbacks=[]
```
tgaddair, line 1458: In general, we should avoid using `[]` as a default argument in Python, as it becomes a mutable constant, which can have unexpected side effects when you call the function multiple times. Better alternative is to use `None`.

ludwig/data/preprocessing.py
```diff
@@ -1089,6 +1157,9 @@ def build_dataset(
         backend
     )
 
+    for callback in callbacks or []:
+        callback.on_build_metadata_end(dataset_df)
+
```
tgaddair, line 1162: Can add callbacks for `on_build_data` below.

ludwig/contribs/whylogs/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+class WhyLogsCallback(Callback):
+    def __init__(self, path_to_config=None):
+        self.session = get_or_create_session(path_to_config)
+        self.logger = self.session.logger()
+
+    def on_build_metadata_start(self, df):
+        self.logger.log({'metadata_start': df})
+
+    def on_build_metadata_end(self, df):
+        self.logger.log({'metadata_end': df})
```
tgaddair, line 14: We probably only need to log once during `on_build_metadata_start`. The other callbacks we can just leave for others who may be interested in overriding them.

tgaddair (commented):
Nice! As discussed, tests and distinguishing between train and predict modes are good next steps, then we should be good to land.

ludwig/contribs/whylogs/__init__.py
anmshkmr: Makes sense. This in combo with change from your previous comment makes sense.

ludwig/data/preprocessing.py
anmshkmr: Done.

anmshkmr:
Took out the WIP tag. Tested locally and verified we're seeing whylog proto bins in the directory where we run tests.

requirements.txt
```diff
@@ -22,6 +22,7 @@ tensorboard
 torchmetrics>=0.6.0
 torchinfo
 filelock
+whylogs
```
tgaddair, line 25: We should remove this from the main requirements, as it is optional.

ludwig/data/preprocessing.py
```diff
@@ -981,7 +998,8 @@ def preprocess_for_training(
             skip_save_processed_input=False,
             preprocessing_params=default_preprocessing_parameters,
             backend=LOCAL_BACKEND,
-            random_seed=default_random_seed
+            random_seed=default_random_seed,
+            callbacks=[]
```
tgaddair, line 1002: Let's replace `[]` with `None` to avoid mutable default issue.

ludwig/data/preprocessing.py
```diff
@@ -1113,7 +1132,8 @@ def build_dataset(
         backend=LOCAL_BACKEND,
         random_seed=default_random_seed,
         skip_save_processed_input=False,
-        callbacks=[]
+        callbacks=[],
+        context=None
```
tgaddair, line 1136: `context` can get overloaded, so maybe something like `mode` will be more natural.

ludwig/callbacks.py
```diff
@@ -116,16 +116,16 @@ def on_test_start(self, trainer, progress_tracker, save_path):
     def on_test_end(self, trainer, progress_tracker, save_path):
         pass
 
-    def on_build_metadata_start(self, df):
+    def on_build_metadata_start(self, df, dataset_name):
```
tgaddair, line 119: `dataset_name` should be consistent with the other param if possible, so maybe we can change this to `mode`.

tgaddair (commented):
Looks great! Last thing we need to nail down is how this works with Dask. We can sync up for a quick meeting to discuss.

ludwig/data/preprocessing.py
anmshkmr: Done.

ludwig/data/preprocessing.py
anmshkmr: Done.

ludwig/callbacks.py
anmshkmr: Done.

tgaddair (approved):
LGTM!

PR #1500, opened 2021-11-19
Add hdf5 support for ground_truth in visualization

hungcs:
Unlike csv ground_truth, `train.hdf5` already has category columns mapped to indices, so there is no need to apply `ground_truth_metadata['str2idx']`. Pass in `ground_truth_apply_idx` to check whether or not we should skip this step.

Additionally, when we pass in `train.hdf5` as `ground_truth`, the number of rows may not match the `_get_cols_from_predictions()` call, so we limit probabilities rows to `len(ground_truth)`

### DIFF
--- a/ludwig/globals.py
+++ b/ludwig/globals.py
@@ -22,6 +22,13 @@ TRAIN_SET_METADATA_FILE_NAME = "training_set_metadata.json"
 TRAINING_PROGRESS_TRACKER_FILE_NAME = "training_progress.json"
 TRAINING_CHECKPOINTS_DIR_PATH = "training_checkpoints"
 
+TEST_STATISTICS_FILE_NAME = "test_statistics.json"
+
+PREDICTIONS_PARQUET_FILE_NAME = "predictions.parquet"
+PREDICTIONS_SHAPES_FILE_NAME = "predictions.shapes.json"
+
+TRAINING_PREPROC_FILE_NAME = "training.hdf5"
+
 DISABLE_PROGRESSBAR = False
 
 
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -201,7 +201,7 @@ def read_stata(data_fp, df_lib):
     return df_lib.read_stata(data_fp)
 
 
-def read_hdf5(data_fp):
+def read_hdf5(data_fp, **kwargs):
     return load_hdf5(data_fp, clean_cols=True)
 
 
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -201,6 +201,10 @@ def read_stata(data_fp, df_lib):
     return df_lib.read_stata(data_fp)
 
 
+def read_hdf5(data_fp):
+    return load_hdf5(data_fp, clean_cols=True)
+
+
 def save_csv(data_fp, data):
     with open_file(data_fp, "w", encoding="utf-8") as csv_file:
         writer = csv.writer(csv_file)
@@ -327,13 +331,15 @@ def save_hdf5(data_fp, data):
             h5_file.create_dataset(column, data=numpy_dataset[column])
 
 
-def load_hdf5(data_fp):
+def load_hdf5(data_fp, clean_cols: bool = False):
     with download_h5(data_fp) as hdf5_data:
         columns = [s.decode("utf-8") for s in hdf5_data[HDF5_COLUMNS_KEY][()].tolist()]
 
         numpy_dataset = {}
         for column in columns:
-            numpy_dataset[column] = hdf5_data[column][()]
+            # Column names from training hdf5 will be in the form 'Survived_a2fv4'
+            np_col = column.rsplit("_", 1)[0] if clean_cols else column
+            numpy_dataset[np_col] = hdf5_data[column][()]
 
     return from_numpy_dataset(numpy_dataset)
 
@@ -783,6 +789,7 @@ external_data_reader_registry = {
     **{fmt: read_sas for fmt in SAS_FORMATS},
     **{fmt: read_spss for fmt in SPSS_FORMATS},
     **{fmt: read_stata for fmt in STATA_FORMATS},
+    **{fmt: read_hdf5 for fmt in HDF5_FORMATS},
 }
 
 
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -53,7 +53,9 @@ _PREDICTIONS_SUFFIX = "_predictions"
 _PROBABILITIES_SUFFIX = "_probabilities"
 
 
-def _vectorize_ground_truth(ground_truth: pd.Series, str2idx: np.array, ground_truth_apply_idx: bool = True) -> np.array:
+def _vectorize_ground_truth(
+    ground_truth: pd.Series, str2idx: np.array, ground_truth_apply_idx: bool = True
+) -> np.array:
     # raw hdf5 files generated during preprocessing don't need to be converted with str2idx
     if not ground_truth_apply_idx:
         return np.vectorize(lambda x, y: x)(ground_truth, str2idx)
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -53,9 +53,7 @@ _PREDICTIONS_SUFFIX = "_predictions"
 _PROBABILITIES_SUFFIX = "_probabilities"
 
 
-def _vectorize_ground_truth(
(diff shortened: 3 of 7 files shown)


tgaddair:
Not sure why determinism test is failing, but looks like it might be transient.

@w4nderlust @jimthompson5802 could you also take a look?

jimthompson5802:
> Not sure why determinism test is failing, but looks like it might be transient.
> 
> @w4nderlust @jimthompson5802 could you also take a look?

@tgaddair   not sure I understand the comment.  Are you referring to the first set of unit tests that had one failure in the visualization set?  Are you saying that one or more of these visualization unit tests fail intermittently with no apparent reason?

tgaddair:
Apologies @jimthompson5802, I meant: would you mind reviewing this PR? I recall you and Piero were looking into this part of the code recently regarding the CSV vs Parquet issue.

jimthompson5802:
@tgaddair   Thank you for the clarification.  I ran a small test from the command line on this branch, it looks like predictions are still parquet file.  Should I assume the predictions will be parquet all the time and not csv?  

Reason for asking is that current user guide reference predictions as csv files.  We'll need to update the docs with that change.

After converting to using parquet files, I ran into this issue.  Right now I'm trying to figure out if I need to make additional updates to my cli example.  I'm using from pre 0.4 days.
```
  File "/usr/local/bin/ludwig", line 33, in <module>
    sys.exit(load_entry_point('ludwig', 'console_scripts', 'ludwig')())
  File "/opt/project/ludwig/cli.py", line 136, in main
    CLI()
  File "/opt/project/ludwig/cli.py", line 71, in __init__
    getattr(self, args.command)()
  File "/opt/project/ludwig/cli.py", line 99, in visualize
    visualize.cli(sys.argv[2:])
  File "/opt/project/ludwig/visualize.py", line 4540, in cli
    vis_func(**vars(args))
  File "/opt/project/ludwig/visualize.py", line 371, in compare_classifiers_performance_from_prob_cli
    compare_classifiers_performance_from_prob(
  File "/opt/project/ludwig/visualize.py", line 1644, in compare_classifiers_performance_from_prob
    ground_truth = vfunc(ground_truth, feature_metadata['str2idx'])
KeyError: 'str2idx'
root@1befdcec1c03:/opt/project/sandbox/debug_issues/iss_1234#

```

hungcs:
@jimthompson5802 Thanks for testing it out. What does your `ground_truth_metadata` file look like? The `ground_truth_metadata` file generated by the latest version of Ludwig should have `str2idx` and `idx2str` in the json file.

Right now we assume:
- `predictions` is `parquet`
- `ground_truth` is`csv` (or now with this PR `hdf5`)
- `ground_truth_metadata` is `json`

This is a sample `ground_truth_metadata` from titanic
[training_set_metadata.json.zip](https://github.com/ludwig-ai/ludwig/files/7614541/training_set_metadata.json.zip)

w4nderlust:
The bottom line should be
- we should make it possible to provide both preprocessed and non preprocessed ground truth (both hdf5, parquet and csv), and determine if to do the transformation or not
- regarding prquet vs csv for outputs, i think we want to save csv for local and parquet for remote backend, and as a consequence, viz should work with both

ludwig/utils/data_utils.py
```diff
@@ -197,6 +197,10 @@ def read_stata(data_fp, df_lib):
     return df_lib.read_stata(data_fp)
 
 
+def read_hdf5(data_fp, df_lib):
+    return load_hdf5(data_fp, clean_cols=True)
```
w4nderlust, line 201: Why do we need `df_lib`? What does `clean_cols` provide? Asking because if both are not useful we could get away with using `load_hdf5`` directly.

ludwig/utils/data_utils.py
```diff
@@ -197,6 +197,10 @@ def read_stata(data_fp, df_lib):
     return df_lib.read_stata(data_fp)
 
 
+def read_hdf5(data_fp, df_lib):
```
w4nderlust, line 200: df_lib is unused, maybe we can have kwargs instead? also please add type hints

hungcs: tried to keep consistent with the other `read` functions, which take in `(data_fp, df_lib)`, but can remove

jimthompson5802:
@hungcs   Turns out my issue was an IAK error (Idiot At Keyboard :-)).  The particular visualization I tested only support categorical output features and I just happened to use a model with a binary output feature.  So it was a false positive on the error.

hungcs:
> The bottom line should be
> 
> * we should make it possible to provide both preprocessed and non preprocessed ground truth (both hdf5, parquet and csv), and determine if to do the transformation or not
> * regarding prquet vs csv for outputs, i think we want to save csv for local and parquet for remote backend, and as a consequence, viz should work with both

Got it - so we should not use just the file type as the factor for whether we should do the transformation or not.

Would this be a matter of automatically attempting to read `ground_truth` as processed, and falling back to unprocessed if we get a key error? In that case, should we still expose the `ground_truth_apply_idx` argument?

w4nderlust:
Added a minor comment but I have 2 general considerations:
1. what is a case when we need to actually truncate the length of the precautions with `len(ground_truth)`? I can't imagine one that is not an error scenario. If the two lengths differ I'd rather raise an error
2. regarding the parameter that governs if to index or not, I think it is serviceable for now, although we should have a task to refactor this in 3 steps functions most likely: process input data (which maps raw predictions and ground truths in all format combinations into something standard and usable), calculate values to visualize, visualize the values. Right now there is some conflating of the first 2, and if we go in this direction we want to think specifically about the desired interface for both CLI and API and work backward from there.

Does this make sense to everyone? @tgaddair @jimthompson5802 @hungcs ?

ludwig/utils/data_utils.py
```diff
@@ -327,13 +331,15 @@ def save_hdf5(data_fp, data):
             h5_file.create_dataset(column, data=numpy_dataset[column])
 
 
-def load_hdf5(data_fp):
+def load_hdf5(data_fp, clean_cols: bool = False):
```
anmshkmr, line 334: Just a reminder to make sure we want to flip the default value for `clean_cols` here compared to the caller.

ludwig/globals.py
```diff
@@ -22,6 +22,13 @@
 TRAINING_PROGRESS_TRACKER_FILE_NAME = "training_progress.json"
 TRAINING_CHECKPOINTS_DIR_PATH = "training_checkpoints"
 
+TEST_STATISTICS_FILE_NAME = "test_statistics.json"
+
+PREDICTIONS_PARQUET_FILE_NAME = "predictions.parquet"
+PREDICTIONS_SHAPES_FILE_NAME = "predictions.shapes.json"
+
+TRAINING_PREPROC_FILE_NAME = "training.hdf5"
+
```
anmshkmr, line 31: +1

ludwig/visualize.py
```diff
@@ -53,6 +53,20 @@
...
 
+def _vectorize_ground_truth(
+    ground_truth: pd.Series, str2idx: np.array, ground_truth_apply_idx: bool = True
+) -> np.array:
+    # raw hdf5 files generated during preprocessing don't need to be converted with str2idx
+    if not ground_truth_apply_idx:
+        return np.vectorize(lambda x, y: x)(ground_truth, str2idx)
+
+    try:
+        return np.vectorize(_encode_categorical_feature)(ground_truth, str2idx)
```
anmshkmr, line 64: `lambda x, y: y[x]`?

anmshkmr: Ah, nvm.

anmshkmr (approved):
LGTM

ludwig/utils/data_utils.py
```diff
@@ -327,13 +331,15 @@ def save_hdf5(data_fp, data):
...
     with download_h5(data_fp) as hdf5_data:
         columns = [s.decode("utf-8") for s in hdf5_data[HDF5_COLUMNS_KEY][()].tolist()]
```
anmshkmr, line 336: For clarification, what happens if the columns key resolves to more hdf5 metadata?

hungcs: We assume `hdf5['columns']` will be flat and only contain the column names - I'm not aware of any cases where it contains extra metadata, but let me know if this is wrong

anmshkmr: I don't know first hand, but from reading it sounds like there could be arbitrary depth to group -> dataset expansion when reading hdf5 data. Let's close this PR and revisit if necessary.

PR #1507, opened 2021-11-21
Add experiment heuristics to automl module (variant of Avanika PR 1362)

amholler:
This PR is a variant of Avanika's PR 1362, generated against the tf-legacy branch.
The changes relative to PR 1362 are:
1) Corrected issue when generating config for concat model type (missing merge_dict, commented on PR)
2) Updated to use tune.with_parameters to support ray tune on large datasets
3) Ensure trial limit is consistently time-based (not epoch based) as intended
4) omitted python style changes that reviewer wanted removed

This PR has been run successfully on the following datasets:
* ames_housing, forest_cover, mushroom_edibility, synthetic_fraud [tabnet]
* higgs[concat, large]
* otto_group_product [concat]

### DIFF
--- a/ludwig/automl/automl.py
+++ b/ludwig/automl/automl.py
@@ -15,13 +15,15 @@ from typing import Dict, Union
 
 import numpy as np
 import pandas as pd
+import copy
 
 from ludwig.api import LudwigModel
-from ludwig.automl.base_config import _create_default_config, DatasetInfo
+from ludwig.automl.base_config import _create_default_config, DatasetInfo, get_dataset_info, infer_type
 from ludwig.automl.auto_tune_config import memory_tune_config
 from ludwig.automl.utils import _ray_init, get_model_name
-from ludwig.constants import COMBINER, TYPE
+from ludwig.constants import COMBINER, NUMERICAL, TYPE
 from ludwig.hyperopt.run import hyperopt
+from ludwig.utils.misc_utils import merge_dict
 
 try:
     import dask.dataframe as dd
@@ -65,6 +67,7 @@ def auto_train(
     time_limit_s: Union[int, float],
     output_directory: str = OUTPUT_DIR,
     tune_for_memory: bool = False,
+    user_specified_config: Dict = None,
     **kwargs
 ) -> AutoTrainResults:
     """
@@ -75,7 +78,7 @@ def auto_train(
     All batch and learning rate tuning is done @ training time.
 
     # Inputs
-    :param dataset: (str) filepath to dataset.
+    :param dataset: (str, pd.DataFrame, dd.core.DataFrame) data source to train over.
     :param target: (str) name of target feature
     :param time_limit_s: (int, float) total time allocated to auto_train. acts
                         as the stopping parameter
@@ -86,7 +89,7 @@ def auto_train(
     :return: (AutoTrainResults) results containing hyperopt experiments and best model
     """
     config = create_auto_config(
-        dataset, target, time_limit_s, tune_for_memory, **kwargs)
+        dataset, target, time_limit_s, tune_for_memory, user_specified_config, **kwargs)
     return train_with_config(
         dataset,
         config,
@@ -100,6 +103,7 @@ def create_auto_config(
     target: str,
     time_limit_s: Union[int, float],
     tune_for_memory: bool,
+    user_specified_config: Dict = None,
 ) -> dict:
     """
     Returns an auto-generated Ludwig config with the intent of training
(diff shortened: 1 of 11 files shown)


ludwig/automl/base_config.py
```diff
@@ -41,10 +41,18 @@
...
+
+combiner_defaults = {
+    "concat": os.path.join(CONFIG_DIR, "combiner/concat_config.yaml"),
+    "tabnet": os.path.join(CONFIG_DIR, "combiner/tabnet_config.yaml"),
+    "transformer": os.path.join(CONFIG_DIR, "combiner/transformer_config.yaml"),
+}
+
+encoder_defaults = {
+    "text": {
+        "bert": os.path.join(CONFIG_DIR, "text/bert_config.yaml"),
```
tgaddair, line 54: While BERT gets very good performance, I feel like it's cost/performance ratio is pretty bad (running on a single v100, for example, the batch size ends up being so small you'd be better off using something else like DistilBERT). Finding a batch size small enough to work for BERT in such cases is also a challenge. But curious if you've observed something different.

ludwig/automl/base_config.py
```diff
@@ -133,17 +141,48 @@ def _create_default_config(
...
+        experiment_resources
+    )
+    base_automl_config["hyperopt"]["executor"][
+        "time_budget_s"
+    ] = time_limit_s
+    if time_limit_s is not None:
+        num_samples = base_automl_config["hyperopt"]["sampler"]["num_samples"]
+        if num_samples is not None:
+            # allow trials to get 2x even division, since some trials perform better than avg
+            base_automl_config["hyperopt"]["sampler"]["scheduler"][
```
tgaddair, line 157: I'm curious what happens if we let the per trial max_t equal the total training time. Would this result in some of the `num_samples` never getting tried? I would hope that hyperband would be fair in its scheduling so that it would try all the samples for a small amount of time before letting the best trials run indefinitely.

ludwig/automl/automl.py
```diff
@@ -177,13 +183,76 @@ def train_with_config(
...
+            # TODO (ASN): add more robust heuristics
+            if input_feature["type"] == "text":
+                input_feature["encoder"] = "bert"
+                base_config = merge_dict(
+                    base_config, default_configs["text"]["bert"])
+
+            # TODO (ASN): add image heuristics
+
+    # override and constrain automl config based on user specified values
+    if user_specified_config is not None:
```
tgaddair, line 242: I wonder if something like OmegaConf would help simplify the merging process, particularly now that we also have a base_config and combiner config separation in addition to the user_config here.

https://omegaconf.readthedocs.io/en/2.1_branch/usage.html#omegaconf-merge

tgaddair (commented):
Love the new heuristics. Added some comments, let me know what you think.

ludwig/automl/automl.py
```diff
@@ -69,6 +71,7 @@ def auto_train(
     time_limit_s: Union[int, float],
     output_directory: str = OUTPUT_DIR,
     tune_for_memory: bool = False,
+    user_specified_config: Dict = None,
```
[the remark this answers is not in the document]
amholler: Updated.

ludwig/automl/base_config.py
amholler: I personally don't know about this; would need to check with Avanika here.
In the meantime, I could pull this update out of the change set, while we get more clarity here,
if you think that would be best.

ludwig/automl/base_config.py
amholler: Well, I thought I tried this and observed starvation of trials, but let me try it again to double-check.

ludwig/automl/automl.py
amholler: Thanks for the reference to OmegaConf, looks nice!

My understanding of the complexity of this code is that it is removing config from
the hyperopt parameters section that is overlapping with the config_section.key
that is included in the user_config.  I don't see a particular feature of OmegaConf
that can handle this specialized case.  Apologies if I am missing something here.

ludwig/automl/base_config.py
amholler: Okay, I just observed that for my Ray Tune async hyperband run, if specify num_samples=20
and set max_t == time_limit_s, then the first 3 trials are run for max_t time and no other trials are run.

ludwig/automl/base_config.py
tgaddair: Sounds good, we can try it out and see how it goes.

ludwig/automl/base_config.py
tgaddair: Thanks for confirming, sounds like this is the right thing to do for now, then. Would be great if we could find a way to both explore all trials and let trials run as long as they can.

ludwig/automl/automl.py
tgaddair: I think it will end up being a refactor of some things. I can take a look in a follow-up. But my goal would be to just be able to eventually reduce the complexity to "merge a bunch of independent configs".

tgaddair (approved):
LGTM!

ludwig/automl/base_config.py
amholler: just posted this question to the ray tune slack channel

PR #1517, opened 2021-11-24
Add 1.0 and 0.0 as bool-like strings and pass kwargs to roc eval_utils.

### LINKED ISSUE #1511: [automl] Binary column type not detected for floats
For example, the titanic dataset has values like `1.0` and `0.0`. These can correctly be interpreted as binary values `1` and `0` respectively, but the auto detect system is treating this as a categorical feature.

justinxzhao:
Fixes #1511.


tgaddair (approved):
LGTM!

PR #1522, opened 2021-11-26
Added decorators for custom combiners, losses, metrics

### LINKED ISSUE #1502: Extensible loss and metric modules
Currently, users need to make changes to the underlying Ludwig internals in order to add a new loss or metric. Similar to what we did for encoders / decoders, it should be possible for users to add and register new losses and metrics using a simple decorator system like the following:

```
@loss("my_loss")
class MyLoss(nn.Module):
    def forward(self, preds: Tensor, target: Tensor) -> Tensor:
        ...

@metric("my_metric")
class MyMetric(Metric):
    def update(self, preds: Tensor, target: Tensor) -> None:
        ...

    def compute(self) -> Tensor:
        ...

    @abstractmethod
    def get_current_value(self, preds: Tensor, target: Tensor) -> Tensor:
        ...

@register_loss("my_loss", [NUMERICAL, CATEGORY])
@register_metric("my_metric", [NUMERICAL])
```

tgaddair:
Any class can now be made a custom combiner, loss, or metric as follows:

```
@register("my_combiner")
class MyCombiner:
    ...

@register_loss("my_loss", [NUMERICAL, VECTOR])
class MyLoss:
    ...

@register_metric("my_metric", [CATEGORY])
class MyMetric:
    ...
```

Also refactors Encoders to use a new `register_encoder` decorator. This allows us to remove the special subclasses of `Encoder` that only serve to put the derived classes into the correct registry. Same for Decoders.

Fixes #1502.


ludwig/combiners/combiners.py
```diff
@@ -482,19 +491,18 @@ class Meta:
         unknown = INCLUDE
 
 
-class TabNetCombiner(CombinerClass):
+@register_combiner(name='tabnet')
+class TabNetCombiner(Combiner):
     def __init__(
             self,
-            input_features: Dict,
+            input_features: Dict[str, "InputFeature"],
```
justinxzhao, line 498: Is wrapping the class name in quotes the standard way to declare a type hint for a Dict input arg?

ludwig/combiners/combiners.py
```diff
@@ -1131,14 +1122,3 @@ def get_combiner_class(combiner_type):
         combiner_type,
         combiner_registry
     )
-
-
-combiner_registry = {
```
justinxzhao, line 1136: Nice!

justinxzhao (approved):
LGTM! Feels like a big improvement over the current registry mechanisms.

ludwig/combiners/combiners.py
tgaddair: It's a way to reference the type without needing to import it (in this case, it would be a circular dependency).

PR #1525, opened 2021-11-29
Fix exception in file type_utils.py

kriziacicchetti:
Herewith the previous error given by the code during training:

  File "\ludwig\utils\type_utils.py", line 10, in <module>
    if dd is not None:
NameError: name 'dd' is not defined


carlogrisetti:
LGTM, awaiting CI

tgaddair (approved):

PR #1533, opened 2021-11-29
Fix test_api.py tests

jeffreyftang:
Set of small changes to get tests in `test_api.py` working. Notably:

1. Disables auto-detection of Ray backend if running under pytest. This is a workaround to allow testing to run properly on AWS even in the presence of a Ray cluster and Pandas dataframes, which we currently don't support on Ray.
2. Use absolute filepaths in output directory.
3. Replace some numpy functions with their torch equivalents.
4. Move the entire ECD module (and by extension its submodules) onto GPU if one is available.

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -87,6 +87,7 @@ class Predictor(BasePredictor):
                                     file=sys.stdout,
                                     disable=is_progressbar_disabled())
 
+            model.to(self.device)
             predictions = defaultdict(list)
             while not batcher.last_batch():
                 batch = batcher.next_batch()
@@ -104,7 +105,8 @@ class Predictor(BasePredictor):
 
         return from_numpy_dataset(predictions)
 
-    def predict_single(self, model, batch):
+    def predict_single(self, model: ECD, batch):
+        model.to(self.device)
         predictions = defaultdict(list)
         preds = self._predict(model, batch)
         self._accumulate_preds(preds, predictions)
@@ -156,6 +158,7 @@ class Predictor(BasePredictor):
                                     file=sys.stdout,
                                     disable=is_progressbar_disabled())
 
+            model.to(self.device)
             predictions = defaultdict(list)
             while not batcher.last_batch():
                 batch = batcher.next_batch()
@@ -213,6 +216,7 @@ class Predictor(BasePredictor):
                                 file=sys.stdout,
                                 disable=is_progressbar_disabled())
 
+            model.to(self.device)
             collected_tensors = []
             while not batcher.last_batch():
                 batch = batcher.next_batch()
--- a/ludwig/backend/__init__.py
+++ b/ludwig/backend/__init__.py
@@ -15,6 +15,8 @@
 # limitations under the License.
 # ==============================================================================
 
+import os
+
 from ludwig.backend.base import Backend, LocalBackend
 from ludwig.utils.horovod_utils import has_horovodrun
 
@@ -35,6 +37,11 @@ ALL_BACKENDS = [LOCAL, DASK, HOROVOD, RAY]
 
 
 def _has_ray():
+    # Temporary workaround to prevent tests from automatically using the Ray backend. Taken from
+    # https://stackoverflow.com/questions/25188119/test-if-code-is-executed-from-within-a-py-test-session
+    if "PYTEST_CURRENT_TEST" in os.environ:
+        return False
+
     if _ray is None:
         return False
 
--- a/ludwig/utils/misc_utils.py
+++ b/ludwig/utils/misc_utils.py
@@ -126,7 +126,7 @@ def get_output_directory(
         output_directory,
         experiment_name + ('_' if model_name else '') + model_name
     )
-    return find_non_existing_dir_by_adding_suffix(base_dir_name)
+    return os.path.abspath(find_non_existing_dir_by_adding_suffix(base_dir_name))
 
 
 def get_file_names(output_directory):
--- a/tests/integration_tests/test_api.py
+++ b/tests/integration_tests/test_api.py
@@ -20,6 +20,7 @@ from unittest import mock
 
 import numpy as np
 import pytest
+import torch
 
 from ludwig.api import LudwigModel
 from ludwig.utils.data_utils import read_csv
@@ -256,13 +257,13 @@ def test_api_training_determinism(csv_filename):
 
         divergence = False
         for weight_1, weight_2 in zip(model_weights_1, model_weights_2):
-            if not np.allclose(weight_1, weight_2):
+            if not torch.allclose(weight_1, weight_2):
                 divergence = True
                 break
         assert divergence, 'model_1 and model_2 have identical weights with different seeds!'
 
         for weight_1, weight_3 in zip(model_weights_1, model_weights_3):
-            assert np.allclose(weight_1, weight_3)
+            assert torch.allclose(weight_1, weight_3)
 
 
 def run_api_commands(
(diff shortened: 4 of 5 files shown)


ludwig/models/ecd.py
```diff
@@ -69,6 +69,8 @@ def __init__(
         # ================ Combined loss metric ================
         self.eval_loss_metric = torchmetrics.MeanMetric()
 
+        self.to('cuda' if torch.cuda.is_available() else 'cpu')
```
ShreyaR, line 72: @jeffreyftang does removing this line cause failures?

I'm unsure about why this line is needed since we already move an initialized ECD object to to CUDA in the trainer [here](https://github.com/ludwig-ai/ludwig/pull/1508/files)

jeffreyftang: Yes it does (although only 1 failure). 
```
>       return torch._C._nn.linear(input, weight, bias)
E       RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_addmm)

/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/functional.py:1848: RuntimeError
```

jeffreyftang: Originally I had modified `generic_decoders.py` to move all the decoders to GPU explicitly, but adding the single line to `ECD` resolved things (as well as cleared up some of the issues I encountered after making the mods to `generic_decoders.py`).

jeffreyftang: Full error trace:
```
tests/integration_tests/utils.py:522: in run_api_experiment
    loaded_model.predict(dataset=data_csv)
ludwig/api.py:759: in predict
    dataset,
ludwig/models/predictor.py:118: in batch_predict
    preds = self._predict(model, batch)
ludwig/models/predictor.py:145: in _predict
    return model.predict_step(inputs)
ludwig/models/ecd.py:214: in predict_step
    return self.predictions(inputs, output_features=None)
ludwig/models/ecd.py:199: in predictions
    outputs = self(inputs)
/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py:1102: in _call_impl
    return forward_call(*input, **kwargs)
ludwig/models/ecd.py:159: in forward
    decoder_outputs = decoder(decoder_inputs, mask=mask)
/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py:1102: in _call_impl
    return forward_call(*input, **kwargs)
ludwig/features/base_feature.py:276: in forward
    logits = self.logits(logits_input, target=target)
ludwig/features/category_feature.py:191: in logits
    LOGITS: self.decoder_obj(hidden),
/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py:1102: in _call_impl
    return forward_call(*input, **kwargs)
ludwig/decoders/generic_decoders.py:166: in forward
    return self.dense(inputs)
/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py:1102: in _call_impl
    return forward_call(*input, **kwargs)
ludwig/utils/torch_utils.py:203: in forward
    output = torch.squeeze(self.dense(input), dim=-1)
/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py:1102: in _call_impl
    return forward_call(*input, **kwargs)
/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/modules/linear.py:103: in forward
    return F.linear(input, self.weight, self.bias)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

input = tensor([[ 0.6498, -3.4886, -3.3236, -2.5369, -1.2582,  0.6344,  3.4147, -1.1249],
        [ 1.1283, -0.5554,  1.0740, ....9143, -2.2151, -2.4031, -3.4755,  0.0326, -0.3848,  1.9021, -2.3640]],
       device='cuda:0', grad_fn=<SumBackward1>)
weight = Parameter containing:
tensor([[-0.3009,  0.5633,  0.1450, -0.3612, -0.3303, -0.0303,  0.3645, -0.1663],
        [-0.37...-0.3957],
        [-0.3995, -0.5854, -0.2124,  0.2221,  0.4184,  0.3013, -0.5777, -0.3927]],
       requires_grad=True)
bias = Parameter containing:
tensor([-0.0010,  0.0010,  0.0010,  0.0010, -0.0010,  0.0010],
       requires_grad=True)

    def linear(input: Tensor, weight: Tensor, bias: Optional[Tensor] = None) -> Tensor:
        r"""
        Applies a linear transformation to the incoming data: :math:`y = xA^T + b`.

        This operator supports :ref:`TensorFloat32<tf32_on_ampere>`.

        Shape:

            - Input: :math:`(N, *, in\_features)` N is the batch size, `*` means any number of
              additional dimensions
            - Weight: :math:`(out\_features, in\_features)`
            - Bias: :math:`(out\_features)`
            - Output: :math:`(N, *, out\_features)`
        """
        if has_torch_function_variadic(input, weight, bias):
            return handle_torch_function(linear, (input, weight, bias), input, weight, bias=bias)
>       return torch._C._nn.linear(input, weight, bias)
E       RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_addmm)

/home/ray/anaconda3/lib/python3.7/site-packages/torch/nn/functional.py:1848: RuntimeError
========================================================================================================================================================================= short test summary info ==========================================================================================================================================================================
FAILED tests/integration_tests/test_api.py::test_api_intent_classification - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_addmm)
```

w4nderlust: Jumping in without the full context, but could it be that the problem is with Dense, and how we initialize it? Like instead of `weights_initializer(self.dense.weight)` we should do `weights_initializer(self.dense.weight).to("coda")`? And the same for the bias?

jeffreyftang: Discussed a bit with @ShreyaR offline, and I think the recommended approach is to move `ECD` to GPU from the predictor (similar to how `ECD` is moved in the trainer). Going to give it a try :)

ShreyaR (approved):
lgtm!

ludwig/models/predictor.py
```diff
@@ -87,6 +87,7 @@ def batch_predict(
                                     file=sys.stdout,
                                     disable=is_progressbar_disabled())
 
+            model.to(self.device)
```
ShreyaR, line 90: One change here could be to pass in the model during initialization of an object of `Predictor`, and move it to the correct device at initialization. Then every method of `Predictor` can use `self.model` directly, instead of moving the model to the correct device.

This should be ok for now however.

jeffreyftang: I considered that as well, but I wasn't sure about modifying the signature of `Predictor` methods, so I went with the least intrusive approach.

PR #1539, opened 2021-11-30
[torch] Fix GPU tests for test_visualizations

ShreyaR:
Fixes tests in:

- `tests/integration_tests/test_visualization.py` -- some tests are fixed using the changes introduced in @jeffreyftang's PR https://github.com/ludwig-ai/ludwig/pull/1533. Other tests are fixed with the changes to `embedding_modules`.

### DIFF
--- a/ludwig/modules/attention_modules.py
+++ b/ludwig/modules/attention_modules.py
@@ -1,4 +1,3 @@
-# coding=utf-8
 # Copyright (c) 2019 Uber Technologies, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -25,6 +24,7 @@ logger = logging.getLogger(__name__)
 
 
 class FeedForwardAttentionReducer(LudwigModule):
+
     def __init__(self, input_size, hidden_size=256, activation='tanh'):
         super().__init__()
         self.fc_layer1 = nn.Linear(input_size, hidden_size)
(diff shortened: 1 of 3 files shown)


ludwig/utils/misc_utils.py
```diff
@@ -111,28 +105,18 @@ def set_default_values(dictionary, default_value_dictionary):
...
-        model_name='run'
-):
-    base_dir_name = os.path.join(
-        output_directory,
-        experiment_name + ('_' if model_name else '') + model_name
-    )
-    return find_non_existing_dir_by_adding_suffix(base_dir_name)
+def get_output_directory(output_directory, experiment_name, model_name='run'):
+    base_dir_name = os.path.join(output_directory,
+                                 experiment_name + ('_' if model_name else '') + model_name)
```
anmshkmr, line 113: [Nit], if `model_name` is `None` it could result in TypeError.

Maybe `experiment_name + ('_' + model_name if model_name else '')`

ludwig/modules/attention_modules.py
```diff
@@ -181,6 +185,11 @@ def forward(self, inputs, mask=None):
             hidden = layer(hidden, mask=mask)
         return hidden
 
+    @property
+    def output_shape(self) -> torch.Size:
+        return torch.Size([self.sequence_size, self.input_size])
```
w4nderlust, line 190: should this be hidden size instead of input size?

ShreyaR: Good catch! Fixed.

ludwig/utils/misc_utils.py
ShreyaR: Thanks! Updated

anmshkmr (approved):

jeffreyftang (approved):

PR #1544, opened 2021-12-01
[torch] Fixed tests in test_encoder.py

ShreyaR:


ludwig/modules/mlp_mixer_modules.py
```diff
@@ -1,4 +1,3 @@
-# coding=utf-8
 # Copyright (c) 2021 Linuf Foundation
```
jeffreyftang, line 1: Nit: s/Linuf/Linux :)

jeffreyftang (approved):

anmshkmr (approved):

ludwig/modules/mlp_mixer_modules.py
ShreyaR: Thanks, fixed!

PR #1551, opened 2021-12-02
Format and fix lint errors for sequence-related files and output features.

justinxzhao:


ludwig/utils/metric_utils.py
```diff
@@ -33,73 +30,17 @@ def dynamic_partition(data: Tensor, partitions: Tensor, num_partitions: int):
...
-        data=correct_preds,
-        partitions=mask,
-        num_partitions=2)
+    mask = sequence_mask(lengths=targets_sequence_lengths, maxlen=correct_preds.shape[1], dtype=torch.int32)
+    _, masked_correct_preds = dynamic_partition(data=correct_preds, partitions=mask, num_partitions=2)
 
     return masked_correct_preds.type(torch.float32)
-
-
-# TODO(shreya): After sequence loss
```
ShreyaR, line 64: @justinxzhao would this commented out code be helpful for when you're migrating the sequence losses? Feel free to remove if not.

ludwig/features/numerical_feature.py
```diff
@@ -54,7 +59,7 @@ def inverse_transform(self, x: np.ndarray) -> np.ndarray:
         return x * self.sigma + self.mu
 
     @staticmethod
-    def fit_transform_params(column: np.ndarray, backend: "Backend") -> dict:
+    def fit_transform_params(column: np.ndarray, backend) -> dict:
```
ShreyaR, line 62: Can you keep the type hint but add a `noqa` instead? I believe it should be `noqa F821` but not 100% sure.

ShreyaR: Here and in the other type hint removals in this file.

ShreyaR (approved):
Lgtm except for removing type hints. Feel free to merge once that is fixed!

ludwig/utils/metric_utils.py
justinxzhao: Now that we have the tf-legacy branch, my plan was to reference that instead.

ludwig/modules/metric_registry.py
```diff
@@ -0,0 +1,29 @@
+from typing import List, Union
+
+from ludwig.utils.registry import Registry
+
+metric_feature_registry = Registry()
+metric_registry = Registry()
+
+
+def register_metric(name: str, features: Union[str, List[str]]):
```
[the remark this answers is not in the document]
justinxzhao: This came from my attempt to address the circular dependency introduced by the `"OutputFeature"` type hint on `LudwigMetric.can_report()`, added in https://github.com/ludwig-ai/ludwig/pull/1522/files. The linter complains (rightly so) that it doesn't know what OutputFeature is.

The circular dependency arises because `base_feature.py` needs the metrics registry, and metrics registry was in `metrics_modules.py`, but `metrics_modules.py` needs to depend on `OutputFeature` in `base_feature.py` for the type hint.

My experience is that circular dependencies are usually an indication of an opportunity for a better design, so that's my hope with splitting out the metrics registry into `metric_registry.py` file, separate from actual metric definitions.

ludwig/features/numerical_feature.py
justinxzhao: Done. 

Adding noqa bypasses the lint error, though I wonder if type hints causing linter errors is an indication of a suboptimal design.

justinxzhao:
> Lgtm except for removing type hints. Feel free to merge once that is fixed!

Sounds good! Thanks for the review!

PR #1552, opened 2021-12-02
Fix minor typo

louixs:
Hi! 
I think there's a small typo in the readme. I add the word "all", based on the rest of the sentence and my best guess of the authors intent :)

We could perhaps remove "is" instead as well. If you prefer this wording please let me know, I'll prepare a new patch!

Many thanks!


justinxzhao:
The new wording seems great, thanks Ryuei!

louixs:
That's great, thank you Justin!

PR #1554, opened 2021-12-02
Fixed embedding modules tests

anmshkmr:


ludwig/modules/embedding_modules.py
```diff
@@ -223,7 +223,7 @@ def forward(self, inputs: torch.Tensor, mask: Optional[torch.Tensor] = None) ->
                     inputs[b, i] indicates that token i is present in sample b.
         """
         # Convert multi-hot input to input of indices
-        inputs = inputs.int() * torch.arange(self.vocab_size)
+        inputs = (inputs.int() * torch.arange(self.vocab_size).to('cuda' if torch.cuda.is_available() else 'cpu'))
```
jeffreyftang, line 226: Can this take advantage of the `device` property on `LudwigModules`?

ShreyaR: @anmshkmr I believe that this change was introduced in the https://github.com/ludwig-ai/ludwig/pull/1544 PR and should be in the `torch-gpu` branch already.

anmshkmr: @ShreyaR I don't see these tests working after rebasing on `torch-gpu`. Also, for some reason the `self.device` property is still `cpu` in this case.

Will investigate later.

ShreyaR: @anmshkmr I believe that's probably happening because you still need the changes made to `tests/ludwig/modules/test_embedding_modules.py`.

Additionally, I believe you'd also need to move the `embed` module in each test to the correct device, along with the inputs.

tests/ludwig/modules/test_embedding_modules.py
```diff
@@ -19,7 +19,7 @@ def test_embed(
         embedding_size=embedding_size,
         representation=representation,
     )
-    inputs = torch.randint(0, 2, size=(2, 1)).bool()
+    inputs = torch.randint(0, 2, size=(2, 1)).bool().to('cuda' if torch.cuda.is_available() else 'cpu')
```
ShreyaR, line 22: Could you create a global `DEVICE` variable, and use that while moving all `inputs` to the correct device?

PR #1556, opened 2021-12-02
Fixed encoders tests except test_h3_rnn_embed

anmshkmr:


ludwig/encoders/h3_encoders.py
```diff
@@ -183,10 +183,10 @@ def forward(self, inputs: torch.Tensor) -> Dict[str, torch.Tensor]:
         # Mask out cells beyond the resolution of interest.
         resolution = input_vector[:, 2]
         mask = torch.unsqueeze(
-            torch_utils.sequence_mask(resolution, 15), -1).type(
+            torch_utils.sequence_mask(resolution, 15).to('cuda' if torch.cuda.is_available() else 'cpu'), -1).to('cuda' if torch.cuda.is_available() else 'cpu').type(
```
ShreyaR, line 186: In general, I would prefer avoiding adding `.to` statements throughout the code. One alternative is that you can register this as a buffer instead (example [here](https://github.com/ludwig-ai/ludwig/blob/00b5fa57b8d46a60faf7479784e913f589cb541e/ludwig/modules/embedding_modules.py#L466)) -- that will make it so that this tensor is automatically moved to the correct device when we perform `model.to` at the top level.

ShreyaR: Same for the other tensors that are edited in this file and `torch_utils.py`.

ShreyaR (changes requested):

PR #1558, opened 2021-12-02
Pin torch==1.9.1 and torchvision==0.10.1 on Windows

tgaddair:
https://github.com/pytorch/pytorch/issues/65473

cc @carlogrisetti

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,8 @@ scipy>=0.18
 tabulate>=0.7
 scikit-learn<1.0
 tqdm
-torch>=1.9.0
+torch==1.9.1 ; platform_system == "Windows"  # https://github.com/pytorch/pytorch/issues/65473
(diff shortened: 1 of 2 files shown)


tgaddair:
Ah I see, I didn't realize this would also require a downgrade to 1.9. That's unfortunate. I will try to see if there's another workaround that doesn't require pinning PyTorch itself, as that would be a very limiting change.

tgaddair:
Although, it does sound like this is Windows-specific, so pinning only on Windows would be an acceptable workaround, I feel. Let me try that.

carlogrisetti:
Oh well... pinning only on windows would be more than acceptable in the meantime that this will be fixed.
Once it's fixed we could bump the minimum version to the fixed one (maybe only for windows, although I don't like that so much hehe)

Thanks again

PR #1559, opened 2021-12-02
Accept and move model to GPU as part of Predictor init

jeffreyftang:
This is a follow-up to https://github.com/ludwig-ai/ludwig/pull/1533#discussion_r759939214. Instead of passing and moving the target model to GPU in every invocation of `Predictor` functions, we now pass the model in at `Predictor` initialization.


ShreyaR (approved):
lgtm!

PR #1574, opened 2021-12-07
Better handle empty fields with distinct_values=[]

hungcs:


tgaddair (approved):
Nice! Can you also cherry-pick into master?

PR #1577, opened 2021-12-08
Better handle empty fields with distinct_values=[]

hungcs:


tgaddair:
@hungcs looks like the whole codebase was formatted, can you rebase on master?

tgaddair (approved):

PR #1581, opened 2021-12-08
GRU, RNN, and LSTM sequence decoders re-implemented in Torch

justinxzhao:
[Design Doc](https://docs.google.com/document/d/1_UkuPvvGZNQ9yUaKPFzZv0GtS9e91HEFfFkO5nQNqG0/edit?usp=drive_web&ouid=111201313049129417866)

### 1. Implementations of GRU, RNN, and LSTM sequence decoders without attention.

- Reference implementations:
    - [PyTorch seq2seq tutorial](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html)
    - [bentrevett/pytorch-seq2seq](https://github.com/bentrevett/pytorch-seq2seq)
- `SequenceGeneratorDecoder` is now a dispatcher for different sequence decoders instead of a monolith for all cell types and decoding techniques. This is also more aligned with `huggingface`'s design, which favors code duplication over consolidation.
- This PR also re-implements `SequenceSoftmaxCrossEntropyLoss`, also based on `huggingface`'s sequence loss functions, and we register this loss exclusively for `TEXT` and `SEQUENCE` features.

### 2. Revamp/Fixing output feature dependencies.

- `base_feature.OutputFeature` now correctly initializes its `FCStack` with the input size of the combiner outputs, concatenated with any dependent features (see `get_input_size_with_dependencies()`).
- To do this, `OutputFeature`'s constructor receives a dictionary of other OutputFeatures as input. This allows an `OutputFeature` to parse output sizes from any dependent output features, to properly compute the input size for itself. This fits well into `ECD`'s existing topological sorting of output feature definitions. It's interesting to speculate that this was probably not an issue with TF/0.4 since the shapes are automatically determined via graph compilation.
- This PR also moves `OutputFeature.concat_dependencies()` to a stateless utility function in `output_feature_utils`, which also helps clean up `test_dependencies.py` integration tests.

### 3. Move all special symbol data adjustments to data pre-processing.

- In 0.4, `UNKNOWN` and `PADDING` symbols are added in pre-processing while `START` and `STOP` symbols are added in modeling. Now, we add all special tokens at data pre-processing time. Doing this all in one place, outside of the model, makes the modeling and evaluation code cleaner and it also matches how other libraries do pre-processing, like [huggingface](https://huggingface.co/docs/transformers/preprocessing).
- Having all of the special symbols formalized and used in pre-processing and `strings_utils` also allows us to clean up magical `!= 0` PADDING symbol references in metrics (`TokenAccuracy`) and sequence (`sequence_length_2D()`) libraries. Now, these libraries and reference `strings_utils.SpecialSymbol.PADDING.value`.
- This PR also adds unit tests for vocabulary and sequence dataset building utilities in `strings_utils`, also fixing a bug with the default character-based tokenizer.

### 4. Other changes

- Add `logits()`, `predictions()`, and `get_prediction_set()` as abstract methods to base_feature, with documentation. 
- Remove `ConfusionMatrix` calculations. Perhaps we can add these back in when we reimplement the `SequenceTaggerDecoder`.
- Simplify `build_output_feature()` and `build_input_feature()`
- Fix implementation of `dynamic_partition()` and `TokenAccuracyMetric`, and add unit tests for them.
- Re-enable unit tests and integration tests that were disabled for sequence decoder migration.

## Follow-up changes

- Re-implementing and testing remaining sequence metrics.
- Re-implementing and testing `SampledSoftmaxCrossEntropyLoss`. There's a nice reference implementation from [AllenNLP](https://github.com/allenai/allennlp/blob/main/allennlp/modules/sampled_softmax_loss.py)
- Beam search decoding. Reference implementations on [github](https://github.com/budzianowski/PyTorch-Beam-Search-Decoding/blob/master/decode_beam.py) and [huggingface](https://github.com/huggingface/transformers/blob/e513c16e82221e0a150b97a31f2c3c864913c454/src/transformers/generation_utils.py). [Relevant blog post](https://huggingface.co/blog/how-to-generate)
- Sequence decoding with attention by passing sequence encoder states through the combiner. This could probably use some re-vamping.
- `SequenceTaggerDecoder`, potentially with a separate output type to build more insulation between sequence generation and sequence classification.

## Proof that it's working

These graphs comes from training models on the `twitterbots` dataset, using `account_type` as a text feature.

Tensorflow LSTM:
image

Pytorch LSTM:
image

The PyTorch implementation has a smoother and lower loss curve, which may also be attributed to changes around unified pre-processing.


ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,227 @@
...
+from ludwig.constants import HIDDEN, LOGITS, SEQUENCE, TEXT
 from ludwig.decoders.base import Decoder
 from ludwig.decoders.registry import register_decoder
+from ludwig.utils import strings_utils
 
 logger = logging.getLogger(__name__)
 
 
-# TODO(Justin): Implement sequence decoders.
+class RNNDecoder(nn.Module):
```
jimthompson5802, line 29: Should the super class be `LudwigModule`?  Same comment will apply to the other classes added.

justinxzhao: LudwigModule has additional methods like `input_shape` and `output_shape`.

I'm open to the idea, but I don't see too much benefit over the standard `torch.nn.Module` for these Decoder-internal classes.

Using `torch.nn.Module` also makes our implementation more easily referenceable against other open source versions of these models.

@ShreyaR @w4nderlust WDYT?

jimthompson5802: @justinxzhao   Recalling our conversation, I think you're saying these classes are completely encapsulated w/in the `SequenceGeneratorDecoder` class, for example, and will never be visible to the "outside world".  If this is the case, then what I wrote may not make sense.

w4nderlust: I second @jimthompson5802 's last comment

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,227 @@
...
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.vocab_size = vocab_size
+        self.embedding = nn.Embedding(vocab_size, hidden_size)
+        self.relu = nn.ReLU()
+        if cell_type == "gru":
+            self.rnn = nn.GRU(hidden_size, hidden_size, batch_first=True)
+        else:
+            self.rnn = nn.RNN(hidden_size, hidden_size, batch_first=True)
+        self.out = nn.Linear(hidden_size, vocab_size)
```
w4nderlust, line 42: In the tf implementation I believe we had the option to have the same weights for embedding and out linear (being one the transpose of the other)

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,227 @@
...
+            hidden: [batch_size, hidden_size] tensor with the previous step's hidden state.
+
+        Returns:
+            Tuple of two tensors:
+            - output: [batch_size, vocab_size] tensor with the logits.
+            - hidden: [batch_size, hidden_size] tensor with the hidden state for the next time step.
+        """
+        # Unsqueeze predicted tokens.
+        input = input.unsqueeze(1).to(torch.int)
+        output = self.relu(self.embedding(input))
```
w4nderlust, line 60: why rely here?

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,227 @@
...
+    """LSTM-based decoder."""
+
+    def __init__(self, hidden_size: int, vocab_size: int):
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.vocab_size = vocab_size
+        self.embedding = nn.Embedding(vocab_size, hidden_size)
+        self.relu = nn.ReLU()
+        self.lstm = nn.LSTM(hidden_size, hidden_size, batch_first=True)
+        self.out = nn.Linear(hidden_size, vocab_size)
```
w4nderlust, line 76: same thing for embedding and out linear optionally using the same weights as for the rundecoder

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,227 @@
...
+
+        Returns:
+            Tuple of 3 tensors:
+            - output: [batch_size, vocab_size] tensor with the logits.
+            - hidden_state: [batch_size, hidden_size] tensor with the hidden state for the next time step.
+            - cell_state: [batch_size, hidden_size] tensor with the cell state for the next time step.
+        """
+        # Unsqueeze predicted tokens.
+        input = input.unsqueeze(1).to(torch.int)
+        output = self.relu(self.embedding(input))
```
w4nderlust, line 98: why rely?

ludwig/features/text_feature.py
```diff
@@ -392,7 +381,7 @@ def postprocess_predictions(
         backend,
     ):
         # todo: refactor to reuse SequenceOutputFeature.postprocess_predictions
-        level_idx2str = "{}_{}".format(self.level, "idx2str")
+        level_idx2str = f"{self.level}_idx2str"
```
w4nderlust, line 384: maybe idx2str and str2idx can become constants

ludwig/utils/strings_utils.py
```diff
@@ -156,19 +162,26 @@ def load_vocabulary(vocab_file):
                 line = line.split(" ")[0]
             vocabulary.append(line)
         return vocabulary
-        # return [line.strip() for line in f]
+
+
+def add_or_move_symbol(vocab_list, vocab_set, symbol, index):
```
w4nderlust, line 167: add type hints

ludwig/utils/strings_utils.py
```diff
@@ -257,22 +262,29 @@ def _get_sequence_vector(
...
+    # https://huggingface.co/docs/transformers/preprocessing
+    if tokenizer_type != "hf_tokenizer":
+        unit_indices_vector = np.append(unit_indices_vector, unit_to_id[STOP_SYMBOL])
+        unit_indices_vector = np.insert(unit_indices_vector, 0, unit_to_id[START_SYMBOL])
     return unit_indices_vector
 
 
 def build_sequence_matrix(
-    sequences,
+    sequences: pd.core.series.Series,
```
w4nderlust, line 276: @tgaddair is this always true? I think it depends on the backend

tests/integration_tests/test_experiment.py
```diff
@@ -187,14 +187,14 @@ def test_experiment_multiclass_with_class_weights(csv_filename):
 def test_experiment_multilabel_with_class_weights(csv_filename):
     # Multiple inputs, Single category output
     input_features = [category_feature(vocab_size=10)]
-    output_features = [set_feature(vocab_size=3, loss={"class_weights": [0, 0, 1, 2, 3]})]
+    output_features = [set_feature(vocab_size=3, loss={"class_weights": [0, 1, 2, 3]})]
```
w4nderlust, line 190: Not 100% sure how padding is done now in set features, as removing that 0 means there's only unk, right?

ludwig/decoders/sequence_decoders.py
justinxzhao: Relu was used in the [reference implementation](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html#simple-decoder).

I'm leaning towards keeping it since it matches the reference implementation, but also happy to take it out if there's a good reason to remove it.

The model seems to train about the same either way.

ludwig/decoders/sequence_decoders.py
justinxzhao: Relu was used in the [reference implementation](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html#simple-decoder).

I'm leaning towards keeping it since it matches the reference implementation, but also happy to take it out if there's a good reason to remove it.

The model seems to train about the same either way.

ludwig/utils/strings_utils.py
justinxzhao: Extended this to `Union[pd.core.series.Series, dd.core.series.Series]`

ludwig/features/text_feature.py
justinxzhao: Sure, happy to follow up on this in a separate PR. `idx2str` and `str2idx` are used in many files.

ludwig/utils/strings_utils.py
justinxzhao: Done.

ludwig/decoders/sequence_decoders.py
w4nderlust: If we want to keep it, I'd make the activation to use a config parameter and set None as default, what do you think?

ludwig/decoders/sequence_decoders.py
justinxzhao: Discussed offline. It looks like the weight sharing isn't present in version 0.4. I've added it back in this torch implementation, with a pointer to the Transformer paper.

Empirically, this gets us a lower loss on the twitterbots dataset, but it might be good to follow up with a PR to make this configurable.

ludwig/decoders/sequence_decoders.py
justinxzhao: Discussed offline, let's remove this completely for now as it's unlikely that the RNN can't learn from not strictly positive weights.

ludwig/decoders/sequence_decoders.py
justinxzhao: Done.

tests/integration_tests/test_experiment.py
justinxzhao: Right -- there's only UNK. As discussed offline, the padding symbol doesn't seem necessarily since we convert set and bag features to a vector of binarys/ints.

ludwig/decoders/sequence_decoders.py
justinxzhao: Discussed offline. Decided to remove the relu.

ludwig/features/sequence_feature.py
```diff
@@ -206,41 +207,25 @@ class SequenceOutputFeature(SequenceFeatureMixin, OutputFeature):
...
-    def _setup_loss(self):
-        pass
-
-    def _setup_metrics(self):
-        pass
+    def logits(self, inputs: Dict[str, torch.Tensor], target=None):
+        return self.decoder_obj(inputs, target=target)
 
-    # overrides super class OutputFeature.update_metrics() method
-    def update_metrics(self, targets, predictions):
```
ShreyaR, line 223: Do we update metrics somewhere else?

ludwig/features/sequence_feature.py
```diff
@@ -340,17 +326,7 @@ def update_config_with_metadata(output_feature, feature_metadata, *args, **kwarg
 
     @staticmethod
     def calculate_overall_stats(predictions, targets, train_set_metadata):
-        overall_stats = {}
```
ShreyaR, line 343: Also curious why we don't need this anymore.

ludwig/modules/loss_modules.py
```diff
@@ -165,6 +165,26 @@ def forward(self, preds: Tensor, target: Tensor) -> Tensor:
         return self.loss_fn(preds, target)
 
 
+@register_loss("sequence_softmax_cross_entropy", [SEQUENCE, TEXT])
```
ShreyaR, line 168: nit: There were a few other losses specific to seq decoders. Are they planned foe a separate PR?

ludwig/utils/metric_utils.py
```diff
@@ -21,25 +20,32 @@ def sequence_mask(lengths: Tensor, maxlen: Optional[int] = None, dtype=torch.boo
...
+        1D tensor of all correct predictions.
     """
-    Params:
-        targets: 2D tensor
-        preds: 2D tensor
-    """
-    truncated_preds = preds[:, : targets.shape[1]]
-    padded_truncated_preds = F.pad(truncated_preds, pad=[0, targets.shape[1] - truncated_preds.shape[1]])
-    correct_preds = padded_truncated_preds == targets
+    correct_preds = preds == targets
```
ShreyaR, line 48: nit: More for my own understanding -- can you share why targets and preds are now guaranteed to have the same shape?

ludwig/utils/strings_utils.py
```diff
@@ -29,8 +29,13 @@
 from ludwig.utils.nlp_utils import load_nlp_pipeline, process_text
 
 UNKNOWN_SYMBOL = "<UNK>"
+UNKNOWN_IDX = 3
```
ShreyaR, line 32: nit: An `Enum` for this might be more appropriate.

tests/ludwig/utils/test_metric_utils.py
```diff
@@ -0,0 +1,45 @@
+import torch
+
+from ludwig.utils import metric_utils
+
+
+def test_dynamic_partition():
```
ShreyaR, line 6: Thanks for adding these tests!

train_twitter.py
```diff
@@ -0,0 +1,101 @@
+import logging
```
ShreyaR, line 1: Not sure if you managed to keep the comments in this file in.

ShreyaR (commented):
This is amazing -- thanks for putting this together.

Left comments. Happy to chat offline about some of them if needed.

ludwig/features/base_feature.py
```diff
@@ -310,74 +346,33 @@ def calculate_overall_stats(predictions, targets, train_set_metadata):
...
-        return hidden
-
-    def output_specific_fully_connected(self, inputs, mask=None):  # feature_hidden
+        If the dependent feature's FCStack doesn't use an FCStack (num_fc_layers = 0), then we use the output feature's
+        input size.
+        """
+        input_size_with_dependencies = combiner_output_size
+        for feature_name in dependencies:
+            if output_features[feature_name].num_fc_layers:
+                input_size_with_dependencies += output_features[feature_name].fc_stack.output_shape[-1]
```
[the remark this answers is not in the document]
justinxzhao: Your understanding is completely correct.

This is a specific complexity that's only necessary for when an output feature depends on other output features, e.g.

predict Survived FROM ([
  (predict Age FROM Price, Name, Sex),
  Price, Name, Sex])

When an output feature depends on other output features, the hidden states of any dependent output features get concatenated with the combiner's output, which has implications for the current output feature's "input" size.

I've revised the documentation and variable names a bit to hopefully make this clearer in the code as well.

justinxzhao (commented):
Thanks for the detailed review! Responded to most comments.

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,235 @@
...
+        self.hidden_size = hidden_size
+        self.vocab_size = vocab_size
+        self.rnn_decoder = RNNDecoder(hidden_size, vocab_size, cell_type)
+        self.max_sequence_length = max_sequence_length
+
+    def forward(self, encoder_output_state: torch.Tensor, target: torch.Tensor):
+        batch_size = encoder_output_state.size()[0]
+
+        # Tensor to store decoder output logits.
+        logits = torch.zeros(batch_size, self.max_sequence_length, self.vocab_size)
```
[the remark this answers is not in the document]
justinxzhao: Done.

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,235 @@
...
+        self.hidden_size = hidden_size
+        self.vocab_size = vocab_size
+        self.lstm_decoder = LSTMDecoder(hidden_size, vocab_size)
+        self.max_sequence_length = max_sequence_length
+
+    def forward(self, encoder_output_state: torch.Tensor, target: torch.Tensor):
+        batch_size = encoder_output_state.size()[0]
+
+        # Tensor to store decoder output logits.
+        logits = torch.zeros(batch_size, self.max_sequence_length, self.vocab_size)
```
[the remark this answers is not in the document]
justinxzhao: Done.

ludwig/utils/strings_utils.py
justinxzhao: Done.

ludwig/decoders/sequence_decoders.py
```diff
@@ -13,18 +13,235 @@
...
+        logits = self.rnn_decoder(encoder_output_state, target)
+        return {LOGITS: logits}
+
+    def get_prediction_set(self):
+        return {LOGITS}
+
+    @property
+    def input_shape(self):
+        # Dummy implementation.
+        return torch.Size([1])
```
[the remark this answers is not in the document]
ShreyaR: Sounds good! As discussed offline, it would be good to add TODOs to track this at a later date.

ludwig/features/base_feature.py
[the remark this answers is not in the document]
ShreyaR: Thanks for the explanation and the offline discussion about the details of this function!

ShreyaR (approved):
Looks great, thanks for addressing all comments!

PR #1584, opened 2021-12-09
Adds unit test for generic encoders (test_generic_encoders.py)

dantreiman:
Adds unit tests for:
- PassthroughEncoder
- DenseEncoder

from ludwig.encoders.generic_encoders

Also adds missing type annotations for parameters in test_category_encoders.py.

### DIFF
--- a/tests/ludwig/encoders/test_generic_encoders.py
+++ b/tests/ludwig/encoders/test_generic_encoders.py
@@ -6,7 +6,7 @@ from ludwig.encoders.generic_encoders import DenseEncoder, PassthroughEncoder
 
 @pytest.mark.parametrize("input_size", [1, 2, 10])
 @pytest.mark.parametrize("categorical", [True, False])
-def test_generic_passthrough_encoder(input_size, categorical):
+def test_generic_passthrough_encoder(input_size: int, categorical: bool):
     passthrough_encoder = PassthroughEncoder(input_size)
     # Passthrough encoder allows categorical input feature (int), dense encoder's input must be float.
     if categorical:
@@ -23,7 +23,7 @@ def test_generic_passthrough_encoder(input_size, categorical):
 @pytest.mark.parametrize("input_size", [1, 2, 10])
 @pytest.mark.parametrize("num_layers", [1, 3, 6])
 @pytest.mark.parametrize("fc_size", [1, 2, 10, 256])
-def test_generic_dense_encoder(input_size, num_layers, fc_size):
+def test_generic_dense_encoder(input_size: int, num_layers: int, fc_size: int):
     dense_encoder = DenseEncoder(input_size, num_layers=num_layers, fc_size=fc_size)
     inputs = torch.rand((10, input_size))
     outputs = dense_encoder(inputs)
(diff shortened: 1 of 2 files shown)


tests/ludwig/encoders/test_generic_encoders.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.mark.parametrize("input_size", [1, 2, 10])
+@pytest.mark.parametrize("num_layers", [1, 3, 6])
+@pytest.mark.parametrize("fc_size", [1, 2, 10, 256])
+def test_generic_dense_encoder(input_size: int, num_layers: int, fc_size: int):
+    dense_encoder = DenseEncoder(input_size, num_layers=num_layers, fc_size=fc_size)
+    inputs = torch.rand((10, input_size))
+    outputs = dense_encoder(inputs)
+    assert "encoder_output" in outputs
```
justinxzhao, line 30: Does it make sense to simplify this to:

```
assert outputs["encoder_output"].shape[1:] == dense_encoder.output_shape
```

Failures due to a missing "encoder_output" on this simplified form are still quite descriptive.

justinxzhao (changes requested):

tests/ludwig/encoders/test_generic_encoders.py
dantreiman: I prefer to err on the side of explicitness and 'one claim per assertion', but thats just my style preference I could go either way.

tests/ludwig/encoders/test_generic_encoders.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+import torch
+
+from ludwig.encoders.generic_encoders import DenseEncoder, PassthroughEncoder
+
+
+@pytest.mark.parametrize("input_size", [1, 2, 10])
+@pytest.mark.parametrize("categorical", [True, False])
+def test_generic_passthrough_encoder(input_size: int, categorical: bool):
```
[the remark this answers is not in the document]
dantreiman: Thats a convention I could get behind.  On the other hand, the Ludwig repo seems to use newlines the way a writer uses paragraphs, to group related ideas together - which is what I'm doing.

I think we should propose additions to the Ludwig style guide, and this discussion can happen at that level and then be applied to all PRs:  https://ludwig-ai.github.io/ludwig-docs/developer_guide/style_guidelines_and_tests/

tests/ludwig/encoders/test_generic_encoders.py
justinxzhao: The concision of having a single assert for one behavior seems nice, but I agree that it's also less explicit. Happy to go with what you and other people think is best here.

tests/ludwig/encoders/test_generic_encoders.py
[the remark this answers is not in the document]
justinxzhao: Glad to hear that using newlines to structure tests seems like a reasonable convention!

In addition to better readability, I think it also encourages good unit test design by discouraging bad patterns where a certain phase (setup/exercise/verify) was overly complicated, like testing several different behaviors in the same test, for example.

I'll probably follow up with a PR to propose we adopt this convention. I don't think it's something we need to enforce religiously, but it's a useful framework for thinking about well-designed unit tests.

justinxzhao (approved):

PR #1587, opened 2021-12-10
Added audio and bag feature tests

anmshkmr:


tests/ludwig/features/test_audio_feature.py
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+
+@pytest.mark.parametrize("encoder", ["rnn", "stacked_cnn", "parallel_cnn", "stacked_parallel_cnn", "rnn", "cnnrnn"])
+def test_audio_input_feature(audio_config: Dict, encoder: str) -> None:
+    audio_config.update({"encoder": encoder})
+    audio_input_feature = AudioInputFeature(audio_config)
+    audio_tensor = torch.randn([BATCH_SIZE, SEQ_SIZE, AUDIO_W_SIZE], dtype=torch.float32)
+    encoder_output = audio_input_feature(audio_tensor)
+    assert encoder_output["encoder_output"].shape == (BATCH_SIZE, DEFAULT_FC_SIZE)
```
justinxzhao, line 46: The encoder output shape might be 3-dimensional depending on the encoder object's parameters like if `reduce_output=None`.

Perhaps this is better, what do you think?
`assert encoder_output["encoder_output"].shape == audio_input_feature.encoder_obj.output_shape`

anmshkmr: Yes, makes sense. Thanks.

anmshkmr: Oh, but the encoder object's shape does not take into account the batch size.

justinxzhao: Ah, right. You could do:

`assert encoder_output["encoder_output"].shape == (BATCH_SIZE, *audio_input_feature.output_shape)`

Or

`assert encoder_output["encoder_output"].shape[1:] == audio_input_feature.output_shape`

anmshkmr: Thanks, updated both tests

justinxzhao (approved):
LGTM! Thanks Animesh!

PR #1590, opened 2021-12-10
Makes SequenceReducer a LudwigModule, adds reduction_modules_test.py

dantreiman:
- Also adds doc strings to torch_utils.py
- Adds SequenceReducer max_sequence_length parameter
- Adds SequenceReducer embedding_size optional parameter

### DIFF
--- a/ludwig/modules/reduction_modules.py
+++ b/ludwig/modules/reduction_modules.py
@@ -31,11 +31,12 @@ class SequenceReducer(LudwigModule):
     A sequence is a tensor of 2 or more dimensions, where the shape is [batch size x sequence length x ...].
 
     :param reduce_mode: The reduction mode, one of {"last", "sum", "mean", "max", "concat", "attention", "none"}
-    :param max_sequence_length The maximum sequence length.
+    :param max_sequence_length The maximum sequence length.  Only used for computation of shapes - inputs passed
+                               at runtime may have a smaller sequence length.
     :param encoding_size The size of each sequence element/embedding vector, or None if input is a sequence of scalars.
     """
 
-    def __init__(self, reduce_mode: str = None, max_sequence_length: int = None, encoding_size: int = None, **kwargs):
+    def __init__(self, reduce_mode: str = None, max_sequence_length: int = 256, encoding_size: int = None, **kwargs):
         super().__init__()
         # save as private variable for debugging
         self._reduce_mode = reduce_mode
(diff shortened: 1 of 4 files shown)


ludwig/modules/reduction_modules.py
```diff
@@ -35,6 +35,22 @@ def __init__(self, reduce_mode=None, **kwargs):
     def forward(self, inputs, mask=None):
         return self._reduce_obj(inputs, mask=mask)
 
+    def infer_output_shape(self, input_shape):
```
ShreyaR, line 38: This method seems to be doing a lot of what the `output_shape` property in a `LudwigModule` does.

You could make the parent class of `SequenceReducer` a `LudwigModule`, and add `input_shape` and `output_shape` properties. The `output_shape` would have the same logic that's implemented in `infer_output_shape`, and the `input_shape` can return a dummy tensor.

ShreyaR: Additionally, can you update parts of the code that use `SequenceReducer` and then manually infer the shape of the output after reduction to use the `output_shape` property of the `SequenceReducer` instead?

E.g. here's some [code](https://github.com/ludwig-ai/ludwig/blob/389c1aa443ddc0b5be92981aa9811421d58491bd/ludwig/encoders/sequence_encoders.py#L793) that can use `SequenceReducer.output_shape`.

ludwig/utils/torch_utils.py
```diff
@@ -15,27 +15,48 @@
 
 
 def sequence_length_2D(sequence: torch.Tensor) -> torch.Tensor:
-    """Returns the number of non-zero elements per sequence."""
+    """Returns the number of non-zero elements per sequence in batch.
```
ShreyaR, line 18: Thanks for adding this!

tests/ludwig/modules/test_reduction_modules.py
```diff
@@ -0,0 +1,24 @@
...
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize("reduce_mode", ["last", "sum", "mean", "avg", "max", "concat", None])
+@pytest.mark.parametrize("test_input_shape", [(16, 1, 4), (4, 10, 16)])
+def test_sequence_reducer(reduce_mode: str, test_input_shape: Tuple[int, ...]):
+    batch_size = test_input_shape[0]
+    max_sequence_length = test_input_shape[1]
+    sequence_reducer = reduction_modules.SequenceReducer(reduce_mode=reduce_mode)
+    inputs = torch.zeros(test_input_shape)
```
ShreyaR, line 18: Can you make sure these tests are GPU compliant? Involves moving the input tensor and module to GPU if available.

E.g. [here](https://github.com/ludwig-ai/ludwig/blob/1daab8714463e63c8f4b4ba4caed3cc1ef97a195/tests/ludwig/encoders/test_h3_encoders.py)

ShreyaR (commented):
Thanks for the PR! Added comments.

ludwig/modules/reduction_modules.py
dantreiman: Haha, I missed LudwigModule._compute_output_shape.   However, the sequence reducer does not know the sequence length ahead of time, so I can’t fully specify output shape having input shape passed in somehow.

justinxzhao (approved):
LGTM! Using defaults for the additional args required by `FeedForwardAttentionReducer` seems reasonable.

ShreyaR (approved):
Looks great!

PR #1591, opened 2021-12-11
Added binary feature test

anmshkmr:


tests/ludwig/features/test_binary_feature.py
```diff
@@ -0,0 +1,26 @@
...
+import torch
+
+from ludwig.features.binary_feature import BinaryInputFeature
+
+SEQ_SIZE = 2
+BINARY_W_SIZE = 1
+
+
+@pytest.fixture(scope="module")
+def binary_config():
```
justinxzhao, line 13: Not sure if one exists, but it would be great if there was a schema we could point to as a comment.

tests/ludwig/features/test_binary_feature.py
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+
+@pytest.mark.parametrize("encoder", ["passthrough"])
+def test_binary_input_feature(binary_config: Dict, encoder: str) -> None:
+    binary_config.update({"encoder": encoder})
+    binary_input_feature = BinaryInputFeature(binary_config)
+    binary_tensor = torch.randn([SEQ_SIZE, BINARY_W_SIZE], dtype=torch.float32)
+    encoder_output = binary_input_feature(binary_tensor)
+    assert encoder_output["encoder_output"].shape[1:] == binary_input_feature.encoder_obj.output_shape
```
justinxzhao, line 26: Why not just `binary_input_feature.output_shape`? Perhaps we need to change `binary_input_feature.output_shape` to return the encoder output shape?

tests/ludwig/features/test_binary_feature.py
```diff
@@ -0,0 +1,26 @@
...
+        "name": "binary_feature",
+        "type": "binary",
+    }
+
+
+@pytest.mark.parametrize("encoder", ["passthrough"])
+def test_binary_input_feature(binary_config: Dict, encoder: str) -> None:
+    binary_config.update({"encoder": encoder})
+    binary_input_feature = BinaryInputFeature(binary_config)
+    binary_tensor = torch.randn([SEQ_SIZE, BINARY_W_SIZE], dtype=torch.float32)
```
ShreyaR, line 24: Can you also make sure that this test is GPU compliant? E.g. having a global `DEVICE` and assigning both the input tensor and the binary feature to cuda if available.

anmshkmr: Yes, thanks for reminding. I should do that for the other feature tests too. Will send out a separate PR.

justinxzhao (approved):

PR #1592, opened 2021-12-11
Fix makedirs call to path_exists to pass url

amholler:
This defect was exposed from auto_train API saving to s3 the hyperparameter search results from ray on K8s cluster.


tgaddair (approved):
Good catch!

PR #1601, opened 2021-12-16
Added timeseries feature test

anmshkmr:


tests/ludwig/features/test_timeseries_feature.py
```diff
@@ -0,0 +1,36 @@
...
+        "embedding_size": EMBEDDING_SIZE,
+        "max_sequence_length": SEQ_SIZE,
+        "fc_size": 8,
+        "state_size": 8,
+        "num_filters": 8,
+        "hidden_size": 8,
+    }
+
+
+@pytest.mark.parametrize("encoder", ["rnn", "stacked_cnn", "parallel_cnn"])
```
justinxzhao, line 30: Instead of defining a list here, could we use `tests.integration_tests.utils.ENCODERS`?

anmshkmr: Not all encoders are needed for timeseries, correct?

justinxzhao: I'm not sure. Perhaps when we revamp the documentation, we should list the encoders that timeseries features use.

justinxzhao (approved):

PR #1606, opened 2021-12-17
feat: Modify Trainer to use marshmallow_dataclass syntax for handling hyperparameters. Add basic scripting for docstring extraction to marshmallow schema. Fix some existing marshmallow issues.

ksbrar:
## Summary
Three areas that this PR covers:
* Shifts hyperparameters previously set by `init` of `Trainer` class into a `TrainerConfig` dataclass of marshmallow fields, enabling json schema style validation of these parameters. Adds similar dataclasses for all known optimizers.
* Expanding our current custom marshmallow definitions and fixing a lot of the issues that were holdovers from the original combiner PR a couple months ago. Tests are fleshed out as well.
* The `marshmallow_dataclass` package does not currently support setting metadata on the generated JSON schemas for the new dataclasses easily. In particular, to keep the `description` in these schemas up to date, this PR adds scripting to extract and apply dataclass docstrings to their corresponding schemas during generation. To make this efficient, the parsed structures of these dataclasses are actually saved as separate JSON files that will also now be distributed.

### Removes:
* `bucketing_field`, re: [discussion](https://github.com/ludwig-ai/ludwig/pull/1606#discussion_r772705787)
* All `sgd` aliases except for `sgd` are now invalid

### Adds:
* New package: `pytkdocs`, used for parsing dataclass structures.
* `scripts/extract_schema.py`
* `TrainerConfig` marshmallow dataclass to handle hyperparameters, while attributes related to device config (e.g. `skip_save_model`) are left on the `init` function.
* `BaseOptimizer` marshmallow dataclass and children that support each optimizer type in `optimization_modules`.
* `ludwig/marshmallow`, which organizes the major marshmallow-related structures.
    * `ludwig/marshmallow/test_classes.py`, which now contains (the moved) `CustomTestCombiner` class and other utilties for testing marshmallow-based classes. These test classes are moved here so that `pytkdocs` can parse inheritance relationships correctly.
    * `ludwig/marshmallow/generated`, which organizes the parsed output of `pytkdocs` into separate JSON files for each marshmallow class. These are also included in the `MANIFEST.in` file.
* docstrings for optimizer, trainer, combiner dataclasses
* docstring extraction/application logic in `schema_utils.py`
* Tests for existing marshmallow schema and utils in `tests/ludwig/marshmallow`.

### Changes:
* `optimizer_registry` is now a `Registry` rather than just a `dict`.
* Keeps `Clipper` as a separate dataclass, renamed to `GradientClippingConfig`
* Renames `clipper` param and all related uses to `gradient_clipping_config`  to be consistent. `gradient_clipping` is also the valid keyword for Ludwig YAML.
* `clip_grads` is moved out of the clipping dataclass and is now a method of `Trainer`
* Moved/split `tests/ludwig/utils/test_schema.py` into different files.
* In `base.py`, the `Trainer` is initialized using `load_config_with_kwargs`
* Subtlety: when `gradient_clipping` is set to `None` the internal trainer representation does **not** disable gradient clipping but rather uses the default values from `GradientClippingConfig()` (in particular, `clipglobalnorm` is not `None`). This is due to the follow-up case below.
* Subtlety: sets all marshmallow dataclasses (trainer, combiners, optimizers) to exclude (`marshmallow.EXCLUDE`) additional properties but helper function manually sets all schemas' `additionalProperties` attribute to True. This is to mirror current Ludwig behavior - in which you can specify any parameter and Ludwig can pick and fill in intelligently. The schema attribute is manually changed because the `marshmallow_dataclass` package does not support INCLUDE/EXCLUDE keywords (since Python dataclasses require declaring all attributes in advance).

### Follow-up PRs/Investigations:
- [ ] Add Github action for extracting schema and diffing against branch.
- [ ] Change docstrings to be less developer-focused and more end-user focused in jargon/vocab.
- [ ] Move marshmallow config classes into `ludwig/marshmallow` to streamline other files and to allow importing into `defaults.py` without circular imports
    - [ ] More generally, remove duplicate values from `defaults.py` - source of truth should now exist on the dataclasses themselves.
    - [ ] Perhaps move json-related `schema.py` methods as well.
- [ ] Flesh out remaining docstrings, marked in the code by `TODO: Document parameters.`
- [ ] Investigate why `nan`s are produced by `test_model_training_options.py/resume training` for `sgd`, if `gradient_clipping.clipglobalnorm` is None.
- [ ] Simplify or further clarify `learning_rate` hand-off flow (LR set to auto vs. LR not set at all) in `Trainer.__init__`. 
    - Somewhat odd from the user's perspective that they not specify a parameter and assume Ludwig will automatically choose its value but for `learning_rate` and certain other parameters you can explicitly set `auto`, which performs a similar behavior. Maybe choose a different word instead of `auto`?
- [ ] Investigate potentially streamlining current marshmallow definitions even further (deserialize/nesting overloading)
- [ ] Consider simplifying usage of `None` in various registries.
    - [ ] `initializer_registry` - removing this as part of this PR
    - [ ] `activations` - can't just delete `None` key here, consider replacing with `identity`
    - [ ] `reduce_mode_registry`


ksbrar:
Failures because learning_rate, batch_size, and eval_batch_size should also support keyword 'auto'

ludwig/models/trainer.py
```diff
@@ -91,40 +94,84 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.shutdown()
 
 
+@dataclass
+class TrainerConfig:
+    optimizer: Dict = schema.OptimizerOptions(default={TYPE: "adam"})
+    epochs: int = schema.PositiveInteger(default=100)
```
justinxzhao, line 100: Totally not required for this PR -- this seems like an appropriate place to document what all of the different parameters mean, at least in comments. WDYT?

ludwig/models/trainer.py
```diff
@@ -91,40 +94,84 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    reduce_learning_rate_on_plateau_rate: float = schema.FloatRange(default=0.5, min=0.0, max=1.0)
+    increase_batch_size_on_plateau: int = schema.NonNegativeInteger(default=0)
+    increase_batch_size_on_plateau_patience: int = schema.NonNegativeInteger(default=5)
+    increase_batch_size_on_plateau_rate: float = schema.FloatRange(default=2.0, min=0.0)
+    increase_batch_size_on_plateau_max: int = schema.PositiveInteger(default=512)
+    decay: bool = False
+    decay_steps: int = schema.PositiveInteger(default=10000)
+    decay_rate: float = schema.FloatRange(default=0.96, min=0.0, max=1.0)
+    staircase: bool = False
+    # "gradient_clipping": None, # TODO: What is this?
```
justinxzhao, line 126: [Gradient clipping](https://deepai.org/machine-learning-glossary-and-terms/gradient-clipping#:~:text=Gradient%20clipping%20is%20a%20technique,input%20into%20a%20specific%20output.) is the practice of forcing the gradient values (element-wise) to a specific minimum or maximum value if the gradient exceeded an expected range. 

It appears that this might have been replaced by the Clipper data class, used by `create_optimizer_with_clipper`, which is configured using the `**optimizer` config. Should we add the attributes of the `Clipper` dataclass to the schema for `OptimizerOptions`?

ludwig/models/trainer.py
```diff
@@ -91,40 +94,84 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        decay=False,
-        decay_rate=0.96,
-        decay_steps=10000,
-        staircase=False,
-        batch_size=128,
-        eval_batch_size=None,
+        # optimizer=None,
+        # epochs=100,
+        # regularization_lambda=0.0,
+        # regularization_type=None,
```
justinxzhao, line 150: Why comment these parameters out? Can we remove them?

ludwig/models/trainer.py
```diff
@@ -91,40 +94,84 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    increase_batch_size_on_plateau_max: int = schema.PositiveInteger(default=512)
+    decay: bool = False
+    decay_steps: int = schema.PositiveInteger(default=10000)
+    decay_rate: float = schema.FloatRange(default=0.96, min=0.0, max=1.0)
+    staircase: bool = False
+    # "gradient_clipping": None, # TODO: What is this?
+    # TODO: Need some more logic here for validating against output features
+    validation_field: str = COMBINED
+    validation_metric: str = LOSS
+    # "bucketing_field": None, # TODO: What is this?
```
justinxzhao, line 130: In NLP, if you have data with a high variance of sequence lengths, you can be a bit smarter about batching your examples by batching together examples with similar sequence lengths. Then, when you eventually pad examples in each batch to the max sequence length (of that batch), the total extra padding that you need for any single batch is relatively minimal.

This trick is a rather last-mile performance optimization, and I don't think it's actually used anywhere, so probably fine to remove completely, imo.

ludwig/models/trainer.py
```diff
@@ -91,40 +94,84 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        # regularization_lambda=0.0,
+        # regularization_type=None,
+        # learning_rate=0.001,
+        # decay=False,
+        # decay_rate=0.96,
+        # decay_steps=10000,
+        # staircase=False,
+        # batch_size=128,
+        # eval_batch_size=None,
         should_shuffle=True,
```
justinxzhao, line 158: Why are some of these parameters defined in the schema/conifg while others are not?

justinxzhao:
Hey Kabir, thanks for the change.

I see a couple of failures in the checks, specifically for `tests/integration_tests/test_cli.py::test_train_cli_horovod` and `tests/integration_tests/test_horovod.py::test_horovod_implicit`. Could you check them?

ludwig/models/trainer.py
ksbrar: Agreed! Going to do this in the next couple of commits.

ludwig/models/trainer.py
ksbrar: Originally was just trying to match `default_training_params` in `defaults.py` but going to move these over as well.

ludwig/models/trainer.py
ksbrar: Just keeping them around for reference, to be taken out in final refactor :P

ludwig/models/trainer.py
ksbrar: Will remove

ludwig/models/trainer.py
ksbrar: For stronger 'typing' I was actually going to make small dataclasses for each optimizer type and remove `OptimizerOptions`. 

Is `Clipper` functionality always going to be tied to optimizer configuration? (Seems to be the case currently.) If so, then yeah I think we might as well move those attributes over, but if it's worth keeping `Clipper` as a separate dataclass then I could just add a `clipperConfig` attribute to each optimizer dataclass that is `None` by default.

justinxzhao: I could go either way on whether to keep Clipper as a separate dataclass vs. copying each attribute over to each optimizer.

ksbrar:
Fixed horovod bug. New failures are because of https://github.com/ludwig-ai/ludwig/pull/1650 - now that `lr` is set for every optimizer while currently only SGD expects it.

ludwig/utils/schema_utils.py
```diff
@@ -159,6 +223,74 @@ def _jsonschema_type_mapping(self):
         }
 
 
+def IntegerOrStringOptionsField(
+    options: List[str],
+    nullable: bool,
+    default: Union[None, int],
+    isIntegeric: bool = True,
```
justinxzhao, line 230: nit: I'm not sure Integeric is a word. Maybe "is_like_integer", "is_integer_like", "is_integer"?

justinxzhao (commented):
Hi Kabir, 

Not sure if you were looking for another review, but I figured I might leave some comments since all the checks were green 🥇 

Mostly LGTM, just some minor nits around casing and wording. Seems like it's coming together nicely!

ludwig/utils/schema_utils.py
```diff
@@ -102,10 +127,49 @@ def Embed():
...
 
+def FloatRangeTupleDataclassField(N=2, default: Tuple = (0.9, 0.999), min=0, max=1):
+    if N != len(default):
+        raise ValidationError(f"Dimension of tuple '{N}' must match dimension of default val. '{default}'")
+
+    class FloatTupleMarshmallowField(fields.Tuple):
+        def _jsonschema_type_mapping(self):
+            return {
+                "type": "array",
+                "prefixItems": [
```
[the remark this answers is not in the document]
ksbrar: Nah this is a json-schema prop

ludwig/utils/schema_utils.py
```diff
@@ -102,10 +127,49 @@ def Embed():
...
+                    {
+                        "type": "number",
+                        "minimum": min,
+                        "maximum": max,
+                    }
+                ]
+                * N,
+            }
+
+    def validateRange(data: Tuple):
```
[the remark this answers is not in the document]
ksbrar: Done

ludwig/utils/schema_utils.py
ksbrar: Haha I was thinking of [this](https://www.npmjs.com/package/is-integeric), but sure - fixed

ludwig/models/trainer.py
```diff
@@ -91,40 +101,91 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    increase_batch_size_on_plateau_rate: float = schema.NonNegativeFloat(default=2.0)
+    increase_batch_size_on_plateau_max: int = schema.PositiveInteger(default=512)
+    increase_batch_size_eval_metric: str = LOSS
+    increase_batch_size_eval_split: str = TRAINING
+    decay: bool = False
+    decay_steps: int = schema.PositiveInteger(default=10000)
+    decay_rate: float = schema.FloatRange(default=0.96, min=0.0, max=1.0)
+    staircase: bool = False
+    gradient_clipping: Optional[Clipper] = ClipperDataclassField()
+    # TODO: Need some more logic here for validating against output features
```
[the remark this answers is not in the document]
ksbrar: Will do!

ksbrar:
@justinxzhao Dumb question, is the `learning_rate` in `TrainerConfig` supposed to match the one that is sent to the optimizer? Right now they can be different

justinxzhao:
> @justinxzhao Dumb question, I may have introduced an erroneous behavior. Is the `learning_rate` in `TrainerConfig` supposed to match the one that is sent to the optimizer? Right now they can be set differently.

I think per @dantreiman's [recent change](https://github.com/ludwig-ai/ludwig/pull/1650/files), the optimizer shouldn't specify it's own LR and the optimizer's LR should instead be set from "learning_rate" in the `TrainerConfig`. Does that make sense?

ludwig/models/trainer.py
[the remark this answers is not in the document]
ksbrar: https://github.com/ludwig-ai/ludwig/issues/1673

ludwig/models/trainer.py
ksbrar: clipper is being kept as a separate dataclass and is now explicitly on the trainer dataclass, `create_optimizer_with_clipper` has been updated

ksbrar:
New failures are because I added training schema to `get_schema`, and the bug there traces to optimizer schema matching multiple of the current `oneOf` JSON construction, fixing

ludwig/api.py
```diff
@@ -200,6 +200,10 @@ def __init__(
         # merge config with defaults
         self.base_config = copy.deepcopy(config_dict)
         self.config = merge_with_defaults(config_dict)
+        import json
```
ksbrar, line 203: Remove

ludwig/combiners/combiners.py
```diff
@@ -97,20 +97,44 @@ def output_shape(self) -> torch.Size:
...
     flatten_inputs: bool = False
+    """Whether to flatten input tensors to a vector (default: False)."""
+
     residual: bool = False
+    """Whether to add a residual connection to each fully connected layer block. All fully connected layers must have
+       the same size (default: False)."""
 
     class Meta:
-        unknown = INCLUDE
+        unknown = EXCLUDE
```
ksbrar, line 137: explain in description

ludwig/combiners/combiners.py
```diff
@@ -514,28 +570,63 @@ def output_shape(self) -> torch.Size:
...
 
 
+# TODO(ksbrar): Refactor transformers into using base class for common attrs?
 @dataclass
 class TransformerCombinerConfig:
     num_layers: int = schema.PositiveInteger(default=1)
+    """TODO (default: 1)."""
+
     hidden_size: int = schema.NonNegativeInteger(default=256)
+    """The size of the hidden representation within the transformer block. It is usually the same as `embedding_size`,
```
ksbrar, line 580: Does the point about `embedding_size` still make sense? This will be sent out in the schema, so does it make sense to keep this in the docstring?

ludwig/combiners/combiners.py
```diff
@@ -514,28 +570,63 @@ def output_shape(self) -> torch.Size:
...
     transformer_output_size: int = schema.NonNegativeInteger(default=256)
+    """Size of the fully connected layer after self attention in the transformer block. This is usually the same as
+       `hidden_size` and `embedding_size` (default: 256)."""
+
     dropout: float = schema.FloatRange(default=0.1, min=0, max=1)
+    """Dropout rate for the transformer block (default: 0.1)."""
+
     fc_layers: Optional[List[Dict[str, Any]]] = schema.DictList()
     num_fc_layers: int = schema.NonNegativeInteger(default=0)
+    """The number of stacked fully connected layers (only applies if `reduce_output` is not null) (default: 0)."""
```
ksbrar, line 596: In a future PR, want to add some more logic for validating conditional fields like this one

ludwig/utils/schema.py
```diff
@@ -142,12 +142,9 @@ def get_combiner_conds(combiner_types):
     for combiner_type in combiner_types:
         combiner_cls = combiner_registry[combiner_type]
         schema_cls = combiner_cls.get_schema_cls()
-        schema = marshmallow_dataclass.class_schema(schema_cls)()
-        schema_json = JSONSchema().dump(schema)
-        combiner_json = schema_json["definitions"][schema_cls.__name__]["properties"]
-
-        # TODO: add type to schema: https://github.com/lovasoa/marshmallow_dataclass/issues/62
-        combiner_cond = create_cond({"type": combiner_type}, combiner_json)
+        combiner_schema = get_custom_schema_from_marshmallow_class(schema_cls)
```
ksbrar, line 145: combiners now also have descriptions

ludwig/utils/schema_utils.py
```diff
@@ -1,29 +1,153 @@
...
+
+def create_cond(if_pred: Dict, then_pred: Dict):
+    """Returns a JSONSchema conditional for the given if-then predicates."""
+    return {
+        "if": {"properties": {k: {"const": v} for k, v in if_pred.items()}},
+        "then": {"properties": {k: v for k, v in then_pred.items()}},
+    }
+
+
+def get_fully_qualified_class_name(cls):
```
ksbrar, line 25: Worth moving this docstring stuff to a separate file?

ksbrar (commented):
Self-review

ludwig/combiners/combiners.py
ksbrar: https://github.com/ludwig-ai/ludwig/issues/1736 - add this to the comment

ludwig/combiners/combiners.py
justinxzhao: Maybe: `The number of hidden units of the TransformerStack as well as the dimension that each incoming input feature is projected to before feeding to the TransformerStack.`

ludwig/utils/schema_utils.py
justinxzhao: Keeping it in this file seems fine for now (imo).

ludwig/utils/schema_utils.py
```diff
@@ -159,6 +346,75 @@ def _jsonschema_type_mapping(self):
...
+    min: Union[None, int] = None,
+    max: Union[None, int] = None,
+    min_exclusive: Union[None, int] = None,
+    max_exclusive: Union[None, int] = None,
+):
+    is_integer = True
+    return NumericOrStringOptionsField(**locals())
+
+
+def NumericOrStringOptionsField(
```
justinxzhao, line 402: The definitions for these custom Marshmallow fields are actually quite long!

It might be useful to add small unit tests for our custom primitive-level Marshmallow validators (like `NumericOrStringOptionsField`) to concisely demonstrate what they do and don't allow, like:

```
@dataclass
class Config:
    parameter: float = schema.NumericOrStringOptionsField(
        default=0.001, min=0.0, max=1.0, options=["auto"], nullable=False
    )

Config(parameter=0.1)
with pytest.raises(ValidationError):
    Config(parameter="not")
with pytest.raises(ValidationError):
    Config(parameter="2.0")
assert Config().parameter == 0.0001
```

In addition, could we add Docstrings explaining each of the different fields and what validation behavior they control/enforce/govern?

It might be a bit tedious now, but I think we'll be happy in a few months that the different parameters of our custom validators are all documented/tested. WDYT?

ludwig/utils/schema_utils.py
```diff
@@ -1,29 +1,153 @@
...
+    `TrainerConfig`."""
+    return ".".join([cls.__module__, cls.__name__])
+
+
+def unload_schema_from_marshmallow_jsonschema_dump(mclass) -> Dict:
+    """Helper method to get a marshmallow class's direct schema without extra wrapping props."""
+    return js().dump(mclass.Schema())["definitions"][mclass.__name__]
+
+
+def get_custom_schema_from_marshmallow_class(mclass: Type[Schema]) -> Dict:
```
justinxzhao: I would suggest adding unit tests for `get_custom_schema_from_marshmallow_class()`, which seems quite complex.

ludwig/utils/schema_utils.py
ksbrar: Yup, working on unit tests right now

ludwig/utils/schema_utils.py
ksbrar: Will do

ludwig/utils/schema_utils.py
ksbrar: And yes, agreed on explaining the validators

ludwig/models/trainer.py
```diff
@@ -91,187 +101,209 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    )
+    """Learning rate specified in configuration, represents how much to scale the gradients by. If 'auto',
+       `tune_learning_rate` must be called before training to estimate the optimal learning rate. (default: 0.001)."""
+
+    batch_size: Union[int, str] = schema.IntegerOrStringOptionsField(
+        default=128, options=["auto"], nullable=False, min_exclusive=0
+    )
+    "Size of batch to pass to the model for training (default: 128)."
+
+    eval_batch_size: Union[None, int, str] = schema.IntegerOrStringOptionsField(
```
ksbrar, line 135: This is an interesting case wrt. setting defaults in the schema (it's causing the latest error), would be interested in your thoughts @tgaddair @justinxzhao.

As the field is coded now, this is what is generated as the schema:
```
{
   "oneOf":[
      {
         "type":"integer",
         "exclusiveMinimum":0
      },
      {
         "type":"string",
         "enum":[
            "auto"
         ]
      },
      {
         "type":"null"
      }
   ]
}
```
With a `oneOf` construction across different types (e.g. integer and string-enum), there is no way to set a single "default" - you have to specify defaults for each option/sub-schema and there is no way in native JSON to set a macro level choice as to what the default value of the field should be. We are fine because marshmallow handles this so e.g. trying to make a TrainerConfig() object without specifying `eval_batch_size` will correctly create an instance with that attr. set to `None`.

But should the sub-schema have defaults at all? They would more or less just be metadata for the user.

ksbrar:
Remaining reworks (comments from further reviews notwithstanding) (CC @tgaddair):
- [ ] Change Trainer/TrainerConfig so that `optimizer` cannot explicitly be set to `None`, which is a difference from current [behavior](https://github.com/ludwig-ai/ludwig/blob/a5339b988fd28fd611cdaa08b64f4c6ff0e3f229/ludwig/models/trainer.py#L100)
- [ ] Fix python 3.6 failures by either getting rid of 3.6 support (dovetailing with `jsonschema` [recently bumping its minimum version](https://github.com/Julian/jsonschema/issues/903) or using a workaround (i.e. downgrade validator from current 'latest'/upstream json spec validator to draft 7 validator, which is the minimum spec version needed for our purposes).
- [ ] Change documentation extraction from making expensive call to pytkdocs (which reads all modules) on each schema generation to relying on a static file that contains latest extracted information. Follow up this PR with a pre-commit hook PR to automatically update this file.

tests/ludwig/utils/marshmallow/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2019 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
ksbrar: Done

tests/ludwig/utils/config_validation/test_config_misc.py
```diff
@@ -133,7 +133,7 @@ def test_config_tabnet():
             "relaxation_factor": 1.5,
             "bn_virtual_bs": 512,
         },
-        "trainer": {
+        "training": {
```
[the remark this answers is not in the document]
ksbrar: Done

ludwig/modules/optimization_modules.py
```diff
@@ -12,33 +12,333 @@
...
+    def wrap(cls: BaseOptimizerConfig):
+        optimizer_registry[name] = (cls.torch_type, cls)
+        return cls
+
+    return wrap
+
+
+@dataclass
+class BaseOptimizerConfig(BaseMarshmallowConfig):
+    """Base class for optimizers. Not meant to be used directly.
```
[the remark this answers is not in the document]
ksbrar: Done

ludwig/modules/optimization_modules.py
```diff
@@ -12,33 +12,333 @@
...
-    "adamax": torch.optim.Adamax,
-    # 'ftrl': tf.keras.optimizers.Ftrl,
-    # 'nadam': tf.keras.optimizers.Nadam,
-    "rmsprop": torch.optim.RMSprop,
-}
+optimizer_registry = Registry()
 
 
+def register_optimizer(name: str):
+    def wrap(cls: BaseOptimizerConfig):
```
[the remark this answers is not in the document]
ksbrar: Done - chose latter

ludwig/modules/optimization_modules.py
```diff
@@ -12,33 +12,333 @@
...
+@dataclass
+class BaseOptimizerConfig(BaseMarshmallowConfig):
+    """Base class for optimizers. Not meant to be used directly.
+
+    The dataclass format prevents arbitrary properties from being set. Consequently, in child classes, all properties
+    from the corresponding `torch.optim.Optimizer` class are copied over: check each class to check which attributes are
+    different from the torch-specified defaults.
+    """
+
+    torch_type: ClassVar[Optional[torch.optim.Optimizer]] = None
```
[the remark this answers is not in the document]
ksbrar: Done

ludwig/modules/optimization_modules.py
```diff
@@ -12,33 +12,333 @@
...
+        optimizer_schemas_json[opt] = get_custom_schema_from_marshmallow_class(schema_cls)
+    return optimizer_schemas_json
+
+
+@dataclass
+class ClipperConfig(BaseMarshmallowConfig):
+    """Dataclass that holds gradient clipping parameters."""
+
     clipglobalnorm: Optional[float] = 0.5
+    "(default: 0.5)"
```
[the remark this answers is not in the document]
ksbrar: Fixed

ludwig/modules/optimization_modules.py
```diff
@@ -12,33 +12,333 @@
...
+    """Return a dict of strings, wherein each key is an optimizer name pointing to its stringified JSON schema."""
+    optimizer_schemas_json = {}
+    for opt in optimizer_registry:
+        schema_cls = optimizer_registry[opt][1]
+        optimizer_schemas_json[opt] = get_custom_schema_from_marshmallow_class(schema_cls)
+    return optimizer_schemas_json
+
+
+@dataclass
+class ClipperConfig(BaseMarshmallowConfig):
```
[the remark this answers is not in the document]
ksbrar: Renamed

ludwig/modules/optimization_modules.py
```diff
@@ -49,16 +349,77 @@ def clip_grads(self, variables: Iterable[torch.Tensor]):
...
+    :param model: Underlying Ludwig model
+    :param optimizer: Instance of `ludwig.modules.optimization_modules.BaseOptimizerConfig` (default:
+           `ludwig.modules.optimization_modules.SGDOptimizerConfig()`).
+    :param clipper: Instance of `ludwig.modules.optimization_modules.ClipperConfig` (default:
+           `ludwig.modules.optimization_modules.ClipperConfig(clipglobalnorm=5.0)`)
+    """
+    optimizer_cls = get_from_registry(optimizer.type.lower(), {k: v[0] for k, v in optimizer_registry.items()})
+    cls_kwargs = {k: optimizer.__dict__[k] for k in optimizer.__dict__ if k != "type"}
+    torch_optimizer: torch.optim.Optimizer = create_optimizer(optimizer_cls, model, horovod, **cls_kwargs)
+    return torch_optimizer, clipper
```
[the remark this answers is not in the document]
ksbrar: Removed create_clipper (unnecessary now), refactored this

ludwig/modules/optimization_modules.py
```diff
@@ -12,33 +12,333 @@
...
+@register_optimizer(name="gradient_descent")
+@dataclass
+class SGDOptimizerConfig(BaseOptimizerConfig):
+    """Parameters for stochastic gradient descent."""
+
+    torch_type: ClassVar[torch.optim.Optimizer] = torch.optim.SGD
+    "Points to `torch.optim.SGD`."
+
+    type: str = StringOptions(
+        ["sgd", "gd", "stochastic_gradient_descent", "gradient_descent"], default="sgd", nullable=False
```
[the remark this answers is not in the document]
ksbrar: Will wait for confirmation on this.

ludwig/modules/optimization_modules.py
```diff
@@ -49,16 +349,77 @@ def clip_grads(self, variables: Iterable[torch.Tensor]):
...
+                return None
+            if isinstance(value, dict):
+                try:
+                    return ClipperConfig.Schema().load(value)
+                except (TypeError, ValidationError):
+                    raise ValidationError(f"Invalid params for clipper: {value}, see ClipperConfig class.")
+            raise ValidationError("Field should be None or dict")
+
+        def _jsonschema_type_mapping(self):
+            return {"oneOf": [{"type": "null"}, get_custom_schema_from_marshmallow_class(ClipperConfig)]}
```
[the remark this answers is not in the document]
ksbrar: Actually, adding this back because it *does* work as expected - (i.e. conceptually saying "gradient_clipping: None" means no clipping will happen!).

ludwig/utils/schema.py
```diff
@@ -157,12 +155,11 @@ def get_custom_definitions():
...
-    }
-
-
 def validate_config(config):
-    validate(instance=config, schema=get_schema())
+    # Add support for tuples (watch this issue: https://github.com/Julian/jsonschema/issues/148):
+    def custom_is_array(checker, instance):
+        return isinstance(instance, list) or isinstance(instance, tuple)
+
+    type_checker = Draft7Validator.TYPE_CHECKER.redefine("array", custom_is_array)
```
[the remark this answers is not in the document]
ksbrar: added ref to https://github.com/ludwig-ai/ludwig/issues/1783

ludwig/utils/schema_utils.py
ksbrar: added

ludwig/models/trainer.py
ksbrar: out of date

ludwig/utils/schema_utils.py
ksbrar: added

ludwig/modules/optimization_modules.py
```diff
@@ -12,34 +12,319 @@
...
+       (default: 'adam')"""
+
+    lr: float = FloatRange(default=0.001, min=0.0, max=1.0)
+    """(default: 0.001)"""
+
+    betas: Tuple[float, float] = FloatRangeTupleDataclassField(default=(0.9, 0.999))
+    """(default: (0.9, 0.999))"""
+
+    eps: float = NonNegativeFloat(default=1e-08)
+    """(default: 1e-08)"""
```
justinxzhao, line 109: It looks like these are torch defaults: https://pytorch.org/docs/stable/generated/torch.optim.Adam.html

ksbrar:
Just noticed I accidentally took out `AdamW`: https://github.com/ludwig-ai/ludwig/commit/e01e7c2e2db35afc5ab5835924889c29905c8772 (to be fixed)

ksbrar:
One last question I have: should some of the defaults (e.g. for optimizers) be deleted from `defaults.py`? They now exist on the relevant dataclasses.

justinxzhao:
> One last question I have: should some of the defaults (e.g. for optimizers) be deleted from `defaults.py`? They can now exist on the relevant dataclasses. CC @tgaddair @justinxzhao

I think so. Going forward the dataclasses can be the source of truth for what Ludwig's defaults are.

ludwig/modules/optimization_modules.py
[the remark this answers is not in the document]
ksbrar: removed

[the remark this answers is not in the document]
justinxzhao: There's now consensus on using a single name for the SGD optimizer. @ksbrar, could you go ahead and make this change?

scripts/extract_schema.py
```diff
@@ -0,0 +1,88 @@
+# !/usr/bin/env python
```
justinxzhao, line 1: nit: Remove the copyright for new files (CC: @w4nderlust), and add docstrings to each function.

tests/ludwig/utils/test_validate_config_misc.py
```diff
@@ -366,7 +243,7 @@ def test_validate_with_preprocessing_defaults():
             vector_feature(preprocessing=VectorFeatureMixin.preprocessing_defaults()),
         ],
         "output_features": [{"name": "target", "type": "category"}],
-        "trainer": {
+        "training": {
```
justinxzhao, line 246: nit: This should be trainer.

ksbrar: fixed this and one other holdover

scripts/extract_schema.py
ksbrar: done

ludwig/modules/optimization_modules.py
ksbrar: also changed the defaults for the other optimizers, as per offline disc.

ludwig/models/trainer.py
```diff
@@ -91,183 +100,201 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        self.regularization_lambda = regularization_lambda
-        self.regularization_type = regularization_type
-        self.learning_rate = learning_rate
+        if config is None:
+            config = TrainerConfig()
+
+        self.epochs = config.epochs
+        self.regularization_lambda = config.regularization_lambda
+        self.regularization_type = config.regularization_type
+        self.learning_rate = config.learning_rate
```
ksbrar, line 268: @dantreiman Asking since you most recently touched `lr`, just wondering if appropriate for this logic to be simplified as part of this PR? In the context of the other changes, it is no longer the case that `learning_rate` can be non-`float`.

ludwig/modules/optimization_modules.py
```diff
@@ -50,20 +330,78 @@ def clip_grads(self, variables: Iterable[torch.Tensor]):
...
+        default_factory=lambda: GradientClippingConfig.Schema().load(default),
+    )
+
+
+def create_clipper(gradient_clipping_config: Optional[GradientClippingConfig]):
+    """Utility function that will convert a None-type gradient clipping config to the correct form."""
+    if isinstance(gradient_clipping_config, GradientClippingConfig):
+        return gradient_clipping_config
+    # none_dict = dict.fromkeys(asdict(GradientClippingConfig()), None)
+    # return GradientClippingConfig.Schema().load(none_dict)
```
justinxzhao, line 376: Remove comments.

tests/ludwig/utils/test_validate_config_trainer.py
```diff
@@ -0,0 +1,171 @@
+#! /usr/bin/env python
```
justinxzhao, line 1: Remove copyright.

tests/ludwig/utils/test_validate_config_combiner.py
```diff
@@ -0,0 +1,183 @@
+#! /usr/bin/env python
```
justinxzhao, line 1: Remove copyright.

tests/ludwig/marshmallow/test_fields_optimization.py
```diff
@@ -0,0 +1,156 @@
+#! /usr/bin/env python
```
justinxzhao, line 1: Remove copyright.

tests/ludwig/utils/test_validate_config_combiner.py
ksbrar: removed

tests/ludwig/utils/test_validate_config_trainer.py
ksbrar: removed

ludwig/modules/optimization_modules.py
ksbrar: removed

justinxzhao (approved):
LGTM! Thanks for all the changes. Go schema enforcement!

w4nderlust:
In the config json files i see references to specific paths in @ksbrar 's file system. i guess they shouldn't be there :)

ludwig/combiners/combiners.py
```diff
@@ -94,23 +122,52 @@ def output_shape(self) -> torch.Size:
...
+    """TODO: Document parameters. (default: 0)."""
+
     output_size: int = schema.PositiveInteger(default=256)
+    """Output size of a fully connected layer (default: 256)."""
+
     use_bias: bool = True
+    """Whether the layer uses a bias vector (default: True)."""
+
     weights_initializer: Union[str, Dict] = schema.InitializerOrDict(default="xavier_uniform")
+    """TODO: Document parameters. (default: 'xavier_uniform')."""
```
tgaddair, line 148: Can we remove TODOs from these docstrings, otherwise they will end up in generated UI elements.

ludwig/backend/horovod.py
```diff
@@ -37,7 +38,8 @@ def initialize_pytorch(self, *args, **kwargs):
         initialize_pytorch(*args, horovod=self._horovod, **kwargs)
 
     def create_trainer(self, **kwargs):
-        return Trainer(horovod=self._horovod, **kwargs)
+        config, kwargs = load_config_with_kwargs(Trainer.get_schema_cls(), kwargs)
```
tgaddair, line 41: Seems redundant to put this in the `create_trainer` Backend implementation. Can we just move this to `api.py` and pass it in as one of the kwargs?

ludwig/marshmallow/generated/AdadeltaOptimizerConfig.json
```diff
@@ -0,0 +1,685 @@
...
+                            }
+                        },
+                        "ludwig.modules.optimization_modules.AdadeltaOptimizerConfig.Schema._declared_fields": {
+                            "attributes": [],
+                            "category": "attribute",
+                            "children": {},
+                            "classes": [],
+                            "docstring": null,
+                            "docstring_sections": [],
+                            "file_path": "/home/ksbrar/work/ludwig/ludwig/modules/optimization_modules.py",
```
tgaddair, line 98: Definitely want to remove this, as Piero mentioned.

requirements.txt
```diff
@@ -19,6 +19,7 @@ jsonschema
 marshmallow
 marshmallow-jsonschema
 marshmallow-dataclass
+pytkdocs
```
tgaddair, line 22: Is this needed at runtime? I thought this was only used to generate the compiled docs json.

tgaddair (commented):
Looks great overall. I think my biggest open question is around the generated files. Do we need to store all of this information generated from pytkdocs? It seems we only need to the docstrings, right?

ludwig/combiners/combiners.py
ksbrar: I was thinking to leave them here and then manually remove TODOs in UI elements.

requirements.txt
ksbrar: It's needed for easily loading back the JSON, see `restloader` and `googleloader` in `marshmallow_schema_utils`.

ksbrar: Also it's used in `extract_schema.py`, so hypothetically if we didn't need it at runtime should I just remove it as a requirement and assume that someone running that script installs it separately?

ksbrar:
> In the config json files i see references to specific paths in @ksbrar 's file system. i guess they shouldn't be there :)

Ha, good catch. Will remove.

> Do we need to store all of this information generated from pytkdocs? It seems we only need to the docstrings, right?

I suppose we don't need to, it's the most frictionless (but yeah, most expensive) solution because it's just using `pytkdocs` to save and then load . Let me look into trying to store smaller, more customized JSON.

ludwig/marshmallow/generated/AdadeltaOptimizerConfig.json
ksbrar: removed

requirements.txt
ksbrar: removed

ludwig/backend/horovod.py
ksbrar: moved

ludwig/models/trainer.py
```diff
@@ -91,183 +100,200 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        device=None,
+        resume: float = False,
+        skip_save_model: bool = False,
+        skip_save_progress: bool = False,
+        skip_save_log: bool = False,
+        callbacks: List = None,
+        random_seed: float = default_random_seed,
+        horovod: Optional[Dict] = None,
+        device: Optional[str] = None,
+        config: Optional[TrainerConfig] = TrainerConfig(),
```
tgaddair, line 227: Let's set this to `None` so the caller must pass it explcitly.

ludwig/combiners/combiners.py
```diff
@@ -60,6 +59,35 @@ def wrap(cls):
     return wrap
 
 
+def get_combiner_jsonschema():
```
tgaddair, line 62: Let's move this into a separate file (can be done in a follow-up PR).

ludwig/models/trainer.py
ksbrar: In description follow-up list

ludwig/combiners/combiners.py
ksbrar: In description follow-up list

ksbrar:
Do not merge this yet, there are a few small json adjustments I'd like to make.

tgaddair (approved):

PR #1612, opened 2021-12-21
Update hyperopt: Choose best model from validation data; For stopped Ray Tune trials, run evaluate at search end

amholler:
Update hyperopt as discussed recently in automl community.
* Choose best model from Validation data
* For stopped Ray Tune trials, run evaluate at search end
* Set epochs appropriately when max_t is expressed in seconds
* Handle case where no trial reports any results [rare]

Details for the first item:
* Update Ludwig hyperopt operation to compute metric_score on training validation stats.
* This means that the best model will be chosen based on validation statistics, rather than on the test statistics corresponding to the best validation statistics as is done currently.
* An error will be reported if no validation set is provided.
* N.B.:
** The metric_score is constrained to stats computed during training; it cannot be drawn from
   those only computed during post-train overall stats model evaluation.  If needed, additional
   stats can be added to the training set.
** The post-train overall stats evaluation for the best model is computed on the validation set.
   An overall stats evaluation for the best model run on the test set can be performed by the
   use as a separate step after the hyperparameter optimization job is completed.

Details for the second item:
* For Ray Tune trials that are stopped before training/evaluation completes, load
and evaluate the trial's best model after the overall Ray Tune run completes.
* Currently, Ludwig calls tune.report to report a trial's intermediate results, which include
training_stats set to train_stats[TRAINING] and eval_stats set to train_stats[VALIDATION].
And when the trial's training completes normally, it then calls tune.report to report the
trial's final results, which include training_stats set to include all 3 train_stats
(train_stats[TRAINING],train_stats[VALIDATION],train_stats[TEST]) and eval_stats set to
the output of running an overall stats evaluation the trial's best model on the eval_set.
* For Ray Tune trials that are stopped before training/evaluation completes, the final
tune.report is never executed, meaning that the overall stats evaluation is not computed
and reported as eval_stats.  Also, the train_stats[TEST] is not reported in training_stats.
* This PR changes the intermediate tune.report calls so that training_stats includes all 3
train_stats (train_stats[TRAINING],train_stats[VALIDATION],train_stats[TEST]), with eval_stats
set to empty.  And when the overall Ray Tune run completes, for any stopped trials, it loads
and evaluates the trial's best model, setting eval_stats for that trial in ordered_trials,
which is returned and persisted in hyperopt_statistics.json.

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -1189,41 +1189,46 @@ class RayTuneExecutor(HyperoptExecutor):
         # results in ordered_trials which is returned & is persisted in hyperopt_statistics.json.
         for trial in temp_ordered_trials:
             if trial['eval_stats'] == '{}' and trial['training_stats'] != '{}':
-                trial_path = trial['trial_dir']
-                self._remove_partial_checkpoints(trial_path) # needed by get_best_checkpoint
-                best_model_path = analysis.get_best_checkpoint(trial_path.rstrip('/'))
-                best_model = LudwigModel.load(
-                    os.path.join(best_model_path, 'model'),
-                    backend=backend,
-                    gpus=gpus,
-                    gpu_memory_limit=gpu_memory_limit,
-                    allow_parallel_threads=allow_parallel_threads,
-                )
-                # Evaluate the best model on the eval_split
-                if best_model.config[TRAINING]['eval_batch_size']:
-                    batch_size = best_model.config[TRAINING]['eval_batch_size']
-                else:
-                    batch_size = best_model.config[TRAINING]['batch_size']
-                try:
-                    eval_stats, _, _ = best_model.evaluate(
-                        dataset=validation_set,
-                        data_format=data_format,
-                        batch_size=batch_size,
-                        output_directory=output_directory,
-                        skip_save_unprocessed_output=skip_save_unprocessed_output,
-                        skip_save_predictions=skip_save_predictions,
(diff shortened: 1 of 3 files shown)


ludwig/hyperopt/execution.py
```diff
@@ -1151,25 +1147,80 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
...
+                                skip_save_unprocessed_output=skip_save_unprocessed_output,
+                                skip_save_predictions=skip_save_predictions,
+                                skip_save_eval_stats=skip_save_eval_stats,
+                                collect_predictions=False,
+                                collect_overall_stats=True,
+                                return_type='dict',
+                                debug=debug
+                            )
+                            trial['eval_stats'] = json.dumps(eval_stats, cls=NumpyEncoder)
+                        except NotImplementedError:
```
w4nderlust, line 1202: Curious, when does this actually happen?

w4nderlust (approved):
Thanks for the @amholler !

ludwig/hyperopt/execution.py
amholler: It doesn't actually happen the way we run by default, which is to use the local backend to do batch evaluation.

If one tries to use the ray backend to do batch evaluation, control goes through this code:
    def batch_evaluation(self, model, dataset, collect_predictions=False, **kwargs):
        raise NotImplementedError(
            'Ray backend does not support batch evaluation at this time.'
        )

PR #1614, opened 2021-12-21
Make a BaseFeatureMixIn base class, with documentation, and have all feature MixIns inherit it.

justinxzhao:
This gives us centralized enforcement for missing abstract methods, as well as a common location for documentation for those methods.


ludwig/features/base_feature.py
```diff
@@ -37,6 +37,72 @@
...
+        Args:
+            column: Pandas column of values.
+            preprocessing_parameters: Preprocessing configuration for this feature.
+            backend: (Union[Backend, str]) Backend to use for feature data processing.
+        """
+        raise NotImplementedError
+
+    @abstractstaticmethod
+    def add_feature_data(
+        feature: Dict[str, Any],
```
w4nderlust, line 84: maybe, as we are making some changes, we can call this feature_cfg for increased readability? In other parts of the code the name feature refers to either an input or output feature object, so maybe whenever we refer to to a dict containing the feature info we can use the _cfg suffix?

w4nderlust (approved):

PR #1620, opened 2021-12-22
Added csv prediction file compatibility with visualize

anmshkmr:
# Code Pull Requests

Fixes issue https://github.com/ludwig-ai/ludwig/issues/1234

Provides the ability for visualize utilities to work with CSV prediction files in case of local backend

### DIFF
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -51,6 +51,8 @@ logger = logging.getLogger(__name__)
 
 _PREDICTIONS_SUFFIX = "_predictions"
 _PROBABILITIES_SUFFIX = "_probabilities"
+_CSV_SUFFIX = "csv"
+_PARQUET_SUFFIX = "parquet"
 
 
 def _vectorize_ground_truth(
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):

tgaddair:
Hey @anmshkmr, there's still a bit more we need to do to make this work. There is a difference between the CSV and Parquet files, in that the CSV format does not contain the raw probabilities vectors. Those we have traditionally stored in separate `.npy` files (see the existing `visualization` docs for details).

Here's the PR that switched from using CSV + NPY to Parquet:

https://github.com/ludwig-ai/ludwig/commit/651db586204097a595df2bbc089d50fa58278288#diff-546dea649dc80a5ffb0bfee77ee4cb8ac59f18e0092c7d3012b2ef125dff2efcR335

In order to get back the previous behavior for local mode, we still need to do a couple things:

1. Use the old serialization format (CSV + NPY) that was removed in the above commit when the backend is LOCAL.
2. When loading the CSV file, we also need to find all the NPY files and load them in as the appropriate columns in the Pandas DataFrame (to match the Parquet format that is expected).

Let me know if you have any questions.

anmshkmr:
> Hey @anmshkmr, there's still a bit more we need to do to make this work. There is a difference between the CSV and Parquet files, in that the CSV format does not contain the raw probabilities vectors. Those we have traditionally stored in separate `.npy` files (see the existing `visualization` docs for details).
> 
> Here's the PR that switched from using CSV + NPY to Parquet:
> 
> [651db58#diff-546dea649dc80a5ffb0bfee77ee4cb8ac59f18e0092c7d3012b2ef125dff2efcR335](https://github.com/ludwig-ai/ludwig/commit/651db586204097a595df2bbc089d50fa58278288#diff-546dea649dc80a5ffb0bfee77ee4cb8ac59f18e0092c7d3012b2ef125dff2efcR335)
> 
> In order to get back the previous behavior for local mode, we still need to do a couple things:
> 
> 1. Use the old serialization format (CSV + NPY) that was removed in the above commit when the backend is LOCAL.
> 2. When loading the CSV file, we also need to find all the NPY files and load them in as the appropriate columns in the Pandas DataFrame (to match the Parquet format that is expected).
> 
> Let me know if you have any questions.

Thanks for pointing out. I will create another PR.

camaya7:
Hi @anmshkmr, Ludwig previously presented results from `ludwig predict` with a final` *.csv ` predictions file. It seems you were trying to achieve the same CSV results from` ludwig predict` before stepping into visualization in the discussion above. Is the output of a CSV prediction file still a supported feature? 

I've been tinkering with the PR @tgaddair posted but only produced sloppy results. Any help here is much appreciated.

PR #1622, opened 2021-12-22
Add random_seed to auto_train API to improve repeatability

amholler:
As titled.  Cross-port from tf-legacy.


justinxzhao (approved):
LGTM!

PR #1627, opened 2021-12-27
Keep search_alg type in dict to record in hyperopt_statistics.json

amholler:
As titled.  Crossport from tf-legacy.


justinxzhao (approved):
Thanks Anne!

PR #1638, opened 2022-01-02
Avoid warning spew in is_image when field contains str that is not filename

amholler:


ludwig/utils/image_utils.py
```diff
@@ -69,10 +69,13 @@ def get_image_from_path(
         return img_entry
     if src_path or os.path.isabs(img_entry):
         return get_abs_path(src_path, img_entry)
-    with open_file(img_entry, "rb") as f:
-        if ret_bytes:
-            return f.read()
-        return f
+    if os.path.exists(img_entry):
```
tgaddair, line 72: May want to use `ludwig.utils.fs_utils.exists` to support remote filesystems.

amholler: Done, thanks!

tgaddair (approved):

PR #1642, opened 2022-01-04
Use tempfile to automatically garbage collect data and modeling artifacts in ludwig integration tests.

justinxzhao:
When running integration tests, there's often leftover generated data or modeling files that stick around in your workspace when the test fails.

Try/finally works, but a more pythonic design would be to use Python's tempfile context manager to generate a temporary directory that will always get cleaned up at the end of the test.

`pytest` has a built in `tmpdir` fixture, so we can just use that.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,70 +19,30 @@ import uuid
 import pytest
 
 from ludwig.hyperopt.run import hyperopt
-from ludwig.utils.data_utils import replace_file_extension
 from tests.integration_tests.utils import category_feature, generate_data, text_feature
-
-
-@pytest.fixture(scope="session", autouse=True)
-def init_tensorflow_cpu(request):
-    """Initialize tensorflow at the start of testing to only use CPUs.
-
-    This fixture runs once before any tests, and ensures that the main process
-    running pytest does not claim any GPU resources.
-
-    This is critical to avoid OOM errors when running subprocesses that need GPUs (e.g., hyperopt),
-    as otherwise the main process will consume all the memory and cause the subprocesses to crash.
-
-    Run most tests eagerly as the cost of graph construction can easily increase runtime by
-    and order of magnitude for small tests. Tests that execute in subprocesses, and tests
-    in `test_graph_execution.py` still run in graph mode.
-    """
-    # tf.config.experimental_run_functions_eagerly(True)
-    # initialize_tensorflow(gpus=-1)
-    pass
 
 
 @pytest.fixture()
 def csv_filename():
-    """This methods returns a random filename for the tests to use for generating temporary data.
-
-    After the data is used, all the temporary data is deleted.
-    :return: None
-    """
+    """Yields a csv filename for holding temporary data."""
     with tempfile.TemporaryDirectory() as tmpdir:
         csv_filename = os.path.join(tmpdir, uuid.uuid4().hex[:10].upper() + ".csv")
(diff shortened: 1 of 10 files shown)


tests/integration_tests/test_experiment.py
```diff
@@ -252,66 +253,64 @@ def test_experiment_multiple_seq_seq(csv_filename, output_features):
...
-        # import image from file and store in dataframe as tensors.
-        df = pd.read_csv(rel_path)
-        image_feature_name = input_features[0]["name"]
-        df[image_feature_name] = df[image_feature_name].apply(lambda x: torchvision.io.read_image(x))
-
-        run_experiment(input_features, output_features, dataset=df, skip_save_processed_input=skip_save_processed_input)
-
-    # Delete the temporary data created
-    shutil.rmtree(image_dest_folder, ignore_errors=True)
+    with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 256: `tmpdir` can also be a fixture you pass in as a param to the test function, so you don't need to use a context manager here.

tests/integration_tests/test_server.py
```diff
@@ -175,15 +180,12 @@ def test_server_integration_with_images(csv_filename):
...
     assert model_output == server_response
 
-    # Cleanup
-    shutil.rmtree(output_dir, ignore_errors=True)
-    shutil.rmtree(image_dest_folder, ignore_errors=True)
-
 
 @pytest.mark.parametrize("single_record", [False, True])
 def test_server_integration_with_audio(single_record, csv_filename):
+    tmpdir = os.path.dirname(csv_filename)
```
tgaddair, line 186: Another option here is to replace `csv_filename` with `tmpdir` and then let `csv_filename = os.path.join(tmpdir, 'dataset.csv`). Would be consistent with other parts of the code, but not a big deal.

tgaddair (commented):
Nice! Much needed changes. One thing I would suggest is to replace context managers with the `tmpdir` fixture to simplify things.

tests/integration_tests/test_server.py
justinxzhao: Done.

tests/integration_tests/test_experiment.py
justinxzhao: Thanks for the suggestion! Done.

tests/conftest.py
```diff
@@ -19,72 +19,32 @@
...
+    """Yields a csv filename for holding temporary data."""
     with tempfile.TemporaryDirectory() as tmpdir:
         csv_filename = os.path.join(tmpdir, uuid.uuid4().hex[:10].upper() + ".csv")
         yield csv_filename
 
 
 @pytest.fixture()
-def yaml_filename():
-    """This methods returns a random filename for the tests to use for generating a config file.
+def tmpdir():
```
tgaddair, line 34: This fixture already exists natively in pytest, just need to use it (no imports etc. needed):

https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

tgaddair (approved):
LGTM! Don't need to implement the `tmpdir` fixture, though, it already exists.

tests/conftest.py
justinxzhao: Good to know! Very sensible that this is a built-in fixture of pytest.

PR #1643, opened 2022-01-05
Fixed KeyError while creating default config

ShreyaR:


jeffreyftang (approved):

ludwig/automl/base_config.py
```diff
@@ -118,6 +118,9 @@ def _create_default_config(
     input_and_output_feature_config = get_features_config(
         dataset_info.fields, dataset_info.row_count, resources, target_name
     )
+    # create set of all feature types appearing in the dataset
+    feature_types = [[feat["type"] for feat in features] for features in input_and_output_feature_config.values()]
+    feature_types = set(sum(feature_types, []))
 
     model_configs = {}
```
justinxzhao, line 125: Does it make sense to make this a defaultdict instead?

ShreyaR: @justinxzhao not sure if that would make sense. `feature_types` is used as the set of distinct feature types present in the dataset, so it wouldn't make sense for it to be in a map.

PR #1650, opened 2022-01-06
Removes lr from default_optimizer_params_registry

### LINKED ISSUE #1621: Inconsistent learning rates shown in output when specifying learning rate in config
When specifying the learning_rate parameter in the training section of the config, `description.json` shows a different default learning rate in the optimizer section:

config.yaml:
Screen Shot 2021-12-21 at 4 46 42 PM

description.json:
Screen Shot 2021-12-21 at 4 47 04 PM

Expected behavior is that there is one learning rate, and if it is specified in the config it should match that.

dantreiman:
Fixes #1621

Removes `training.optimizer.lr` defaults from `default_optimizer_params_registry`.

Use `training.learning_rate` as default value for`training.optimizer.lr` instead.

`training.learning_rate` will always be present, either from config or `default_training_params`
```
default_training_params = {
    "learning_rate": 0.001,
```

### DIFF
--- a/ludwig/utils/defaults.py
+++ b/ludwig/utils/defaults.py
@@ -247,6 +247,7 @@ def merge_with_defaults(config):
 
     # ===== Training Optimizer =====
     optimizer = config[TRAINING]["optimizer"]
+    set_default_value(optimizer, 'lr', config[TRAINING]["learning_rate"])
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):
LGTM

PR #1652, opened 2022-01-06
Remove sampled softmax loss.

justinxzhao:
Tensorflow has built-in support for [sampled softmax loss](https://www.tensorflow.org/api_docs/python/tf/nn/sampled_softmax_loss), but Pytorch doesn’t, which means that in Pytorch, we’d have to implement, test, and maintain our own implementation. There are reference Pytorch implementations that we can adapt, but it might be more strategic to wait until either:

1) Pytorch natively supports it, or
2) we know for certain it’s going to make a substantial impact for an imminent client use case.

@w4nderlust and I discussed this briefly offline, re-sharing here:

- The fact that sampled softmax loss isn’t built into Pytorch yet, nor used for any huggingface model, might be an indication that perhaps this loss technique isn’t as necessary as it used to be for models with large vocabulary sizes, perhaps due to better availability of compute.
- Practical experience tells us that sampled softmax loss speeds things up things substantially for 1000+ classes and allows for much larger batch sizes. That said, the industry practice seems to be moving towards smaller batch sizes (and increase them by increasing cluster sizes).
- We can revisit this decision when model speed / batch size starts becoming the dominant limitation / bottleneck.


anmshkmr:
Thanks for the detailed description. Helpful!

anmshkmr (approved):

PR #1656, opened 2022-01-07
Downgrade input and output shape from properties to regular base class methods.

justinxzhao:


justinxzhao:
Closing this in favor of leaving shapes as properties, but extending shapes to return formalized tuples (dict) of tensor sizes instead of single tensor sizes.

This extension and formalization will make the upfront input/output declarative paradigm fit with ludwig modules that have multi-tensor inputs or outputs, which currently have dummy shape implementations to bypass enforcement by torchscript.

PR #1659, opened 2022-01-07
Reset MeanMetric's running average on reset()

dantreiman:
MeanMetric instances do not reset their running average when the .reset() method is called.

This is causing the epoch eval loss to show a cumulative average instead of the average at each epoch.

This chart shows training set metrics.  Note the step loss vs epoch loss:

148465005-38b4f7f5-40b2-41ed-82a1-77049d15e631

After the change, epoch loss average is reset properly, and epoch loss vs batch loss match up:
Screen Shot 2022-01-07 at 11 42 35 AM

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -961,7 +961,7 @@ class Trainer(BaseTrainer):
 
             # training step loop
             progress_bar = tqdm(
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):
Thanks for the change!

PR #1660, opened 2022-01-07
Improve Ludwig resilience to Ray Tune issues

amholler:
*) A trial can intermittently fail in Ray Tune, e.g., https://github.com/ray-project/ray/issues/21458
    In general, it seems more resilient to specify the option to have Ray Tune retry failed trials once.
    
*) Ray Tune can intermittently fail to upload checkpoints, e.g., https://github.com/ray-project/ray/issues/21469
    In general, it seems more resilient to have the Ludwig post-search evaluation process warn on & skip
    trials with missing checkpoints, to provide value for the completed search.  Note that unless
    the checkpoints from the best trial are missing, the lack of other checkpoints is not critical.

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -799,6 +799,64 @@ class RayTuneExecutor(HyperoptExecutor):
                 # Remove checkpoint marker on incomplete directory
                 os.remove(marker_path)
 
+    def _get_best_model_path(self, trial_path, analysis):
+        sync_info = self._get_sync_client_and_remote_checkpoint_dir(Path(trial_path))
+        if sync_info is not None:
+            sync_client, remote_checkpoint_dir = sync_info
+            sync_client.sync_down(remote_checkpoint_dir, trial_path)
+            sync_client.wait()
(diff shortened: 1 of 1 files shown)


ludwig/hyperopt/execution.py
```diff
@@ -1178,37 +1179,42 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
...
-                                batch_size=batch_size,
-                                output_directory=trial_path,
-                                skip_save_unprocessed_output=skip_save_unprocessed_output,
-                                skip_save_predictions=skip_save_predictions,
-                                skip_save_eval_stats=skip_save_eval_stats,
-                                collect_predictions=False,
-                                collect_overall_stats=True,
-                                return_type='dict',
-                                debug=debug
+                        if best_model_path is not None:
```
justinxzhao, line 1182: Not totally necessary for this PR, but the level of nesting here is a bit intimating. It might be worth abstracting out some of this logic into helper methods, or adding comments to describe stages of the overall flow.

amholler: Good call out, thanks!  PR updated.

justinxzhao (approved):

Issue #1613, opened 2021-12-21
Training differences between PyTorch and Tensorflow versions of Ludwig


dantreiman:
Train a ludwig model using concat combiner on Titanic, Forest Cover, or Mushroom Edibility.

The loss curves as well as final results are significantly different.

Desired behavior:
Given the same dataset and config, models trained with TF version of Ludwig vs. Torch may not produce identical results but they should be very similar.

For example, here are the results of running `examples/titanic/multiple_model_training.py` on both Ludwig 0.4 and Ludwig 0.5dev (e065e03c63c6f43a229a2af92ac4cfedaf2b569e)

Ludwig 0.4
![loss_titanic_ludwig0 4](https://user-images.githubusercontent.com/687280/146861212-51c32e4a-78ea-43bc-abc2-8cf37aa97590.png)
![accuracy_titanic_ludwig0 4](https://user-images.githubusercontent.com/687280/146861418-cdef8717-0f27-4070-8ae9-8a08a84a1c90.png)

Ludwig 0.5dev
![loss_titanic_luswig0 5dev](https://user-images.githubusercontent.com/687280/146861222-6df5bc18-5ed5-494d-af6f-4ac689390958.png)
![accuracy_titanic_ludwig0 5dev](https://user-images.githubusercontent.com/687280/146861431-cf940c29-370c-4c47-93b4-dec011cdf97a.png)

w4nderlust:
@dantreiman One thing I would make sure to confirm is first of all the optimizer. Please doublecheck that the same optimizer, with the same parameters is used (also doublecheck the parameters themselves, as sometimes they may be one the flip of the other, like dropout -> dropout_rate). If big differences still persist, I would try using a regular sgd optimizer, in my experience Adam behaves VERY differently in TF and PT (in TF it converges super fast but is really unstable and bouncy, in PT it seems more stable and slower to converge, ut it's just anecdotal). Finally, I would also doublecheck initializations, if they happen correctly for all modules in PT and if the ranges and distributions of the weights after initialization are roughly similar.

w4nderlust:
Anyway, results look different, in particular model1 training, but it's difficult o set a bar really. I'd say the goals here should be to:
1. make sure there are no errors in model building, loss computation (for instance regularization and other special losses like the ones in tablet), optimization and initialization
2. make sure we are not losing performance
3. make sure that in the case where we do, maybe wiggling a bit parameters and config compensates it

[...]

PR #1663, opened 2022-01-09
TabNet training issues

dantreiman:
- Fixes loss metrics using MeanMetric not resetting.  Descendants of MeanMetric (any LossMetric) were not getting reset between epoch, and thus were reporting a global cumulative average, not an epoch average.

- Changes LudwigModule's `_callable_losses` property to just `_losses`, and changes `add_loss` to `update_loss`.  This is only used for one thing: TabNet's Sparsity loss term.  In tensorflow, `add_loss` made sense because the call operator is only really called once to build the graph, as opposed to Torch's `forward` method which is called on every forward pass.  So, now we update the sparsity loss on every forward pass and include it in the overall train loss.

- Sets model into evaluation mode when generating predictions  using `model.train()` and `model.eval()`.  TorchModule's `training` property controls whether batch norm updates its running statistics, and whether drop out drops channels.  If we generate predictions while `training` is True, the predictions will be wrong if dropout or batch norm are used.

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -144,6 +144,8 @@ class Predictor(BasePredictor):
             predictions[key] = torch.cat(pred_value_list, dim=0).clone().detach().cpu().numpy()
 
     def batch_evaluation(self, dataset, collect_predictions=False, dataset_name=None):
+        model_training_mode = self.model.training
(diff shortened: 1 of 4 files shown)


dantreiman:
Should fix #1613

tgaddair (approved):

PR #1668, opened 2022-01-10
Fix MeanMetric reset to fix GPU device placement

ShreyaR:


tgaddair:
I suspect the issue is here: https://github.com/ludwig-ai/ludwig/pull/1663/files#diff-0cb21d063f67f426e91ce2cbdec1043e46add7b733a995c0a9616d1cea0f9e36L150

This tensor is now disconnected and will not be placed on gpu with the rest of the model.

Alternatively, _losses would probably need to be a ModuleDict: https://github.com/ludwig-ai/ludwig/pull/1663/files#diff-66a93229b58d1cc81e738d4472f50350428dac9e46d16b6c8752d662eae2a46bR131

ludwig/modules/metric_modules.py
```diff
@@ -163,7 +163,7 @@ def compute(self) -> Tensor:
 
     def reset(self):
         super().reset()
-        self.avg = _MeanMetric()
+        self.avg.reset()
```
tgaddair, line 166: @dantreiman any reason you recreated the `MeanMetric` instance here instead of calling `reset()`? The problem appears to be that the new instance is not placed on the correct device. Calling reset here should (I hope) have the same effect while preserving the device placement, but wanted to double-check with you.

justinxzhao (approved):

ShreyaR:
Retriggering checks because of some failures in the last round of tests. The failures were unrelated to GPUs or the metrics change, and were caused by an HTTP timeout error.

PR #1674, opened 2022-01-13
Added torch.no_grad() during eval to prevent OOM

tgaddair:
Context here: https://discuss.pytorch.org/t/out-of-memory-error-during-evaluation-but-training-works-fine/12274/11

Two changes incorporated from that threads:

- Use `torch.no_grad()` during eval to prevent gradient accumulation
- Pull training loop out into a separate function to prevent holding onto references

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -738,93 +738,7 @@ class Trainer(BaseTrainer):
                 self.callback(lambda c: c.on_epoch_start(self, progress_tracker, save_path))
 
                 # training step loop
-                while not batcher.last_batch():
-                    self.callback(lambda c: c.on_batch_start(self, progress_tracker, save_path))
-
-                    # Set learning rate for this batch
-                    current_learning_rate = progress_tracker.learning_rate
-
-                    if self.decay:
-                        current_learning_rate = exponential_decay(
-                            current_learning_rate,
-                            self.decay_rate,
-                            self.decay_steps,
-                            progress_tracker.steps,
-                            self.staircase,
-                        )
-
-                    if self.horovod:
-                        current_learning_rate = (
-                            learning_rate_warmup_distributed(
-                                current_learning_rate,
-                                progress_tracker.epoch,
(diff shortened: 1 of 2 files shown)


ludwig/models/trainer.py
```diff
@@ -738,93 +738,7 @@ def train(self, training_set, validation_set=None, test_set=None, save_path="mod
...
-                    if self.is_coordinator():
-                        progress_bar.update(1)
-                        logger.debug(
-                            f"training: completed batch {progress_bar.n} "
-                            f"memory used: "
-                            f"{psutil.Process(os.getpid()).memory_info()[0] / 1e6:0.2f}MB"
-                        )
-
-                    self.callback(lambda c: c.on_batch_end(self, progress_tracker, save_path))
+                self._train_loop(batcher, progress_tracker, save_path, train_summary_writer, progress_bar)
```
justinxzhao, line 741: At some point, we may want to re-think the epoch-based training-eval loops as we begin to work with large datasets (O(10) or O(100) million examples).

A nice setup, in my experience, was:
1. The trainer outputs checkpoints every few thousand steps/batches to a checkpoint directory.
2. A separate continuous eval process (binary?) waits for new files in the checkpoint directory.
3. When a new checkpoint gets written, the eval process runs eval on the eval set. This can be configured to either:
    - predict on a fixed number of examples (or the entire eval set), output metrics, wait for / grab the latest checkpoint, repeat
    - predict as many batches as possible until the next checkpoint gets written, output metrics, repeat

    The latter is nice for having eval metrics available for every checkpoint, but the former is better for comparing across modeling runs.

4. In either mode, at the end of training, metrics are computed over the entire eval set using the final checkpoint.

@w4nderlust @ShreyaR FYI

justinxzhao (approved):
Thanks for the fix! LGTM, with minor comments/questions.

ludwig/models/trainer.py
w4nderlust: Yes makes sense, let's do a brainstorming session on this after v0.5

PR #1690, opened 2022-01-18
Add WMT15 (text -> text) as a kaggle dataset, with sample modeling code.

justinxzhao:
Also, tweak print_boxed statements.

### DIFF
--- a/examples/wmt15/config_large.yaml
+++ b/examples/wmt15/config_large.yaml
@@ -7,3 +7,4 @@ input_features:
 output_features:
   - name: fr
     type: text
+    tokenizer: french_tokenize
--- a/examples/wmt15/config_large.yaml
+++ b/examples/wmt15/config_large.yaml
@@ -1,0 +1,9 @@
+input_features:
+  - name: en
+    type: text
+    encoder: bert
+    pretrained_model_name_or_path: bert-base-uncased
+
+output_features:
+  - name: fr
+    type: text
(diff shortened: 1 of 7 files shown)


examples/wmt15/config_large.yaml
```diff
@@ -0,0 +1,9 @@
+input_features:
+  - name: en
+    type: text
+    encoder: bert
```
w4nderlust, line 4: what about using GPT2 if it's computationally comparable? Just so that we have a different encoder example

examples/wmt15/config_large.yaml
```diff
@@ -0,0 +1,9 @@
...
+    pretrained_model_name_or_path: bert-base-uncased
+
+output_features:
+  - name: fr
+    type: text
```
w4nderlust, line 9: What about tokenization? we could either tokenize with space for French or with a workpiece tokenizer similar to the one used for the transformer encoder

ludwig/datasets/wmt15/config.yaml
```diff
@@ -0,0 +1,5 @@
+version: 2.0
```
w4nderlust, line 1: Any specific reason for 2.0?

examples/wmt15/config_large.yaml
justinxzhao: Done.

examples/wmt15/config_large.yaml
justinxzhao: `french_tokenize` seems like a reasonable one to use.

ludwig/datasets/wmt15/config.yaml
justinxzhao: Nope. Reverted.

w4nderlust (approved):

PR #1691, opened 2022-01-19
For ray, move all tensors back to CPU before returning to the head node.

justinxzhao:
This addresses an error encountered when training text (with HF) to category models on Ray: `(BackendExecutor pid=None) RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.`

Also, set `recursive=True` for deleting the cache. Preprocessed data in the form of Parquet files may be directories rather than individual files.

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -215,7 +215,6 @@ class RayTrainerV2(BaseTrainer):
 
         # load state dict back into the model
         state_dict, *args = results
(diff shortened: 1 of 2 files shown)


ShreyaR (approved):
looks good!

tgaddair (approved):

PR #1693, opened 2022-01-19
Added end-to-end Torchscript compilation for tabular types

tgaddair:
The scripted module takes in a `Dict[str, Union[List[str], Tensor]]` as input.

More specifically, for every input feature, we provide either a Tensor of `batch_size` inputs or a list of strings `batch_size` in length.

Note that the dimensions of all Tensors and lengths of all lists must match.

Similarly, the output will be a dictionary of dictionaries, where each feature has its own dictionary of outputs. The outputs will be a list of strings for predictions with string types, while other outputs will be tensors of varying dimensions for probabilities, logits, etc.

Example:

```
inputs = {
    'cat_feature': ['foo', 'bar']
    'num_feature': torch.tensor([42, 7])
    'bin_feature1': torch.tensor([True, False])
    'bin_feature2': ['No', 'Yes']
}

scripted_model = model.to_torchscript()
output = scripted_model(inputs)
```


ludwig/api.py
```diff
@@ -1438,26 +1439,10 @@ def save_config(self, save_path: str) -> None:
...
-        # Example usage
-
-        ```python
-        ludwig_model.save_for_serving(save_path)
-        ```
-        """
+    def to_torchscript(self):
         self._check_initialization()
-        self.model.save_savedmodel(save_path)
+        inf_model = InferenceModule(self.model, self.config, self.training_set_metadata)
```
justinxzhao, line 1444: nit: Could we use "inference_module" here? Minor preference for spelled out variable names.

ludwig/features/base_feature.py
```diff
@@ -103,6 +103,14 @@ def add_feature_data(
         raise NotImplementedError
 
 
+class PredictModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.predictions_key = PREDICTIONS
```
justinxzhao, line 109: What's the reasoning behind using `self.predictions_key` instead of referencing constants directly?

It looks like all this base class is doing is storing common keys. If we instead reference global constants directly, then perhaps this PredictModule isn't needed at all, or is this required by torchscript?

ludwig/features/binary_feature.py
```diff
@@ -53,6 +53,63 @@
...
+        predictions = preds[self.predictions_key]
+        if self.bool2str is not None:
+            predictions = predictions.to(dtype=torch.int32)
+            predictions = [self.bool2str.get(pred, self.bool2str[0]) for pred in predictions]
+
+        probs = preds[self.probabilities_key]
+        probs = torch.dstack(1 - probs, probs)
+
+        return {
+            self.predictions_key: predictions,
```
justinxzhao, line 92: Similar question here. It seems simpler to reference global constants directly instead of first setting them as class attributes, i.e.:

```
return {
    PREDICTIONS: predictions,
    PROBABILITIES: probs,
}
```

tests/integration_tests/test_torchscript.py
```diff
@@ -190,3 +193,100 @@ def test_torchscript(csv_filename, should_load_model):
...
+    )
+
+    # Obtain predictions from Python model
+    preds_dict, _ = ludwig_model.predict(dataset=training_data_csv_path, return_type=dict)
+
+    # Create graph inference model (Torchscript) from trained Ludwig model.
+    script_module = ludwig_model.to_torchscript()
+
+    def to_input(s: pd.Series) -> Union[List[str], torch.Tensor]:
+        if s.dtype == "object":
```
justinxzhao, line 262: This looks like a non-trivial conversion. Is this something that users of end-to-end torchscript would need to run on their data?

tests/integration_tests/test_torchscript.py
```diff
@@ -190,3 +193,100 @@ def test_torchscript(csv_filename, should_load_model):
...
+    # TODO: these are the only outputs we provide from Torchscript for now
+    ts_outputs = {PREDICTIONS, PROBABILITIES, LOGITS}
+
+    # Compare results from Python trained model against Torchscript
+    for feature_name, feature_outputs_expected in preds_dict.items():
+        assert feature_name in outputs
+
+        feature_outputs = outputs[feature_name]
+        for output_name, output_values_expected in feature_outputs_expected.items():
+            if output_name not in ts_outputs:
```
justinxzhao, line 279: What other output tensor names are returned by `ludwig_model.predict` that aren't returned by torchscript? Shouldn't they be the same?

ludwig/api.py
```diff
@@ -1438,26 +1439,10 @@ def save_config(self, save_path: str) -> None:
...
-
-        :return: `None`
-
-        # Example usage
-
-        ```python
-        ludwig_model.save_for_serving(save_path)
-        ```
-        """
+    def to_torchscript(self):
```
justinxzhao, line 1442: It might be worth adding a comment to clarify that this `to_torchscript()` includes preprocessing and post processing.

justinxzhao (commented):
Thanks for the PR Travis. Looks pretty good -- just a few minor nits and questions!

ludwig/features/set_feature.py
```diff
@@ -169,11 +184,10 @@ class SetOutputFeature(SetFeatureMixin, OutputFeature):
     default_validation_metric = JACCARD
 
     def __init__(self, feature, output_features: Dict[str, OutputFeature]):
-        super().__init__(feature, output_features)
-
         self.num_classes = 0
         self.threshold = 0.5
 
+        super().__init__(feature, output_features)
```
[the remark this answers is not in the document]
tgaddair: This is because we create the PredictModule (which uses these values) in the parent class' constructor. So if these values aren't set before then, it will blow up.

ludwig/features/numerical_feature.py
```diff
@@ -132,6 +171,51 @@ def fit_transform_params(column: np.ndarray, backend: "Backend") -> dict:  # noq
...
+
+
+class _NumericalPostprocessing(torch.nn.Module):
+    def __init__(self, metadata: Dict[str, Any]):
+        super().__init__()
+        self.numeric_transformer = get_transformer(metadata, metadata["preprocessing"])
+        self.predictions_key = PREDICTIONS
+
+    def forward(self, preds: Dict[str, torch.Tensor]) -> Dict[str, Any]:
+        return {self.predictions_key: self.numeric_transformer.inverse_transform_inference(preds[self.predictions_key])}
```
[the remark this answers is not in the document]
tgaddair: Discussed over Zoom. This is a good practice in general, but I don't feel comfortable having a single function that essentially does two completely different things based on a flag. If we were to go this route, I would rather break the transform and inverse transform into separate modules, but this may require some more design first.

ludwig/features/base_feature.py
tgaddair: Without setting these constants to properties, Torchscript will complain during scripting. It seems scripting is unable to pull in global variables.

tgaddair: Will add a comment.

tests/integration_tests/test_torchscript.py
tgaddair: If they want to do inference on a Pandas dataframe, yes. However, I'm not sure that's a very common scenario. More likely they would have a JSON structure or similar. We can definitely revisit if this turns out to be a useful helper function.

tests/integration_tests/test_torchscript.py
tgaddair: Yes, we should try to support all outputs. This is just a starting point from which we can expand in the future.

ludwig/api.py
tgaddair: Added docstring.

ludwig/models/inference.py
```diff
@@ -0,0 +1,60 @@
...
+
+        model.cpu()
+        self.model = model.to_torchscript()
+
+        input_features = {
+            feature[NAME]: get_from_registry(feature[TYPE], input_type_registry) for feature in config["input_features"]
+        }
+        self.preproc_modules = nn.ModuleDict(
+            {
+                feature_name: feature.create_preproc_module(training_set_metadata[feature_name])
```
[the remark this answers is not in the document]
tgaddair: Good catch!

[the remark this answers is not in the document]
tgaddair: Looks like we're going to need to make some changes to LudwigFeatureDict to make to scriptable.

ludwig/features/set_feature.py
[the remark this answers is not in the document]
justinxzhao: Ah, I see. `create_predict_module()` is called by the super constructor, with `self.threshold`. I suppose `self.num_classes` could technically be initialized after the call to the super constructor, but it's not crucial.

ludwig/features/numerical_feature.py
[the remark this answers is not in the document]
justinxzhao: +1. Having separate functions seems totally sufficient and is quite clear.

justinxzhao (approved):

PR #1694, opened 2022-01-19
Removes skimage and imageio dependencies replacing imsave with torchvision.utils.save_image

w4nderlust:


ludwig/datasets/mnist/__init__.py
```diff
@@ -47,6 +47,18 @@ class Mnist(CSVLoadMixin, GZipDownloadMixin, BaseDataset):
     processed_dataset_path: str
 
     def __init__(self, cache_dir=DEFAULT_CACHE_LOCATION):
+        try:
```
ShreyaR, line 50: Curious about importing `torchvision` during initialization of this class, instead of importing it directly with the other imports and import guarding it?

ShreyaR (approved):
lgtm! Left a small comment

ludwig/datasets/mnist/__init__.py
w4nderlust: the point is that if you import it directly you won0t get the message with what to do to fix the issue if the image dependencies are not installed, you would just get an import error

tgaddair: This is true, but calling `sys.exit` is not very good for the user, as it could, for example, crash their IPython kernel. Better to raise a new `ImportError` with the more descriptive error message.

w4nderlust: That's a good point, although the same pattern is used in other parts of the code where we do this safeguarded import, so maybe we can have another PR that replaces the system exit with import errors all around.

PR #1696, opened 2022-01-19
Require that there must be 2 non-fill distinct values for binary features.

justinxzhao:
Fill values may appear as a distinct value if the original binary column contained missing values. This can create a situation where there are 3 distinct values, where multiple elements map to a single boolean value, i.e. "T", "F", and "False" (filled).

We now permit 3 distinct values, if one of them is the fill value, while ensuring that the elements used for `bool2str` postprocessing list use non-fill values.

Table describing the [implemented behavior](https://docs.google.com/spreadsheets/d/1EAJCDX1qLlToRYzsMesmgASsZDfwmXYyeBDbDdHGSvQ/edit#gid=0).

Verified that IEEE successfully trains with this change.

A potentially cleaner solution would be to set the default true label and fill_value explicitly during automl config generation. This would be good to revisit when extending automl to preprocessing in automl 2.0.

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -215,6 +215,7 @@ class RayTrainerV2(BaseTrainer):
 
         # load state dict back into the model
         state_dict, *args = results
+        # print(f"state_dict: {state_dict}")
         self.model.load_state_dict(state_dict)
         results = (self.model, *args)
 
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -215,7 +215,6 @@ class RayTrainerV2(BaseTrainer):
 
         # load state dict back into the model
         state_dict, *args = results
-        # print(f"state_dict: {state_dict}")
         self.model.load_state_dict(state_dict)
         results = (self.model, *args)
 
--- a/ludwig/features/binary_feature.py
+++ b/ludwig/features/binary_feature.py
@@ -62,7 +62,7 @@ class BinaryFeatureMixin(BaseFeatureMixin):
     def preprocessing_defaults() -> Dict[str, Any]:
         return {
             "missing_value_strategy": FILL_WITH_CONST,
-            "fill_value": 0,
+            "fill_value": False,  # Default False.
         }
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


tgaddair (changes requested):
Not sure how I feel about allowing the col to have multiple conflicting values for true / false. Feels like something that should be cleaned up in the data outside Ludwig, but curious to get others' thoughts (cc @w4nderlust).

In particular, I worry about what this means for postprocessing (`bool2str`). For example, with this change, let's imagine that the user has only two distinct values in the col: `F` and `T`. However, they have some missing values that get coerced to `False`. What does this do to `bool2str`? I worry that it could result in getting outputs that were not in the original data, or even getting True outputs postprocessed into a value for False (note that bool2str is a list, so this does seem possible).

My suggestion would be to keep the 2 distinct values rule, but change the default `fill_value` to `None`. Then if we detect that `fill_value` is `None` here, we set it to be the element of `distinct_values` that corresponds to the "false" value.

What do you think?

w4nderlust:
I agree with the potential problems for `bool2str`, and I prefer the strategy of default fill value set to `None` which means figure out what is the 0/False from the data distribution, while if the user overrides it with a specific value, then we respect that.

justinxzhao:
One challenge with dynamically choosing `fill_value` is that we currently `handle_missing_values` before computing `str2bool` (computed in `get_feature_metadata`). See [the full preprocessing flow](https://github.com/ludwig-ai/ludwig/blob/4a92ac3f28d40c82a94856b9493ce36990dc9435/ludwig/data/preprocessing.py#L1146-L1159).

The code there is cleanly feature-agnostic so it'd be nice to not tinker with it.

As long as the `False` element of `bool2str` is the element of `distinct_values` that corresponds to the "false" value, then post-processing should be good to go.

I'll also add back in the requirement that there should be strictly two (non-fill) distinct values, which means we won't accommodate cases like `["t", "T", "True", "FALSE"]`.

Here's a [table summarizing](https://docs.google.com/spreadsheets/d/1EAJCDX1qLlToRYzsMesmgASsZDfwmXYyeBDbDdHGSvQ/edit#gid=0) the final binary feature preprocessing behavior that I'm imagining.

tgaddair:
@justinxzhao you can also do this at the automl level. So when we detect the feature type (from the distinct values) we also set the fill value.

justinxzhao:
I like the idea of solving this at the automl level. Right now, this would require extending automl to cover setting preprocessing parameters, which would need some restructuring (in scope for automl 2.0). I've included this idea in the PR description.

w4nderlust:
The table summary looks good to me.

For the future (v0.6 focus on preprocessing) we can revisit this though, and potentially add a `column_analysis` step for all features that happens in preprocessing before filling missing values and before computing metadata. During this step we can do a bunch of additional cleaning, like for instance dropping outliers, or feature engineering that creates new deviated columns, and, like in this case, figuring out a god fill value could be part of this additional step. We need some more brainstorming to design a great solution here, for the time being following your table summary is the way to go :)

justinxzhao:
Went with an alternative solution, https://github.com/ludwig-ai/ludwig/pull/1698 - closing.

PR #1699, opened 2022-01-21
Added pretrained_kwargs to BertEncoder

connor-mccorm:
Added pretrained_kwargs constructor variable. Then added same variable to .from_pretrained() which will capture kwargs if present.


tgaddair:
Thanks @connor-mccorm, can you add this to all the huggingface text encoders?

connor-mccorm:
Absolutely @tgaddair, I’ll get right on it!

PR #1712, opened 2022-01-25
Learning rate fix: don't copy learning rate to optimizer.lr in defaults.py

dantreiman:
In https://github.com/ludwig-ai/ludwig/pull/1650, I changed defaults.py to copy `training.learning_rate` to `training.optimizer.lr` by default.

However, `training.learning_rate` isn't always a number or None - it also may be set to 'auto'.  This causes a type mismatch error as 'auto' gets copied to `optimizer.lr`

This change removes the copy in defaults.py, and instead delay's setting optimizer lr until trainer's `set_learning_rate`  is called.


justinxzhao (approved):
Thanks for the change!

w4nderlust:
The approach seems correct. I would check other places where we change/set the learning rate during training and make sure they all make the right thing. Also I believe some of the failing tests are due to the use of the auto string, let's make sure that is taken into account too (probably making sure that when auto is used, the actual values passed around are the numerical ones the auto process figures out)

dantreiman:
Ah yes, the test is failing in tune_batch_size (which happens before tune_learning_rate, so if `batch_size: auto` and `learning_rate`: auto this fails.  Hmmmmm.  I see why the config learning rate and the optimizer learning rate must be fully separate, I can't set them jointly in the same setter method.

w4nderlust:
@dantreiman  I'm trying to understand why `target_learning_rate`. To me it seems that it's the value that is set at the beginning, while the one computed every epoch is `current_learning_rate`, so maybe `initial_learning_rate` or `starting_learing_rate` could be better names? Or am I missing something?

w4nderlust:
also, as this code touches something she worked on, @ANarayan could you also take a look (maybe after ICML if you are submitting)?

dantreiman:
@w4nderlust I converted this back to a draft PR, I'm still experimenting with this.

In the autoML case:
- `trainer.learning_rate` is the learning rate specified in the config (which might be 'auto')
- `trainer.target_learning_rate` is a number, and acts like a control point (end point of LR warmup, start point of LR decay)
- `trainer.set_optimizer_learning_rate` sets the learning rate during training, which varies from epoch to epoch

In the case where the config specifies a learning rate, `trainer.learning_rate` and `trainer.target_learning_rate` are identical.

I separated the parameters this way primarily to make it easier to debug.

w4nderlust:
> @w4nderlust I converted this back to a draft PR, I'm still experimenting with this.
> 
> * `trainer.learning_rate` is the learning rate specified in the config (which might be 'auto')
> * `trainer.target_learning_rate` is a number, and acts like a control point (end point of LR warmup, start point of LR decay)
> * `trainer.set_optimizer_learning_rate` sets the learning rate during training, which varies from epoch to epoch
> 
> If the config specifies a learning rate, `trainer.learning_rate` and `trainer.target_learning_rate` are identical.
> 
> I separated the parameters this way primarily to make debugging easier.

This makes sense. to me, and I understand the logic.
The name target still doesn't convince me, as, for instance, if both warmup and decay are on, you may actually never hit that learning rate (already decayed by one epoch when warmup ends).

Maybe an alternative name could be `computed_learning_rate` as in computed by the auto lr process (and it is the same of the `learning_rate` if there's no computation involved).

I think finding a good clear name is important, otherwise with all these learning rates we could get confused a month from now :)

curious of @tgaddair @ShreyaR @justinxzhao and @jimthompson5802 's opinion here.

ludwig/models/trainer.py
```diff
@@ -155,9 +155,9 @@ def __init__(
         :param epochs: Number of epochs the algorithm is intended to be run over
         :type epochs: Integer
         :param learning_rate: Learning rate for the algorithm, represents how
-               much to scale the gradients by
-        :type learning_rate: Integer
-        :param batch_size: Size of batch to pass to the model for training.
+               much to scale the gradients by.  May be 'auto' to auto-tune learning rate.
```
justinxzhao, line 158: Move/copy the comment: "The learning_rate specified in configuration" here.

Does `auto` specify to auto-tune the learning rate, per-epoch? Or to choose a default `target_learning_rate`, or both?

ludwig/models/trainer.py
```diff
@@ -241,7 +241,13 @@ def __init__(
         self.epochs = epochs
         self.regularization_lambda = regularization_lambda
         self.regularization_type = regularization_type
-        self.learning_rate = learning_rate
+        self.learning_rate = learning_rate  # The learning_rate specified in configuration, may be 'auto'.
+        try:
+            target_learning_rate = float(learning_rate)
+        except ValueError:
+            # TODO (ASN): Circle back on how we want to set default placeholder value
+            target_learning_rate = 0.001  # Default initial learning rate for autoML.
```
justinxzhao, line 249: Is this a general default or is this the default for automl's "auto"?

justinxzhao (commented):
The separation of `target_learning_rate` and `current_learning_rate` seems good to me.

Happy with `target_learning_rate` as a name. One alternative to consider would be `peak_learning_rate` to convey that the learning rate will never exceed this value.

Just a couple of questions.

ludwig/models/trainer.py
w4nderlust: Either case we should probably place it somewhere in defaults.py and import it

w4nderlust:
Peak may not be correct as we may have other schedulers in the future, like oscillating ones or in general schedulers that may end up increasing the learning rate.

Another option I'm thinking is `base` because it's the learning rate before any scheduler modification. Wdyt?

dantreiman:
I like this `base_learning_rate` wording better actually.

ludwig/models/trainer.py
dantreiman: This constant was in `tune_learning_rate()` before, I just moved it up here.  It happens to be the same value as `default_training_params.learning_rate` in `defaults.py`, but has different semantics.

dantreiman:
I'm going to merge this to get the fix in, but I think there is some more work to do here.  Opened a new issue https://github.com/ludwig-ai/ludwig/issues/1720 to discuss refactoring and design.

PR #1714, opened 2022-01-25
Added eval mode to model prediction

ShreyaR:


tgaddair (approved):
LGTM, but is it necessary to set back to training at the end? Seems like this is something we can just handle in the Trainer.

PR #1715, opened 2022-01-26
Unpin dask[dataframe] and fix error reading unreachable image urls

hungcs:
- Original issue with ray has been resolved so can unpin dask.
- in `tf-legacy`, trying to read from an unreachable url would cause an `HTTPError`. In torch, this is a `RuntimeError`, so change catch statement in `read_image_from_str`


justinxzhao:
Can you remove `from urllib.error import HTTPError` in `image_utils`? Pre-commit is complaining that it's imported but not used.

justinxzhao (approved):

PR #1724, opened 2022-01-28
Build ray gpu image with latest ray and cuda 11.1 version of PyTorch

tgaddair:


docker/ludwig-ray-gpu/Dockerfile
```diff
@@ -23,8 +22,8 @@ RUN sudo apt-get update && DEBIAN_FRONTEND="noninteractive" sudo apt-get install
     rsync \
     vim
 
-# Upgrade to the latest ray
-RUN pip install --upgrade ray
+# Install PyTorch with CUDA 11.1 (compatible with 11.2 image)
+RUN pip install torch==1.10.1+cu111 -f https://download.pytorch.org/whl/torch_stable.html
```
noyoshi, line 26: It looks like this would pin us to the ray version specified by the base image, whereas before we were always using the latest ray version. Is this intended?

noyoshi (approved):
lgtm

docker/ludwig-ray-gpu/Dockerfile
tgaddair: In general, I prefer using the version of ray built into the parent image instead of overriding it (the one in the image has some assurance that it plays nicely with all the other dependencies the image brings in). The only reason we were doing what we were doing before was that we needed to pin to an older image to pull in the correct version of CUDA, then upgrade ray to avoid using the old version. This is no longer necessary.

PR #1726, opened 2022-01-28
Squeeze extra dimensions explicitly when preparing input symbols for the next time step.

justinxzhao:
From [torch.squeeze](https://pytorch.org/docs/stable/generated/torch.squeeze.html): 

> If the tensor has a batch dimension of size 1, then squeeze(input) will also remove the batch dimension, which can lead to unexpected errors.

When the `batch_size = 1`, we would pass in a 0-dim tensor as the next time step. This causes an error: `IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)`.

The solution is to explicitly squeeze out the extra multi-layer and vocabulary dimensions with `tensor.squeeze(1).squeeze(1)` instead of `tensor.squeeze()` which squeezes `[batch_size, 1, 1]` -> `[]`.


tgaddair (approved):
Nice catch!

justinxzhao:
It looks like rebasing + force push automatically closed this PR.

PR #1731, opened 2022-01-31
Upgrade Docker Python 3.7 -> 3.8

hungcs:


docker/ludwig/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #   model serving
 #
 
-FROM python:3.7.12-slim
+FROM python:3.8.12-slim
```
tgaddair, line 12: We should check what the latest patch version is on DockerHub for 3.8

tgaddair (commented):
We should also upgrade the other Docker images to use 3.8.

docker/ludwig/Dockerfile
hungcs: seems like latest is 3.8.12

https://hub.docker.com/_/python

docker/ludwig-dev/Dockerfile
```diff
@@ -10,6 +10,7 @@
 #   unit testing
 #
 
+FROM python:3.8.12-slim
```
tgaddair, line 13: Hmm, I forgot about this image. Looks like we need to update the parent on line 15 to match `ludwig-gpu` (really the whole image should match except line 33, which should indeed be using test).

docker/ludwig-gpu/Dockerfile
```diff
@@ -9,6 +9,8 @@
 #   model serving
 #
 
+FROM python:3.8.12-slim
```
tgaddair, line 12: We should see what version of Python is used by the parent pytorch image.

docker/ludwig-ray-gpu/Dockerfile
```diff
@@ -9,6 +9,8 @@
 #   model serving
 #
 
+FROM python:3.8.12-slim
```
tgaddair, line 12: Just need to change parent to py38 below.

docker/ludwig-ray/Dockerfile
```diff
@@ -9,6 +9,8 @@
 #   model serving
 #
 
+FROM python:3.8.12-slim
```
tgaddair, line 12: Should make this match with the ray gpu image, so something like `rayproject/ray:1.9.2-py38` probably.

tgaddair (commented):
In general, I don't think adding the additional parent image will have the desired effect. What we need to do is update the existing parent images to ones that support python 3.8.

tgaddair (changes requested):
Changes requested regarding comments above.

docker/ludwig-ray/Dockerfile
hungcs: this can be deleted, since we have `FROM rayproject/ray:latest` below right?

docker/ludwig-gpu/Dockerfile
hungcs: pytorch docker is using `Python 3.7.11`, does this block us from upgrading?

PR #1737, opened 2022-02-03
Fix TypeError: cannot assign 'int' object to buffer 'pos_weight' (torch Tensor or None required)

justinxzhao:


dantreiman (approved):
LGTM!

PR #1742, opened 2022-02-04
Changed --config_file to --config

tgaddair:
Reasoning here is that for most real applications of Ludwig, it's not practical to pass the config as a JSON string, so the most common usage (a YAML file) should be the simpler and more intuitive form of the parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ You can use the following config:
 and start the training typing the following command in your console:
 
 ```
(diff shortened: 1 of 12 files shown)


w4nderlust:
Are there other places where this pattern applies? Like the preprocess command on the synthesize data command? if so let's make sure there's consistency

tgaddair:
Yes, good catch. Added the preprocessing config.

PR #1743, opened 2022-02-05
Renamed "training" to "trainer" and "numerical" to "number"

tgaddair:


tests/ludwig/utils/test_schema.py
```diff
@@ -134,7 +135,7 @@ def test_config_tabnet():
             "use_keras_batch_norm": False,
             "bn_virtual_bs": 512,
         },
-        "training": {
+        TRAINER: {
```
w4nderlust, line 138: maybe here and in a couple other instances within the text-schema module like this we may want to replace with "trainer" instead of using the constant

tgaddair: Updated.

PR #1748, opened 2022-02-11
[crossport from tf-legacy] Address missing batch_size specification if autoML is used for text dataset

amholler:


justinxzhao (approved):
LGTM!

PR #1751, opened 2022-02-12
Address issues when AutoML tune_for_memory parameter set true

amholler:
Previous AutoML validation testing has been done with its tune_for_memory parameter set to false.

Running AutoML validation workloads with tune_for_memory set true exposed several
minor but blocking problems that are addressed in this PR.  Will cross-port to master.


ludwig/automl/auto_tune_config.py
```diff
@@ -125,7 +126,7 @@ def memory_tune_config(config, dataset):
     current_param_values = {}
     max_memory = get_machine_memory()
 
-    while param_list is not None:
+    while param_list is not None and len(param_list) > 0:
```
justinxzhao, line 129: Can this be simplified to `while param_list`?

justinxzhao (approved):
Thanks for the changes Anne. Just one small comment.

PR #1755, opened 2022-02-16
Use percentage cutoff of distinct values to assign categorical type instead of an arbitrary number of values.

justinxzhao:
Verified that this change fixes typing assignment discrepancies for KDD and Mercedes Benz.

### DIFF
--- a/ludwig/automl/base_config.py
+++ b/ludwig/automl/base_config.py
@@ -43,8 +43,8 @@ encoder_defaults = {"text": {"bert": os.path.join(CONFIG_DIR, "text/bert_config.
 # The highest percentage of distinct values that we might still assign a CATEGORY type.
 CATEGORY_TYPE_DISTINCT_VALUE_PERCENTAGE_CUTOFF = 0.5
 
-# Set >= SMALL_DISTINCT_COUNT & >= MAX_DISTINCT_BOOL_PERMUTATIONS
-MAX_DISTINCT_VALUES_TO_RETURN = 100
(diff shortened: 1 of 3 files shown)


ludwig/automl/base_config.py
```diff
@@ -40,10 +40,12 @@
...
 
-SMALL_DISTINCT_COUNT = 20
+# For a given feature, the highest percentage of distinct values out of the total number of rows that we might still
+# assign the CATEGORY type.
+CATEGORY_TYPE_DISTINCT_VALUE_PERCENTAGE_CUTOFF = 0.5
 
-# Set >= SMALL_DISTINCT_COUNT & >= MAX_DISTINCT_BOOL_PERMUTATIONS
-MAX_DISTINCT_VALUES_TO_RETURN = 100
+# Cap for number of distinct values to return.
+MAX_DISTINCT_VALUES_TO_RETURN = 10000
```
amholler, line 48: This is much higher than the current value.  I recall @tgaddair being sensitive to this from the scaling standpoint; is this new number okay from that standpoint?

amholler (commented):
Overall looks good!  One question on the increase in distinct values being returned.

ludwig/automl/base_config.py
justinxzhao: Reverted back to the original value as discussed. As long as `FieldInfo.num_distinct_values` contains the true, non-capped number of distinct values, then we can simply use that.

amholler:
Looks great!

Issue #1749, opened 2022-02-11
Ray integration tests are flaky / timing out with higher frequency.


justinxzhao:
`tests/integration_tests/test_ray*` tests seem to be timing out more recently.

It'll be worth diving into this more, as the flaky tests are adding noise to PR checks.

justinxzhao:
Ray's [testing tips](https://docs.ray.io/en/latest/auto_examples/testing-tips.html) offer some nice hints that are worth exploring, namely:

1. `ray.init(local_mode=True)` if possible. This seems like a nice option.
2. If local mode isn't sufficient, we can share the ray cluster across tests. We currently start a new ray cluster for each test, which is probably most of the overhead.

PR #1757, opened 2022-02-16
Class imbalance

connor-mccorm:
Raising this PR to test in a non-local env.


connor-mccorm:
Ready for review @ShreyaR. Not sure why test_ray_tabular and test_ray_audio are failing in the unit tests since they are passing for me when I spin up a ray cluster for testing. Let me know if these failures are expected.

justinxzhao:
The failing ray tests are likely not related to your PR. These tests have been timing out/flaky for a hot second. See #1749.

tests/ludwig/utils/test_class_balancing.py
```diff
@@ -0,0 +1,178 @@
...
+        input_df = backend.df_engine.from_pandas(input_df)
+        test_df = balance_data(input_df, input_features, preprocessing_parameters, backend)
+
+        new_class_balance = round(
+            test_df[target].value_counts().compute()[test_df[target].value_counts().compute().idxmin()]
+            / test_df[target].value_counts().compute()[test_df[target].value_counts().compute().idxmax()],
+            1,
+        )
+
+        assert input_df.compute().shape == (200, 6)
```
justinxzhao, line 74: I would only keep the last assert on the new class balance, which is the single behavior you are trying to verify. The other asserts like about dataset size aren't relevant.

tests/ludwig/utils/test_class_balancing.py
```diff
@@ -0,0 +1,178 @@
...
+        assert input_df[input_df["Label"] == 1].compute().shape == (20, 6)
+        assert input_df[input_df["Label"] == 0].compute().shape == (180, 6)
+        assert test_df.compute().shape == (270, 6)
+        assert isinstance(backend, RayBackend)
+        assert new_class_balance == 0.5
```
justinxzhao, line 79: Could we pass in `new_class_balance` as a variable to this test helper functions and add additional test cases for class balance percentages other than `0.5`.

justinxzhao (commented):
Hey Connor, thanks for the change.

My main comments are around the use of `output_flag` and if we can simplify things by passing in the config directly, as well as some thoughts around unit testing.

ludwig/data/dataframe/dask.py
```diff
@@ -61,6 +61,9 @@ def parallelize(self, data):
     def persist(self, data):
         return data.persist() if self._persist else data
 
+    def concatenate(self, dfs):
```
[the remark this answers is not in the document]
connor-mccorm: I agree, keeping consistent with the df_lib naming is probably best to avoid confusion.

ludwig/data/preprocessing.py
```diff
@@ -1205,6 +1208,43 @@ def build_data(
     return proc_cols
 
 
+def balance_data(dataset_df, features, preprocessing_parameters, backend):
```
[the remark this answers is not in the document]
connor-mccorm: Absolutely!

ludwig/data/preprocessing.py
```diff
@@ -1205,6 +1208,43 @@ def build_data(
...
+    if backend.df_engine.partitioned:
+        majority_class = backend.df_engine.compute(dataset_df[target[PROC_COLUMN]].value_counts()).idxmax()
+        minority_class = backend.df_engine.compute(dataset_df[target[PROC_COLUMN]].value_counts()).idxmin()
+    else:
+        majority_class = dataset_df[target[PROC_COLUMN]].value_counts().idxmax()
+        minority_class = dataset_df[target[PROC_COLUMN]].value_counts().idxmin()
+    majority_df = dataset_df[dataset_df[target[PROC_COLUMN]] == majority_class]
+    minority_df = dataset_df[dataset_df[target[PROC_COLUMN]] == minority_class]
+
+    if preprocessing_parameters["oversample_minority"]:
```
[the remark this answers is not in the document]
connor-mccorm: I felt the same way, but interestingly enough, I read through the sampling post travis put in the ml channel last night and someone mentioned a point about using undersampling for both cost reduction purposes and also to combat overfitting since oversampling too much could cause some overfitting. I agree that 95-99% of the time the users will want oversampling, but I think it's nice to have in there in the off chance someone wants to use it for those reasons. WDYT?

tests/ludwig/utils/test_class_balancing.py
connor-mccorm: Sounds good, I'll make the change!

tests/ludwig/utils/test_class_balancing.py
```diff
@@ -0,0 +1,178 @@
...
+    assert input_df[input_df["Label"] == 0].shape == (180, 5)
+    assert test_df.shape == (270, 5)
+    assert new_class_balance == 0.5
+    assert isinstance(backend, LocalBackend)
+
+
+@pytest.mark.parametrize("balance", ["oversample_minority", "undersample_majority"])
+@pytest.mark.distributed
+def test_balance_data_ray(balance):
+    config = CONFIGS["test_config_1"].copy()
```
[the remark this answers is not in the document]
connor-mccorm: That makes sense, I'll make the change!

tests/ludwig/utils/test_class_balancing.py
connor-mccorm: Sounds good, I'll test for 0.25 and 0.75 as well

ludwig/constants.py
```diff
@@ -128,6 +128,7 @@
 NAME = "name"
 COLUMN = "column"
 TYPE = "type"
+OUTPUT_FLAG = "output_flag"
```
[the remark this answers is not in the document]
connor-mccorm: I agree that it's better to get the output class straight from the config instead of looping through the features, I guess my initial thought was to change as little pre-existing code as possible so I tried to just add the flag so I could pick it up from the flattened feature list. If it's ok to pipe the config into all the preprocessors so I can access it at the `_preprocess_df_for_training` and `_preprocess_file_for_training` functions then I'll go ahead and make that change!

tests/integration_tests/test_class_imbalance_feature.py
```diff
@@ -0,0 +1,189 @@
...
+import pandas as pd
+import pytest
+import ray
+
+from ludwig.api import LudwigModel
+from ludwig.backend import LocalBackend
+from ludwig.backend.ray import RayBackend
+from tests.integration_tests.utils import create_data_set_to_use, spawn
+
+RAY_BACKEND_CONFIG = {
```
justinxzhao, line 16: My sense is that we can probably remove ray backends from all of these tests since class imbalance should work independent of which backend we're using, and there's also no ray-specific configuration parameters.

Ray tests are slower and the additional overhead for setting up makes these tests harder to read.

Using local backends here will keep the tests simple, and if we wanted to test the ray+imbalance combination, we should do so in `test_ray.py`.

examples/class_imbalance/credit_fraud.py
```diff
@@ -0,0 +1,71 @@
+fraud_inputs = [
+    {"name": "Time", "type": "numerical", "output_flag": False},
```
justinxzhao, line 2: Remove the `output_flag` from these input maps.

Also, where/how is this file used?

tests/ludwig/utils/test_class_balancing.py
```diff
@@ -0,0 +1,200 @@
...
+    with ray_start(num_cpus=num_cpus, num_gpus=num_gpus):
+        backend = create_ray_backend()
+        input_df = backend.df_engine.from_pandas(input_df)
+        test_df = balance_data(input_df, config["output_features"], config["preprocessing"], backend)
+
+        majority_class = test_df[target].value_counts().compute()[test_df[target].value_counts().compute().idxmax()]
+        minority_class = test_df[target].value_counts().compute()[test_df[target].value_counts().compute().idxmin()]
+        new_class_balance = round(minority_class / majority_class, 2)
+
+        assert (target_balance - 0.02) <= new_class_balance <= (target_balance + 0.02)
```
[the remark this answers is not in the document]
connor-mccorm: Good idea, will do!

examples/class_imbalance/credit_fraud.py
connor-mccorm: Having this file in here is a mistake, it's actually a file that was going to be used in the example with the credit card fraud dataset. I accidentally committed it so I can remove in the next commit.

ludwig/data/preprocessing.py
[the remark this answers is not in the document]
connor-mccorm: Great idea, I've added in that check!

ludwig/data/preprocessing.py
```diff
@@ -1205,6 +1241,55 @@ def build_data(
...
+        sample_fraction = (len(majority_df) * preprocessing_parameters["oversample_minority"]) / len(minority_df)
+        minority_df = minority_df.sample(frac=sample_fraction, replace=True)
+    elif preprocessing_parameters["undersample_majority"]:
+        sample_fraction = int(len(minority_df) / preprocessing_parameters["undersample_majority"]) / len(majority_df)
+        majority_df = majority_df.sample(frac=sample_fraction, replace=False)
+
+    if backend.df_engine.partitioned:
+        balanced_df = backend.df_engine.concat([minority_df, majority_df])
+    else:
+        balanced_df = pd.concat([minority_df, majority_df])
```
w4nderlust, line 1288: Wouldn't `backend.df_engine.concat` work even if the df engine was not partitioned?

connor-mccorm: Yes actually it would. I think what happened here was I had named the backend concat function for dask concatenate which Justin pointed out was inconsistent so I changed it to concat, not realizing that it parallels the local backend setup. I'll change that now.

ludwig/data/preprocessing.py
```diff
@@ -1282,7 +1367,7 @@ def get_split(
             if backend.df_engine.partitioned:
                 # This approach is very inefficient for partitioned backends, which
                 # can split by partition
-                return None
+                return
```
w4nderlust, line 1370: This suggests that we are not creating a split at all in this case?

connor-mccorm: So what happens here is the split for dask dataframes happens at a higher up level since splitting dask dataframes in the get_split() function is very slow because you have to call set_index() which is an expensive dask method. So i decided to pull the function out of the build_dataset function altogether and just operate on the training split of the dataset after its been defined in the _preprocess_df_for_training and _preprocess_file_for_training functions. The reason there is a change here at all was because I had made a splitting function for dask dataframes initially, but then deleted it all after I realized how slow it would be but forgot to put back the None in the return. I can add it back in now.

tests/integration_tests/test_class_imbalance_feature.py
```diff
@@ -0,0 +1,70 @@
...
+        skip_save_progress=True,
+        skip_save_processed_input=True,
+        skip_save_training_description=True,
+        skip_save_training_statistics=True,
+    )
+
+    input_train_set = input_df.sample(frac=0.7, replace=False)
+    processed_len = output_dataset[0].size
+    processed_target_pos = sum(output_dataset[0].dataset["Label_mZFLky"])
+    processed_target_neg = len(output_dataset[0].dataset["Label_mZFLky"]) - processed_target_pos
```
w4nderlust, line 28: Replace `"Label_mZFLky"` with a programamtic way to obtain that string. Like `model.config.output_feature[0][PROC_COLUMN]` or something like that

tests/integration_tests/test_class_imbalance_feature.py
```diff
@@ -0,0 +1,70 @@
...
+@pytest.mark.parametrize(
+    "balance",
+    ["oversample_minority", "undersample_majority"],
+)
+def test_imbalance(balance):
+    config = {
+        "input_features": [
+            {"name": "Index", "column": "Index", "type": "numerical"},
+            {"name": "random_1", "column": "random_1", "type": "numerical"},
+            {"name": "random_2", "column": "random_2", "type": "numerical"},
```
w4nderlust, line 54: numerical -> number

tests/ludwig/utils/test_class_balancing.py
```diff
@@ -0,0 +1,124 @@
...
+
+    run_test_balance_data(df, config, target, balance, backend)
+
+
+def test_non_binary_failure():
+    config = {
+        "input_features": [
+            {"name": "Index", "proc_column": "Index", "type": "numerical"},
+            {"name": "random_1", "proc_column": "random_1", "type": "numerical"},
+            {"name": "random_2", "proc_column": "random_2", "type": "numerical"},
```
w4nderlust, line 74: numerical -> number

connor-mccorm: done

tests/integration_tests/test_class_imbalance_feature.py
connor-mccorm: done

tests/integration_tests/test_class_imbalance_feature.py
connor-mccorm: Done

tests/integration_tests/test_class_imbalance_feature.py
connor-mccorm: Great idea, consolidating all the tests now!

justinxzhao (approved):
Nice! The tests look a lot cleaner - thanks for the changes. :)

PR #1758, opened 2022-02-17
Added faster AUROC

ShreyaR:
Attached some screenshots below of the comparison with `torchmetrics`' AUROC metric, and the maximum delta is around 0.005.

Screen Shot 2022-02-17 at 3 38 06 PM
Screen Shot 2022-02-17 at 3 39 35 PM
Screen Shot 2022-02-17 at 3 38 39 PM
Screen Shot 2022-02-17 at 3 38 29 PM

### DIFF
--- a/ludwig/modules/metric_modules.py
+++ b/ludwig/modules/metric_modules.py
@@ -134,8 +134,7 @@ class ROCAUCMetric(LudwigMetric):
         super().__init__(dist_sync_fn=gather_all_tensors)
         self.num_thresholds = num_thresholds
         self.epsilon = epsilon
-        # Add an extra dimension to each state to make it compatible with Horovod all gather.
-        self.add_state('summary_stats', torch.zeros(num_thresholds, 4)[None], dist_reduce_fx='sum')
+        self.add_state('summary_stats', torch.zeros(num_thresholds, 4), dist_reduce_fx='sum')
 
     def _get_thresholds(self, device, dtype) -> Tensor:
         thresholds = torch.linspace(0, 1, self.num_thresholds, device=device, dtype=dtype)
--- a/ludwig/modules/metric_modules.py
+++ b/ludwig/modules/metric_modules.py
@@ -122,32 +122,6 @@ class RMSEMetric(MeanSquaredError, LudwigMetric):
         return PREDICTIONS
 
 
-# @register_metric(ROC_AUC, [BINARY])
-# class ROCAUCMetric(AUROC, LudwigMetric):
-#     def __init__(self, **kwargs):
-#         super().__init__(dist_sync_fn=gather_all_tensors, compute_on_step=True)
-
-#     """ Metric for area under ROC curve. """
-
-#     def update(self, preds: Tensor, target: Tensor) -> None:
-#         # Currently only supported for binary tasks.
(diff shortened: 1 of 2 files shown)


ludwig/modules/metric_modules.py
```diff
@@ -123,11 +123,27 @@ def get_inputs(cls):
...
 
-    """ Metric for area under ROC curve. """
+    def __init__(
+        self,
+        num_thresholds: int = 201,
+        epsilon: float = 1e-7,
+        compute_on_step: Optional[bool] = None,
+        dist_sync_on_step: bool = False,
+        process_group: Optional[Any] = None,
+        dist_sync_fn: Optional[Callable] = None,
```
tgaddair, line 136: Should be `gather_all_tensors` (see `MeanMetric` just below).

ludwig/modules/metric_modules.py
```diff
@@ -123,11 +123,27 @@ def get_inputs(cls):
...
-        super().__init__(dist_sync_fn=gather_all_tensors)
+class ROCAUCMetric(LudwigMetric):
+    """Fast implementation of metric for area under ROC curve."""
 
-    """ Metric for area under ROC curve. """
+    def __init__(
+        self,
+        num_thresholds: int = 201,
+        epsilon: float = 1e-7,
+        compute_on_step: Optional[bool] = None,
```
tgaddair, line 133: The generic torchmetrics params should be capable of being passed through the `**kwargs`, so shouldn't need to specify explicitly.

tgaddair (commented):
Amazing! Just a couple points regarding the `__init__` params.

ludwig/modules/metric_modules.py
ShreyaR: Done!

ludwig/modules/metric_modules.py
ShreyaR: Updated the code to follow a similar pattern as what the other metrics are doing. It looks like we pass in some common arguments to each metric module in **kwargs, so I'm not passing in all kwargs to the torchmetrics constructor.

ludwig/modules/metric_modules.py
```diff
@@ -123,11 +123,26 @@ def get_inputs(cls):
...
+    def __init__(
+        self,
+        num_thresholds: int = 201,
+        epsilon: float = 1e-7,
+        **kwargs,
+    ) -> None:
         super().__init__(dist_sync_fn=gather_all_tensors)
+        self.num_thresholds = num_thresholds
+        self.epsilon = epsilon
+        # Add an extra dimension to each state to make it compatible with Horovod all gather.
```
tgaddair, line 138: What do you think about moving this fix to `gather_all_tensors`? It seems the correct behavior for all metrics should be something like:

```
t = t.reshape([1, *t.shape])
hvd.allgather(t)
```

ShreyaR: That makes sense -- Ip updated the code to reflect this.

I guess we weren't seeing issues from this previously as each metric being gathered was a scalar value, and so concatenating and reducing them wasn't causing an issue.

I haven't tested out this change yet but I can update here to make sure that it doesn't cause an issue with other metrics.

tgaddair (approved):
LGTM!

PR #1759, opened 2022-02-17
Update tune_for_memory code

amholler:
Update tune_for_memory code [crossporting to master]:

    Include optimizer memory use per model weight in model memory use estimation.
    Support lower minimum batch size for text models.
    Add logging of estimated model size and machine memory.
    Issue a warning if the model configuration has not been estimated to fit in memory.
    Restructure the code to make the get_model_size functionality a method in the ECD class.
    Simplify memory_tune_config path and allow it to setup gpu options as expected by ludwig.

Details on last item above:
   
    Currently, when tune_for_memory is set True for Ludwig AutoML running against a Ray cluster,
    a Ray remote call to memory_tune_config is invoked with num_cpus=1. That method in turn
    engenders an additional Ray remote call via get_machine_memory, with either num_gpus=1 or
    num_cpus=1, depending on whether the ray cluster resources include gpu or not.
   
    This PR updates this path so that the Ray remote call to memory_tune_config is invoked with
    num_cpus=1, optionally also including num_gpus=1 if ray cluster resources include gpu. This
    not only simplifies this code path, it also provides gpu-visibility across memory_tune_config,
    allowing the gpu initialization code (initialize_tensorflow) to be called to set up the gpus
    as ludwig expects and hence avoid their getting set up with the default gpu configuration,
    which can cause later ludwig gpu setup to fail, with the message "Intra op parallelism cannot
    be modified after initialization." when run across a multi-node ray cluster. 
 
Tested on Ray cluster on various ludwig text datasets.

### DIFF
--- a/ludwig/automl/auto_tune_config.py
+++ b/ludwig/automl/auto_tune_config.py
@@ -14,10 +14,11 @@ except ImportError:
     )
 
 from ludwig.api import LudwigModel
-from ludwig.automl.utils import get_available_resources, get_model_name
+from ludwig.automl.utils import get_model_name
 from ludwig.data.preprocessing import preprocess_for_training
 from ludwig.features.feature_registries import update_config_with_metadata
 from ludwig.utils.defaults import merge_with_defaults
+from ludwig.utils.tf_utils import initialize_tensorflow
 from ludwig.constants import COMBINER, HYPEROPT, BATCH_SIZE, TRAINING, TYPE, PREPROCESSING, SPACE
 
 # maps variable search space that can be modified to minimum permissible value for the range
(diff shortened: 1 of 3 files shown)


ludwig/automl/auto_tune_config.py
```diff
@@ -144,7 +136,7 @@ def memory_tune_config(config, dataset):
         if param in modified_hyperparam_search_space.keys():
             param_space = modified_hyperparam_search_space[param]["space"]
             if param_space == "choice":
-                if len(modified_hyperparam_search_space[param]['categories']) > 2 and \
+                if len(modified_hyperparam_search_space[param]['categories']) >= 2 and \
```
justinxzhao, line 139: Good catch!

justinxzhao (approved):
LGTM.

PR #1760, opened 2022-02-17
[crossport from tf-legacy] Update tune_for_memory code

amholler:
Cross port of https://github.com/ludwig-ai/ludwig/pull/1759

### DIFF
--- a/ludwig/automl/auto_tune_config.py
+++ b/ludwig/automl/auto_tune_config.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 from collections import OrderedDict
 
 import psutil
@@ -15,6 +16,8 @@ from ludwig.data.preprocessing import preprocess_for_training
 from ludwig.features.feature_registries import update_config_with_metadata
 from ludwig.utils.defaults import merge_with_defaults
 from ludwig.utils.torch_utils import initialize_pytorch
+
+logger = logging.getLogger(__name__)
 
 # maps variable search space that can be modified to minimum permissible value for the range
 RANKED_MODIFIABLE_PARAM_LIST = {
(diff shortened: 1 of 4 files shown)


ludwig/automl/auto_tune_config.py
```diff
@@ -1,20 +1,23 @@
...
 from ludwig.api import LudwigModel
-from ludwig.automl.utils import get_available_resources, get_model_name
+from ludwig.automl.utils import get_model_name
 from ludwig.constants import BATCH_SIZE, HYPEROPT, PREPROCESSING, SPACE, TRAINER
 from ludwig.data.preprocessing import preprocess_for_training
 from ludwig.features.feature_registries import update_config_with_metadata
 from ludwig.utils.defaults import merge_with_defaults
+from ludwig.utils.torch_utils import initialize_pytorch
+
+logger = logging.getLogger(__name__)
```
justinxzhao, line 20: I may be wrong here -- I believe with https://github.com/ludwig-ai/ludwig/pull/1669, we don't need to set up extra logging objects in torch -- `logging.info()` should just work.

ludwig/automl/auto_tune_config.py
```diff
@@ -115,25 +100,37 @@ def get_new_params(current_param_values, hyperparam_search_space, params_to_modi
...
     fits_in_memory = False
     raw_config = merge_with_defaults(config)
     training_set_metadata = get_trainingset_metadata(raw_config, dataset)
     modified_hyperparam_search_space = copy.deepcopy(raw_config[HYPEROPT]["parameters"])
     current_param_values = {}
     param_list = []
-    model_name = get_model_name(raw_config)
-    if model_name in RANKED_MODIFIABLE_PARAM_LIST:
-        params_to_modify = RANKED_MODIFIABLE_PARAM_LIST[model_name]
+    if (
```
justinxzhao, line 111: Does it make sense to have this logic live in `get_model_name()`, perhaps renaming `get_model_name()` to `get_model_type()`?

amholler: Done.

ludwig/automl/auto_tune_config.py
amholler: Yes, that works, thanks.  Done.

justinxzhao (approved):
Thanks for the changes!

PR #1768, opened 2022-02-24
Fixed MlflowCallback to work with distributed training

tgaddair:
The issue here is that mlflow has global state that tracks the "active run". If we call `start_run` from the driver, then attempt to do training on a remote worker, the active run information is lost. To workaround this, we can reset the active run by calling `start_run` again immediately after deserializing the callback.


anmshkmr (approved):
LGTM!

PR #1771, opened 2022-02-25
Fix stalling issue while early stopping

ShreyaR:


ludwig/models/trainer.py
```diff
@@ -1355,8 +1358,9 @@ def callback(self, fn, coordinator_only=True):
 
 class RemoteTrainer(Trainer):
     def __init__(self, gpus=None, gpu_memory_limit=None, allow_parallel_threads=True, **kwargs):
-        horovod = initialize_horovod()
-        super().__init__(horovod=horovod, **kwargs)
+        # horovod = initialize_horovod()
```
ShreyaR, line 1361: Noticed that we initialize horovod twice, once in `train_fn` and once in `RemoteTrainer`. Not sure if this is intentional. If not, I can also remove this from the legacy trainer.

tgaddair: The only thing here is you will need to make sure you plumb `horovod` from the caller that creates the `RemoteTrainer` to here. I think it's probably safer to leave this as is (calling `hvd.init()` multiple times is safe).

ludwig/models/trainer.py
```diff
@@ -1144,7 +1144,10 @@ def check_progress_on_validation(
                 )
 
         # ========== Early Stop logic ==========
-        if 0 < early_stop <= progress_tracker.last_improvement:
+        should_early_stop = torch.Tensor([0 < early_stop <= progress_tracker.last_improvement])
```
tgaddair, line 1147: Might want to use an int type tensor here just to be safe:

```
torch.as_tensor([...], dtype=torch.int)
```

ludwig/models/trainer.py
ShreyaR: I'd added the plumbing in the `RemoteTrainer` initialization inside the `train_fn`, but hadn't added it yet inside `legacy_train_fn`.

I can remove this if unnecessary.

tgaddair: Let's try doing the plumbing for the legacy trainer as well, just need to put it in the `executable_kwargs` here:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L293

tgaddair: Also, we should delete the commented out code.

ShreyaR: @tgaddair double checking this, looks like hvd is independently inside `legacy_train_fn` that the `RayTrainExecutor` executes. Given that, do we need to initialize hvd inside `RayLegacyTrainer` and then pass it to `executable_kwargs`?

ludwig/models/trainer.py
ShreyaR: Done

tgaddair (approved):

PR #1772, opened 2022-02-26
Fix loss discrepancy between single feature and combined

w4nderlust:


justinxzhao (approved):

justinxzhao:
## Sticking with batch-wise mean loss

An earlier snapshot of this PR attempts to migrate Ludwig with example-wise mean loss.

After some thinking, we've come to the conclusion that it’s perfectly functional, and much cleaner, to stick with batch-wise mean loss.

The batch-wise mean loss (average of averages) is the same as example-wise mean loss unless we are using batches of different sizes. Our batches are almost always the same size, with the exception of the very last batch, so the practical difference between the batch-wise mean loss and example-wise mean loss, is really small.

This is probably why in keras, tf, and torch, the default reduction behavior for all of the canonical loss functions is `mean` (instead of `none`).

Changing all of our loss functions to work with un-reduced loss vectors would require us to use non-default `reduction="none"`. This is ok, but then we’ll need to have special handling for when there’s a mix of batch-wise non-`MeanMetric` losses like `RMSPELoss` and regular `MeanMetric` losses like `SigmoidCrossEntropyLoss`, which can happen when there are multiple output features. Sticking with having all loss metrics always return the batch-wise mean is simpler and completely mitigates this incompatibility.

## List of changes

This PR now contains:

1. Fix for the loss discrepancy between training loss and combined loss when there’s a single output feature.

    The culprit is in `base_feature.py`, which calls the eval loss metric module directly. This is fine for non-`MeanMetric` metrics, but not for `MeanMetric`-based losses like `*CrossEntropyLoss`: 

    a) torchmetrics.MeanMetric's forward() returns the running average of the metric, not the imminent metric value
    b) torchmetrics.MeanMetric's forward() implicitly updates the running average. ECD makes an explicit call to `update_metrics()` after gathering losses from all of the output features, resulting in a double-update to output feature's  avg loss metrics.

    The fix for both issues s to call `get_current_value` for `MeanMetric`-based losses. This returns the imminent metric value, independent of the metric's running avg state.

2. Home-baked implementation of Jaccard metric calculation.

3. Change `SigmoidCrossEntropyLoss` to use torch’s default parameters (`reduction="mean"`). For some reason, we had been trying to do the reduction manually, (and incorrectly: not dividing by the overall batch size). Corresponding unit tests have also been updated.

4. Fix for `ValueError: all input arrays must have the same shape` for set output features by using the `PROBABILITIES` prediction key instead of `PROBABILITY` when calculating metrics. The latter’s tensor shape, for set features, is example-dependent, and cannot be consistently grouped with other examples.

## Testing

We've verified that the loss metrics discrepancy and Jaccard metrics have been fixed on a modified ATIS dataset, with both `SET` and `TEXT` output feature types, with and without regularization, with and without tabnet, with and without multiple output features.

PR #1776, opened 2022-03-01
Added Neuropod Torchscript export

tgaddair:


.github/workflows/pytest.yml
```diff
@@ -88,6 +89,12 @@ jobs:
           pip list
         shell: bash
 
+      - name: Install Neuropod backend
+        run: |
+          sudo mkdir -p "$NEUROPOD_BASE_DIR"
+          curl -L https://github.com/uber/neuropod/releases/download/v0.3.0-rc6/libneuropod-cpu-linux-v0.3.0-rc6-torchscript-1.10.2-backend.tar.gz | sudo tar -xz -C "$NEUROPOD_BASE_DIR"
```
brightsparc, line 95: Suggest including ENVARS in this curl request  for future upgrades:
* `NEUROPOD_VERISON:  "0.3.0-rc6"`
* `TORCHSCRIPT_VERISON: "1.10.2"`

```
https://github.com/uber/neuropod/releases/download/v${{ env.NEUROPOD_VERISON }}/libneuropod-cpu-linux-v${{ env.NEUROPOD_VERISON }}-torchscript-${{ env.TORCHSCRIPT_VERISON }}-backend.tar.gz
```

.github/workflows/pytest.yml
```diff
@@ -226,6 +234,12 @@ jobs:
...
```
brightsparc, line 240: Same env vars as above.

ludwig/utils/neuropod_utils.py
```diff
@@ -1,208 +1,119 @@
...
+        super().__init__()
+        self.inference_module = inference_module
+
+    def forward(self, {input_signature}):
+        inputs = {input_dict}
+        results = self.inference_module(inputs)
+        #for name, preds in results.items():
+        #    for k, v in preds.items():
+        #        if v.dtype == torch.bool:
+        #            preds[k] = v.to(dtype=torch.int32)
```
brightsparc, line 30: Is this comment here for a reason?

ludwig/utils/neuropod_utils.py
```diff
@@ -1,208 +1,119 @@
...
+        return {output_dicts}
+"""
+
+
+def _get_input_signature(config: Dict[str, Any]) -> str:
+    args = []
+    for feature in config["input_features"]:
+        name = feature[NAME]
+        args.append(f"{name}: Union[List[str], torch.Tensor]")
```
brightsparc, line 39: Optionally remove Union[] with explicit typing:

```
        name = feature[NAME]
        dt = 'List[str]' if feature[TYPE] in ['category'] else 'torch.Tensor'
        args.append(f"{name}: {dt}")
```

brightsparc (approved):
Suggested some minor improvements for env vars.

ludwig/utils/neuropod_utils.py
tgaddair: Experimental code for binary types. Will remove for now.

ludwig/utils/neuropod_utils.py
tgaddair: Some of these can be both (like bools), category should also support a tensor type (as ints) in the future. So will leave as is for now, but we can revisit.

PR #1779, opened 2022-03-02
Tune train path for items observed using py-spy wrt text AutoML (Bert) on dbpedia

amholler:
Items observed using py-spy for text AutoML (Bert) on dbpedia:

1) In reg_loss() in the current code, l1_reg and l2_reg are unconditionally computed, but both
    may not be used; it depends on the regularizer value (which is l2 by default).  The time to
    loop through the bert model parameters was non-trivial according to py-spy.  Updated the
    code to compute only the l1_reg and l2_reg values that will be consumed.
    
2) Additionally in reg_loss(), if 0 is passed in for lX, and the associated lX_reg value is to be computed,
    py-spy shows that the loop through the model parameters was still executed, which is uninteresting
    since the result is multiplied by 0, and regularization_lambda is 0 by default.  Updated the call site
    in train_loss to avoid calling reg_loss when regularization_lambda is 0.

3) Use set_to_none functionality for zero_grad, which py-spy had showed as taking non-trivial time,
    and can have good secondary effects as well (cited in pytorch tuning guide).  Note that I updated
    this PR to use the lower-level expression of this, because github testing showed failures otherwise.

With these changes (and limiting max_sequence_length), a dbpedia epoch actually completed in 2
hours (first time I have seen any dbpedia epoch complete), with good accuracy.

Bonus cleanup: specify learning rate as simple choice categories that match the values recommended
on the Google bert github page.

### DIFF
--- a/ludwig/automl/defaults/text/bert_config.yaml
+++ b/ludwig/automl/defaults/text/bert_config.yaml
@@ -2,10 +2,8 @@ hyperopt:
   # goal: maximize
   parameters:
     trainer.learning_rate:
-      type: float
-      lower: 0.00002
-      upper: 0.00005
-      space: loguniform
+      space: choice
+      categories: [0.00002, 0.00003, 0.00005]
     trainer.batch_size:
       space: choice
       categories: [8, 16, 32, 64, 128]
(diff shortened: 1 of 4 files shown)


justinxzhao (approved):
LGTM!

PR #1784, opened 2022-03-03
Additional tuning for Text AutoML using Bert

amholler:
Additional tuning for Text AutoML using Bert:
- Reduce Text AutoML batch size range, based on results from recent runs and from the LBT paper.
- Set skip_save_log to True for AutoML, given its non-trivial per-step overhead reported by py-spy.
- Add AdamW to Ludwig.  Set Text AutoML default optimizer to AdamW, often used to fine-tune Bert, with citations to https://arxiv.org/abs/1711.05101.

Tested on 2 hour Text AutoML runs of sst2, sst3, sst5, agnews, dbpedia.


justinxzhao (approved):
LGTM!

PR #1786, opened 2022-03-03
Temporarily disable gpu tests

tgaddair:


connor-mccorm (approved):

justinxzhao (approved):

tgaddair:
Looks like we're still having issues with detecting nvcc. Let's install from Horovod master for now and then revert once Let's wait until https://github.com/horovod/horovod/pull/3446 is released.

PR #1815, opened 2022-03-14
Improve AutoML heuristics for text classification

amholler:
AutoML heuristics improvements when tune_for_memory set True:
- Improve text model memory estimate accuracy: Replace worst-case max input (512) w/ dataset effective max input.
- When text models est. size > mem, consider reduce dataset max input before reduce batch size range, as per LBT data.
- Extend range of batch sizes for text models to 128, as per LBT data.

For time-based hyperparameter search, also allow specifying train epoch count limit, as discussed in AutoML meeting.
- This is particularly relevant for model fine-tuning, when only a small number of epochs are needed/useful.

Tested on unit tests and on automl_nlp workloads in ludwig/experiments repo


justinxzhao (approved):
LGTM

Issue #1802, opened 2022-03-10
Problem logging an experiment to wandb


danielduckworth:
**Describe the bug**
Logging a training run works fine, however when logging an experiment run (the evaluation part) with the default parameters, there is a OSError 22 relating to "probabilities_<UNK>.csv"

`OSError: [Errno 22] Invalid argument: '/results/classification_Distilbert_2/label_probabilities_<UNK>.csv`

**To Reproduce**

1. Run an experiment from the CLI with default parameters.
2. Configure the model for text classification.
3. Add the wandb flag

**Model config**

```
preprocessing:
    text:
        word_tokenizer: hf_tokenizer

input_features:
    -
        name: feature1
        type: text
        level: word
        encoder: distilbert
        pretrained_model_name_or_path: distilbert-base-uncased
    -
        name: feature2
        type: text
        level: word
        encoder: distilbert
        pretrained_model_name_or_path: distilbert-base-uncased
        tied_weights: feature1
    -
        name: feature3
        type: category

output_features:
    -
        name: label
        top_k: 5
        type: category

training:
    epochs: 3
    batch_size: 16
    learning_rate: 0.000005
```

**wandb metadata**

```
{
    "os": "Windows-10-10.0.19041-SP0",
    "python": "3.9.0",
    "heartbeatAt": "2022-03-10T14:21:23.463980",
    "startedAt": "2022-03-10T14:21:17.632459",
    "docker": null,
    "gpu": "NVIDIA GeForce RTX 2070",
    "gpu_count": 1,
    "cpu_count": 6,
    "cuda": null,
    "args": [
        "experiment",
        "--output_directory",
        "results",
        "--experiment_name",
        "classification",
        "--model_name",
        "Distilbert",
        "--training_set",
        "TRAIN_en.arrow",
        "--validation_set",
        "VAL_en.arrow",
        "--test_set",
        "EVAL_en.arrow",
        "--data_format",
        "feather",
        "--config_file",
        "model1.yaml",
        "--wandb"
    ],
    "state": "running",
    "program": "-m __main__",
    "executable": "I:\\Anaconda\\envs\\ludwig\\python.exe"
}
```

**Expected behavior**
The csv file for the UNK token should be created.

**Environment**

- OS: Windows-10-10.0.19041-SP0
- Python version: 3.9
- Ludwig version: 0.4.1

**Additional context**
The problem can be avoided with the ludwig API and passing the parameter skip_save_predictions=True (not possible with the CLI).

[...]

PR #1817, opened 2022-03-15
Extending os.matrix to support `windows-latest`

brightsparc:
The github action for pytest currently only runs on linux os which may be failing to catch errors in the windows environment as related to git issue #1802.  This PR is created to surface any such errors, and also to identifier libraries of dependencies that may not be fully functional in non-linux environments.

The neuropod library is excluded as it is not supported, and tests are [skipped](https://github.com/ludwig-ai/ludwig/blob/master/tests/integration_tests/test_neuropod.py#L37) for windows.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


brightsparc:
Close this in favour of #1830

PR #1829, opened 2022-03-18
AutoML: When single output feature, select default metric for hyperopt/trainer

amholler:
Validated for AutoML on GoEmotions text classification dataset predicting set output type

### DIFF
--- a/ludwig/automl/utils.py
+++ b/ludwig/automl/utils.py
@@ -7,8 +7,11 @@ from typing import Dict, List
 from dataclasses_json import dataclass_json, LetterCase
 from pandas import Series
 
-from ludwig.constants import COMBINER, CONFIG, HYPEROPT, NUMBER, PARAMETERS, SAMPLER, TRAINER, TYPE
+from ludwig.constants import COMBINER, CONFIG, HYPEROPT, NAME, NUMBER, PARAMETERS, SAMPLER, TRAINER, TYPE
+from ludwig.features.feature_registries import output_type_registry
+from ludwig.modules.metric_registry import metric_registry
 from ludwig.utils.defaults import default_combiner_type
+
 
 try:
     import ray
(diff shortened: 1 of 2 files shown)


ludwig/automl/utils.py
```diff
@@ -150,3 +152,23 @@ def _add_option_to_evaluate(
             if option_val not in hyperopt_params[option_param]["categories"]:
                 bisect.insort(hyperopt_params[option_param]["categories"], option_val)
     return point_to_evaluate
+
+
+# Set trainer and hyperopt metric and goal for first output feature if not already set
+def set_output_feature_metric(base_config):
```
justinxzhao, line 158: nit: Could we fold in the `== 1` logic into this function:

```
if len(base_config["output_features"]) > 1:
    # Multiple output features. Rely on default minimize `combined.loss goal`.
    # In the future, we could be more clever about choosing a combination of metrics.
    return
# if single output feature, set relevant metric and goal if not already set
    ...
```

ludwig/automl/utils.py
```diff
@@ -150,3 +152,23 @@ def _add_option_to_evaluate(
             if option_val not in hyperopt_params[option_param]["categories"]:
                 bisect.insort(hyperopt_params[option_param]["categories"], option_val)
     return point_to_evaluate
+
+
+# Set trainer and hyperopt metric and goal for first output feature if not already set
```
justinxzhao, line 157: nit: Turn this into a docstring.

justinxzhao (approved):

ludwig/automl/utils.py
amholler: done.

ludwig/automl/utils.py
amholler: done

PR #1830, opened 2022-03-21
fix: Updated fs_utils to get valid windows path from url

### LINKED ISSUE #1802: Problem logging an experiment to wandb

brightsparc:
Resolves #1802 which was having an issue writing to a windows file path after calling `get_fs_and_path`

```
File I:\Anaconda\envs\ludwig5\lib\site-packages\ludwig\utils\fs_utils.py:133, in open_file(url, *args, **kwargs)
    [130](file:///i%3A/Anaconda/envs/ludwig5/lib/site-packages/ludwig/utils/fs_utils.py?line=129) @contextlib.contextmanager
    [131](file:///i%3A/Anaconda/envs/ludwig5/lib/site-packages/ludwig/utils/fs_utils.py?line=130) def open_file(url, *args, **kwargs):
    [132](file:///i%3A/Anaconda/envs/ludwig5/lib/site-packages/ludwig/utils/fs_utils.py?line=131)     fs, path = get_fs_and_path(url)
--> [133](file:///i%3A/Anaconda/envs/ludwig5/lib/site-packages/ludwig/utils/fs_utils.py?line=132)     with fs.open(path, *args, **kwargs) as f:
    [134](file:///i%3A/Anaconda/envs/ludwig5/lib/site-packages/ludwig/utils/fs_utils.py?line=133)         yield f
```

The `fs_utils.get_fs_and_path` method that calls `fsspec.split_protocol` which does to support conversion to windows paths [windows paths](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/core.py#L518).

Updated this method to parse the url path, and decode any url escaping before using [pathlib.PurePosixPath](https://docs.python.org/3/library/pathlib.html) to convert the url into relative os path.

### DIFF
--- a/tests/ludwig/utils/test_fs_utils.py
+++ b/tests/ludwig/utils/test_fs_utils.py
@@ -1,17 +1,3 @@
-# Copyright (c) 2019 Uber Technologies, Inc.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-# ==============================================================================
 import logging
 import os
 import platform
--- a/tests/ludwig/utils/test_fs_utils.py
+++ b/tests/ludwig/utils/test_fs_utils.py
@@ -1,0 +1,48 @@
+# Copyright (c) 2019 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import os
+import tempfile
+
+from ludwig.utils.fs_utils import get_fs_and_path
+
+
+def assert_and_create_file(url, expected_path):
+    _, path = get_fs_and_path(url)
+    assert path == expected_path, f"Expected path {expected_path}"
+    try:
+        with tempfile.TemporaryDirectory() as tmpdir:
+            file_path = os.path.join(tmpdir, path)
+            os.makedirs(os.path.dirname(file_path))
+            with open(file_path, "w"):
+                assert True
+    except OSError as exc:
+        # OSError if file exists or is invalid
+        assert False, f"OS exception {exc}"
+
+
+def test_get_fs_and_path_simple():
+    assert_and_create_file("http://a/b.jpg", os.path.join("a", "b.jpg"))
+
+
+def test_get_fs_and_path_query_string():
+    assert_and_create_file("http://a/b.jpg?c=d", os.path.join("a", "b.jpg"))
+
+
+def test_get_fs_and_path_decode():
+    assert_and_create_file("http://a//b%20c.jpg", os.path.join("a", "b c.jpg"))
+
+
+def test_get_fs_and_path_unicode():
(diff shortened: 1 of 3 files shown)


tests/ludwig/utils/test_fs_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    assert path == expected_path, f"Expected path {expected_path}"
+    try:
+        with tempfile.TemporaryDirectory() as tmpdir:
+            file_path = os.path.join(tmpdir, path)
+            os.makedirs(os.path.dirname(file_path))
+            with open(file_path, "w"):
+                assert True
+    except OSError as exc:
+        # OSError if file exists or is invalid
+        assert False, f"OS exception {exc}"
```
justinxzhao, line 32: Do we need / could we remove try/except logic here?

An OSError from this code would be rather surprising, and also doesn't appear to be relevant for the tests.

brightsparc: I'm trying to catch the situation where errors might arise from operating systems that don't support special characters, I wasn't able to test this on my Posix machine, but on windows there are a bunch of un supported characters like `"` which could be work special tests for.  In this case perhaps catching the exception in the actual test might be more useful? https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

justinxzhao: Makes sense. Though, since this is logic is in the unit test file, I would expect the exception to be referred to in at least one of the unit tests.

Should we add a test that runs if the OS is windows and expects to fail on the `OSError`?

```
@pytest.mark.skipif(platform.system() != "Windows", reason="Special test case for Windows where certain characters aren't permitted."):
def test_get_fs_and_path_fails_with_special_characters():
    with pytest.raises(OSError):
        assert_and_create_file("http://a/æ.jpg", "a/æ.jpg")
```

brightsparc: I've changed the way i'm doing assertions, and added all special characters for linux/windows.  I've also attempted to add a new windows runner, but not sure if this is going to work (or is worth the compute cycles to run).  Instead perhaps we could leave it to the end user to run the "filesystem" marker test locally.

tests/ludwig/utils/test_fs_utils.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2019 Uber Technologies, Inc.
```
justinxzhao, line 1: nit: Remove the header for new files.

justinxzhao (approved):
LGTM!

PR #1836, opened 2022-03-22
feat: unit test for reproducibility

jimthompson5802:
# Code Pull Requests

Start of unit test for reproducibility.  This contains two tests:
* `test_preprocess` that test `LudwingModel.preprocess()`  splitting of data into train/validation/test data sets.  This is working.
*  `test_experiment` test reproducibility of  data splitting, training and evaluation statistics for `LudwigModel.experiment()`.   At this time, data splitting is reproducible.  Training statistics is not reproducible.  Currently looking into this.  Example error: 
image

### DIFF
--- a/tests/integration_tests/test_reproducibility.py
+++ b/tests/integration_tests/test_reproducibility.py
@@ -1,0 +1,86 @@
+import logging
+from typing import Union
+
+import os
+import numpy as np
+import pandas as pd
+import pytest
+import torch
+import tempfile
+import random
+
+from ludwig.data.dataset_synthesizer import cli_synthesize_dataset
+from ludwig.api import LudwigModel
+
+RANDOM_SEED = 1919
+
+INPUT_FEATURES = [
+    {'name': 'num_1', 'type': 'number'},
+    {'name': 'num_2', 'type': 'number'},
+]
+
+OUTPUT_FEATURES = [
+    {'name': 'y', 'type': 'number'}
+]
+
+CONFIG = {
+    'input_features': INPUT_FEATURES,
+    'output_features': OUTPUT_FEATURES,
+    'trainer': {'epochs': 10}
+}
+
+
+@pytest.fixture(scope='module')
+def raw_dataset_fp():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        raw_fp = os.path.join(tmpdir, 'raw_data.csv')
+        random.seed(RANDOM_SEED)
+        cli_synthesize_dataset(100, INPUT_FEATURES + OUTPUT_FEATURES, raw_fp)
+        yield raw_fp
+
+
+def test_preprocess(raw_dataset_fp):
+    # define Ludwig model
+    model = LudwigModel(config=CONFIG)
+
+    # preprocess the raw data set, specify seed
+    training_set1, validation_set1, test_set1, _ = model.preprocess(raw_dataset_fp, random_seed=RANDOM_SEED)
+
+    # invoke torch random functions
+    torch.manual_seed(RANDOM_SEED + 1)
+    torch.rand((5,))
+
+    # preprocess same raw data set with same seed
+    training_set2, validation_set2, test_set2, _ = model.preprocess(raw_dataset_fp, random_seed=RANDOM_SEED)
+
+    # confirm same split occurs before and after the calls to torch
+    assert np.all(pd.DataFrame(training_set1.dataset) == pd.DataFrame(training_set2.dataset))
+    assert np.all(pd.DataFrame(validation_set1.dataset) == pd.DataFrame(validation_set2.dataset))
+    assert np.all(pd.DataFrame(test_set1.dataset) == pd.DataFrame(test_set2.dataset))
+
+
+def test_experiment(raw_dataset_fp):
+    # define Ludwig model
+    model = LudwigModel(config=CONFIG, logging_level=logging.WARN)
+
+    evaluation_statistics1, training_statistics1, preprocessed_data1, _ = model.experiment(
+        dataset=raw_dataset_fp,
+        random_seed=RANDOM_SEED
+    )
+
+    evaluation_statistics2, training_statistics2, preprocessed_data2, _ = model.experiment(
+        dataset=raw_dataset_fp,
+        random_seed=RANDOM_SEED
+    )
+
+    # training data set split
+    assert np.all(pd.DataFrame(preprocessed_data1[0].dataset) == pd.DataFrame(preprocessed_data2[0].dataset))
+
+    # validation data set split
+    assert np.all(pd.DataFrame(preprocessed_data1[1].dataset) == pd.DataFrame(preprocessed_data2[1].dataset))
+
+    # test data set split
+    assert np.all(pd.DataFrame(preprocessed_data1[2].dataset) == pd.DataFrame(preprocessed_data2[2].dataset))
+
+    # check for equality of training statistics
+    assert training_statistics1['training'] == training_statistics2['training']


jimthompson5802:
Summary of changes:
git diff --name-status master
M	ludwig/models/ecd.py
  * embeded todo message for further discussion

M	ludwig/utils/checkpoint_utils.py
  * modified logging messages to conform with Ludwig setup
 
M	ludwig/utils/misc_utils.py
  * set `torch` random seed
 
A	tests/integration_tests/test_reproducibility.py
  * reproducibile train/validation/test split
  * reproducible `train()`
  * reproducible `experiment()`
  
Next step is to add tests for cli reproducibility

jimthompson5802:
git diff --name-status master reproducibility-unit-test
M	ludwig/models/ecd.py
 * embeded todo message for further discussion

M	ludwig/utils/misc_utils.py
  * set `torch` random seed
 
M	tests/integration_tests/test_cli.py, added these unit tests
* added cli train and experiment reproducibility tests

A	tests/integration_tests/test_reproducibility.py, added these unit tests
 * reproducibile train/validation/test split `preprocess()`  api
  * reproducible `train()` api
  * reproducible `experiment()` api

jimthompson5802:
git diff --name-status master reproducibility-unit-test
M	ludwig/models/ecd.py
 * embeded todo message for further discussion

M	ludwig/utils/misc_utils.py
  * set `torch` random seed
 
M	tests/integration_tests/test_cli.py, added these unit tests
* added cli train and experiment reproducibility tests
* test for reproducibility with horovod backend 

A	tests/integration_tests/test_reproducibility.py, added these unit tests
 * reproducibile train/validation/test split `preprocess()`  api
  * reproducible `train()` api
  * reproducible `experiment()` api

justinxzhao:
@jimthompson5802 looks like there's been good progress on this PR. Is it ready for review?

ludwig/utils/misc_utils.py
```diff
@@ -33,6 +34,11 @@ def set_random_seed(random_seed):
     os.environ["PYTHONHASHSEED"] = str(random_seed)
     random.seed(random_seed)
     numpy.random.seed(random_seed)
+    torch.manual_seed(random_seed)
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed(random_seed)
+        torch.backends.cudnn.deterministic = True
```
tgaddair, line 40: We should be careful about setting cudnn to deterministic, as that will restrict parallelism (most non-determinism in CUDA comes from mutli-threading). My opinion is that we should move this line (and the benchmark line just below) to the `initialize_pytorch` function in `torch_utils.py`, and only set this when `allow_parallel_threads` is `False`. This is consistent with the desired effect of setting that param, which is to restrict parallelism to enforce determinism.

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/torch_utils.py#L236

jimthompson5802:
@justinxzhao   The PR is ready for review.

ludwig/utils/misc_utils.py
jimthompson5802: @tgaddair   I understand you want me to move the setting of the `torch.backend.cudnn....` options to be in the section of code in `torch_utils.py` that is controlled by the `allow_parallel_threads`.  

Is this correct understanding?

tgaddair: Hey @jimthompson5802, yes that would be my preference. It might make sense at some point to change `allow_parallel_threads` to `deterministic` or similar (though truly guaranteeing determinism, particularly with gpus or distributed systems, is extremely difficult). But for now I believe this is the most correct place to ensure we aren't incurring a performance penalty.

jimthompson5802: @tgaddair   Got will make the change later tonight.  Thank you for the clarification.

jimthompson5802: incorporated change.

tgaddair (approved):
LGTM! Thanks @jimthompson5802.

jimthompson5802:
@justinxzhao   Just checking to see if you had any comments.  Looks like @tgaddair  and @w4nderlust   are OK with the new unit test.  If everything looks OK, I'll merge it.

tests/integration_tests/test_cli.py
```diff
@@ -300,3 +301,57 @@ def test_preprocess_cli(csv_filename):
...
+        # run first model
+        command_to_run(
+            type_of_run,
+            dataset=dataset_filename,
+            config=config_filename,
+            output_directory=tmpdir,
+            skip_save_processed_input="",  # skip saving preprocessed inputs for reproducibility
+            experiment_name="reproducible",
+            model_name="run1",
+            random_seed="1919",
```
justinxzhao, line 329: Consider parameterizing this random seed to make sure that we see the reproducibility for at least one other random seed value.

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,109 @@
...
+    for i in range(3):
+        for k in preprocessed_data1[i].dataset:
+            assert np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
+
+
+def test_train(raw_dataset_fp):
+    # define Ludwig model
+    model1 = LudwigModel(config=CONFIG, logging_level=logging.WARN)
+    training_statistics1, preprocessed_data1, _ = model1.train(
+        dataset=raw_dataset_fp, random_seed=RANDOM_SEED, skip_save_progress=True, skip_save_processed_input=True
```
justinxzhao, line 64: Consider parameterizing this random seed to make sure that we see the reproducibility for at least one other random seed value.

justinxzhao (commented):
Thanks for the changes Jim.

My main comment is to parameterize the random seed value so that we can check that we see reproducibility with at least one other seed value.

Additionally I'm wondering if we could also add an "un-reproducing" test to make sure that the data/models are certainly not the same when we use a different random seed, as a sanity check that using a different random seed certainly has the expected effect of producing something different.

tests/integration_tests/test_cli.py
jimthompson5802: @justinxzhao 
incorporated comments:
* added second random seed
* add test that a different seed generate different results

tests/integration_tests/test_reproducibility.py
jimthompson5802: @justinxzhao 
incorporated comments:
* second seed 
* confirm that different seed generates different results

w4nderlust:
> Additionally I'm wondering if we could also add an "un-reproducing" test to make sure that the data/models that use a different random seed has produces something different.

I like this idea

jimthompson5802:
> > Additionally I'm wondering if we could also add an "un-reproducing" test to make sure that the data/models that use a different random seed has produces something different.
> 
> I like this idea

I incorporated this comment with the `second_seed_offset`.  If it is zero, both seeds are the same.  If non-zero, the second seed is different from the first seed.
https://github.com/jimthompson5802/ludwig/blob/494aee0cf00928a75704003846e1c1eacb4a35e5/tests/integration_tests/test_reproducibility.py#L126-L170

tests/integration_tests/test_cli.py
```diff
@@ -300,3 +301,86 @@ def test_preprocess_cli(csv_filename):
...
+        with open(os.path.join(tmpdir, "reproducible_run2", "training_statistics.json")) as f:
+            training2 = json.load(f)
+
+        if second_seed_offset == 0:
+            # same seeds should result in same output
+            assert training1 == training2
+        else:
+            # non-zero second_seed_offset uses different seeds and should result in different output
+            with pytest.raises(AssertionError):
+                assert training1 == training2
```
justinxzhao, line 371:
```suggestion
            assert training1 != training2
```

tests/integration_tests/test_cli.py
```diff
@@ -300,3 +301,86 @@ def test_preprocess_cli(csv_filename):
...
+        else:
+            command_to_run = _run_ludwig_horovod
+
+        # run first model
+        command_to_run(
+            type_of_run,
+            dataset=dataset_filename,
+            config=config_filename,
+            output_directory=tmpdir,
+            skip_save_processed_input="",  # skip saving preprocessed inputs for reproducibility
```
justinxzhao, line 341: This should be `False`, right?

tests/integration_tests/test_cli.py
```diff
@@ -300,3 +301,86 @@ def test_preprocess_cli(csv_filename):
...
+        backend (str): backend to use
+        type_of_run(str): type of run, either train or experiment
+        csv_filename(str): file path of dataset to use
+        random_seed(int): random seed integer to use for test
+        second_seed_offset(int): zero to use same random seed for second test, non-zero to use a different
+            seed for the second run.
+
+    Returns: None
+    """
+    with tempfile.TemporaryDirectory() as tmpdir:
```
justinxzhao, line 326: Instead of setting up a `tmpdir` manually, tmpdir is a built-in fixture of pytest:

https://github.com/ludwig-ai/ludwig/blob/master/tests/integration_tests/test_api.py#L188

tests/integration_tests/test_cli.py
```diff
@@ -300,3 +301,86 @@ def test_preprocess_cli(csv_filename):
...
+            with open(os.path.join(tmpdir, "reproducible_run2", "test_statistics.json")) as f:
+                test2 = json.load(f)
+
+            if second_seed_offset == 0:
+                # same seeds should result in same output
+                assert test1 == test2
+            else:
+                # non-zero second_seed_offset uses different seeds and should result in different output
+                with pytest.raises(AssertionError):
+                    assert test1 == test2
```
justinxzhao, line 386:
```suggestion
                assert test1 != test2
```

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+    # confirm data splits are reproducible
+    for i in range(3):
+        for k in preprocessed_data1[i].dataset:
+            if second_seed_offset == 0:
+                # same seeds should result in same output
+                assert np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
+            else:
+                # non-zero second_seed_offset uses different seeds and should result in different output
+                with pytest.raises(AssertionError):
+                    assert np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
```
justinxzhao, line 77:
```suggestion
                assert not np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
```

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
```
justinxzhao, line 120:
```suggestion
                assert np.all(preprocessed_data1[i].dataset[k] != preprocessed_data2[i].dataset[k])
```

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # confirm reproducibility/non-reproducibility of results
+    if second_seed_offset == 0:
+        # same seeds should result in same output
+        assert training_statistics1 == training_statistics2
+    else:
+        # non-zero second_seed_offset uses different seeds and should result in different output
+        with pytest.raises(AssertionError):
+            assert training_statistics1 == training_statistics2
```
justinxzhao, line 129:
```suggestion
        assert training_statistics1 != training_statistics2
```

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+    # confirm data splits are reproducible
+    for i in range(3):
+        for k in preprocessed_data1[i].dataset:
+            if second_seed_offset == 0:
+                # same seeds should result in same output
+                assert np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
+            else:
+                # non-zero second_seed_offset uses different seeds and should result in different output
+                with pytest.raises(AssertionError):
+                    assert np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
```
justinxzhao, line 170:
```suggestion
                assert not np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
```

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+    if second_seed_offset == 0:
+        # same seeds should result in same output
+        assert training_statistics1 == training_statistics2
+        assert evaluation_statistics1 == evaluation_statistics2
+    else:
+        # non-zero second_seed_offset uses different seeds and should result in different output
+        with pytest.raises(AssertionError):
+            assert training_statistics1 == training_statistics2
+        with pytest.raises(AssertionError):
+            assert evaluation_statistics1 == evaluation_statistics2
```
justinxzhao, line 182:
```suggestion
        assert training_statistics1 != training_statistics2
        assert evaluation_statistics1 != evaluation_statistics2
```

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    # define Ludwig model
+    model1 = LudwigModel(config=CONFIG)
+
+    # preprocess the raw data set, specify seed
+    preprocessed_data1 = model1.preprocess(raw_dataset_fp, random_seed=random_seed)
+
+    # invoke torch random functions
+    torch.manual_seed(random_seed + 1)
+    torch.rand((5,))
```
justinxzhao, line 61: Is it necessary to set these seeds manually in the test for the test to pass?

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    # define Ludwig model
+    model1 = LudwigModel(config=CONFIG, logging_level=logging.WARN)
+    training_statistics1, preprocessed_data1, _ = model1.train(
+        dataset=raw_dataset_fp, random_seed=random_seed, skip_save_progress=True, skip_save_processed_input=True
+    )
+
+    # invoke torch random functions
+    torch.manual_seed(random_seed + 1)
+    torch.rand((5,))
```
justinxzhao, line 101: Same question: Is it necessary to set these seeds manually in the test for the test to pass?

tests/integration_tests/test_reproducibility.py
```diff
@@ -0,0 +1,182 @@
...
+    # define Ludwig model
+    model1 = LudwigModel(config=CONFIG, logging_level=logging.WARN)
+
+    evaluation_statistics1, training_statistics1, preprocessed_data1, _ = model1.experiment(
+        dataset=raw_dataset_fp, random_seed=random_seed, skip_save_processed_input=True
+    )
+
+    # invoke torch random functions
+    torch.manual_seed(random_seed + 1)
+    torch.rand((5,))
```
justinxzhao, line 154: Same question: Is it necessary to set these seeds manually in the test for the test to pass?

tests/integration_tests/test_reproducibility.py
jimthompson5802: @w4nderlust wanted to confirm that an unrelated torch random operation, using an unrelated seed, did not affect Ludwig reproducibility.  That is the reason for this particular set of code.  Though I realize that with the introduction of the `second_seed_offset`, I need to account for this new construct.

tests/integration_tests/test_cli.py
jimthompson5802: Given the way the function that generates the cli command, we only need `--skip_save_preprocessed_input`, which is assumed to be true.  For cli this is not a keyword parameter, like the api version. 

Based on my testing, I believe this is correct.

tests/integration_tests/test_cli.py
jimthompson5802: converted to use pytest fixture `tmpdir`.

tests/integration_tests/test_reproducibility.py
jimthompson5802: see this response: https://github.com/ludwig-ai/ludwig/pull/1836#discussion_r841142594

tests/integration_tests/test_cli.py
jimthompson5802: @justinxzhao  a bit more specifics on why `skip_save_processed_input=""` is used.

From a cli perspective the argument is defined as follows:
```
    parser.add_argument(
        "-sspi",
        "--skip_save_processed_input",
        help="skips saving intermediate HDF5 and JSON files",
        action="store_true",
        default=False,
```

The presence or absence of the argument on the command line determines `True` or `False`, respectively.

Second in `test_cli.py`, this function generates the cli arguments and runs the cli command
```
def _run_commands(commands, **ludwig_kwargs):
    for arg_name, value in ludwig_kwargs.items():
        commands += ["--" + arg_name, value]
    cmdline = " ".join(commands)
    print(cmdline)
    completed_process = subprocess.run(cmdline, shell=True, stdout=subprocess.PIPE, env=os.environ.copy())
    assert completed_process.returncode == 0


    return completed_process
```

The relevant code is `command += ["--" + arg_name, value]` is always looking for a value.  So by using `skip_save_processed_input=""`, the `_run_commands()` function generates this cli argument:
 ```
--skip_save_processed_input
```

Does this answer the question?

File: tests/integration_tests/test_reproducibility.py (hunk 9 above)
Comment (jimthompson5802) [reply tests/integration_tests/test_reproducibility.py:120]: done

File: tests/integration_tests/test_reproducibility.py (hunk 10 above)
Comment (jimthompson5802) [reply tests/integration_tests/test_reproducibility.py:129]: done

File: tests/integration_tests/test_reproducibility.py (hunk 13 above)
Comment (justinxzhao) [reply tests/integration_tests/test_reproducibility.py]: Thanks for the clarification.

If it is important to test that changing torch random seeds should not affect data preprocessing, then I would suggest putting this in it's own test:

```
@pytest.mark.parametrize("random_seed", [1919, 31])
def test_preprocess_ignores_torch_seed(raw_dataset_fp: str, random_seed: int) -> None:
    # define Ludwig model
    model1 = LudwigModel(config=CONFIG)

    # preprocess the raw data set, specify seed
    preprocessed_data1 = model1.preprocess(raw_dataset_fp, random_seed=random_seed)

    # Override torch seed.
    torch.manual_seed(random_seed + 1)
    torch.rand((5,))

    # define Ludwig model
    model2 = LudwigModel(config=CONFIG)
    
    # preprocess same raw data set with same seed
    preprocessed_data2 = model2.preprocess(raw_dataset_fp, random_seed=random_seed)

    # confirm data splits are reproducible
    for i in range(3):
        for k in preprocessed_data1[i].dataset:
            # same seeds should result in same output
            assert np.all(preprocessed_data1[i].dataset[k] == preprocessed_data2[i].dataset[k])
```

The decoupling makes the behavior that we are trying to test clearer (imo).

justinxzhao (approved):
Nice! Thanks for all of the changes @jimthompson5802

PR #1837, opened 2022-03-22
Add callbacks for on_eval_start and on_eval_end

justinxzhao:
- Add callbacks for on_eval_start and on_eval_end.
- Parameterize callback unit tests to cover step-based training.
- Copy existing `on_epoch_end` callbacks to `on_eval_end` callbacks.

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -601,9 +601,10 @@ class Trainer(BaseTrainer):
         Returns whether the trainer should early stop, based on validation metrics history.
         """
         start_time = time.time()
+        self.callback(lambda c: c.on_eval_start(self, progress_tracker, save_path))
+
         if self.is_coordinator():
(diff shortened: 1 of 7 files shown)


ludwig/models/trainer.py
```diff
@@ -601,6 +601,8 @@ def run_evaluation(
         Returns whether the trainer should early stop, based on validation metrics history.
         """
         start_time = time.time()
+        self.callback(lambda c: c.on_eval_start(self, progress_tracker, save_path))
```
jeffreyftang, line 604: Just curious: the docstring for `on_eval_start` says it's called from the coordinator, but this invocation doesn't check `is_coordinator()`. Is that intentional?

tgaddair: Check the implementation of `self.callback` ;)

justinxzhao: I missed this detail initially as well. Keeping the `self.is_coordinator` centralized makes sense to avoid code duplication.

w4nderlust:
lgtm

PR #1838, opened 2022-03-22
Fix python 3 compatibility issues

tirkarthi:
# Code Pull Requests

* `time.clock` was removed in https://github.com/python/cpython/pull/13270.
* Import ABC from `collections` was deprecated and removed in Python 3.10. Use `collections.abc` .


justinxzhao:
Thanks for the changes @tirkarthi !

PR #1840, opened 2022-03-22
Added annotations to spread Dask data loading across Ray workers

tgaddair:
By default, Dask on Ray will try to place workers on the head node instead of spreading the across the cluster.

This is the new API to address the issue described here: https://docs.ray.io/en/latest/data/dask-on-ray.html#best-practice-for-large-scale-workloads.

cc @clarkzinzow


clarkzinzow (approved):
LGTM! 🙌

PR #1844, opened 2022-03-23
Hyperopt warning

connor-mccorm:
Added a warning in CLI that asks you if you're sure you want to run ludwig train even though you have a hyperopt section present. The user is prompted with a [Y/n] decision in the command line. This change only affects cli since interacting with the api could interfere with Predibase operations.

### DIFF
--- a/ludwig/train.py
+++ b/ludwig/train.py
@@ -23,8 +23,8 @@ import pandas as pd
 from ludwig.api import LudwigModel
 from ludwig.backend import ALL_BACKENDS, Backend, initialize_backend
 from ludwig.callbacks import Callback
-from ludwig.contrib import add_contrib_callback_args
 from ludwig.constants import HYPEROPT, HYPEROPT_WARNING
+from ludwig.contrib import add_contrib_callback_args
 from ludwig.globals import LUDWIG_VERSION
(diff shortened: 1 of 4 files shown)


justinxzhao (approved):
Assuming this is only for the CLI, LGTM.

ludwig/utils/print_utils.py
```diff
@@ -58,3 +58,33 @@ def print_boxed(text, print_fun=logger.info):
 
 def repr_ordered_dict(d: OrderedDict):
     return "{" + ",\n  ".join(f"{x}: {pformat(y, indent=4)}" for x, y in d.items()) + "}"
+
+
+def query_yes_no(question: str, default="yes"):
```
w4nderlust, line 63: add type hint to default

ludwig/train.py
```diff
@@ -153,6 +154,10 @@ def train_cli(
 
     :return: (`None`)
     """
+    if HYPEROPT in config:
```
w4nderlust, line 157: this runs on the train cli. It should also be run in the experiment cli.
Also, both train and experiment API should have a simila waring, but without the blocking yes/no question, just a warning.

PR #1850, opened 2022-03-26
Handle when progress_trainer.log_metrics() is called before any metrics are added.

justinxzhao:
Other changes:
- Refactor and move ProgressTracker to `trainer_utils`, and add unit tests.
- Introduce a new trainer parameter, `should_evaluate_train` that dictates whether to run evaluation on the entire training set. Default False.


justinxzhao:
Closing in favor of split PRs, #1851, #1852, and #1853.

PR #1852, opened 2022-03-27
Refactor ProgressTracker and other stateless methods into separate utils files, and add unit tests to handle empty log_metrics() case.

justinxzhao:

PR #1853, opened 2022-03-27
Add evaluate_training_set, which dictates whether to run evaluation on the entire training set.

justinxzhao:
Default to True.


justinxzhao:
Tests should be fixed after #1852 merges.

ludwig/models/trainer.py
```diff
@@ -101,6 +101,7 @@ def __init__(
         optimizer=None,
         epochs=100,
         steps_per_checkpoint=0,
+        should_evaluate_train=False,
```
tgaddair, line 104: Maybe `evaluate_training_set`?

tgaddair (approved):

ludwig/models/trainer.py
justinxzhao: Done.

PR #1859, opened 2022-03-29
Remove the concept of a "level" for text features preprocessing.

justinxzhao:
Each text feature is tokenized according to a specific `tokenizer_type`, be it `space_punct`, `characters`, or `hf_tokenizer`.

Also, rename `sequence_length_limit` -> `max_sequence_length` everywhere.

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -277,7 +277,7 @@ class RemotePredictor(Predictor):
 def calculate_overall_stats(output_features, predictions, dataset, training_set_metadata):
     overall_stats = {}
     for of_name, output_feature in output_features.items():
-        feature_metadata = {}
+        feature_metadata = training_set_metadata[output_feature.feature_name]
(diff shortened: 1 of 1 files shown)


ludwig/models/predictor.py
```diff
@@ -277,7 +277,7 @@ def __init__(self, model: ECD, gpus=None, gpu_memory_limit=None, allow_parallel_
 def calculate_overall_stats(output_features, predictions, dataset, training_set_metadata):
     overall_stats = {}
     for of_name, output_feature in output_features.items():
-        feature_metadata = output_feature.overall_statistics_metadata()
+        feature_metadata = {}
```
dantreiman, line 280:
```suggestion
        feature_metadata = training_set_metadata[output_feature.feature_name]
```

dantreiman (approved):
This is much simpler than before.  Well done!

I tested it on my Spam Assassin experiment, which uses `space`, `characters`, and `hf_tokenizer` - everything worked as expected.

PR #1861, opened 2022-03-30
Adds preprocessing to agnews dataset to add readable class names.

dantreiman:


ludwig/datasets/agnews/__init__.py
```diff
@@ -42,3 +42,12 @@ def read_file(self, filetype, filename, header=0):
         # 1-World, 2-Sports, 3-Business, 4-Science/Tech
         file_df.columns = ["class_index", "title", "description"]
         return file_df
+
+    def process_downloaded_dataset(self):
+        super().process_downloaded_dataset(header=None)
+        processed_df = pd.read_csv(os.path.join(self.processed_dataset_path, self.csv_filename))
+        # Maps class_index to class name.
+        class_names = ["", "world", "sports", "business", "sci/tech"]
```
w4nderlust, line 50: Not super familair with the dataset, but does class 0 exist in the dataset? I believe no, which makes sense, but just doublechecking

dantreiman: Thats right, there are no instances of class 0 in the dataset.

PR #1867, opened 2022-04-01
feat: Add github workflow to check extracted marshmallow files are up-to-date.

ksbrar:
Adds a github workflow that checks out the PR, runs `python scripts/extract_schema.py` and then `git diff` - returning an error if there are changes that require the marshmallow extracted files to be regenerated by the user before a PR can be merged.

### DIFF
--- a/.github/workflows/extract_schema.yml
+++ b/.github/workflows/extract_schema.yml
@@ -1,0 +1,84 @@
+#
+name: extract_schema
+
+on:
+  push:
+    branches: [master]
+  pull_request:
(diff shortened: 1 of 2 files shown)


ksbrar:
Workflow works, now testing that an error occurs if indeed there is some change.

ksbrar:
Did not work

hungcs (approved):

ksbrar:
Going to remove extraction script, so this is now unnecessary.

PR #1869, opened 2022-04-02
Use Tune checkpoint naming that corresponds to Tune reporting [Fix for issue 1866]

amholler:


tests/ludwig/utils/test_trainer_utils.py
```diff
@@ -61,6 +62,7 @@ def test_progress_tracker():
         "num_increases_bs": 0,
         "num_reductions_lr": 0,
         "steps": 0,
+        "tune_ckpts": 0,
```
justinxzhao, line 65: nit: Use the full name, `tune_checkpoints`.

justinxzhao (approved):

tests/ludwig/utils/test_trainer_utils.py
amholler: Done.  Changed name to tune_checkpoint_num as per your suggestion in slack.

PR #1872, opened 2022-04-04
Move eval end callback trigger after model_weights save to ensure complete/consistent tune checkpoint 

amholler:
Fix for https://github.com/ludwig-ai/ludwig/issues/1871
Validated using manual repro of problem in debugger.

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -716,13 +716,13 @@ class Trainer(BaseTrainer):
 
         elapsed_time = (time.time() - start_time) * 1000.0
 
-        self.callback(lambda c: c.on_eval_end(self, progress_tracker, save_path))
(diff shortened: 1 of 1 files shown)


ludwig/models/trainer.py
```diff
@@ -744,14 +744,15 @@ def run_evaluation(
...
             )
-            if should_break:
-                return should_break
         else:
             # There's no validation, so we save the model.
             if self.is_coordinator() and not self.skip_save_model:
                 torch.save(self.model.state_dict(), model_weights_path)
 
-        return False
+        # Trigger eval end callback after any model weights save for complete checkpoint
```
justinxzhao, line 752: Thanks for adding a comment about why the callback should be after model saving.

justinxzhao (approved):

PR #1873, opened 2022-04-04
Added wrapper to allow caching DataFrames with custom checksum

tgaddair:
One issue with using Pandas or Dask DataFrames as inputs is preprocessing caching. Prior to this PR, DataFrame inputs would always require redoing the preprocessing even if the input DF hasn't changed.

This PR does not transparently solve this issue, but instead introduces a new wrapper class called `CacheableDataframe` which can be used to provide the necessary caching metadata for the DF manually.

For example:

```
dataset = CacheableDataframe(df=df, name=name, checksum=checksum)
model.train(dataset)
```

The `name` is the key that will be used to identify the entry in the cache. The `checksum` is what will be compared to determine whether the dataset has changed since it was last cached.

One application of this would be to load a table in a database into a Dask DF and then give it a `name` equal to the table name and a `chcksum` equal to the last modification time of the table.


ludwig/data/cache/types.py
```diff
@@ -0,0 +1,104 @@
...
+    if dataset is None:
+        return None
+
+    if isinstance(dataset, CacheableDataset):
+        return dataset
+    if isinstance(dataset, str):
+        return CacheablePath(path=dataset)
+
+    # TODO(travis): could try hashing the in-memory dataset, but this is tricky for Dask
+    checksum = str(uuid.uuid1())
```
anmshkmr, line 102: For Dask dataframe, won't checksum will resolve to a unique str each time? That can prevent cache lookup even for the same source. For dask df should we count on some graph properties that will resolve to the same unique checksum?

anmshkmr:
Looks very clean to me. The only issue I can foresee is the checksum for dask dataframes. I understand it is tricky to implement it. Maybe we can provide a manual override for checksums for manual preprocessing workloads? That way the user can precisely control reloading cached dataset once preprocessing has been finished.

ludwig/data/cache/types.py
anmshkmr: Never mind, I see that you have an override mechanism. Thanks.

anmshkmr (approved):

PR #1877, opened 2022-04-05
Prevent loading pre-trained encoder weights when loading a trained Ludwig model

dantreiman:
After training a model, adds a new `saved_weights_in_checkpoint` key to input feature config, which indicates that input feature encoder weights are stored in the model checkpoint.

If `saved_weights_in_checkpoint` is true, encoders which were initialized from pretrained weights i.e. Huggingface should not load their pre-trained weights, as they will be overwritten by the saved weights in the checkpoint anyway.


justinxzhao (approved):
LGTM.

Is my understanding correct:

1. when a user trains a model for the first time, `input_feature["saved_weights_in_checkpoint"]` will be `False` for all input features
2. `input_feature["saved_weights_in_checkpoint"]` will be set to `True` at the end of training and the config is saved to disk (JSON).
3. Re-loading the model with `evaluate`/`predict`/`model.load()` will retain `saved_weights_in_checkpoint=True` for all input features.
4. huggingface weights downloads are skipped since `saved_weights_in_checkpoint=True`

w4nderlust:
> 2. `input_feature["saved_weights_in_checkpoint"]` will be set to `True` at the end of training and the config is saved to disk (JSON).

I would say that should be set to True at the beginning of training / the first time a checkpoint is saved (they are equivalent because if you start training and kill it before saving the first checkpoint it doesn't matter anyway). Also if you resume from a checkpoint (created before the end of training) the value should still be True.

ludwig/api.py
```diff
@@ -607,6 +607,10 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                     if not path_exists(weights_save_path):
                         with open_file(weights_save_path, "wb") as f:
                             torch.save(self.model.state_dict(), f)
+                    # Adds a flag to all input features indicating that the weights are saved in the checkpoint.
+                    for input_feature in self.config["input_features"]:
+                        input_feature["saved_weights_in_checkpoint"] = True
```
w4nderlust, line 612: Would generalize it to all output features and combiner as well

w4nderlust: Maybe we could have a separate pr for it though, we can add to the board as a reminder

w4nderlust: I would doublecheck the placement of this code for setting it to true, need to make sure it is hit by every possible path that involves training (cli, api, train, experiment, etc.)

dantreiman: I thought about that, and I was curious how you would implement this: either trainer needs to inform LudwigModel when it saves,

dantreiman: Oh hey what if we move saving down to `LudwigModule`.  Define load and save at that level, implement it for ECD only for today.  That is a push in the direction of modular saving and loading.  On the downside, its a bigger refactor than I wanted to do here to address this one bug.

dantreiman: Moved the setting of the flag to the first save_config before training starts (instead of after training).

I would like to push saving logic down to the model (ECD), will do that in a separate PR.  Opened an issue for this: https://github.com/ludwig-ai/ludwig/issues/1909

PR #1880, opened 2022-04-06
Modified figure_data_format_dataset to work with recently introduced CacheableDataFrame

anmshkmr:


ludwig/utils/data_utils.py
```diff
@@ -682,7 +684,9 @@ def clear_data_cache():
 
 
 def figure_data_format_dataset(dataset):
-    if isinstance(dataset, pd.DataFrame):
+    if isinstance(dataset, ludwig.data.cache.types.CacheableDataframe):
```
tgaddair, line 687: Nice, we can make this a little bit more general by changing to:

```
if isinstance(dataset, CacheableDataset):
    return figure_data_format_dataset(dataset.unwrap()
```

ludwig/utils/data_utils.py
```diff
@@ -31,6 +31,7 @@
 from pandas.errors import ParserError
 from sklearn.model_selection import KFold
 
+import ludwig.data.cache.types
```
tgaddair, line 34: Would change to:

```
from ludwig.data.cache.types import CacheableDataset
```

tgaddair (approved):

PR #1884, opened 2022-04-08
merge_with_defaults now populates image feature defaults

connor-mccorm:
When running ludwig train with a ray backend, during preprocessing the ray backend makes a check to see if a preprocessing param `in_memory` is True. However, the `merge_with_defaults` call in the LudwigModel constructor does not actually populate the config with any defaults. This PR adds those defaults into the preprocessing section of the input features if they are not already specified.

### DIFF
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -471,5 +471,6 @@ class ImageInputFeature(ImageFeatureMixin, InputFeature):
     @staticmethod
     def populate_defaults(input_feature):
         set_default_value(input_feature, TIED, None)
+        set_default_value(input_feature, PREPROCESSING, {})
         defaults = ImageFeatureMixin.preprocessing_defaults()
         set_default_values(input_feature[PREPROCESSING], defaults)
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -57,7 +57,7 @@ from ludwig.utils.image_utils import (
     read_image,
     resize_image,
 )
-from ludwig.utils.misc_utils import set_default_value, set_default_values
+from ludwig.utils.misc_utils import set_default_value
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


jeffreyftang (approved):

jeffreyftang:
Are the defaults for other feature types being set correctly, or should a change like this be made for them as well?

connor-mccorm:
That's a great question. I haven't checked, but I'm guessing they are not. Since they aren't blocking the other model types in predibase (though audio might be blocked with the same `in_memory` key error) I think we should probably merge this asap and then make another change after that fixes the rest.

w4nderlust:
Other features don't do it. it should be taken care of by defaults.py:L275 that sets the global, er type, preprocessing defaults. And then by preprocessing.py:1175 that merges the feature preprocessing with the global preprocessing. Those two lines should take care of this. Can you check the confg before and after those calls?

connor-mccorm:
Ya i'll check on that right now!

connor-mccorm:
So when `merge_preprocessing` is called in build metadata, it does what you described correctly. The issue here is that it never updates the `feature_config` itself which is the argument that's passed into the `check_lazy_load_supported` downstream a bit. So while in `add_feature_data` for image_feature.py,   `preprocessing_parameters` contains the default preprocessing parameters but the feature_configs themselves do not still. Since image features make the same check to `in_memory` I tried to mirror the way that they are setting that preprocessing parameter before this check and they are doing it with the line I have in the latest commit here. What are your thoughts @w4nderlust?

w4nderlust:
sounds good. Although please add a task to the project board to revise how preprocessing default parameters are set for all features and link to this PR.

connor-mccorm:
Will do! Going to merge this PR so that model training with image features is not blocked for now. Once I get time I can go back and work on the project created and revise how preprocessing default parameters are set for all features.

PR #1895, opened 2022-04-09
fix: Update notebook to use ray instead of deprecated parallel

brightsparc:
# Code Pull Requests

Update the example hyperopt notebook to target the latest code, and include ray hyperopt config.
* add ray executor and hyperopt config, including updating hyperopt parameters like so:
  - `low` -> `lower`
  - `high` -> `upper`
  - type:`float` space:`log` -> space: `loguniform`
  - type: `int`, space: `log` -> space: `qlograndint` for int type
  - type: `int`/w steps -> space: `qrandint` 

The `hyperopt` method now returns object `HyperoptResults`, so I introduced a method to convert to a dict such that results cloud be passed to visualize functions, and also compatibility is maintained when loading from json files. 

```
def hyperopt_results_dict(results):
    return [{"metric_score": t.metric_score, "parameters": t.parameters} for t in results.ordered_trials]
```

Additional refactor could be done to add load json into objects and change the method sig to require objects.

Testing was performed by running the notebook to completion, and inspecting results.

`NOTE`: ray tune process is >10x faster with 3 samples completing in 12 seconds, compared with serial version that vs 2 mins 42 seconds completing only 2 samples.


tgaddair (approved):

brightsparc:
This is merged. 

In wonder in future if we should be adding nbclean and black-no to only commit input cells as whilst the output looks nice, it’s large and could bloat the git repo.

PR #1904, opened 2022-04-12
Fix TextOutputFeature postprocessing predictions

geoffreyangus:
Fixes a bug in the calculation of sentence-level probability in TextOutputFeature.postprocess_predictions. Adds regression test and unit test for SequenceOutputFeature as a whole.


tests/ludwig/features/test_sequence_features.py
```diff
@@ -77,6 +102,39 @@ def test_sequence_input_feature(
     assert encoder_output["encoder_output"].shape == (BATCH_SIZE, *input_feature_obj.output_shape)
 
 
-# todo: add unit test for sequence output feature
-def test_sequence_output_feature():
-    pass
+@pytest.mark.parametrize("sequence_type", [SequenceOutputFeature, TextOutputFeature])
+def test_sequence_output_feature(sequence_type: Union[SequenceOutputFeature, TextOutputFeature]):
```
justinxzhao, line 106: Thanks for adding this.

tests/ludwig/features/test_sequence_features.py
```diff
@@ -77,6 +102,39 @@ def test_sequence_input_feature(
...
+    )
+    output_feature_obj = sequence_type(output_feature_defn, {}).to(DEVICE)
+    combiner_outputs = {}
+    combiner_outputs["combiner_output"] = torch.randn([BATCH_SIZE, SEQ_SIZE, VOCAB_SIZE], dtype=torch.float32).to(
+        DEVICE
+    )
+
+    text_output = output_feature_obj(combiner_outputs, {})
+
+    assert "last_hidden" in text_output
```
justinxzhao, line 118: nit: Use the official ludwig.constants

tests/ludwig/features/test_sequence_features.py
```diff
@@ -77,6 +102,39 @@ def test_sequence_input_feature(
...
+    postprocessed_result = output_feature_obj.postprocess_predictions(result, {"idx2str": idx2str}, "", LOCAL_BACKEND)
+
+    assert len(postprocessed_result) == BATCH_SIZE
+    assert f"{output_feature_obj.feature_name}_{LENGTHS}" not in postprocessed_result.keys()
+    assert f"{output_feature_obj.feature_name}_{PREDICTIONS}" in postprocessed_result.keys()
+    assert f"{output_feature_obj.feature_name}_{PROBABILITY}" in postprocessed_result.keys()
+    assert np.allclose(
+        probability_expected,
+        np.stack(postprocessed_result[f"{output_feature_obj.feature_name}_{PROBABILITY}"].to_numpy()),
+    )
```
justinxzhao, line 140: I'm not sure how much value this test is adding beyond a sanity check of implementation details and that postprocessing produces something. Maybe the key checks are useful to keep around, but I would decouple the sequence probability calculation into its own test.

In the current snapshot, the expected_probability comes from a close copy/mirror of the real implementation, so we're really just testing that the real implementation produces the same result as a slightly different implementation.

Instead, could we factor out the sentence probability calculation into a separate method (perhaps in `feature_utils.py`), call that method from the `text_feature`, and then add a test for it with real numbers?

In feature_utils:
```
def compute_sequence_probability(sequence_probabilities: Union[list, tuple, np.ndarray], backend) -> float:
    def compute_prob(probs):
        if isinstance(probs, (list, tuple, np.ndarray)):
            for i in range(len(probs)):
                probs[i] = np.max(probs[i])
            return np.prod(probs)
        else:
            return np.prod(probs, axis=-1)

    return backend.df_engine.map_objects(
        sequence_probabilities,
        compute_prob,
    )
```

In test_feature_utils:
```
def test_compute_sequence_probability():
    sequence_probabilities_list = [
        [0.2, 0.3, 0.5],
        [0.1, 0.9, 0.0]
    ]
    sequence_probabilities_tuple = tuple(sequence_probabilities_list)
    sequence_probabilities_ndarray = np.array(sequence_probabilities_tuple)
    expected_probability = 0.5 * 0.9

    assert feature_utils.compute_sequence_probability(
        sequence_probabilities_list) == expected_probability
    assert feature_utils.compute_sequence_probability(
        sequence_probabilities_tuple) == expected_probability
    assert feature_utils.compute_sequence_probability(
        sequence_probabilities_ndarray) == expected_probability
```

tests/ludwig/features/test_sequence_features.py
geoffreyangus: Done.

tests/ludwig/features/test_sequence_features.py
geoffreyangus: Np!

tests/ludwig/features/test_sequence_features.py
geoffreyangus: Agreed, good idea!

tests/ludwig/features/test_feature_utils.py
```diff
@@ -35,3 +37,20 @@ def test_ludwig_feature_dict_with_periods():
...
+@pytest.mark.parametrize("sequence_type", [list, tuple, np.ndarray])
+def test_compute_sequence_probability(sequence_type):
+    inputs = [
+        [0.1, 0.2, 0.7],
+        [0.3, 0.4, 0.3],
+        [0.6, 0.3, 0.2],
+    ]
+    if sequence_type == tuple:
+        inputs = tuple(inputs)
+    elif sequence_type == np.ndarray:
```
justinxzhao, line 51: nit: Avoid logic in tests:
https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests

Since there's only 3 types of input, I would prefer removing the parameterization and list out the different possible inputs explicitly, or change `np.ndarray` to `np.array` and then do

```
inputs = sequence_type([
        [0.1, 0.2, 0.7],
        [0.3, 0.4, 0.3],
        [0.6, 0.3, 0.2],
    ])

...
```

justinxzhao (commented):
Thanks for the changes! One last nit for me.

tests/ludwig/features/test_feature_utils.py
geoffreyangus: Got it, good tip! I made the latter change.

tests/ludwig/features/test_sequence_features.py
```diff
@@ -38,6 +40,27 @@ def input_sequence() -> Tuple[torch.Tensor, List]:
     return input_tensor, idx2str
 
 
+@pytest.fixture(scope="module")
+def output_result():
```
justinxzhao, line 44: This can be removed now.

justinxzhao (approved):

tests/ludwig/features/test_sequence_features.py
geoffreyangus: Done.

PR #1913, opened 2022-04-13
Adds Torchscript-enabled BPE tokenizers

geoffreyangus:
This PR adds the final two tokenizers officially supported by torchtext, `CLIPTokenizer` and `GPT2BPETokenizer`.

The primary challenge in this work is that these two tokenizers behave slightly different than all other tokenizers supported by Ludwig. Out of the box, they return a `List[List[str]]` of _token ids_ directly (e.g. `[['23', '54', '76'], ['98', '1']]`). The rest of the Ludwig ecosystem expects tokenizers to return a `List[List[str]]` of tokens (e.g. `[['quick', 'brown', 'fox'], ['jump', 'ed']]`).

I work around this issue by making `vocab: List[str]` (which is passed to the tokenizer as a filepath) a tokenizer attribute, similar to [HFTokenizer](https://github.com/ludwig-ai/ludwig/blob/8309bc8f5ee8a92244ab25b2bf8ab70e9fdaa778/ludwig/utils/strings_utils.py#L241). During the forward pass, `vocab` is used to convert the token ids into the original string tokens. This vocab is also used as the vocab in `strings_utils.create_vocabulary`.


ludwig/utils/strings_utils.py
```diff
@@ -258,7 +258,8 @@ def create_vocabulary(
             vocab = [padding_symbol] + vocab
         else:
             padding_symbol = pad_token
-
+    elif tokenizer_type in {"clip_tokenizer", "gpt2bpe_tokenizer"}:
```
tgaddair, line 261: Perhaps we can make this a little more generic:

```
elif hasattr(tokenizer, "get_vocab"):
    vocab = tokenizer.get_vocab()
```

tests/integration_tests/test_torchscript.py
```diff
@@ -228,6 +226,7 @@ def test_torchscript_e2e(csv_filename, tmpdir):
         # set_feature(vocab_size=3),
         # bag_feature(vocab_size=3),
     ]
+    input_features += torchscript_enabled_text_features
```
tgaddair, line 229: Nit: more pythonic would be:

```
input_features = [
    ...,
    *torchscript_enabled_text_features
]
```

tgaddair (commented):
Looks great! Just minor nits.

ludwig/utils/strings_utils.py
geoffreyangus: Done.

tests/integration_tests/test_torchscript.py
geoffreyangus: Done.

tgaddair (approved):

Issue #1914, opened 2022-04-13
Fresh install of ludwig does not work without torchtext and torchvision.  Also psutil missing from requirements.txt


dantreiman:
TL;DR: I recommend we move `torchvision`, `torchtext` and `ray` into the default requirements.txt

My reasoning is this: new users should be able to use all of the features of Ludwig without extra steps, including text, images, and ray parallelism.

---

Ludwig's code today does not isolate the text and vision dependencies, and this easily goes unnoticed because most DL environments with torch have torchtext and torchvision installed already.  Though we could untangle and encapsulate these python deps I do not think it is worth the added complexity.

I think the cost of installing `torchvision` and `torchtext` is trivial compared to the engineering cost of maintaining optional dependencies on these packages and gracefully handling errors when these packages are not installed.

If someone is developing a container, and they aren't using images and they really need to forgo installing `torchvision` to save space, that is a 1% case, and we can provide a way to do that but it should not be the default.  Maybe we could make a `ludwig-serve` package that strips away all but the minimial dependencies for serving?  Just a thought.

---

Here is the log of a new install I just ran today:

In a new conda environment `temp39`, I installed ludwig `pip install -e .` then try to train higgs:

```
(temp39) daniel@Daniel-M1-Pro higgs % python train_higgs_small.py
Note: NumExpr detected 10 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
NumExpr defaulting to 8 threads.
Traceback (most recent call last):
  File "/Users/daniel/Desktop/github/dantreiman-ludwig/examples/tabnet/higgs/train_higgs_small.py", line 3, in <module>
    from ludwig.api import LudwigModel
  File "/Users/daniel/Desktop/github/dantreiman-ludwig/ludwig/api.py", line 38, in <module>
    from ludwig.backend import Backend, initialize_backend
  File "/Users/daniel/Desktop/github/dantreiman-ludwig/ludwig/backend/__init__.py", line 20, in <module>
    from ludwig.backend.base import Backend, LocalBackend
  File "/Users/daniel/Desktop/github/dantreiman-ludwig/ludwig/backend/base.py", line 25, in <module>
    from ludwig.models.ecd import ECD
  File "/Users/daniel/Desktop/github/dantreiman-ludwig/ludwig/models/ecd.py", line 10, in <module>
    from ludwig.combiners.combiners import Combiner, get_combiner_class
[...]

PR #1933, opened 2022-04-15
Adds Torchscript-compatible image transforms

geoffreyangus:
This PR adapts many of the image preprocessing functions in `image_utils.py` (namely, `crop`, `pad`, `crop_or_pad`, and `resize_image`) in order to make them torchscript-compatible. Namely, it (1) removes their dependency on Numpy, (2) enforces stricter typing throughout, and (3) brings constants into local scope in the case of `resize_image`.

Note: this PR assumes `torchvision` is a part of the minimal Ludwig installation because `try` blocks and `import` statements are not permitted in torchscript code. Concretely, I've removed the `torchvision` checks in `resize_image` and `grayscale`– however, those checks are not currently doing anything in any case because `import torchvision.transforms.functional as F` is currently already located at the top of the file. **That said, we still probably should NOT merge this in until we reach a resolution for #1914.**

### DIFF
--- a/ludwig/utils/image_utils.py
+++ b/ludwig/utils/image_utils.py
@@ -163,22 +163,20 @@ def read_image_from_str(img: str, num_channels: Optional[int] = None) -> torch.T
 
 def pad(
     img: torch.Tensor,
-    size: Union[int, Tuple[int, int]],
+    new_size: Union[int, Tuple[int, int]],
 ) -> torch.Tensor:
     """torchscript-compatible implementation of pad.
 
     Args:
         img (torch.Tensor): image with shape [..., height, width] to pad
-        size (Union[int, Tuple[int, int]]): size to pad to. If int, resizes to square image of that size.
+        new_size (Union[int, Tuple[int, int]]): size to pad to. If int, resizes to square image of that size.
 
     Returns:
         torch.Tensor: padded image of size [..., size[0], size[1]] or [..., size, size] if size is int.
     """
-    new_size = to_tensor_pair(size)
-
-    old_size = torch.tensor(img.shape[-2:])
-
-    pad_size = (new_size - old_size) / 2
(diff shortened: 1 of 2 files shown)


tgaddair (approved):

ludwig/utils/image_utils.py
```diff
@@ -163,68 +163,110 @@ def read_image_from_str(img: str, num_channels: Optional[int] = None) -> torch.T
...
+    """torchscript-compatible implementation of pad.
+
+    Args:
+        img (torch.Tensor): image with shape [..., height, width] to pad
+        size (Union[int, Tuple[int, int]]): size to pad to. If int, resizes to square image of that size.
+
+    Returns:
+        torch.Tensor: padded image of size [..., size[0], size[1]] or [..., size, size] if size is int.
+    """
+    if torch.jit.isinstance(size, int):
```
ShreyaR, line 177: Since this conditional seems to be common across all preprocessing methods, could we abstract this out? Not sure if that affects torchscript compatibility.

ShreyaR (commented):
Looks good to me! Left a small comment.

ludwig/utils/image_utils.py
geoffreyangus: I think we probably could! Depends somewhat on the use of `pad` and `crop`. I assumed that because they were public functions, they could conceivably be called independently from one another.

On the other hand, if the only entry point into these functions is `resize_image`, then I believe we could abstract it out so the check only happens once in `resize_image`. What do you think?

tgaddair: Maybe a utility function like `to_tensor_pair` whose implementation is these four lines?

geoffreyangus: Ah, I see what you were saying now. Made those changes. Thanks!

PR #1935, opened 2022-04-15
torchscript-enabled ImageInputFeature

geoffreyangus:
This PR adds torchscript-compatible preprocessing to `ImageInputFeature`. Users of `inference.InferenceModule` objects can now pass in a `torch.Tensor` or `List[torch.Tensor]` as image feature input to the `forward` function. Note that this is different than the usual flow which uses image paths for image input features.

In order to make this change, `inference.InferenceModule.forward` must now additionally take `List[torch.Tensor]` as input. This caused minor downstream changes to existing torchscript-compatible preprocessing modules `forward` function headers, such as those for binary, category, number, and text features.

This PR makes use of torchscript-compatible image preprocessing functions, which are implemented in #1933.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -36,7 +36,7 @@ jobs:
           - python-version: 3.9
             pytorch-version: nightly
             torchscript-version: 1.10.2
-            ray-version: 1.12.0rc1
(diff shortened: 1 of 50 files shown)


tgaddair (approved):
LGTM!

PR #1887, opened 2022-04-08
Adds Torchscript-enabled Torchtext Sentence Piece Tokenizer

geoffreyangus:
This PR adds the torchscript-enabled Torchtext Sentence Piece Tokenizer to enable scripted text input feature preprocessing. The new code uses the pattern established #1693. 

Note: this PR requires torchtext 0.12.0 and torch 1.11.0. When updating torch, you may have to update torchvision and torchaudio as well to prevent abi errors.

To test the changes in this PR, one can check out the branch and run the following tests:

```
pytest -rP tests/integration_tests/test_torchscript.py::test_torchscript_e2e
pytest -rP tests/ludwig/features/test_sequence_features.py::test_text_preproc_module
```

PR #1939, opened 2022-04-18
Deletes requirements_[audio|image|text].txt

geoffreyangus:
Broadly, this PR (1) deletes `requirements_audio.txt`, `requirements_image.txt` and `requirements_text.txt`, (2) adds dependencies in those files to `requirements.txt`.

The primary way I tested this PR was by (1) creating a new virtual environment, (2) `pip install -e .`, then ran the text classification CLI [tutorial](http://127.0.0.1:8000/ludwig-docs/examples/text_classification/) from the 0.5 docs for a couple of epochs.

It is possible that this PR may cause downstream issues that I am not entirely familiar with, so please look through this PR carefully. Small details I want to highlight:

1. Removed `sentencepiece` as a dependency. I couldn't find it being used anywhere in the repository. If users are interested in using a sentencepiece tokenizer, we recently added `torchtext`-based implementation (#1887) (requires `torch>=1.11.0` and `torchtext>=0.12.0`).
2. Added `psutil` as a dependency. I attempted to train a model using a fresh install `pip install -e .` and ran into a `ModuleNotFoundError` for this module.

I could see this affecting tutorials and/or existing documentation, as well as potentially any automated installation workflows used in production. Please take a look and confirm.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,11 @@ scikit-learn
 tqdm
 torch==1.9.1 ; platform_system == "Windows"  # https://github.com/pytorch/pytorch/issues/65473
 torch>=1.10.0 ; platform_system != "Windows"
+torchtext
+torchvision==0.10.1 ; platform_system == "Windows"  # https://github.com/pytorch/pytorch/issues/65473
+torchvision>=0.10.1 ; platform_system != "Windows"
+transformers>=4.10.1
+spacy>=2.3
 PyYAML>=3.12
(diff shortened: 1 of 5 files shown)


w4nderlust:
Looks good to me. I would remove also the audio package and add its dependencies. In the future we may remove soundfile entirely if we can.

@hungcs @jeffreyftang @brightsparc for visibility

tgaddair (approved):

PR #1941, opened 2022-04-19
Adds heuristic for deciding number of image channels

dantreiman:
Addresses issue 1940: [Twitter bots dataset fails to load most images because 4 channels not supported for JPEG](https://github.com/ludwig-ai/ludwig/issues/1940)

When inferring number of channels for an image dataset, my assumption is that 3 channels is the most common use case.

This PR changes the behavior to use:

1 channel if the majority of samples are 1 channel
2 channel if the majority of samples are 2 channel
4 channel if the majority are 4 channel
3 otherwise

Also added logging to explain channel inference.  Sample output (twitter bots):
```
Inferring num_channels from the first 100 images for column profile_image_path.
  images with 1 channels: 4
  images with 3 channels: 88
  images with 4 channels: 8
Using 3 channels because it is the majority in sample. If an image with a different depth is read, will attempt to convert to 3 channels.
To explicitly set the number of channels, define num_channels in the preprocessing dictionary of the image input feature config.
```

### DIFF
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -334,9 +334,16 @@ class ImageFeatureMixin(BaseFeatureMixin):
             user_specified_num_channels = False
             if preprocessing_parameters[INFER_IMAGE_DIMENSIONS]:
                 user_specified_num_channels = True
-                # Use the maximum num_channels found across all sampled images. torchvision has built-in support for
-                # upsampling images.
-                num_channels = max(num_channels_in_image(x) for x in inferred_sample)
+                channels_in_sample = np.array([num_channels_in_image(x) for x in inferred_sample])
+                if sum(channels_in_sample == 1) > len(inferred_sample) / 2:
+                    # If the majority of images in sample are 1 channel, use 1.
+                    num_channels = 1
+                elif sum(channels_in_sample == 4) > len(inferred_sample) / 2:
+                    # If the majority of images in sample are 4 channel, use 4.
+                    num_channels = 4
+                else:
+                    # Default case: use 3 channels.
+                    num_channels = 3
             elif first_image is not None:
                 num_channels = num_channels_in_image(first_image)
             else:
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -253,6 +253,43 @@ class ImageFeatureMixin(BaseFeatureMixin):
         return img.numpy()
 
     @staticmethod
+    def _infer_number_of_channels(image_sample: List[torch.Tensor]):
+        """Infers the channel depth to use from a group of images.
+
+        We make the assumption that the majority of datasets scraped from the web will be RGB, so if we get a mixed bag
+        of images we should default to that. However, if the majority of the sample images have a specific channel depth
+        (other than 3) this is probably intentional so we keep it, but log an info message.
+        """
+        n_images = len(image_sample)
+        channel_frequency = Counter([num_channels_in_image(x) for x in image_sample])
+        if channel_frequency[1] > n_images / 2:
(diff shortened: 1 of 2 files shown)


ludwig/features/image_feature.py
```diff
@@ -287,9 +287,19 @@ def _finalize_preprocessing_parameters(
...
-                # Use the maximum num_channels found across all sampled images. torchvision has built-in support for
-                # upsampling images.
-                num_channels = max(num_channels_in_image(x) for x in inferred_sample)
+                channels_in_sample = np.array([num_channels_in_image(x) for x in inferred_sample])
+                if sum(channels_in_sample == 1) > len(inferred_sample) / 2:
+                    # If the majority of images in sample are 1 channel, use 1.
+                    num_channels = 1
+                elif sum(channels_in_sample == 2) > len(inferred_sample) / 2:
+                    # If the majority of images in sample are 2 channel, use 2.
+                    num_channels = 2
```
geoffreyangus, line 296: What is the expected behavior of the image preprocessing step if the majority of images in the sample have 2 channels, but the model comes across an image with 3 channels?

My understanding was that we only cropped channels if num_channels == 1. In that case, we convert it to grayscale: https://github.com/ludwig-ai/ludwig/blob/aacabd3f5406701ecc336692c27c487ed043517c/ludwig/features/image_feature.py#L158

Not sure how we would deal with it otherwise. This is perhaps why we currently take the `max` number of channels (because padding is relatively safe operation, whereas dropping channels is not).

Let me know what you think– I might be missing some context here.

dantreiman: @geoffreyangus If I understand this right, you are describing a situation where the majority of images are 2 channels (grayscale + alpha), and we come across an image with 3 channels (RGB)?

In this situation, dropping the RGB image (along with a warning) is the behavior I would expect.  What do you think the expected behavior should be?

dantreiman: Good ideas.  I'll integrate these changes.  I also think you bring up a generally good idea:  giving users config advice based on the data.  Maybe we should even designate a special log category for this so we can surface this type of useful feedback in predibase.  I'm going to think about this and maybe open a new issue for it, meanwhile I'll use `logging.info`.

dantreiman: Implemented Counter-based checks and logging.  

On my twitter bots example, the output now reads:

```
Inferring num_channels from the first 100 images for column profile_image_path.
  images with 1 channels: 4
  images with 3 channels: 88
  images with 4 channels: 8
Using 3 channels because it is the majority in sample. If an image with a different depth is read, will attempt to convert to 3 channels.
To explicitly set the number of channels, define num_channels in the preprocessing dictionary of the image input feature config.
```

ludwig/features/image_feature.py
```diff
@@ -287,9 +288,37 @@ def _finalize_preprocessing_parameters(
             user_specified_num_channels = False
             if preprocessing_parameters[INFER_IMAGE_DIMENSIONS]:
                 user_specified_num_channels = True
-                # Use the maximum num_channels found across all sampled images. torchvision has built-in support for
-                # upsampling images.
-                num_channels = max(num_channels_in_image(x) for x in inferred_sample)
+                channel_frequency = Counter([num_channels_in_image(x) for x in inferred_sample])
```
justinxzhao, line 291: This control flow is already happening in a rather nested chunk of code. Perhaps consider factoring this out into a separate function, i.e. `get_majority_num_channels()`

dantreiman: Good idea.  Also did the same for image size inference

justinxzhao (approved):

PR #1943, opened 2022-04-19
Use explicit date dtype if provided in automl config

hungcs:
if FieldInfo contains explicit dtype=DATE, we should honor it instead of inferring it as text. It's difficult to infer dates organically since they can be both string timestamps and unix epoch seconds.

In general, maybe we should be honoring user supplied field dtypes if they match ludwig's types

### DIFF
--- a/tests/ludwig/automl/test_base_config.py
+++ b/tests/ludwig/automl/test_base_config.py
@@ -51,7 +51,7 @@ def test_infer_type(num_distinct_values, distinct_values, img_values, missing_va
     assert infer_type(field, missing_vals, ROW_COUNT) == expected
 
 
-def test_infer_type_explicit_date(num_distinct_values, distinct_values, img_values, missing_vals, expected):
+def test_infer_type_explicit_date():
     field = FieldInfo(
         name="foo",
         distinct_values=["1", "2"],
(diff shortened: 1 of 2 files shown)


tests/ludwig/automl/test_base_config.py
```diff
@@ -51,6 +51,46 @@ def test_infer_type(num_distinct_values, distinct_values, img_values, missing_va
...
+        # Mostly random strings.
+        (90, [generate_string(5) for _ in range(90)], 0, 0.0, TEXT),
+        # Mostly random strings with capped distinct values.
+        (90, [generate_string(5) for _ in range(10)], 0, 0.0, TEXT),
+        # All random strings.
+        (ROW_COUNT, [generate_string(5) for _ in range(ROW_COUNT)], 0, 0.0, TEXT),
+        # Images.
+        (ROW_COUNT, [], ROW_COUNT, 0.0, IMAGE),
+    ],
+)
```
justinxzhao, line 82: Remove the copy-pasted the parameterizations. We can just demonstrate this behavior on its own with a single case.

```
    field = FieldInfo(
        name="foo",
        distinct_values=["1", "2"]
        num_distinct_values=2,
        dtype=DATE,
    )
    assert infer_type(field, 0, ROW_COUNT) == DATE
```

PR #1945, opened 2022-04-20
Fixes export.cli_export_torchscript functionality

geoffreyangus:
This PR fixes the `export.cli_export_torchscript` function. 

Prior to this PR, the `LudwigModel.save_torchscript` was called from `export.export_torchscript`, but the method did not exist:
https://github.com/ludwig-ai/ludwig/blob/aacabd3f5406701ecc336692c27c487ed043517c/ludwig/export.py#L47

This PR implements `LudwigModel.save_torchscript`, which creates an `InferenceModule`, converts it to Torchscript, and writes it to `output_path`.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -1451,6 +1451,19 @@ class LudwigModel:
         inference_module = InferenceModule(self.model, self.config, self.training_set_metadata)
         return torch.jit.script(inference_module)
 
+    def save_torchscript(self, save_path: str):
+        """Saves the Torchscript model to disk.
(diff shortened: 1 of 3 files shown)


justinxzhao (commented):
Good catch! Is there a test to write here, potentially in test_api.py?

geoffreyangus:
Good point! Added a test to `test_api.py`.

justinxzhao (approved):

tgaddair (approved):

PR #1946, opened 2022-04-20
Fixes InferenceModule handling of TorchScript-reserved keywords

geoffreyangus:
This PR fixes how `InferenceModule` handles features with keys that make use of TorchScript-reserved keywords such as "class".n

Initially, calls to `torch.jit.load(save_path)` would fail if the model stored at `save_path` contained `nn.ModuleDict` keys that conflicted with the Torchscript-reserved keywords. Our solution is to replace the `nn.ModuleDict` keys with compatible strings.

A first version of this PR attempted to use `feature_utils.LudwigFeatureDict` as a wrapper around the `nn.ModuleDict` objects in `InferenceModule`– however, running `torch.jit.script` with these changes raised errors. My guess is that the errors had to do with the auxiliary methods implemented in the `feature_utils.LudwigFeatureDict` class.

This PR uses the `feature_utils.get_module_dict_key_from_name` and `feature_utils.get_name_from_module_dict_key` functions directly in `InferenceModule` to go back and forth between the `nn.ModuleDict` keys and the feature names required.


ludwig/models/inference.py
```diff
@@ -24,42 +27,41 @@ def __init__(self, model: ECD, config: Dict[str, Any], training_set_metadata: Di
         input_features = {
             feature[NAME]: get_from_registry(feature[TYPE], input_type_registry) for feature in config["input_features"]
         }
-        self.preproc_modules = nn.ModuleDict(
-            {
-                feature_name: feature.create_preproc_module(training_set_metadata[feature_name])
-                for feature_name, feature in input_features.items()
-            }
-        )
+        self.preproc_modules = nn.ModuleDict()
```
justinxzhao, line 30: Instead of using `ModuleDict` and calling `get_module_dict_key_from_name()`, would it be simpler to use `LudwigModuleDict()`, which should work exactly like a regular `ModuleDict` that takes care of suffixing/desuffixing for you.

Example usage:
https://github.com/ludwig-ai/ludwig/blob/master/tests/ludwig/features/test_feature_utils.py#L8-L15

geoffreyangus: Hi Justin– yup that's what I initially had, but it seems that some of the methods in `LudwigFeatureDict` are not supported by TorchScript. I did some experimenting both in how I was calling it and in how that class was implemented, but I wasn't able to get it to work. 

I checked out the TorchScript documentation, and it seems like TorchScript only partially supports emulating container types and does not support magic methods like `__iter__` (3.3.7 here: https://pytorch.org/docs/stable/jit_python_reference.html).

justinxzhao: Ah, I see. I'm curious why torchscript seems to be perfectly happy with `ModuleDict`, which also has `__iter__`.

I noticed that some of the magic methods in the `ModuleDict` [source implementation](https://pytorch.org/docs/stable/_modules/torch/nn/modules/container.html#ModuleDict) use a special decorator: `@_copy_to_script_wrapper`.

If you haven't already tried this, could you see if the code works if we add the same decorators to `LudwigFeatureDict`?

geoffreyangus: Hi Justin, I tried to use LudwigFeatureDict with the decorators and it didn't help.

I think there is something more fundamental in the support of nn.ModuleDict that we are not able to tap into. For example, I tried this bit of code:
```
preproc_inputs = {}
for feature_name in self.preproc_modules.keys():
    preproc_inputs[feature_name] = self.preproc_modules[feature_name](inputs[feature_name])
```
where `self.preproc_modules` was a `LudwigFeatureDict` and I received the following error:
```
RuntimeError: 
E       Only ModuleList, Sequential, and ModuleDict modules are subscriptable:
```

Another thing I tried was enumerating the items directly:
```
preproc_inputs = {}
for feature_name, preproc in self.preproc_modules.items():
    preproc_inputs[feature_name] = preproc(inputs[feature_name])
```
But TorchScript doesn't support `torch.nn.Module` as a type annotation, so I can't get `preproc` out of `LudwigFeatureDict`.

I think that we may want to just stick with the implementation as it is, but I can add a TODO comment for myself to figure out how to implement a TorchScript-compatible `LudwigFeatureDict`.

justinxzhao: Ah, alright, looks like we've exhausted several options -- thanks for the extra digging.

Let's stick with the implementation as is, and adding a TODO to look into this more SGTM.

justinxzhao (approved):

PR #1949, opened 2022-04-21
Adds mechanism for calibrating probabilities for category and binary features

dantreiman:
Implements:
 temperature scaling for binary and category outputs
 matrix scaling for category outputs

Example using Twitter Bots w categorical output feature:
_On this example, matrix scaling does not improve ECE, though it improves NLL (probably due to overfitting) so matrix scaling gets rolled back and we get the original uncalibrated probs.  On much larger datasets matrix scaling may yield better results._

Validation Set (used to determine softmax temperature):
calibration_1_vs_all_1

Test set:
calibration_1_vs_all_1

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -561,13 +561,16 @@ class LudwigModel:
                         )
                         if validation_set is not None:
                             # Use backend.createPredictor to ensure we get ray predictor with ray backend
-                            calibrator.train_calibration(validation_set, VALIDATION, save_path=model_dir)
+                            calibrator.train_calibration(validation_set, VALIDATION)
                         else:
                             logger.warning(
                                 "Calibration uses validation set, but not validation split specified. "
                                 "Will use training set for calibration."
                             )
-                            calibrator.train_calibration(training_set, TRAINING, save_path=model_dir)
+                            calibrator.train_calibration(training_set, TRAINING)
+                        if not self.skip_save_model:
+                            model_weights_path = os.path.join(model_dir, MODEL_WEIGHTS_FILE_NAME)
+                            torch.save(self.model.state_dict(), model_weights_path)
 
                     # Unpack train()'s return.
                     # The statistics are all nested dictionaries of TrainerMetrics: feature_name -> metric_name ->
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -557,7 +557,6 @@ class LudwigModel:
                             trainer.model,
                             self.backend,
                             batch_size=trainer.eval_batch_size,
-                            skip_save_model=skip_save_model,
                         )
                         if validation_set is not None:
                             # Use backend.createPredictor to ensure we get ray predictor with ray backend
@@ -568,7 +567,7 @@ class LudwigModel:
                                 "Will use training set for calibration."
                             )
                             calibrator.train_calibration(training_set, TRAINING)
-                        if not self.skip_save_model:
+                        if not skip_save_model:
                             model_weights_path = os.path.join(model_dir, MODEL_WEIGHTS_FILE_NAME)
                             torch.save(self.model.state_dict(), model_weights_path)
 
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -556,7 +556,6 @@ class LudwigModel:
                         calibrator = Calibrator(
                             trainer.model,
                             batch_size=trainer.eval_batch_size,
-                            horovod=trainer.horovod,
                             skip_save_model=skip_save_model,
                         )
                         calibrator.calibration(validation_set, VALIDATION, save_path=model_dir)
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -549,7 +549,10 @@ class LudwigModel:
                         save_path=model_dir,
                     )
 
-                    trainer.calibration(validation_set, VALIDATION)
+                    # Calibrates output feature probabilities on validation set if calibration is enabled.
+                    # Must be done after training, and before final model parameters are saved.
+                    if self.backend.is_coordinator() and validation_set is not None:
+                        trainer.calibration(validation_set, VALIDATION)
 
                     # Unpack train()'s return.
                     # The statistics are all nested dictionaries of TrainerMetrics: feature_name -> metric_name ->
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -552,7 +552,7 @@ class LudwigModel:
                     # Calibrates output feature probabilities on validation set if calibration is enabled.
                     # Must be done after training, and before final model parameters are saved.
                     if self.backend.is_coordinator() and validation_set is not None:
-                        trainer.calibration(validation_set, VALIDATION)
+                        trainer.calibration(validation_set, VALIDATION, save_path=model_dir)
 
                     # Unpack train()'s return.
                     # The statistics are all nested dictionaries of TrainerMetrics: feature_name -> metric_name ->
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -555,18 +555,19 @@ class LudwigModel:
                     if self.backend.is_coordinator():
                         calibrator = Calibrator(
                             trainer.model,
+                            self.backend,
                             batch_size=trainer.eval_batch_size,
                             skip_save_model=skip_save_model,
                         )
                         if validation_set is not None:
                             # Use backend.createPredictor to ensure we get ray predictor with ray backend
-                            calibrator.calibration(validation_set, VALIDATION, save_path=model_dir)
+                            calibrator.train_calibration(validation_set, VALIDATION, save_path=model_dir)
                         else:
                             logger.warning(
                                 "Calibration uses validation set, but not validation split specified. "
                                 "Will use training set for calibration."
                             )
-                            calibrator.calibration(training_set, TRAINING, save_path=model_dir)
+                            calibrator.train_calibration(training_set, TRAINING, save_path=model_dir)
 
                     # Unpack train()'s return.
                     # The statistics are all nested dictionaries of TrainerMetrics: feature_name -> metric_name ->
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -549,6 +549,8 @@ class LudwigModel:
                         save_path=model_dir,
                     )
 
+                    trainer.calibration(validation_set, VALIDATION)
+
                     # Unpack train()'s return.
                     # The statistics are all nested dictionaries of TrainerMetrics: feature_name -> metric_name ->
                     # List[TrainerMetric], with one entry per training checkpoint, according to steps_per_checkpoint.
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -63,6 +63,7 @@ from ludwig.globals import (
     set_disable_progressbar,
     TRAIN_SET_METADATA_FILE_NAME,
 )
+from ludwig.models.calibrator import Calibrator
 from ludwig.models.ecd import ECD
 from ludwig.models.inference import InferenceModule
 from ludwig.models.predictor import (
@@ -552,7 +553,13 @@ class LudwigModel:
                     # Calibrates output feature probabilities on validation set if calibration is enabled.
                     # Must be done after training, and before final model parameters are saved.
                     if self.backend.is_coordinator() and validation_set is not None:
-                        trainer.calibration(validation_set, VALIDATION, save_path=model_dir)
+                        calibrator = Calibrator(
+                            trainer.model,
+                            batch_size=trainer.eval_batch_size,
+                            horovod=trainer.horovod,
+                            skip_save_model=skip_save_model,
+                        )
+                        calibrator.calibration(validation_set, VALIDATION, save_path=model_dir)
 
                     # Unpack train()'s return.
                     # The statistics are all nested dictionaries of TrainerMetrics: feature_name -> metric_name ->
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -552,13 +552,21 @@ class LudwigModel:
 
                     # Calibrates output feature probabilities on validation set if calibration is enabled.
                     # Must be done after training, and before final model parameters are saved.
-                    if self.backend.is_coordinator() and validation_set is not None:
+                    if self.backend.is_coordinator():
                         calibrator = Calibrator(
                             trainer.model,
                             batch_size=trainer.eval_batch_size,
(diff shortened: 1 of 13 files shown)


w4nderlust:
In the image I can't see the blue points, do they overlap with the green ones?

dantreiman:
> In the image I can't see the blue points, do they overlap with the green ones?

Thats right, in this example the matrix scaling got rolled back to the original uncalibrated probabilities.

ludwig/features/base_feature.py
```diff
@@ -308,11 +310,20 @@ def _setup_metrics(self):
             },
         }
 
+    def create_calibration_module(self, feature) -> CalibrationModule:
+        """Creates and returns a CalibrationModule that converts raw model outputs (logits) to probabilities."""
+        return None
+
+    @property
+    def calibration_module(self) -> torch.nn.Module:
+        """Returns the PredictModule used to convert model outputs to predictions."""
```
w4nderlust, line 319: This docstring is probably copypasted, needs to be changed

ludwig/features/base_feature.py
```diff
@@ -441,6 +452,15 @@ def forward(
...
         pass
 
+    def calibrate(self, logits, labels):
+        """Calibrate this feature's output probabilities given the feature logits and labels.
+
+        The default behavior is to do nothing, calibration (i.e. temperature scaling, vector scaling, etc..) may be
+        optionally supported for some feature types.
+        """
+        if self.calibration_module:
+            self.calibration_module.calibrate(logits, labels)
```
w4nderlust, line 462: Here we probably need some form of warning at the very least that says that this output feature doesn't have a calibration module, so calling calibrate has no effect. Could even be a valueerror, depending on how strict do we want to be.

ludwig/features/binary_feature.py
```diff
@@ -284,8 +290,14 @@ def loss_kwargs(self):
             confidence_penalty=self.loss["confidence_penalty"],
         )
 
+    def create_calibration_module(self, feature) -> torch.nn.Module:
+        calibration_type = feature.get("calibration")
+        if calibration_type == "temperature_scaling":
+            return calibration.TemperatureScaling(binary=True)
+        return None
```
w4nderlust, line 297: here and in the category feature too we can create a registry like we do for most other things and use the functions to get values from registries because they already deal with wrong keys by returning errors. That would help 1) consistency and 2) avoiding silent errors like misspelled calibration type names.

ludwig/features/category_feature.py
```diff
@@ -234,8 +243,16 @@ def logits(self, inputs, **kwargs):  # hidden
         # hidden: shape [batch_size, size of final fully connected layer]
         return {LOGITS: self.decoder_obj(hidden), PROJECTION_INPUT: hidden}
 
+    def create_calibration_module(self, feature) -> torch.nn.Module:
+        calibration_type = feature.get("calibration")
+        if calibration_type == "temperature_scaling":
+            return calibration.TemperatureScaling(num_classes=feature["num_classes"])
+        elif calibration_type == "matrix_scaling":
+            return calibration.MatrixScaling(num_classes=feature["num_classes"])
+        return None
```
w4nderlust, line 252: here and in the binary feature too we can create a registry like we do for most other things and use the functions to get values from registries because they already deal with wrong keys by returning errors. That would help 1) consistency and 2) avoiding silent errors like misspelled calibration type names.

ludwig/models/calibrator.py
```diff
@@ -0,0 +1,57 @@
...
+
+class Calibrator:
+    """Calibrator calibrates the output probabilities of a model."""
+
+    def __init__(self, model: ECD, batch_size: int = 128, skip_save_model=False):
+        self.model = model
+        self.batch_size = batch_size
+        self.skip_save_model = skip_save_model
+
+    def calibration(self, dataset, dataset_name: str, save_path: str):
```
w4nderlust, line 35: Not 100% sure, but it seems to me that in some cases calibration is the act of calibrating outputs, in some cases calibration is the object/function that can calibrate and in some other cases it is the act of training the calibrator.

We probably should be a bit more explicit to avoid confusion.
Wdyt about `train_calibrator`, `get_calibrated_probabilities_from_logits` or something else?
Those may be too verbose, but we can brainstorm a bit on this

ludwig/models/calibrator.py
```diff
@@ -0,0 +1,57 @@
...
+            return
+        predictor = Predictor(self.model, batch_size=self.batch_size)
+        metrics, predictions = predictor.batch_evaluation(
+            dataset, collect_predictions=True, collect_logits=True, collect_labels=True, dataset_name=dataset_name
+        )
+        for output_feature in self.model.output_features.values():
+            feature_logits_key = "%s_logits" % output_feature.feature_name
+            if feature_logits_key in predictions:
+                feature_logits = predictions[feature_logits_key]
+                feature_labels = predictions["%s_labels" % output_feature.feature_name]
```
w4nderlust, line 51: I believe in most of the rest of the code we use f"" for formatting strings, maybe we can do the same here for consistency

ludwig/models/calibrator.py
```diff
@@ -0,0 +1,57 @@
...
+                output_feature.calibrate(
+                    np.stack(feature_logits.values, axis=0), np.stack(feature_labels.values, axis=0)
```
w4nderlust, line 53: Couple notes:
- what's the type of feature_logits.values and feature_labels.values here? Wouldn't it be better to use pytorch bjects here instead of numpy?
- this looks like a 1 time operation on all the dataset after all the predictions on all the datasets have been computed. There could be a scalability bottleneck here for big datasets. it's true that both logits and labels will be small, but I'm curious to test this on, say, criteo, to make sure we don't go out of memory, and also that this works with the remote7ray backend (ideally with raydatasets and out of memory datasets). In case, maybe there's a way to make this process work on minibatches instead than on the full dataset?

ludwig/models/calibrator.py
```diff
@@ -0,0 +1,57 @@
...
+                )
+        if not self.skip_save_model:
+            model_weights_path = os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME)
+            torch.save(self.model.state_dict(), model_weights_path)
```
w4nderlust, line 57: Trying to understand the logic here (maybe add a comment about it too).
Why are we saving the model here?
Did we, at this stage, already save the model and this overrides it to add the calibration module?
Can't we just do this before we save the model so that we don't have to save it inside this function?
if we can make this function only about calibrating and then we make sure that outside of here the model is saved correctly with and without calibration, it seems like a cleaner approach to me.

ludwig/models/predictor.py
```diff
@@ -194,9 +209,14 @@ def batch_evaluation(self, dataset, collect_predictions=False, dataset_name=None
...
                             for pred_name, pred_values in of_preds.items():
-                                if pred_name not in EXCLUDE_PRED_SET:
+                                if collect_logits or pred_name not in EXCLUDE_PRED_SET:
                                     key = f"{of_name}_{pred_name}"
                                     predictions[key].append(pred_values)
+                    # accumulate labels from batch for each output feature
+                    if collect_labels:
+                        for of_name, of_labels in targets.items():
+                            key = f"{of_name}_labels"
+                            predictions[key].append(of_labels)
```
w4nderlust, line 219: I understand why this is needed, but there's an alternative. Outside of this function you could collect the labels by appending the labels column from the dataset to the output of this function. it doesn't need to live inside this predict function imho, but it could live after line 44-46 in calibrator.py

ludwig/models/trainer.py
```diff
@@ -1098,6 +1098,30 @@ def append_metrics(self, dataset_name, results, metrics_log, tables, progress_tr
 
         return metrics_log, tables
 
+    def calibration(self, dataset, dataset_name: str, save_path: str):
```
w4nderlust, line 1101: This function look suspiciously similar to `Calibrator.calibration`. Do we need both? Can't we use only one?

ludwig/utils/calibration.py
```diff
@@ -0,0 +1,261 @@
...
+    In each bin, we compute the confidence gap:
+    bin_gap = | avg_confidence_in_bin - accuracy_in_bin |
+    We then return a weighted average of the gaps, based on the number
+    of samples in each bin
+    See: Naeini, Mahdi Pakdaman, Gregory F. Cooper, and Milos Hauskrecht.
+    "Obtaining Well Calibrated Probabilities Using Bayesian Binning." AAAI.
+    2015.
+    """
+
+    def __init__(self, n_bins: int = 15):
```
w4nderlust, line 39: if i understand correctly these bins are intervals of confidence. Any specific reason for choosing 15? Why not 20 so that they are at increments of .05?

ludwig/api.py
```diff
@@ -539,6 +540,16 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                         save_path=model_dir,
                     )
 
+                    # Calibrates output feature probabilities on validation set if calibration is enabled.
+                    # Must be done after training, and before final model parameters are saved.
+                    if self.backend.is_coordinator() and validation_set is not None:
```
w4nderlust, line 545: This calibrates on the val set if it exists, which is ok, but if there's no val set, i think calibrating on the training set is better than not calibrating at all. May be there's a message that say something like "calibrating on val set" or "no val set, calibrating on train set instead"

ludwig/utils/calibration.py
dantreiman: 15 the number of bins used in all of the benchmarks in "On Calibration of Modern Neural Networks". 

BTW "Obtaining well calibrated probabilities using bayesian binning" uses 10 bins.

Added a reference to the docstring.

ludwig/models/trainer.py
dantreiman: Good catch.  I moved this from Trainer to Calibrator, hadn't deleted it from Trainer.

ludwig/models/calibrator.py
dantreiman: done

ludwig/models/calibrator.py
dantreiman: My logic was to save early and save often... train model, save, train calibration, save again

You are right that this shouldn't live in calibrator.  I moved this code from trainer, I think I meant to factor saving out but I forgot...  Moved saving logic our into the caller (api.py)

ludwig/models/calibrator.py
dantreiman: feature_logits is a `pd.Series` whose elements are numpy arrays (vectors).  `feature_logits.values` yields a numpy array (rows) of numpy arrays (classes), and `np.stack` turns it into a matrix of (rows, classes).

As to the memory concern, as soon as this is in I'm planning to work on another version which uses batched training for optimizing the calibration temperature.

ludwig/models/predictor.py
dantreiman: I'm not sure how to get the labels column without: initializing a batcher, reading through the dataset batch-by-batch, and accumulating the prediction columns.  Seemed to me that tacking this onto the predict loop in Predictor would be better code re-use, and not add much cost to batch_evaluation.

Is there a faster way to just get the feature's label column from dataset?

ludwig/models/calibrator.py
dantreiman: I like `train_calibrator`, that is clear.  `get_calibrated_probabilities_from_logits` is a good name IMO, where are you thinking that should go (which method would be renamed to this)?  Right now it happens in the PredictModule which just calls the `forward` method of the calibration module.

ludwig/models/predictor.py
dantreiman: Aha, I figured out how to do this.

ludwig/models/predictor.py
```diff
@@ -155,7 +155,19 @@ def _concat_preds(self, predictions):
             # is a tensor that requires grad.
             predictions[key] = torch.cat(pred_value_list, dim=0).clone().detach().cpu().numpy()
 
-    def batch_evaluation(self, dataset, collect_predictions=False, dataset_name=None):
+    def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=False, dataset_name=None):
+        """Batch evaluate model on dataset.
+
+        Params:
```
justinxzhao, line 161: Nice.

ludwig/utils/calibration.py
```diff
@@ -0,0 +1,308 @@
...
+            in_bin = confidences.gt(bin_lower.item()) * confidences.le(bin_upper.item())
+            prop_in_bin = in_bin.float().mean()
+            if prop_in_bin.item() > 0:
+                accuracy_in_bin = accuracies[in_bin].float().mean()
+                avg_confidence_in_bin = confidences[in_bin].mean()
+                ece += torch.abs(avg_confidence_in_bin - accuracy_in_bin) * prop_in_bin
+        return ece
+
+
+@dataclass()
```
justinxzhao, line 106: `@dataclass()` or `@dataclass`

ludwig/utils/calibration.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+
+    def __init__(self, n_bins: int = 15):
+        """n_bins (int): number of confidence interval bins."""
+        super().__init__()
+        bin_boundaries = torch.linspace(0, 1, n_bins + 1)
+        self.bin_lowers = bin_boundaries[:-1]
+        self.bin_uppers = bin_boundaries[1:]
+
+    def forward(self, logits, one_hot_labels):
```
justinxzhao, line 89: nit: Add type hints

ludwig/utils/calibration.py
```diff
@@ -0,0 +1,308 @@
...
+@register_calibration("temperature_scaling", [BINARY, CATEGORY], default=True)
+class TemperatureScaling(CalibrationModule):
+    """Implements temperature scaling of logits. Based on results from "On Calibration of Modern Neural Networks":
+    https://arxiv.org/abs/1706.04599. Temperature scaling scales all logits by the same constant factor. Though it
+    may modify output probabilities it will never change argmax or categorical top-n predictions. In the case of
+    binary classification with a threshold, however, calibration may change predictions.
+
+    Implementation inspired by https://github.com/gpleiss/temperature_scaling
+
+        Args:
```
justinxzhao, line 132: nit: out-indent 1 level.

justinxzhao (approved):
This looks great! Also, it's awesome that it's all inherently torchscriptable, too.

PR #1950, opened 2022-04-21
Torchaudio Refactor

connor-mccorm:


ludwig/data/dataset_synthesizer.py
```diff
@@ -270,14 +261,16 @@ def generate_audio(feature):
     sampling_rate = 16000
     num_samples = int(audio_length * sampling_rate)
     audio = np.sin(np.arange(num_samples) / 100 * 2 * np.pi) * 2 * (np.random.random(num_samples) - 0.5)
+    audio_tensor = torch.tensor(np.array([audio]).astype("float32"))
```
geoffreyangus, line 264: nit question: should we use torch.float32 here instead of the string argument?

connor-mccorm: I feel like we should, I'll go ahead and change

geoffreyangus (approved):

PR #1952, opened 2022-04-22
Skip checkpoints when skip_save_progress=True

tgaddair:


justinxzhao (approved):

ludwig/models/trainer.py
```diff
@@ -978,7 +978,7 @@ def train(self, training_set, validation_set=None, test_set=None, save_path="mod
                     progress_tracker.epoch += 1
                     self.callback(lambda c: c.on_epoch_end(self, progress_tracker, save_path))
 
-                    if self.is_coordinator():
+                    if self.is_coordinator() and not self.skip_save_progress:
```
w4nderlust, line 981: we would want to print the epoch time even if we are not saving, so probably two sub ifs

PR #1953, opened 2022-04-22
Adds twitter bots dataset and training example

dantreiman:


ludwig/datasets/twitter_bots/config.yaml
```diff
@@ -0,0 +1,5 @@
+version: 1.0
+competition: danieltreiman/twitter-human-bots-dataset
```
justinxzhao, line 2: Nice.

justinxzhao, line 2: Nice.

justinxzhao (approved):

Issue #1956, opened 2022-04-23
[bug[[distributed] Fix distributed metrics aggregation for custom metrics


tgaddair:
Currently, for metrics we define ourselves, most of the aggregations are using a typical "mean" aggregation. However, for metrics such as RMSPE, this is not correct. We should instead be collecting sufficient statistics, aggregating over those across workers, and then reporting the true value.

It also appears that loss metrics are not being aggregated correctly.

PR #1957, opened 2022-04-23
Added distributed evaluation on Ray

tgaddair:
Also adds tests to ensure approximate equivalent between distributed and local eval metrics computation.

Also uncovered a few issues:

#1956
https://github.com/ray-project/ray/issues/24139

For some reason, Ray 1.9.2 produces different results for metric aggregations, which may be related to some data loading bugs we uncovered in older versions of Ray Datasets. As such, we are also dropping support for Ray < 1.11 in this PR.

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import copy
 import logging
 from distutils.version import LooseVersion
 from functools import partial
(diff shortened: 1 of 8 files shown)


ludwig/backend/ray.py
```diff
@@ -407,8 +452,34 @@ def to_tensors(df: pd.DataFrame) -> pd.DataFrame:
...
-    def batch_evaluation(self, dataset, collect_predictions=False, **kwargs):
-        raise NotImplementedError("Ray backend does not support batch evaluation at this time.")
+    def batch_evaluation(self, dataset: RayDataset, collect_predictions: bool = False, **kwargs):
+        runner = Trainer(**{**get_trainer_kwargs(), **self.trainer_kwargs})
+        runner.start()
+        try:
+            datasets = {"eval": dataset.pipeline(shuffle=False, **self.data_loader_kwargs)}
+            predictor_kwargs = {
+                **self.predictor_kwargs,
+                "collect_predictions": False,
```
w4nderlust, line 462: Not 100% sure why we set collect_predictions to False here, but later (480), it it was true, then we run inference again to collect them. What's the rationale?

tgaddair: We need to be a Horovod context to collect the aggregated metrics, since it relies on collective communication ops. However, Horovod is not suitable for transforming one big dataset to another. For that you need something like Dask or Ray Datasets. So we need to break these up into two different steps. I can add a comment to this effect.

w4nderlust: Got it, this may make the process quite slower when collect predictions is true I imagine, but that's fine, if the batch predict at line 480 as runs in parallel, this will already be faster with 3 machines than it would be with one.

Probably a clarifying comment here and around 480 would be useful as other people encountering these may thing it's a redundant thing.

tgaddair: Comment added. Yeah, will be interesting to see how much of penalty it imposes to break these steps up. One nice thing about Ray Datasets is we can typically cache the whole dataset in the cluster, so two passes over the data ends up being a lot less expensive than going to remote files each time. But still would be nice to find a way to do it in one go.

PR #1961, opened 2022-04-25
refactor: Remove marshmallow extraction scripting, move docstrings into marshmallow field classes.

ksbrar:
Removes marshmallow extraction entirely, instead insert docstrings directly into marshmallow fields. Tradeoff: no more editor introspection without duplicating docstrings, but no more extraction/scripting pains either.


justinxzhao (approved):
LGTM! Passing in descriptions as strings directly seems like a great improvement over using docstrings and having to parse them ourselves.

tgaddair (approved):

PR #1963, opened 2022-04-26
refactor: Rename, reorganize schema module

ksbrar:
Redo of https://github.com/ludwig-ai/ludwig/pull/1936 (which was reverted by 6d8474abeb85a2edb91c9e95ce17bd55e63eb135) . This ends up being a significant reorganization:

#### Changes
* (1) Moves `ludwig.utils.schema` into `ludwig.marshmallow`, and (2) Renames `ludwig.marshmallow` module to `ludwig.schema` and `ludwig/marshmallow/marshmallow_schema_utils.py` to `ludwig/schema/utils.py`. Also, for convenience, all of the functionality inside of the current `schema.py` (namely, `get_schema()` which returns the full Ludwig schema and `validate_config()`) lives inside `ludwig/schema/__init__.py` to make imports elsewhere more readable.
* Moves `TrainerConfig` and `get_trainer_jsonschema()` into `ludwig/schema/trainer.py`.
* Moves all optimizer configs out of `ludwig/modules/optimization_modules.py` and into `ludwig/schema/optimizers.py`'
* Moves `sequence_encoder_registry` out of `ludwig/combiners/combiners.py`, converts it into a full-fledged `Registry()`, and places it in `ludwig/encoders/registry.py`.
* Moves all combiner configs out of `ludwig/combiners/combiners.py` and into a new `ludwig/schema/combiners/` directory (module) with a file per individual combiner. Again, to make things more readable elsewhere, `__init__.py` is set so that a user can import directly from this submodule (e.g. a user can write `from ludwig.schema.combiners import ConcatCombinerConfig`).

#### Misc.
* Small change to `train_with_backend` integration test.
* Some docstring/comment changes.

#### Overview:
* The structure of new `schema` module is the following:
```
ludwig.schema               <-- Meant to contain all schemas, utilities, helpers related to describing and validating Ludwig configs.
├── __init__.py             <-- Contains fully assembled Ludwig schema (`get_schema()`), `validate_config()` for YAML validation, and all "top-level" schema functions.
├── utils.py                <-- An extensive set of marshmallow-related fields, methods, and schemas that are used elsewhere in Ludwig.
├── trainer.py              <-- Contains `TrainerConfig()` and `get_trainer_jsonschema`
├── optimizers.py           <--- Contains every optimizer config (e.g. `SGDOptimizerConfig`, `AdamOptimizerConfig`, etc.) and related marshmallow fields/methods.
└── combiners/
    ├── __init__.py         <-- Imports for each combiner config file (making imports elsewhere more convenient).
    ├── utils.py            <-- New location of `combiner_registry`, `get_combiner_jsonschema()`, `get_combiner_conds()`
    ├── base.py             <-- New location of `BaseCombinerConfig`
    |
    ├──  comparator.py      <-- New location of `ComparatorCombinerConfig`
    ... <file for each combiner> ...
    └──  transformer.py     <-- New location of `TransformerCombinerConfig`
```         

### ~~Do not merge until after https://github.com/ludwig-ai/ludwig/pull/1961 goes through!~~ Ready for review.


ludwig/combiners/combiners.py
```diff
@@ -122,7 +122,7 @@ def output_shape(self) -> torch.Size:
         return output_tensor["combiner_output"].size()[1:]
 
 
-class BaseCombinerConfig(schema.BaseMarshmallowConfig):
+class BaseCombinerConfig(utils.BaseMarshmallowConfig):
```
justinxzhao, line 125: nit: Consider renaming the file to `schema_utils.py` because `utils` is collision-prone:

https://google.github.io/styleguide/cppguide.html#Namespace_Names

ksbrar: Rename the file itself or just alias to to schema_utils?

justinxzhao: Both, I think. It seems like an improvement if we can avoid having multiple files with the same name (`utils.py`) in the ludwig repo.

ksbrar: Sounds good to me. CC @tgaddair @w4nderlust ?

justinxzhao:
Could you add your directory structure description as header `"""` comments in the relevant files (perhaps replacing the copyright header, which we shouldn't add for new files)?

```
ludwig.schema               <-- Meant to contain all schemas, utilities, helpers related to describing and validating Ludwig configs.
├── utils.py                <-- An extensive set of marshmallow-related fields, methods, and schemas that are used elsewhere in Ludwig.
├── schema.py               <-- Contains the fully assembled Ludwig schema and validate() function that is used for user-input YAML validation. Users should generally only need to look at this.
└── __init__.py
```

ksbrar:
> Could you add your directory structure description as header `"""` comments in the relevant files (perhaps replacing the copyright header, which we shouldn't add for new files)?
> 
> ```
> ludwig.schema               <-- Meant to contain all schemas, utilities, helpers related to describing and validating Ludwig configs.
> ├── utils.py                <-- An extensive set of marshmallow-related fields, methods, and schemas that are used elsewhere in Ludwig.
> ├── schema.py               <-- Contains the fully assembled Ludwig schema and validate() function that is used for user-input YAML validation. Users should generally only need to look at this.
> └── __init__.py
> ```

Added

ludwig/combiners/combiners.py
tgaddair: Since it's correctly scoped to the `schema` module, I would leave the filename as `utils.py`, but change the import to:

```
```

Does that work?

tgaddair: Alternatively, you can import the relevant classes and functions from `utils.py` into `ludwig/schema/__init__.py` and then do:

```
from ludwig import schema
```

I personally like this route a bit more.

ludwig/models/trainer.py
```diff
@@ -442,11 +450,11 @@ def train_step(
     def clip_grads(self, variables):
         """Applies gradient clipping."""
         if self.gradient_clipping_config.clipglobalnorm:
-            torch.nn.utils.clip_grad_norm_(variables, self.gradient_clipping_config.clipglobalnorm)
+            torch.nn.schema_utils.clip_grad_norm_(variables, self.gradient_clipping_config.clipglobalnorm)
```
tgaddair, line 453: Looks like a replace-all error ;)

ludwig/schema/schema.py
```diff
@@ -1,18 +1,13 @@
-#! /usr/bin/env python
-# Copyright (c) 2020 Uber Technologies, Inc.
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
+# Module structure:
```
tgaddair, line 2: `ludwig/schema/schema.py` is weird. can you move this function into `ludwig/schema/__init__.py` instead?

tgaddair (commented):
Looks great, just two things to be fixed.

ludwig/models/trainer.py
ksbrar: fixed

ludwig/schema/schema.py
ksbrar: moved

ksbrar: this actually causes a circular import error

>  I don't know enough about how python does submodule initialization but it seems as though to import schema.utils (which is used in e.g. the combiner and trainer files) it will also pull in that init file (which itself pulls in schemas from those trainer/combiner files)

> something I talked about with @justinxzhao in the past was moving the trainer, combiner, optimizer schemas into this new ludwig.schema module as well. I was going to do that in a follow up PR, but should I do this now?

justinxzhao (approved):

ludwig/schema/schema.py
ksbrar: Going to move the other schemas over into this module, per conversation with @tgaddair

tgaddair (approved):

PR #1966, opened 2022-04-26
add AutoML regression testing mechanism

jppgks:
This introduces 2 ways of verifying that a change in AutoML logic did not introduce errors in type inference on known datasets:
1) Using the pytest in test_type_inference.py
2) The expected types for each of the tested datasets are generated with AutoML by scripts/update_golden_types.py and checked in to the folder golden/. Running scripts/update_golden_types.py should result in a clean git diff.


tests/regression_tests/automl/test_type_inference.py
```diff
@@ -0,0 +1,26 @@
+import json
+
+import pytest
+
+from ludwig.automl.automl import create_auto_config
+from tests.regression_tests.automl.utils import get_dataset_golden_types_path, get_dataset_object, REGISTRY
+
+
+@pytest.mark.parametrize("dataset_name", REGISTRY)
+def test_ludwig_auto_config(dataset_name):
```
justinxzhao, line 10: nit: Add "type inference" to the test name, i.e. `test_auto_type_inference_regression`

tests/regression_tests/automl/utils.py
```diff
@@ -0,0 +1,14 @@
+from pathlib import Path
+
+from ludwig.datasets import dataset_registry
+from ludwig.datasets.base_dataset import BaseDataset
+
+REGISTRY = {"adult_census_income", "mnist"}
```
justinxzhao, line 6: nit: rename to `TEST_DATASET_REGISTRY` and add a comment about what this registry is used for, i.e. "Subset of Ludwig Zoo datasets use for automl type inference regression tests."

tests/regression_tests/automl/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def get_dataset_golden_types_path(dataset_name: str) -> str:
```
justinxzhao, line 9: nit: Add docstring.

tests/regression_tests/automl/utils.py
```diff
@@ -0,0 +1,14 @@
...
+    return str(Path(__file__).resolve().parent / "golden" / f"{dataset_name}.types.json")
+
+
+def get_dataset_object(dataset_name: str) -> BaseDataset:
```
justinxzhao, line 13: nit: Add docstring.

justinxzhao (commented):
Looks like a solid regression test setup for type inference checking to me. Just a few nits.

tests/regression_tests/automl/utils.py
w4nderlust: agree, there are very many registries in the codebase so this helps dismbiguate ;)

w4nderlust:
After nits are addressed, this looks good to me.
Will make sure to mark this as SLOW, or with a new specific decorator.
Downloading Mnist (and other datasets like amazon reviews, yahoo etc.) could be quite slow and potentially flaky depending on connections speed, so would not include these tests as default, but suggest to run them only when modifying automl type inference

justinxzhao (approved):
LGTM after some minor comments!

jppgks:
@justinxzhao existing comments are addressed in the latest changes. Can you resolve if you're happy with the changes?

justinxzhao:
@w4nderlust Agreed that minimizing flakiness is really important. At the moment, we have a few non-`@slow` tests in `test_text_encoders` that downloads weights from huggingface (O(10-100 MB)). We run into some issues with the tests rarely, but they have been largely stable. Perhaps we should rethink some of those tests as well.

@jppgks As Piero suggests, let's introduce a new decorator for these automl type inference regression tests, and only run them if there are changes to any files under `ludwig/automl`.

https://github.community/t/is-it-possible-to-run-the-job-only-when-a-specific-file-changes/115484

jppgks:
Thanks for clarifying, Justin! On it 🙂

justinxzhao (approved):

jppgks:
As agreed with @justinxzhao, will work in separate PR on the specific decorator/conditionally running test.

PR #1968, opened 2022-04-26
Adds torchscript-compatible Bag and Set input features

geoffreyangus:
This PR implements the preprocessing module required for Bag and Set input features.

This PR depends on the `ts-whitespace-tokenizer` branch.


tests/integration_tests/test_torchscript.py
```diff
@@ -273,6 +273,8 @@ def to_input(s: pd.Series) -> Union[List[str], List[torch.Tensor], torch.Tensor]
         if "image" in s.name:
             return [image_utils.read_image(v) for v in s]
         if s.dtype == "object":
+            if "bag" in s.name or "set" in s.name:
```
justinxzhao, line 276: nit: Add a short docstring to this function as it will likely be growing in size.

I'm also wondering if it makes sense for this to live in `integration_tests/utils.py`.

justinxzhao (approved):

tests/integration_tests/test_torchscript.py
geoffreyangus: Good idea. In a future PR, I will be unifying this function with https://github.com/ludwig-ai/ludwig/blob/9471e600bc5cf1fb9245c707240993fa3ab3cb25/ludwig/models/inference.py#L94

That way the test behavior exactly replicates what would happen in a real-world setting. Thanks!

PR #1972, opened 2022-04-26
Renames torchtext tokenizers

geoffreyangus:
This PR renames the torchtext tokenizers to follow the convention of the other tokenizers by removing the "_tokenizer" suffix from each.


justinxzhao (approved):
Thanks!

PR #1976, opened 2022-04-26
Torch audio refactor

connor-mccorm:
Refactor audio feature flow for torchaudio support.

### DIFF
--- a/ludwig/features/audio_feature.py
+++ b/ludwig/features/audio_feature.py
@@ -19,6 +19,7 @@ import sys
 
 import numpy as np
 import torch
+import torchaudio
 
 from ludwig.constants import (
     AUDIO,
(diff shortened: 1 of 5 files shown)


ludwig/features/audio_feature.py
```diff
@@ -359,7 +356,7 @@ def add_feature_data(
 
         if feature_config[PREPROCESSING]["in_memory"]:
             audio_features = AudioFeatureMixin._process_in_memory(
-                input_df[feature_config[NAME]],
+                input_df[column],
```
geoffreyangus, line 359: Why has this changed? What is in `input_df[feature_config[NAME]]`?

ludwig/utils/audio_utils.py
```diff
@@ -176,7 +251,7 @@ def _short_time_fourier_transform(
     window_length_in_samp = get_length_in_samp(window_length_in_s, sampling_rate_in_hz)
     window_shift_in_samp = get_length_in_samp(window_shift_in_s, sampling_rate_in_hz)
     preprocessed_data_matrix = _preprocess_to_padded_matrix(
-        data, window_length_in_samp, window_shift_in_samp, zero_mean_offset=zero_mean_offset
+        data[0], window_length_in_samp, window_shift_in_samp, zero_mean_offset=zero_mean_offset
```
geoffreyangus, line 254: I know this was in the last PR, but I'm curious– why does this change? Is it because `torchaudio.load` loads with an extra `batch_size` dim?

geoffreyangus:
Looks good! Just have a couple of general questions since I'm starting to work on the torchscript implementation of audio features.

geoffreyangus (approved):

ludwig/utils/audio_utils.py
connor-mccorm: Right, so basically, the soundfile package loads the audio as a 1D np.array but the torchaudio load function loads it as a 2D torch.Tensor with shape(1, n) so I had to index into the tensor to actually get the data here.

ludwig/features/audio_feature.py
connor-mccorm: I just changed that for consistency since it was getting defined above already and used in a few other places. `COLUMN` and `NAME` are really the same thing in most cases, but in the few edge cases I believe it should be using column instead of name to index the df.

Issue #1991, opened 2022-04-30
Evaluation failing because `target` is not integer Tensor.


geoffreyangus:
When evaluating with Ray, the user may get the following error:

```
 File "/home/ray/anaconda3/lib/python3.7/site-packages/torchmetrics/classification/accuracy.py", line 222, in update
    mode = _mode(preds, target, self.threshold, self.top_k, self.num_classes, self.multiclass, self.ignore_index)
  File "/home/ray/anaconda3/lib/python3.7/site-packages/torchmetrics/functional/classification/accuracy.py", line 66, in _mode
    ignore_index=ignore_index,
  File "/home/ray/anaconda3/lib/python3.7/site-packages/torchmetrics/utilities/checks.py", line 265, in _check_classification_inputs
    _basic_input_validation(preds, target, threshold, multiclass, ignore_index)
  File "/home/ray/anaconda3/lib/python3.7/site-packages/torchmetrics/utilities/checks.py", line 44, in _basic_input_validation
    raise ValueError("The `target` has to be an integer tensor.")
ValueError: The `target` has to be an integer tensor.
```

This error can be reproduced by running the following:

```
from ludwig.api import LudwigModel
from ludwig.backend import RAY

model = LudwigModel(config, backend="ray")
model.train(
    dataset=data_csv,
    skip_save_training_description=True,
    skip_save_training_statistics=True,
    skip_save_model=True,
    skip_save_progress=True,
    skip_save_log=True,
    skip_save_processed_input=True,
    output_directory=output_dir,
)
model.evaluate(
    dataset=data_csv,
    skip_save_unprocessed_output=True,
    skip_save_predictions=True,
    skip_save_eval_stats=True,
    collect_predictions=False,
    collect_overall_stats=False,
    output_directory=output_dir,
)
```

Alternatively, simply run a test which calls `model.evaluate` such as `tests/integration_tests/test_api.py::test_api_skip_parameters_evaluate` where all skip params are True, and set the `LudwigModel` created in `run_api_commands` to have `backend="ray"`.

geoffreyangus:
#1992 attempts to fix the bug. Two things, however:

1. The regression test included in that PR unfortunately doesn't actually surface the bug. Follow-up testing showed that category output features needs to have `vocab_size > 2` in order for the bug to appear. Not currently sure why this is the case. 
2. Piero brings up the point that the casting fix in `dask.py` may cause a loss of precision. Will investigate and update here.

[...]

PR #1992, opened 2022-04-30
Fixes unexpected dtype change in preprocessing step

geoffreyangus:
This PR implements a fix and adds a regression test for #1991.

It seems that the dask df_engine may silently change the dtype of `proc_cols` during column assignment. The proposed solution ensures that the dtype of the new dask DataFrame column matches the dtype of each `proc_col` in `proc_cols`.

Confirmed that this fix works in staging with an engine running ludwig code from this PR. I believe the user will have to re-upload datasets for which they are seeing this error. This is in order to re-trigger the preprocessing step. Datasets uploaded before this change will likely have been cached with the wrong dtype.


ludwig/features/number_feature.py
```diff
@@ -288,7 +288,9 @@ def normalize(series: pd.Series) -> pd.Series:
             return pd.Series(transformed_values, index=series.index)
 
         input_series = input_df[feature_config[COLUMN]].astype(np.float32)
-        proc_df[feature_config[PROC_COLUMN]] = backend.df_engine.map_partitions(input_series, normalize)
+        proc_df[feature_config[PROC_COLUMN]] = backend.df_engine.map_partitions(
+            input_series, normalize, meta=input_series
```
tgaddair, line 292: Does add the `input_series` as meta help address the issue, or was it something tried during debugging that was leftover?

geoffreyangus: This helped address the issue– before adding this, `tests/integration_tests/test_class_imbalance_feature.py::test_imbalance_ray` was failing because proc_df was passed around as an "object" column instead of a float32 column. IIRC, later calls then presumed the array was a `Double` instead of a `Float` type.

tgaddair (approved):

PR #1999, opened 2022-05-05
Fix ray hyperopt

ShreyaR:


tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -81,25 +81,25 @@ def mock_storage_client(path):
...
+    #         "time_attr": "training_iteration",
+    #         "reduction_factor": 4,
+    #     },
+    #     "num_samples": 3,
+    # },
 ]
 
 EXECUTORS = [
-    {"type": "ray"},
+    {"type": "ray", "cpu_resources_per_trial": 0.1},
```
justinxzhao, line 97: Why `0.1`? Could we use an integer like 1?

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -222,7 +222,6 @@ def run_hyperopt_executor(
         )
 
 
-@pytest.mark.skip(reason="https://github.com/ludwig-ai/ludwig/issues/1441")
```
justinxzhao, line 225: Nice.

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -81,25 +81,25 @@ def mock_storage_client(path):
...
-        "type": "ray",
-        "search_alg": {"type": "bohb"},
-        "scheduler": {
-            "type": "hb_bohb",
-            "time_attr": "training_iteration",
-            "reduction_factor": 4,
-        },
-        "num_samples": 3,
-    },
+    # {
```
justinxzhao, line 84: Why are these tests commented out? If we want to leave them disabled like this for now, could you add a TODO with a reference to an issue to re-enable them?

ShreyaR: Updated in the latest commit.

ludwig/api.py
```diff
@@ -30,6 +30,7 @@
 from pprint import pformat
 from typing import Dict, List, Optional, Tuple, Union
 
+import dask
```
tgaddair, line 33: Dask is not a required import. We shouldn't be pulling it into api.

ludwig/api.py
```diff
@@ -89,6 +90,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Set dask config to not use placement groups
+dask.config.set(annotations={"ray_remote_args": {"placement_group": None}})
```
tgaddair, line 94: This should be done somewhere in the RayBackend or DaskEngine classes.

ludwig/backend/ray.py
```diff
@@ -426,24 +426,30 @@ def to_tensors(df: pd.DataFrame) -> pd.DataFrame:
...
+        ).map_batches(
+            batch_predictor,
+            batch_size=self.batch_size,
+            compute="actors",
+            batch_format="pandas",
+            num_cpus=num_cpus,
+            num_gpus=num_gpus,
         )
 
+        with dask.annotate(ray_remote_args={"num_cpus": num_cpus, "num_gpus": num_gpus}):
```
tgaddair, line 450: Looks like these changes are reverted in the next PR.

tests/integration_tests/test_hyperopt_ray_horovod.py
tgaddair: Running a test cluster with 4 cpus exhausts all the resources if every trial as a full cpu.

ludwig/api.py
ShreyaR: Fixed.

ludwig/api.py
ShreyaR: Moved to DaskEngine's initialization.

ludwig/backend/ray.py
ShreyaR: Removed these changes from this PR.

ludwig/api.py
ShreyaR: Actually, setting dask config from within either `RayBackend` or `DaskEngine` classes wasn't working. I've now moved this to `__main__` inside `backend/ray.py`.

ludwig/backend/ray.py
```diff
@@ -571,24 +574,29 @@ def to_tensors(df: pd.DataFrame) -> pd.DataFrame:
...
-                num_gpus=num_gpus,
-            )
-            .to_dask()
+        num_cpus = resources_per_worker.get("CPU", (1 if num_gpus == 0 else 0))
+
+        dask_dataset = dataset.ds.map_batches(
+            to_tensors,
+            batch_format="pandas",
+            num_cpus=num_cpus,
+            num_gpus=num_gpus,
```
tgaddair, line 588: The `to_tensors` conversion shouldn't need to be done on GPU. Can we remove the `num_cpus` and `num_gpus` requests here?

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -137,9 +138,9 @@ def on_hyperopt_start(self, *args, **kwargs):
 
 
 @contextlib.contextmanager
-def ray_start_4_cpus():
+def ray_start_10_cpus():
     res = ray.init(
-        num_cpus=4,
+        num_cpus=10,
```
tgaddair, line 143: I don't think GH runners have 10 cpus. Do we really need this many? I think if it can't run with like 6 at most, there's probably a bug somewhere.

ShreyaR: @tgaddair we would need at least one more CPU than the CPUs used in hyperopt trials (discussed in this [slack thread](https://predibase.slack.com/archives/C02J6QF4T0A/p1652751752635729)).

We could artificially restrict hyperopt to run less than 6 trials so that there are CPUs available for dask computation by setting `max_concurrent_trials` to something less than 6. For now, I've changed the number of CPUs to 7.

ludwig/backend/ray.py
```diff
@@ -46,6 +46,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Disable placement groups on dask
+dask.config.set(annotations={"ray_remote_args": {"placement_group": None}})
```
tgaddair, line 50: I believe this wasn't working before, is that not the case?

ludwig/data/dataset/ray.py
```diff
@@ -101,6 +101,9 @@ def __len__(self):
     def size(self):
         return len(self)
 
+    def to_df(self):
+        return self.ds.to_dask()
```
tgaddair, line 105: For this we should delegate to the `df_engine`, which could be Dask or Modin. Geoffrey has an implementation of this in his PR, which is still in review, but would make sense to adopt the same convention:

https://github.com/ludwig-ai/ludwig/pull/2040/files#diff-fec65d5ff8d68e96ebf11e5114a25e0ffa187739296f1e89d9949ae04cef0cc1R105

ShreyaR: Done.

ludwig/backend/ray.py
```diff
@@ -571,28 +584,25 @@ def to_tensors(df: pd.DataFrame) -> pd.DataFrame:
...
+        dask_predictions = dataset.ds.map_batches(to_tensors, batch_format="pandas").map_batches(
+            batch_predictor,
+            batch_size=self.batch_size,
+            compute="actors",
+            batch_format="pandas",
+            num_cpus=num_cpus,
+            num_gpus=num_gpus,
         )
 
+        dask_predictions = dask_predictions.to_dask()
```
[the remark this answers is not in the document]
ShreyaR: Done

ludwig/backend/ray.py
ShreyaR: Removed, yet to test out the alternative fix.

ludwig/data/dataframe/pandas.py
```diff
@@ -62,6 +62,9 @@ def to_ray_dataset(self, df):
 
         return from_pandas(df)
 
+    def from_ray_dataset(self, dataset) -> pd.DataFrame:
+        return dataset.to_pandas
```
tgaddair, line 66: Missing parens.

tgaddair (commented):
Looks good, only one small thing.

ludwig/backend/ray.py
ShreyaR: Confirmed.

ludwig/data/dataframe/pandas.py
ShreyaR: Thanks, fixed!

tgaddair (approved):

PR #2008, opened 2022-05-09
Fixes tokenizer type hints during torchscript saving/loading

geoffreyangus:
This PR adds a fix and a regression test for the new tokenizers. Error-handling in Torchscript when the scripted module stays in memory is different than when the scripted module is first saved, then loaded from disk.

### DIFF
--- a/ludwig/utils/tokenizers.py
+++ b/ludwig/utils/tokenizers.py
@@ -57,10 +57,12 @@ class SpaceStringToListTokenizer(torch.nn.Module):
     def forward(self, v: Union[str, List[str], torch.Tensor]) -> Any:
         if isinstance(v, torch.Tensor):
             raise ValueError(f"Unsupported input: {v}")
-        elif isinstance(v, str):
-            inputs = [v]
+
+        inputs: List[str] = []
+        if isinstance(v, str):
+            inputs.append(v)
         else:
-            inputs = v
+            inputs.extend(v)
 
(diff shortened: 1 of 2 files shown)


ludwig/utils/tokenizers.py
```diff
@@ -57,10 +57,12 @@ def __init__(self, **kwargs):
     def forward(self, v: Union[str, List[str], torch.Tensor]) -> Any:
         if isinstance(v, torch.Tensor):
             raise ValueError(f"Unsupported input: {v}")
-        elif isinstance(v, str):
-            inputs = [v]
+
+        inputs: List[str] = []
+        if isinstance(v, str):
```
justinxzhao, line 62: When would we expect `v` to be a `str` vs. not a `str`? Does this depend on whether we are loading the model from torchscript? Might be good to include a short comment about this.

justinxzhao (approved):

ludwig/utils/tokenizers.py
geoffreyangus: Good question– Here, for example: https://github.com/ludwig-ai/ludwig/blob/5f3a26b28e3569ffca3ddda3c2df1d98ef00efa6/ludwig/utils/strings_utils.py#L267

Here, the `map` function uses the tokenizer on each individual line, which is of type `str` instead of `List[str]`. 

Added a comment to call this out in each tokenizer.

PR #2009, opened 2022-05-09
Add train_steps as an option for configuring training length.

justinxzhao:
If `train_steps = None`, then `epochs` is used to determine training length.

By default train_steps is `None` -- the default training behavior is unchanged.


ksbrar (approved):
LGTM

PR #2015, opened 2022-05-09
Fixes #1992 regression test and implements `dd.concat` solution

geoffreyangus:
Ensures that the regression test added in #1992 actually checks for the bug in #1991. Also adds an alternative solution to the problem which might be cleaner, but this is up for discussion.

### DIFF
--- a/ludwig/data/dataframe/dask.py
+++ b/ludwig/data/dataframe/dask.py
@@ -46,7 +46,7 @@ class DaskEngine(DataFrameEngine):
 
     def df_like(self, df, proc_cols):
         # TODO: address if following results in fragmented DataFrame
-        dataset = dd.from_pandas(pd.DataFrame(proc_cols, index=df.index), npartitions=1)
+        dataset = self.from_pandas(pd.DataFrame(proc_cols, index=df.index))
         return dataset
 
     def parallelize(self, data):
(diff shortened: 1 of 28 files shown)


ludwig/data/dataframe/dask.py
```diff
@@ -43,15 +44,15 @@ def __init__(self, parallelism=None, persist=True, **kwargs):
...
         # Our goal is to preserve the index of the input dataframe but to drop
         # all its columns. Because to_frame() creates a column from the index,
         # we need to drop it immediately following creation.
         dataset = df.index.to_frame(name=TMP_COLUMN).drop(columns=[TMP_COLUMN])
         # TODO: address if following results in fragmented DataFrame
-        for k, v in proc_cols.items():
-            dataset[k] = v
-            dataset[k] = dataset[k].astype(v.dtype)
+        col_names, cols = zip(*proc_cols.items())
+        dataset = dd.concat([dataset] + list(cols), axis=1)
```
w4nderlust, line 54: If this preserves the cols datatype as i imagine it looks like an interesting workaround for preserving the index and not having to do the assignment and casting explicitly :)

tgaddair: +1

tgaddair: This may also help with the large-number-of-columns issue we see using the other approach.

tgaddair (approved):

PR #2016, opened 2022-05-10
Calculate stats for distributed datasets

ShreyaR:


ludwig/models/predictor.py
```diff
@@ -270,18 +271,21 @@ def __init__(self, model: ECD, gpus=None, gpu_memory_limit=None, allow_parallel_
         self.batch_evaluation = return_first(self.batch_evaluation)
 
 
-def calculate_overall_stats(output_features, predictions, dataset, training_set_metadata):
+def calculate_overall_stats(output_features, predictions, dataset, training_set_metadata, df_engine):
```
justinxzhao, line 274: At least in the docstring, we should call out that the "overall stats" that we are calculating is only on 10K rows.

This doesn't seem great though, as 10K seems like an arbitrary number.

Is there a way we could make this parameter configurable, like using an `eval_steps` parameter in the config (which parallels `train_steps`)?

This is along the lines of an API change that I proposed [here](https://docs.google.com/document/d/1T_CJqvnzH_TAA57-4U-E6vK5IjN8zHKoRHChNUGjR1I/edit#).

tgaddair, line 274: Looks like `df_engine` param is unused.

ludwig/models/predictor.py
```diff
@@ -270,18 +271,21 @@ def __init__(self, model: ECD, gpus=None, gpu_memory_limit=None, allow_parallel_
...
     overall_stats = {}
     for of_name, output_feature in output_features.items():
         feature_metadata = training_set_metadata[output_feature.feature_name]
         feature_metadata.update(training_set_metadata[output_feature.feature_name])
 
         feature_df = predictions.loc[:, predictions.columns.str.startswith(of_name)]
         feature_df = feature_df.rename(columns=lambda c: c[len(of_name) + 1 :])
+        feature_df = feature_df.head(n=STATS_SAMPLE_SIZE, npartitions=-1, compute=True)
```
tgaddair, line 282: I would only take the sample if the `feature_df` is a Dask DF. I would also log a warning stating "Sampling the first 10k rows from Dask DataFrame" or something to that effect.

ludwig/models/predictor.py
tgaddair: I would consider this a stopgap for Dask until we have the full stats collection working. We could make it configurable, but ideally we should remove the need for this soon.

ludwig/api.py
```diff
@@ -878,15 +878,12 @@ def evaluate(
...
             # calculate the overall metrics
             if collect_overall_stats:
-                # TODO ray: support calculating stats on partitioned datasets
                 if self.backend.df_engine.partitioned:
-                    raise ValueError(
-                        "Cannot calculate overall stats on a partitioned DataFrame at this time. "
-                        "Set `calculate_overall_stats=False`."
-                    )
+                    # If using a partitioned engine, we need to convert the RayDataset into a Dask dataframe.
+                    dataset = dataset.ds.to_dask()
```
tgaddair, line 883: Reaching into the `ds` property here violates the interface (you could imagine in the future supporting other types of partitioned backends in the future other than RayDataset).

I would suggest instead we do something like have the `Dataset` base class provide a `to_df()` method or similar we can call, which would also allow us to obtain a Pandas DF directly from the `PandasDataset`.

ludwig/models/predictor.py
ShreyaR: Done. I'm sampling in the case when `feature_df` is not a pandas Series, as I didn't want to add an extra dask import in `predictor.py`.

ludwig/models/predictor.py
ShreyaR: Thanks, removed.

PR #2018, opened 2022-05-10
Adds regression tests for #2007

geoffreyangus:


connor-mccorm (commented):
Test audio feature looks good. Could you augment the integration tests in test_ray.py as well since we want to make sure to test with a ray backend too.

tgaddair (approved):

geoffreyangus:
Added test to ray @connor-mccorm. Let me know if there's anything else you think it needs!

PR #2019, opened 2022-05-10
Fix redundant import

tgaddair:


justinxzhao (approved):

ksbrar (approved):

tests/integration_tests/test_custom_components.py
```diff
@@ -7,7 +7,7 @@
 from torch import nn, Tensor
 
 from ludwig.api import LudwigModel
-from ludwig.combiners.combiners import BaseCombinerConfig, Combiner, register_combiner
+from ludwig.combiners.combiners import Combiner, register_combiner
```
justinxzhao, line 10: Should we delete line 18 instead?

ksbrar: No, this is the correct change - `BaseCombinerConfig` is not exposed by `ludwig.combiners.combiners`

tgaddair: I liked line 18 more because it's the true location of that class, as opposed to an indirect import.

PR #2021, opened 2022-05-11
Adds regression tests for #2020

geoffreyangus:
Adds regression tests for #2020, a PR that implemented a fix for NaNs introduced via an outer join `concat` in the dask df engine. 

While writing these tests, I found that `PandasEngine.df_like` was _also_ doing an outer join of `proc_cols` (implicitly through `pd.DataFrame` init) instead of an inner join, similarly leading to NaN values in columns whose preprocessing step called for dropping rows (typically OutputFeature features). This is remedied through an inner join `concat`. We would like to be able to implement an inner join `concat` in the dask df engine in the future. It is not currently possible in the dask df engine due to the parallel nature of dask dataframes.


tgaddair (approved):

tests/integration_tests/utils.py
```diff
@@ -520,7 +520,18 @@ def run_api_experiment(input_features, output_features, data_csv):
         shutil.rmtree(output_dir, ignore_errors=True)
 
 
-def create_data_set_to_use(data_format, raw_data):
+def read_csv_with_nan(path, nan_percent=0.0):
```
geoffreyangus, line 523: This function converts `nan_percent` of samples in each row of the CSV into NaN. This is important for tests that drop rows for exactly one feature– with this change, one has guarantees about the number of rows that will be dropped.

Example: 
We are predicting over `targets` column which is missing

PR #2022, opened 2022-05-11
fix: Move `combiner_registry`

ksbrar:
Moves the `combiner_registry` object back to its original place in `ludwig/combiners/combiners.py`.

The current placement can yield unexpected results - in particular it is possible to import the `combiner_registry` object directly and get an empty set but otherwise import it *indirectly* and get the full (expected) set of combiners.


ksbrar:
Will make another refactor PR to give registries their own module. CC @tgaddair

PR #2024, opened 2022-05-12
Added support for batch size and learning rate tuning using Ray backend

tgaddair:
This PR also changes the batch size algorithm to only do exponential increases. It has been observed that in general powers of 2 tend to perform best for maximizing GPU throughput.


ludwig/backend/ray.py
```diff
@@ -251,6 +327,49 @@ def train(
     def train_online(self, *args, **kwargs):
         raise NotImplementedError()
 
+    def tune_batch_size(
+        self,
+        config: Dict[str, Any],
+        training_set: RayDataset,
+        **kwargs,
+    ) -> int:
+        trainer_kwargs = {**get_trainer_kwargs(), **self.trainer_kwargs}
```
ShreyaR, line 336: nit: We run this code snippet of setting CPU and GPU resources at a few other places inside the ray backend, so it would be good to abstract this out into a function.

ShreyaR (approved):
lgtm!

PR #2032, opened 2022-05-13
Fix SIGINT handler to modify the number of remaining training steps.

justinxzhao:
- First SIGINT will try to finish training at the very next training step.
- Another SIGINT will stop training abruptly.

Also, removes `quit_training` -- this is no longer called anywhere.

### DIFF
--- a/ludwig/models/trainer.py
+++ b/ludwig/models/trainer.py
@@ -341,6 +341,8 @@ class Trainer(BaseTrainer):
 
         self.epochs = config.epochs
         self.train_steps = config.train_steps
+        self.total_steps = 0  # Computed during training, after batcher has been initialized.
+
         self.regularization_lambda = config.regularization_lambda
         self.regularization_type = config.regularization_type
         self.learning_rate = config.learning_rate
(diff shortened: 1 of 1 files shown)


connor-mccorm (approved):
Looks good to me, only thing I would maybe add is possibly initializing self.total_steps in the constructor if were switching to using the class variable. WDYT?

ludwig/models/trainer.py
```diff
@@ -1418,22 +1418,22 @@ def check_progress_on_validation(
...
         else:
             logger.critical("\nReceived a second SIGINT, will now quit")
             if self.original_sigint_handler:
                 signal.signal(signal.SIGINT, self.original_sigint_handler)
             sys.exit(1)
 
-    def quit_training(self, signum, frame):
-        logger.critical("Received SIGQUIT, will kill training")
-        sys.exit(1)
-
```
w4nderlust, line 1436: Just curious, why removing this? Was it dangling / not used?

justinxzhao:
> Looks good to me, only thing I would maybe add is possibly initializing self.total_steps in the constructor if were switching to using the class variable. WDYT?

Good idea. Done!

ludwig/models/trainer.py
w4nderlust: But it was there for a reason, if I remember correctly that function was associated with another signal instead of sigint to still give users a mechanism to kill with one single signal. Probably it got lost during the porting. I would add a sigquit handler like we do for sigint. Not a deal breaker though

PR #2040, opened 2022-05-18
Parallelizes URL reads using Ray / Multithreading

geoffreyangus:
This PR implements parallelized URL reads for audio features with both Ray and local backends. Compared to `master`, we see preprocessing times for a benchmark audio dataset of approximately 7k files decrease from >90 minutes (anecdotally, please confirm @connor-mccorm) to ~18 minutes. Additionally, because of the dedicated URL read functionality, we see a significant reduction in read errors originating from `torchaudio.load`.

Future work could include using `ray.data.read_binary_files` or ray Tasks directly. We opt for the method introduced in this PR instead due to errors related to Dask indexing. A follow-up PR will implement parallelized URL reads for image features.


ludwig/utils/fs_utils.py
```diff
@@ -15,16 +15,22 @@
...
+from typing import Optional
 from urllib.parse import unquote, urlparse
 
 import fsspec
 import h5py
+import requests
 from filelock import FileLock
 from fsspec.core import split_protocol
 
+logger = logging.getLogger(__name__)
```
justinxzhao, line 32: nit: Remove this and use `logging` directly, #2045

justinxzhao (approved):
Nice change!

connor-mccorm (approved):

geoffreyangus:
Hi all, thanks so much for the reviews so far @connor-mccorm @justinxzhao 😄 

I made some non-trivial changes that enable (1) support for NaN values in the audio column and (2) a new unit test that validate the new behavior introduced by this PR.

A couple of directed feedback questions I thought I'd ask:
1. @tgaddair this change currently only supports `dask` and `pandas`, as reflected by the new unit test `tests/integration_tests/test_ray.py::test_ray_read_binary_files`. There is strange indexing behavior occurring with modin causing errors, and I wasn't sure if it was worth investigating or just moving on to the PR for image URLs.
2. @connor-mccorm are there other use cases for `read_audio` and `read_audio_from_str` functions? I left them in for now, but wanted to check in in case we can delete them.

connor-mccorm:
No I just put those in for reading audio features in this pipeline. So if you've replaced them with read_binary_files, I think we're good to delete them. One thing I'll mention is that I created the function `get_default_audio` to fill in Nan values in audio datasets by replacing them with an averaged audio tensor and sampling rate, so if you weren't using that yet, it could be useful. Otherwise, if you have another solution, we can get rid of that probably.

geoffreyangus:
> No I just put those in for reading audio features in this pipeline. So if you've replaced them with read_binary_files, I think we're good to delete them. 

Sounds good!

> One thing I'll mention is that I created the function get_default_audio to fill in Nan values in audio datasets by replacing them with an averaged audio tensor and sampling rate, so if you weren't using that yet, it could be useful.

I think we should keep `get_default_audio`– I've just added code to skip NaNs at audio loading time so that it doesn't try to read files that do not exist. Once the audio tensors are loaded, we still need a way to handle those NaN rows, which `get_default_audio` is still useful for.

connor-mccorm:
Sounds good, let's keep it in then!

ludwig/features/audio_feature.py
```diff
@@ -214,7 +216,7 @@ def reduce(series):
             merged_stats["cropped"],
             audio_file_length_limit_in_s,
         )
-        logger.debug(print_statistics)
+        logging.debug(print_statistics)
```
tgaddair, line 219: Why not use logger?

ludwig/features/audio_feature.py
```diff
@@ -148,19 +146,23 @@ def _process_in_memory(
...
-        def read_audio_file(path):
-            return read_audio(path, src_path)
 
         df_engine = backend.df_engine
-        raw_audio = df_engine.map_objects(column, read_audio_file)
+        raw_audio = backend.read_binary_files(column, map_fn=read_audio_if_bytes_obj)
+
+        def is_torch_audio_tuple(audio):
+            if isinstance(audio, tuple):
+                if isinstance(audio[0], torch.Tensor) and isinstance(audio[1], int):
```
tgaddair, line 155: Do we want to check tuple length as well to be safe?

ludwig/backend/ray.py
```diff
@@ -758,6 +760,19 @@ def check_lazy_load_supported(self, feature):
...
             )
 
+    def read_binary_files(self, column: Series, map_fn: Optional[Callable] = None) -> Series:
+        ds = self.df_engine.to_ray_dataset(column.to_frame(name=column.name))
+
+        def map_batches_fn(df: pd.DataFrame, fn: Callable) -> pd.DataFrame:
+            df[column.name] = df[column.name].map(fn)
+            return df
+
+        ds = ds.map_batches(partial(map_batches_fn, fn=get_bytes_obj_if_path), batch_format="pandas")
```
tgaddair, line 770: This approach makes sense. One concern I had in the back of my mind is regarding partition size. For example, if the input dataset is like 1MB, and so it ends up in a single partition as a Dask DF, then it could explode in size if every image is 10MB being shoved into a single partition.

This is likely something we'll want to wait to see before we prematurely optimize it, but it's something we should be aware of.

ludwig/utils/fs_utils.py
```diff
@@ -53,6 +57,46 @@ def upgrade_http(urlpath):
...
+    """Gets bytes string if `path` is a path (e.g. a string).
+
+    If it is not a path, return as-is.
+    """
+    if not isinstance(path, str):
+        return path
+    return get_bytes_obj_from_path(path)
+
+
+@functools.lru_cache(maxsize=32)
```
tgaddair, line 70: Does this get called repeatedly for the same inputs?

tgaddair (approved):
Looks great! Couple small questions.

ludwig/features/audio_feature.py
geoffreyangus: Oh, I think we are cleaning up the use of `logger` in favor of `logging`. Open issue: https://github.com/ludwig-ai/ludwig/issues/2045

ludwig/backend/ray.py
geoffreyangus: I see, that makes sense. We may want to repartition if that is the case, but we want to be careful in doing that to make sure that our processed columns can still be coerced into a dataframe again by the `df_like` call at the bottom of `data.preprocessing.build_dataset`: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1137

Happy to follow-up on this later on if this becomes an issue.

ludwig/utils/fs_utils.py
geoffreyangus: The `lru_cache` decorator was present in the original `audio_utils.read_audio` function, so I just preserved that functionality here. Might be useful if there are duplicate paths in the dataset, but happy to remove it if you think it is unnecessary.

ludwig/features/audio_feature.py
geoffreyangus: Sure! Added.

ludwig/features/audio_feature.py
tgaddair: Ah, gotcha!

ludwig/utils/fs_utils.py
tgaddair: We can leave it in for now and revisit down the road, I suppose.

PR #2048, opened 2022-05-21
Parallelizes URL reads for images using Ray/Multithreading

geoffreyangus:
This PR parallelizes URL reads for images, similar to what is done for audio in #2040. Because there are multiple image reads before the whole column is read (namely, the first image and the sample set of images), this PR also refactors the process of inferring image feature properties to unify the byte reading and loading workflow introduced in #2040.

### DIFF
--- a/ludwig/features/image_feature.py
+++ b/ludwig/features/image_feature.py
@@ -17,11 +17,9 @@ import logging
 import os
 from collections import Counter
 from functools import partial
-from multiprocessing import Pool
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
-import requests
 import torch
 import torchvision
 
@@ -49,16 +47,16 @@ from ludwig.constants import (
 from ludwig.data.cache.types import wrap
 from ludwig.features.base_feature import BaseFeatureMixin, InputFeature
 from ludwig.utils.data_utils import get_abs_path
-from ludwig.utils.fs_utils import has_remote_protocol, upload_h5
+from ludwig.utils.fs_utils import get_bytes_str_from_path, has_remote_protocol, upload_h5
 from ludwig.utils.image_utils import (
     get_gray_default_image,
-    get_image_from_path,
     grayscale,
     num_channels_in_image,
-    read_image,
+    read_image_if_bytes_obj,
     resize_image,
 )
 from ludwig.utils.misc_utils import set_default_value
+from ludwig.utils.types import Series
 
 logger = logging.getLogger(__name__)
 
@@ -170,7 +168,7 @@ class ImageFeatureMixin(BaseFeatureMixin):
 
     @staticmethod
     def _read_image_and_resize(
-        img_entry: Union[str, torch.Tensor],
+        img_entry: Union[bytes, np.ndarray, torch.Tensor],
         img_width: int,
         img_height: int,
         should_resize: bool,
@@ -199,10 +197,13 @@ class ImageFeatureMixin(BaseFeatureMixin):
         images to the specifications by dropping channels/padding 0 channels
         """
 
-        img = read_image(img_entry, num_channels)
+        img = read_image_if_bytes_obj(img_entry, num_channels)
         if img is None:
             logger.info(f"{img_entry} cannot be read")
             return None
+        if isinstance(img, np.ndarray):
+            img = torch.tensor(img)
+
         img_num_channels = num_channels_in_image(img)
         # Convert to grayscale if needed.
         if num_channels == 1 and img_num_channels != 1:
@@ -313,11 +314,23 @@ class ImageFeatureMixin(BaseFeatureMixin):
         return num_channels
 
     @staticmethod
+    def read_image_from_entry(
+        entry: Union[str, np.ndarray, torch.Tensor], num_channels: Optional[int] = None
+    ) -> Optional[torch.Tensor]:
+        image = None
+        if isinstance(entry, str):
+            image_bytes = get_bytes_str_from_path(entry)
+            image = read_image_if_bytes_obj(image_bytes, num_channels=num_channels)
+        elif isinstance(entry, np.ndarray):
+            image = torch.tensor(entry)
+        elif isinstance(entry, torch.Tensor):
+            image = entry
+        return image
+
+    @staticmethod
     def _finalize_preprocessing_parameters(
         preprocessing_parameters: dict,
-        first_img_entry: Optional[Union[str, torch.Tensor]],
-        src_path: str,
-        input_feature_col: np.array,
+        column: Series,
     ) -> Tuple:
         """Helper method to determine the height, width and number of channels for preprocessing the image data.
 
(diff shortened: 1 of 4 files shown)


ludwig/utils/image_utils.py
```diff
@@ -163,6 +163,71 @@ def read_image_from_str(img: str, num_channels: Optional[int] = None) -> torch.T
...
+def read_image_from_bytes_obj(
+    bytes_obj: Optional[bytes] = None, num_channels: Optional[int] = None
+) -> Optional[torch.Tensor]:
+    mode = get_image_read_mode_from_num_channels(num_channels)
+
+    try:
+        with BytesIO(bytes_obj) as buffer:
+            buffer_view = buffer.getbuffer()
+            image = decode_image(torch.frombuffer(buffer_view, dtype=torch.uint8), mode=mode)
+            del buffer_view
```
dantreiman, line 223: Why is it necessary to `del buffer_view`?  Seems like it will go out-of-scope on the next line.  If I'm missing something, please add a comment here explaining why this is needed.

ludwig/utils/image_utils.py
```diff
@@ -163,6 +163,71 @@ def read_image_from_str(img: str, num_channels: Optional[int] = None) -> torch.T
...
+        mode = ImageReadMode.RGB
+    elif num_channels == 4:
+        mode = ImageReadMode.RGB_ALPHA
+    return mode
+
+
+def read_image_if_path(path: Any, num_channels: Optional[int] = None) -> Union[Any, torch.Tensor]:
+    """Gets an image if `path` is a path (e.g. a string).
+
+    If it is not a path, return as-is.
```
dantreiman, line 186: What are the possible data types of path (if not a string)?

ludwig/features/image_feature.py
```diff
@@ -467,55 +423,24 @@ def add_feature_data(
...
-                    input_df[feature_config[COLUMN]], _get_processed_image
-                )
-        else:
+            metadata[name]["reshape"] = (num_channels, height, width)
 
-            all_img_entries = [
-                get_abs_path(src_path, img_entry) if isinstance(img_entry, str) else img_entry
-                for img_entry in input_df[feature_config[COLUMN]]
-            ]
+            proc_col = backend.read_binary_files(abs_path_column, map_fn=read_image_if_bytes_obj_and_resize)
```
dantreiman, line 430: This is a significant improvement.  Nice!

ludwig/utils/image_utils.py
geoffreyangus: This function is intended to be usable by `map`-like workflows, so `path` could be any value stored in a column. For image columns, I believe we officially support image features being passed in as either path (str) or torch.Tensor objects (pre-loaded images). So, in this case, `path` could be a `torch.Tensor`. In that case, we would want this function to be a no-op.

ludwig/utils/image_utils.py
geoffreyangus: Hm, not quite sure why it's needed. This was a bit of code that was already in `read_image`: https://github.com/ludwig-ai/ludwig/blob/d3eea1347d394b363ebf2cdfe973bd9adbd84a76/ludwig/utils/image_utils.py#L121

geoffreyangus:
Thanks for the review @dantreiman 😄 One thing I wanted to ask: 

1. There are 4 functions in `image_utils.py` that are effectively replaced by the new functions: `get_image_from_http_bytes`, `get_image_from_path`, `read_image`, `read_image_from_str`. Are you aware of anywhere these are being used external of Ludwig (i.e. in the Predibase app)? If not, I may go ahead and delete these in this PR.

ludwig/utils/image_utils.py
dantreiman: AFAICT it would be safe to remove this line.

ludwig/utils/image_utils.py
dantreiman: I'd recommend adding your response:

> ...path could be a torch.Tensor. In that case, we would want this function to be a no-op.

 as a comment, to make the intent clear here.

dantreiman:
> Thanks for the review @dantreiman 😄 One thing I wanted to ask:
> 
> 1. There are 4 functions in `image_utils.py` that are effectively replaced by the new functions: `get_image_from_http_bytes`, `get_image_from_path`, `read_image`, `read_image_from_str`. Are you aware of anywhere these are being used external of Ludwig (i.e. in the Predibase app)? If not, I may go ahead and delete these in this PR.

I'm not aware of any usage outside the ludwig package.  Lets double-check with @hungcs first before deleting.

ludwig/utils/image_utils.py
dantreiman: How about the name the parameter `path_or_image` (instead of just `path`)

dantreiman:
Removing unused functions from image_utils should be fine.

ludwig/utils/image_utils.py
geoffreyangus: Hm, it seems like removing this is causing tests to fail with the following error:
```
Existing exports of data: object cannot be re-sized
```
Seems related to this open issue: https://github.com/python/cpython/issues/85269

I've put it back for now, but let me know if you think there's an alternative you think is worth exploring.

ludwig/features/feature_utils.py
```diff
@@ -20,14 +20,26 @@
...
+from ludwig.utils.data_utils import get_abs_path, hash_dict
+from ludwig.utils.fs_utils import has_remote_protocol
 from ludwig.utils.strings_utils import tokenizer_registry, UNKNOWN_SYMBOL
 
 SEQUENCE_TYPES = {SEQUENCE, TEXT, TIMESERIES}
 FEATURE_NAME_SUFFIX = "__ludwig"
 FEATURE_NAME_SUFFIX_LENGTH = len(FEATURE_NAME_SUFFIX)
 
 
+def get_abs_path_if_entry_is_str(entry, src_path):
```
jeffreyftang, line 32: Would be great to have type annotations if possible :)

ludwig/utils/image_utils.py
```diff
@@ -104,62 +66,69 @@ def is_image_score(src_path, img_entry, column: str):
...
+    elif num_channels == 2:
+        mode = ImageReadMode.GRAY_ALPHA
+    elif num_channels == 3:
+        mode = ImageReadMode.RGB
+    elif num_channels == 4:
+        mode = ImageReadMode.RGB_ALPHA
+    return mode
+
+
+def read_image_if_path(item: Any, num_channels: Optional[int] = None) -> Union[Any, torch.Tensor]:
```
jeffreyftang, line 86: How important is the "return as-is behavior" here and elsewhere? To me it's more intuitive and self-documenting to return `None` or an exception if the input isn't the type expected, then handle appropriately from the caller.

geoffreyangus: Removed in latest change.

PR #2050, opened 2022-05-21
Adds rule of thumb for determining embeddings size

w4nderlust:
Also changes the default behavior for dense category embed encoder. This has repercussions on documentation, where we should document the rule of thumb for determining size if unspecified: min(512, round(1.6 * x**0.56))


jppgks (approved):
Do we have a reference for the rule of thumb?

w4nderlust:
> Do we have a reference for the rule of thumb?

Just some empirical use: https://forums.fast.ai/t/size-of-embedding-for-categorical-variables/42608/2?u=knesgood

justinxzhao (approved):

PR #2051, opened 2022-05-23
Fix progress bar ray

### LINKED ISSUE #1528: [ray] Fix training progress bars
Progress bars are not rendering properly in the terminal for Ray, instead printing one line per step.

This solution here is to use something like Callbacks to send updates back to the head / driver to print.

magdyksaleh:
# Code Pull Requests

Please provide the following:

- Fix #1528 
- Moves all the tqdm logic into the ray driver and uses a Callback from ray train to trigger tqdm stuff. Need to remove some of the coordinator logic into the LudwigProgressbar Class since all child processes need to run ray.train.report if one of them does (otherwise it hangs)


ludwig/progress_bar.py
```diff
@@ -0,0 +1,48 @@
+import os
+import uuid
+
+import ray.train
```
tgaddair, line 4: Ray is technically an optional import. Since this progress bar is on the critical import path, we should move it. There are a few ways to accomplish this. The one that would require the least refactoring would be to do something like this:

```
try:
    import ray.train as rt
except ImportError:
    rt = None

...

class LudwigProgressBar:
    def __init__(...):
        if report_to_ray and rt is None:
            raise ValueError("Set report_to_ray=True but ray is not installed. Run `pip install ray`")
```

ludwig/progress_bar.py
```diff
@@ -0,0 +1,73 @@
...
+    rt = None
+
+
+class LudwigProgressBarActions:
+    CREATE = "create"
+    UPDATE = "update"
+    CLOSE = "close"
+
+
+class LudwigProgressBar:
```
justinxzhao, line 17: Add docstrings for new classes.

ludwig/progress_bar.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, report_to_ray, config, is_coordinator):
```
justinxzhao, line 18: Add type hints and a docstring for the constructor.

ludwig/progress_bar.py
```diff
@@ -0,0 +1,73 @@
...
+            rt.report(
+                progress_bar={
+                    "id": self.id,
+                    "config": self.config,
+                    "action": LudwigProgressBarActions.CREATE,
+                    "is_coordinator": self.is_coordinator,
+                }
+            )
+
+    def update(self, steps):
```
justinxzhao, line 49: Add type hints and docstring.

ludwig/progress_bar.py
```diff
@@ -0,0 +1,73 @@
...
+            rt.report(
+                progress_bar={
+                    "id": self.id,
+                    "update_by": steps,
+                    "is_coordinator": self.is_coordinator,
+                    "action": LudwigProgressBarActions.UPDATE,
+                }
+            )
+
+    def close(self):
```
justinxzhao, line 63: Add type hints and docstring.

ludwig/backend/ray.py
```diff
@@ -272,6 +273,30 @@ def tune_learning_rate_fn(
         hvd.shutdown()
 
 
+class TqdmCallback(rt.TrainingCallback):
```
justinxzhao, line 276: Add docstring for new classes.

ludwig/backend/ray.py
```diff
@@ -272,6 +273,30 @@ def tune_learning_rate_fn(
...
+    def __init__(self):
+        super().__init__()
+        self.progess_bars = {}
+
+    def process_results(self, results, **info):
```
justinxzhao, line 281: Add type hints and a docstring. What is results?

ludwig/models/trainer.py
```diff
@@ -1039,12 +1041,13 @@ def train_online(self, dataset):
...
-                file=sys.stdout,
-                disable=is_progressbar_disabled(),
-            )
+            progress_bar_config = {
+                "desc": "Training online",
+                "total": batcher.steps_per_epoch,
+                "file": sys.stdout,
+                "disable": is_progressbar_disabled(),
+            }
+            progress_bar = LudwigProgressBar(self.report_tqdm_to_ray, progress_bar_config, True)
```
justinxzhao, line 1050: Why is this constructed with `True` as the argument for `is_coordinator` directly instead of `self.is_coordinator()`?

ludwig/progress_bar.py
```diff
@@ -0,0 +1,73 @@
+import uuid
+
+import tqdm
+
+try:
+    import ray.train as rt  # noqa: E402
```
justinxzhao, line 6: Is the `noqa: E402` necessary?

magdyksaleh: I though E402 also applies if it is wrapped in a try except

ludwig/models/trainer.py
magdyksaleh: good find

ludwig/progress_bar.py
justinxzhao: It seems like it may not be, i.e. https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/data_utils.py#L45-L52

One way to test whether the annotation is needed (I think) is to try removing the annotation and see if flake8 (should be enabled automatically in VSCode for the Ludwig repo) complains.

magdyksaleh: You are right my friend

tgaddair (approved):
LGTM!

justinxzhao (approved):
Nice!

PR #2052, opened 2022-05-23
Update AutoML to check for imbalanced binary or category output features

amholler:
Update AutoML to check for imbalanced binary or category output features.

Currently detection of such features produces an informational log message.
In the future, AutoML could use the detection to trigger specific functionality.

Tested on the AutoML datasets in the experiments repo.


ludwig/automl/utils.py
```diff
@@ -178,3 +193,20 @@ def set_output_feature_metric(base_config):
...
+        if output_feature[TYPE] == BINARY or output_feature[TYPE] == CATEGORY:
+            for feature_metadata in features_metadata:
+                if output_feature[NAME] == feature_metadata.name:
+                    if feature_metadata.imbalance_ratio < IMBALANCE_DETECTION_RATIO:
+                        logging.info(
+                            f"Imbalance in {output_feature[NAME]}: minority/majority={feature_metadata.imbalance_ratio}"
+                        )
+                        imbalanced_output = True
+                    break
+    return imbalanced_output
```
justinxzhao, line 212: Would it be more simple to `return True` directly and then `return False` at the end of the function?

ludwig/automl/utils.py
```diff
@@ -178,3 +193,20 @@ def set_output_feature_metric(base_config):
         base_config[HYPEROPT]["metric"] = output_metric
         base_config[HYPEROPT]["goal"] = output_goal
     return base_config
+
+
+def check_imbalanced_output(base_config, features_metadata):
```
justinxzhao, line 198: nit: Add type hint that this returns bool.

justinxzhao, line 198: nit: Rename to has_imbalanced_output()

ludwig/automl/utils.py
amholler: The function is written so if there is more than one output feature (the break is only out of the inner loop)
and the function returns True if any of the output_features is imbalanced.

ludwig/automl/utils.py
amholler: done

amholler: done

ludwig/automl/utils.py
justinxzhao: I see. If there's value in logging each imbalanced output feature, then I agree that we shouldn't short-circuit.

justinxzhao (approved):

Issue #2054, opened 2022-05-24
Differences in model output between Ray and Local backends.


geoffreyangus:
**Describe the bug**
Given some trained model with weights M and some dataset D, M evaluated on D using a Ray backend and M evaluated on D using a Local backend produces different numerical outputs.

**To Reproduce**

1. Add print statements to look at the predictions from the model evaluation calls in `tests.integration_tests.utils.train_with_backend`.
2. Run a test that calls that function, e.g. `tests/integration_tests/test_ray.py::test_ray_image`.

**Expected behavior**
I expect evaluation calls on the same model with the same dataset to produces the same numerical outputs regardless of the backend.

**Environment (please complete the following information):**

- OS: macOS
- Version: 12.3
- Python version: 3.8
- Ludwig version: 0.5

**Additional context**

We currently workaround this by having some degree of tolerance when comparing the models in unit tests.

geoffreyangus:
Update: I am seeing unit tests fail because AUC values at model evaluation time are different between Ray and Local backends.

The test proceeds as following:

1. A CSV of 100 samples with 10% NaNs in each column is generated as dataset `D`. 
2. A model `M_ray` is instantiated with a Ray backend.
2. `M_ray` is trained on `D`.
3. `M_ray` is evaluated on `D`, creating `metric_ray`.
4. `M_ray` is saved, then reloaded with a Local backend as `M_local`.
5. `M_local` is evaluated on `D`, creating `metric_local`.
6. `metric_ray` is compared with `metric_local`.

Sometimes the test passes and sometimes it fails. I confirmed the following:

1. The data looks identical between the two backends. This is right before the forward pass.
2. The output probabilities look identical between the two backends. This is a change from what I initially thought– the reason being that I was not considering that batches were being split between two Ray processes.
3. Tests that were passing were processing 90 samples. This is expected because drop_rows should be removing 10% of samples (the ones with NaNs).
3. Tests that were failing were processing 100 samples. This is unexpected for the same reason as above.

## Inconsistencies in Dask type inference

It seems that the issue is related to WHERE exactly the first NaN is located in the binary feature. Such an explanation is supported by the [Dask docs](https://docs.dask.org/en/stable/generated/dask.dataframe.read_csv.html): 

[...]

PR #2058, opened 2022-05-25
Fixes NaN handling in boolean dtypes

geoffreyangus:
Addresses #2054. CSVs are now read without type inference to preserve NaN values. These NaNs are then filtered out by `handle_missing_values`. After that, each column is casted to its respective dtype based on its feature type.


ludwig/data/preprocessing.py
```diff
@@ -1101,6 +1102,16 @@ def build_dataset(
     for callback in callbacks or []:
         callback.on_build_metadata_end(dataset_df, mode)
 
+    # Happens after metadata is built so we can use precomputed fill values.
+    logger.debug("handle missing values")
+    for feature_config in feature_configs:
+        preprocessing_parameters = metadata[feature_config[NAME]][PREPROCESSING]
+        handle_missing_values(dataset_cols, feature_config, preprocessing_parameters)
```
justinxzhao, line 1109: I just noticed that `handle_missing_values` is also called in `build_metadata()`, so we'll be making the call to handling missing values twice. Is that expected?

justinxzhao (approved):

ludwig/data/preprocessing.py
w4nderlust: I may be misremembering, but I believe @tgaddair originally introduced the double missing values call for a reason

geoffreyangus: Just pushed up a refactor that makes the call to `handle_missing_values` once. I couldn't find a good reason for calling it twice, but super happy to revert this change @tgaddair if I missed something.

geoffreyangus: Just pushed up a refactor that makes the call to `handle_missing_values` once. I couldn't find a good reason for calling it twice, but super happy to revert this change @tgaddair if I missed something.

w4nderlust: the hdf5 is supposed to be output of internally processing, you cannot throw a random unprocessed hdf5 at it

tgaddair (approved):

PR #2060, opened 2022-05-25
Make text encoder trainable property default to False for pre-trained HF encoders

dantreiman:
Our documentation says that the default value of `trainable` is False.  Updating code to match that:
https://ludwig-ai.github.io/ludwig-docs/0.5/configuration/features/text_features/#huggingface-models

This also matches expectations learned from user feedback: if using a pretrained model, default behavior should be to freeze the pretrained model and train the prediction layers using the pretrained model's embeddings.
\


justinxzhao (approved):
Thanks!

PR #2061, opened 2022-05-25
Fix type hints for dropout, dropout parameter references, and add docs for FCLayer and FCStack.

justinxzhao:


ludwig/encoders/h3_encoders.py
```diff
@@ -411,7 +411,7 @@ def __init__(
             num_layers=num_layers,
             bidirectional=bidirectional,
             use_bias=use_bias,
-            dropout=dropout,
+            dropout=recurrent_dropout,
```
geoffreyangus, line 414: I think that we're missing `recurrent_dropout` in the docstring for this class. Could you add that in?

ludwig/encoders/sequence_encoders.py
```diff
@@ -1327,8 +1327,7 @@ def __init__(
             weights_initializer=weights_initializer,
             recurrent_initializer=recurrent_initializer,
             bias_initializer=bias_initializer,
-            dropout=dropout,
-            recurrent_dropout=recurrent_dropout,
+            dropout=recurrent_dropout,
```
geoffreyangus, line 1330: The docstring of this class is also missing an entry for `recurrent_dropout`. I guess I could imagine a case for punting the description of recurrent_dropout downstream into the RecurrentStack class– just wanted to call it out. Its just weird that there are two dropouts, so it could be good as a user to be able to read up on the difference without inspecting the code too closely.

tests/ludwig/modules/test_fully_connected_modules.py
```diff
@@ -36,9 +36,26 @@ def test_fc_layer(
...
+        first_layer_input_size=first_layer_input_size,
+        layers=layers,
+        num_layers=num_layers,
+        default_output_size=output_size,
+    ).to(DEVICE)
+    input_tensor = torch.randn(BATCH_SIZE, first_layer_input_size, device=DEVICE)
+
+    output_tensor = fc_stack(input_tensor)
+
+    assert list(output_tensor.shape[1:]) == [first_layer_input_size]
```
geoffreyangus, line 60: nit: if this is meant to test if the FC stack is a no-op, it might be worth doing an `assert np.isclose` between the inputs and outputs to validate that nothing is happening to the values themselves.

w4nderlust:
It's not super clear to me why in some cases we are renaming parameters from dropout to recurrent dropout and in some cases we are dropping them altogether (no pun intended). Can you give a bit more context @justinxzhao ?

tests/ludwig/modules/test_fully_connected_modules.py
justinxzhao: Thanks for the suggestion! Added.

justinxzhao:
> It's not super clear to me why in some cases we are renaming parameters from dropout to recurrent dropout and in some cases we are dropping them altogether (no pun intended). Can you give a bit more context @justinxzhao ?

The changes in this PR is mainly to make the use of separate dropout parameters, where there are already multiple dropout parameters in the constructor, consistent across all Ludwig modules. New consistency in this PR:
- `H3RNN`:
  - `dropout` -> `H3Embed`
  - `recurrent_dropout` -> `RecurrentStack`
- `StackedRNN`:
  - `dropout` -> `EmbedSequence`
  - `recurrent_dropout` -> `RecurrentStack`
  - `fc_dropout` -> `FCStack`
- `StackedCNNRNN`: 
  - `dropout` -> `EmbedSequence`
  - `conv_dropout` -> `Conv1DStack`
  - `recurrent_dropout` -> `RecurrentStack`
  - `fc_dropout` -> `FCStack`

As for removing `recurrent_dropout=...` -- the constructor of `RecurrentStack` doesn't actually have a `kwarg` for `recurrent_dropout`. It's simply called `dropout=...`.

On a side note, I wonder if it's reasonable to simplify/consolidate all of the different dropout parameters into a more global `dropout` parameter that we can use for everything. Curious to get people's thoughts @w4nderlust @dantreiman @geoffreyangus @ShreyaR

w4nderlust:
@justinxzhao got it, make sense.
Although it seems to me that `StackedCNNRNN` has a couple issues.
it doesn't have type hints, the order where the dropout parameter is placed is wrong (it should be close to the embedding parameter at the beginning), in some cases the default value for dropouts is 0, in others it is 0.0 (not a big issue, but consistency is great), and finally in the recurrent stack we have `dropout=dropout`, but it should be `dropout=recurrent_dropout`.

A side note: in all these cases, dropout is the one used in embeddings, while other modules have their own. It could be better to rename it to `embedding_dropout` or `embed_dropout` for clarity and consistency potentially. What do you think?

geoffreyangus (approved):
LGTM! Left a comment re: dropout here: https://github.com/ludwig-ai/ludwig/issues/2080

Issue #2066, opened 2022-05-26
Comprehensive Configs: When saving the model, export the fullest Ludwig config that includes module-specific parameters


justinxzhao:
The current "expanded" Ludwig config (saved as `model_hyperparameters.json`) contains user config-specified parameters merged with preprocesing and metadata parameters. However, there aren't any module-specific parameters (unless they were explicitly specified by the initial config).

When saving the model, we should export an even fuller Ludwig config that includes module-specific parameters. This allows older models to load safely if there's new default behavior within a module, like when we [add a rule of thumb for embedding size](https://github.com/ludwig-ai/ludwig/pull/2050/files).

PR #2069, opened 2022-05-27
Revert "Adds rule of thumb for determining embeddings size"

justinxzhao:
Reverts ludwig-ai/ludwig#2050

### DIFF
--- a/ludwig/encoders/category_encoders.py
+++ b/ludwig/encoders/category_encoders.py
@@ -31,7 +31,7 @@ class CategoricalEmbedEncoder(Encoder):
     def __init__(
         self,
         vocab: List[str],
-        embedding_size: Optional[int] = None,
+        embedding_size: int = 50,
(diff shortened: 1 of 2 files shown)


jeffreyftang (approved):

tgaddair:
Can you add an explanation for why this is being reverted?

w4nderlust:
> Can you add an explanation for why this is being reverted?

There was a cat on slack, the tl;dr is that the previous PR introduced a new default that brake loading of older models. The solution would be to save the fully materialized config to avoid it, but that would require more though so the quick fix is reverting and reintroduce this feature later

tgaddair:
Test was fixed upstream.

justinxzhao:
Until we start saving the fully materialized config, there's an implicit dependency on whatever the default behavior is at the time of model creation. Filed #2066 to track.

PR #2074, opened 2022-05-29
Revert "Update AutoML to check for imbalanced binary or category outp…

tgaddair:
Temporarily reverts changes in #2052. Some of the new functions turned out to be expensive to run on large datasets, so


tgaddair:
Can revisit in a more tactical PR, there are workarounds for this issue in the current implementation.

PR #2081, opened 2022-06-02
Adds step tracking at epoch level

geoffreyangus:
This PR fixes evaluation continuing despite being at the end of steps-based training. This happens when the user trains on <1 epoch (typically while debugging).


justinxzhao (approved):
Thanks!

PR #2083, opened 2022-06-02
Fixes for large scale hyperopt

ShreyaR:
- Enable windowing for large datasets
- Add support for tune.report while using ray backend

### DIFF
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -19,7 +19,7 @@ import queue
 import threading
 from distutils.version import LooseVersion
 from functools import lru_cache
-from typing import Any, Dict, Iterator, Union
+from typing import Any, Dict, Iterator, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -72,7 +72,20 @@ class RayDataset(Dataset):
         #     return df
         # self.ds = self.ds.map_batches(to_tensors, batch_format="pandas")
 
-    def pipeline(self, shuffle=True, fully_executed=True) -> DatasetPipeline:
+    def pipeline(
+        self,
+        shuffle: bool = True,
+        fully_executed: bool = True,
+        enable_windowing: bool = False,
+        window_size_bytes: Optional[int] = None
+    ) -> DatasetPipeline:
+        """
+        Args:
+            shuffle: If true, the entire dataset is shuffled in memory before batching.
+            fully_executed: If true, force full evaluation of the Ray Dataset by loading all blocks into memory.
+            enable_windowing: If true, split dataset into windows of fixed memory size.
+            window_size_bytes: If not None, specifies the window size in bytes for the dataset.
+        """
         if not fully_executed and not _ray112:
             raise ValueError(f"Cannot set fully_execute=False in ray {ray.__version__}")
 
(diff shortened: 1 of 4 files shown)


ludwig/data/dataset/ray.py
```diff
@@ -72,15 +72,33 @@ def __init__(
...
 
         if fully_executed and _ray112:
             # set instance state so calls to __len__ will also use the fully_executed version
             self.ds = self.ds.fully_executed()
 
-        pipe = self.ds.repeat()
+        if not enable_windowing:
+            pipe = self.ds.repeat()
+        else:
+            if window_size_bytes is None:
```
ShreyaR, line 99: An alternative to requiring `window_size_bytes` to always be set is to calculate window size based on the following variables:
- Total object store memory in the ray cluster, and
- If running from within a tune trial and piping through the max concurrent trials parameter.
Not sure if that would be overkill tho so I've implemented a simpler solution for now.

ludwig/data/dataset/ray.py
```diff
@@ -72,15 +72,33 @@ def __init__(
...
+        """
         if not fully_executed and not _ray112:
             raise ValueError(f"Cannot set fully_execute=False in ray {ray.__version__}")
 
         if fully_executed and _ray112:
             # set instance state so calls to __len__ will also use the fully_executed version
             self.ds = self.ds.fully_executed()
 
-        pipe = self.ds.repeat()
+        if not enable_windowing:
```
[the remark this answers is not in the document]
ShreyaR: Thanks, updated the code to go with option #1.

ludwig/hyperopt/execution.py
```diff
@@ -491,24 +493,28 @@ def _run():
...
                     results = ray_queue.get_nowait_batch(qsize)
-                    sync_client.sync_down(remote_checkpoint_dir, str(trial_dir.absolute()))
-                    sync_client.wait()
+                    if sync_client is not None:
+                        sync_client.sync_down(remote_checkpoint_dir, str(trial_dir.absolute()))
+                        sync_client.wait()
                     for progress_tracker, save_path in results:
-                        checkpoint(progress_tracker, str(trial_dir.joinpath(Path(save_path))))
+                        # TODO(shreya): Figure out why this doesn't work
+                        # checkpoint(progress_tracker, str(trial_dir.joinpath(Path(save_path))))
```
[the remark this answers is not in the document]
ShreyaR: Uncommented -- the error was because the trial was trying to copy the checkpoint while checkpointing was turned off for the experiment.

ludwig/hyperopt/execution.py
```diff
@@ -474,7 +476,7 @@ def on_trainer_train_teardown(self, trainer, progress_tracker, save_path, is_coo
                 "use_gpu": use_gpu,
                 "resources_per_worker": {
                     "CPU": num_cpus,
-                    "GPU": num_gpus,
+                    "GPU": 1,
```
tgaddair, line 479:
```
1 if use_gpu else 0
```

tgaddair (approved):

ludwig/hyperopt/execution.py
ShreyaR: Done

PR #2090, opened 2022-06-02
Enables saving Torchscript modules during training

geoffreyangus:
This PR enables automatic saving Torchscript modules to LudwigModel.train and tests to confirm this behavior.

There is a slight change in the API– namely `LudwigModel.save_torchscript` now takes a full path instead of just a directory. This is for added flexibility.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -1478,41 +1478,33 @@ class LudwigModel:
         save_json(model_hyperparameters_path, self.config)
 
     def to_torchscript(self, model_only: bool = False):
-        """Returns a scripted model.
+        """Converts the trained LudwigModule, including preprocessing and postprocessing, to Torchscript.
 
-        The input to this model is a dictionary of input features. For every input feature, the user must provide a
-        tensor, a list of tensors or a list of strings.
+        The scripted module takes in a `Dict[str, Union[List[str], Tensor]]` as input.
+
+        More specifically, for every input feature, we provide either a Tensor of batch_size inputs, a list of Tensors
+        batch_size in length, or a list of strings batch_size in length.
+
+        Note that the dimensions of all Tensors and lengths of all lists must match.
 
         Similarly, the output will be a dictionary of dictionaries, where each feature has its own dictionary of
         outputs. The outputs will be a list of strings for predictions with string types, while other outputs will be
         tensors of varying dimensions for probabilities, logits, etc.
 
-        :param model_only: (bool, default: `False`) if `True`, only the ECD model will be converted to torchscript. If
-            `False`, the preprocessing and postprocessing modules will be included in  the torchscript module as well.
-
-        :return: (torch.nn.Module) a scripted model.
+        Args:
+            model_only (bool, optional): If True, only the ECD model will be converted to Torchscript. Else,
+                preprocessing and postprocessing will also be converted to Torchscript.
         """
         self._check_initialization()
-
         if model_only:
-            scripted_model = self.model.to_torchscript()
+            return self.model.to_torchscript()
         else:
-            inference_module = InferenceModule(
-                self.model,
-                config=self.config,
-                training_set_metadata=self.training_set_metadata,
-            )
-            scripted_model = torch.jit.script(inference_module)
-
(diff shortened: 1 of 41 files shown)


ludwig/api.py
```diff
@@ -322,6 +323,12 @@ def train(
             dataset is provided it is preprocessed and cached by saving an HDF5
             and JSON files to avoid running the preprocessing again. If this
             parameter is `False`, the HDF5 and JSON file are not saved.
+        :param skip_save_inference_module: (bool, default: `False`) disables
```
tgaddair, line 326: Comment seems to be outdated, as it looks like the TS model is only saved once after training.

ludwig/api.py
```diff
@@ -615,6 +631,19 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
...
+
+                    # Save inference module
+                    if not skip_save_inference_module:
+                        logger.info("Attempting to save inference module with best weights from saved checkpoint...")
+                        inference_module_path = os.path.join(model_dir, INFERENCE_MODULE_FILE_NAME)
+                        try:
+                            self.save_torchscript(inference_module_path)
+                            logger.info(f'Saved inference module to: "{inference_module_path}"')
+                        except Exception:
+                            logger.warning(
```
tgaddair, line 643: Swallowing the exception in all cases feels off to me, as the user may wish to explicitly save the model and fail loudly if it cannot.

An alternative we might consider here:

- `skip_save_inference_module` defaults to `None`, and if it's set to `None`, then it will attempt to save the inference module, but fail softly if it cannot.
- When set to `False`, fail loudly if the inference model cannot be saved. Also fail fast before training if set to `False` and `skip_save_model=True`.

What do you think?

ludwig/features/binary_feature.py
```diff
@@ -72,7 +72,7 @@ def forward(self, v: Union[List[str], List[torch.Tensor], torch.Tensor]):
         if self.should_lower:
             v = [s.lower() for s in v]
         indices = [self.str2bool.get(s, False) for s in v]
-        return torch.tensor(indices, dtype=torch.bool)
+        return torch.tensor(indices, dtype=torch.float32)
```
tgaddair, line 75: Not sure I understand why this being done. It is also at odds with line 69 above, which explicitly casts to a bool type.

ludwig/api.py
geoffreyangus: This sounds good to me. One thing I'd add is a case for if train is called with `skip_save_model=True` and `skip_save_inference_module=None`. 

In order to not break the API for current users/tests, I feel that in this case, `skip_save_inference_module` should be set to the same value as `skip_save_model` (i.e. True).

ludwig/features/binary_feature.py
geoffreyangus: Upon saving and loading the torchscript module, it seems to raise an error on binary-only features with the following message:

```
...
E           Traceback of TorchScript, original code (most recent call last):
E             File "/Users/geoffreyangus/repositories/predibase/ludwig/ludwig/models/inference.py", line 71, in forward
E                           feature_name = get_name_from_module_dict_key(module_dict_key)
E                           preproc_inputs[feature_name] = preproc(inputs[feature_name])
E                       outputs = self.model(preproc_inputs)
E                                 ~~~~~~~~~~ <--- HERE
E               
E                       predictions: Dict[str, Dict[str, torch.Tensor]] = {}
E           /usr/local/lib/python3.8/site-packages/torch/nn/modules/linear.py(103): forward
...
E           RuntimeError: expected scalar type Float but found Bool
```

This change probably doesn't belong in its own PR, though. I'll put up another one with the change and a regression test. Good catch!

ludwig/api.py
geoffreyangus: Updated the docstring to reflect behavior described in other PR comment.

ludwig/features/binary_feature.py
geoffreyangus: Put this change+regression test in a new PR here: #2103

justinxzhao (approved):

PR #2091, opened 2022-06-02
Updates the export_torchscript CLI to match the API

geoffreyangus:
Updates export_torchscript CLI to be more in-line with API behavior. If `output_path` is not set, the torchscript model is now saved directly into the `model/` directory as `inference_module`.


justinxzhao (approved):

ludwig/backend/ray.py
```diff
@@ -771,6 +773,19 @@ def check_lazy_load_supported(self, feature):
                 f"Set preprocessing config `in_memory: True` for feature {feature[NAME]}"
             )
 
+    def read_binary_files(self, column: Series, map_fn: Optional[Callable] = None) -> Series:
```
tgaddair, line 776: Looks like this includes changes for image reading. Is that intentional?

PR #2100, opened 2022-06-06
Changes batch norm momentum defaults to 1-momentum

dantreiman:
tensorflow.keras Batch Norm momentum is parameterized as 1 - torch Batch Norm momentum:

tf.keras.layers.BatchNormalization: (default momentum: 0.99)
`moving_mean = moving_mean * momentum + mean(batch) * (1 - momentum)`

torch.nn.BatchNorm1d:  (default momentum: 0.1)
` x_hat = (1 - momentum) * x_hat + momentum * x_t`,
where `x_hat` is the estimated statistic and `x_t` is the new observed value.

Personally I agree with Tensorflow's definition, nevertheless since we heavily reference the torch docs I do think we should change defaults to match torch.

Going deeper: https://discuss.pytorch.org/t/batch-norm-momentum-default-value/11955


justinxzhao (approved):

ludwig/schema/combiners/tabnet.py
```diff
@@ -46,7 +46,7 @@ class TabNetCombinerConfig(BaseCombinerConfig):
     )
 
     bn_momentum: float = schema_utils.FloatRange(
-        default=0.95, description="Momentum of the batch norm. `m_B` in the paper."
+        default=0.05, description="Momentum of the batch norm. `m_B` in the paper."
```
w4nderlust, line 49: I would update the description to reflect the "inversion" otherwise someone looking at the paper may be confused

Issue #2102, opened 2022-06-06
Torchscript module throws error when inputs are exclusively binary features


geoffreyangus:
Upon saving and loading the torchscript module, it seems to raise an error on binary-only features with the following message:

```
...
E           Traceback of TorchScript, original code (most recent call last):
E             File "/Users/geoffreyangus/repositories/predibase/ludwig/ludwig/models/inference.py", line 71, in forward
E                           feature_name = get_name_from_module_dict_key(module_dict_key)
E                           preproc_inputs[feature_name] = preproc(inputs[feature_name])
E                       outputs = self.model(preproc_inputs)
E                                 ~~~~~~~~~~ <--- HERE
E               
E                       predictions: Dict[str, Dict[str, torch.Tensor]] = {}
E           /usr/local/lib/python3.8/site-packages/torch/nn/modules/linear.py(103): forward
...
E           RuntimeError: expected scalar type Float but found Bool
```

_Originally posted by @geoffreyangus in https://github.com/ludwig-ai/ludwig/pull/2090#discussion_r890526821_

geoffreyangus:
Fixed in #2103

PR #2103, opened 2022-06-06
Fix Torchscript for exclusively binary feature inputs

geoffreyangus:
This PR implements a fix and regression test for a bug in the torchscript implementation (#2102).


tgaddair (approved):

ludwig/features/binary_feature.py
```diff
@@ -69,13 +69,13 @@ def forward(self, v: TorchscriptPreprocessingInput):
             v = torch.stack(v)
 
         if torch.jit.isinstance(v, torch.Tensor):
-            return v.to(dtype=torch.bool)
+            return v.to(dtype=torch.float32)
```
justinxzhao, line 72: Curious why do these need to be explicitly casted to float32?

justinxzhao (approved):

ludwig/features/binary_feature.py
geoffreyangus: It seems that if the input features are comprised exclusively of binary features, torchscript fails.

My best guess: torchscript does not know to upcast these booleans to float32 during the forward pass, therefore causing an error.

PR #2105, opened 2022-06-07
enh: Implements `InferenceModule` as a pipelined module with separate preprocessor, predictor, and postprocessor modules

brightsparc:
In order to be able to create an ensemble or torchscript models, there is a benefit to having each of the steps in the inference pipeline their own module, such that these could be [ensembled](https://github.com/triton-inference-server/server/blob/main/docs/architecture.md#ensemble-models) together to form an inference graph in triton.

I have re-written the `InferenceModule` to be an `InferencePipelineModule` with the following steps

```
class InferencePipelineModule(nn.Module):
    """Wraps preprocessing, model forward pass, and postprocessing modules into a single module.

    The purpose of the module is to be scripted into Torchscript for native serving. The nn.ModuleDict attributes of
    this module use keys generated by feature_utils.get_module_dict_key_from_name in order to prevent name collisions
    with keywords reserved by TorchScript.
    """

    def __init__(self, model: "ECD", config: Dict[str, Any], training_set_metadata: Dict[str, Any]):
        super().__init__()

        self.preprocess_module = torch.jit.script(PreprocessModule(config, training_set_metadata))
        self.predict_module = torch.jit.script(PredictModule(model))
        self.postprocess_module = torch.jit.script(PostprocessModule(config, training_set_metadata))

    def forward(self, inputs: Dict[str, Union[List[str], List[torch.Tensor], torch.Tensor]]):
        with torch.no_grad():
            preproc_outputs = self.preprocess_module(inputs)
            predictions = self.predict_module(preproc_outputs)
            postproc_outputs = self.postprocess_module(predictions)
            return postproc_outputs
```

I have compared the performance of this pipeline over the original all in one, and identifier that it has very similar performance whilst being more composable.
* `75.9 µs ± 266 ns per loop` - InferenceModule
* `78.3 µs ± 255 ns per loop` - InferencePipelineModule


brightsparc:
This is looking good, is there a plan to depreciate `_InferenceModuleV0` as part of this release?

ludwig/models/inference.py
```diff
@@ -12,37 +12,40 @@
...
+PREPROCESSOR = "preprocessor"
+PREDICTOR = "predictor"
+POSTPROCESSOR = "postprocessor"
+INFERENCE_STAGES = [PREPROCESSOR, PREDICTOR, POSTPROCESSOR]
 
 FEATURES_TO_CAST_AS_STRINGS = {BINARY, CATEGORY, BAG, SET, TEXT, SEQUENCE, TIMESERIES, VECTOR}
 
 
-class InferenceModule(nn.Module):
+class InferenceModuleV0(nn.Module):
```
tgaddair, line 37: Can we remove this before landing?

tgaddair (approved):

ludwig/models/inference.py
geoffreyangus: Yup! Just here for benchmarking.

PR #2107, opened 2022-06-07
First pass at trainer schema separation.

ksbrar:
Still needs some local testing

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -492,13 +492,7 @@ class LudwigModel:
                 logger.info("Warnings and other logs:")
                 self.model = LudwigModel.create_model(self.config, random_seed=random_seed)
 
-            # init trainer
-            # schema_cls = (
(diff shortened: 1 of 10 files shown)


ksbrar:
@jppgks With commit([8e53d79](https://github.com/ludwig-ai/ludwig/pull/2107/commits/8e53d793ad2ae58a66a0cdebee68af1bcef26546) I fixed the major bug with the incorrect default not being set for GBM. But since each trainer schema also has a ray-specific impl. if you choose a ray trainer then the default in that case is still not correctly. 

There are a couple ways to fix this, I tried my hand at one. in the next commit [aa5e502](https://github.com/ludwig-ai/ludwig/pull/2107/commits/aa5e502441e547aa712f5e2894e80eaf34e1b135) - checkout `load_trainer_with_kwargs`.

I tagged a few locations with `TODO(joppe)` with some more notes.

jppgks (approved):

jppgks:
@ksbrar I merged so that I can rebase the gbt branch, but please do review the changes in this PR so we can address anything we missed :)

PR #2108, opened 2022-06-07
Added ability to stop and resume hyperopt / automl runs

tgaddair:
This PR adds two new API features that work with both hyperopt and automl:

1. `should_stop_hyperopt` callback function
2. `resume` parameter to automl and hyperopt entrypoint functions

The idea is that the user can configure a callback that will trigger stopping of the entire job when some criteria is met. Then at a later point in time, the user can re-run the job and the state will be picked up from where it left off.

This PR also makes a small change to the output name of hyperopt results. Previously the directory name was randomized like `trainable_func_12345`. Now the directory name matches that of the `experiment_name` param, so that if the user calls the same hyperopt function twice, the default behavior will be to resume the previous job from where it left off. This can be disabled by setting `resume=False`, which will create a new job and write trial results to the same directory.


ludwig/hyperopt/execution.py
```diff
@@ -698,9 +694,20 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
         run_experiment_trial_params = tune.with_parameters(run_experiment_trial, local_hyperopt_dict=hyperopt_dict)
         register_trainable(f"trainable_func_f{hash_dict(config).decode('ascii')}", run_experiment_trial_params)
 
+        class CallbackStopper(Stopper):
```
justinxzhao, line 697: Could we move this class declaration outside, since this function is already very long?

i.e.


```python
class CallbackStopper(Stopper):
     """Blanket callback for whether to stop hyperopt."""

     def __init__(self, callbacks: List[Callback]):
         self.callbacks = callbacks

     def __call__(self, trial_id, result):
                return False

            def stop_all(self):
                for callback in self.callbacks or []:
                    if callback.should_stop_hyperopt():
                        return True
                return False

"""
inside execute()
"""

callback_stopper = CallbackStopper(callbacks)

analysis = tune.run(
    ...
    stop=callback_stopper,
    ...
)
```

justinxzhao (commented):
Nice! Thanks for the changes.

ludwig/hyperopt/execution.py
```diff
@@ -719,7 +726,9 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
                 trial_name_creator=lambda trial: f"trial_{trial.trial_id}",
                 trial_dirname_creator=lambda trial: f"trial_{trial.trial_id}",
                 callbacks=tune_callbacks,
+                stop=CallbackStopper(),
                 verbose=hyperopt_log_verbosity,
+                resume="AUTO" if resume is None else resume,
```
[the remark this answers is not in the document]
tgaddair: Done.

ludwig/hyperopt/execution.py
tgaddair: Done.

justinxzhao (approved):
Thanks!

PR #2109, opened 2022-06-07
Deflakes and refactors torchscript tests

geoffreyangus:
This PR refactors the torchscript tests in a manner similar to `test_ray.py::test_ray_{tabular|text|sequence|audio|image|timeseries}` in order to surface output parity issues in a faster and more isolated manner. This should also reduce the time taken by each individual test.

This PR deflakes the torchscript tests by increasing the tolerance when comparing ludwig models and torchscript models trained on audio features. Audio features introduce some numerical mismatch (#2110), likely due to the complex preprocessing. Next steps include investigating this issue further.


justinxzhao (approved):
This new structure works for me.

PR #2111, opened 2022-06-07
fix: Move `combiner_registry` back to old location.

ksbrar:


ksbrar:
Closed in favor #2114

PR #2113, opened 2022-06-07
Use the Backend to check for dask dataframes, instead of a hard check.

justinxzhao:
Dask is not a required dependency for Ludwig.

To maintain this abstraction, we pipe in the Ludwig `backend` and use the backend's `df_engine` to check if the dataframe should be treated as a Dask dataframe.

### DIFF
--- a/ludwig/utils/dataframe_utils.py
+++ b/ludwig/utils/dataframe_utils.py
@@ -1,8 +1,10 @@
+from typing import Optional
+
 import numpy as np
 import pandas as pd
 
-from ludwig.backend import Backend, LOCAL_BACKEND
 from ludwig.constants import DASK_MODULE_NAME
+from ludwig.utils.types import DataFrame
 
 
 def is_dask_lib(df_lib):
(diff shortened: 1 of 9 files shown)


ludwig/utils/dataframe_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def is_dask_lib(df_lib):
+    """Returns whether the dataframe library is dask."""
+    return df_lib.__name__ == DASK_MODULE_NAME
+
+
+def is_dask_backend(backend: "Backend"):  # noqa: F821
+    """Returns whether the backend's dataframe is dask."""
+    return backend.df_engine.df_lib.__name__ == DASK_MODULE_NAME
```
w4nderlust, line 17: very minor nit, but why not do this instead?
`is_dask_lib(backend.df_engine.df_lib)`

w4nderlust:
I really like this refactoring, cleans up and separates things neatly

hungcs (approved):

ludwig/utils/dataframe_utils.py
justinxzhao: Good point. Done.

PR #2114, opened 2022-06-08
Fixed combiner schema creation

tgaddair:
Prior import was not running before combiners had been added to the registry.


ksbrar (approved):
90% sure this will pass tests

PR #2115, opened 2022-06-08
Changes for enabling checkpoint syncing for hyperopt

ShreyaR:
Changes necessary in order to use a KubernetesSyncer in order to sync checkpoints across pods.

The proposed changes use the storage on the node running the main process as the default storage location where all checkpoints are synced to.

I've left some comments in there right now mostly for debugging, which I plan to remove before merging this PR.

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -1,5 +1,6 @@
 import copy
 import datetime
+from functools import lru_cache
 import glob
 import json
 import logging
@@ -11,7 +12,7 @@ import traceback
 import uuid
 from inspect import signature
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from packaging import version
 
@@ -32,7 +33,6 @@ from ludwig.utils.misc_utils import get_from_registry
 logger = logging.getLogger(__name__)
 
 try:
-    import kubernetes
     import ray
     from ray import tune
     from ray.tune import register_trainable, Stopper
@@ -55,7 +55,6 @@ except ImportError as e:
     ray = None
     Stopper = object
     get_horovod_kwargs = None
-    kubernetes = None
 
 
 try:
@@ -175,6 +174,7 @@ class RayTuneExecutor:
         self.max_concurrent_trials = max_concurrent_trials
         self.sync_config = None
         self.sync_client = None
+        # Head node is the node to which all checkpoints are synced if running on a K8s cluster.
         self.head_node_ip = ray.util.get_node_ip_address()
 
     def _get_search_space(self, parameters: Dict) -> Tuple[Dict, Dict]:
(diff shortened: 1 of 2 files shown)


ludwig/hyperopt/execution.py
```diff
@@ -31,6 +31,7 @@
 logger = logging.getLogger(__name__)
 
 try:
+    import kubernetes
```
tgaddair, line 34: This is not a listed dependency. We should also not make it required, even for ray users, as not everyone will be running on k8s.

ShreyaR: Fixed.

ludwig/hyperopt/execution.py
```diff
@@ -258,15 +280,32 @@ def _cpu_resources_per_trial_non_none(self):
...
             return None
 
-        remote_checkpoint_dir = os.path.join(
-            self.sync_config.upload_dir, *_get_relative_checkpoints_dir_parts(trial_dir)
-        )
-        return get_cloud_sync_client(remote_checkpoint_dir), remote_checkpoint_dir
+        if self.sync_config.upload_dir is not None:
+            # Cloud storage sync config
+            return os.path.join(self.sync_config.upload_dir, *_get_relative_checkpoints_dir_parts(trial_dir))
+        else:
```
tgaddair, line 291: Should this be the following?

```
elif self.kubernetes_namespace is not None:
    ...
else:
    return None
```

ludwig/hyperopt/execution.py
```diff
@@ -258,15 +280,32 @@ def _cpu_resources_per_trial_non_none(self):
...
+            # Kubernetes sync config. Returns driver node name and path.
+            # When running on kubernetes, each trial is rsynced to the node running the main process.
+            assert self.kubernetes_namespace is not None
+            node_name = self._get_node_address_by_ip()(self.head_node_ip)
+
+            return (node_name, trial_dir)
+
+    @lru_cache(maxsize=1)
```
tgaddair, line 299: Can the node address change? I imagine if the ray train job can be made to be fault tolerance to head failure, it could.

ShreyaR: This is an interesting question. I expect that fault tolerance to head node failure for tune jobs is the only scenario in which head node IP will change.

However, in that case I would expect the `HyperoptExecutor` class to be initialized again, and so would have the correct head node address?

Re: caching here -- the returned value of this function doesn't depend on the head node IP at all, and so even if the head node were to change and we use the cached return val of this function, we'd still have the correct behavior.

Let me know if this makes sense or if there's something I can do to make this more robust.

ShreyaR: This is an interesting question. I expect that fault tolerance to head node failure for tune jobs is the only scenario in which head node IP will change.

However, in that case I would expect the `HyperoptExecutor` class to be initialized again, and so would have the correct head node address?

Re: caching here -- the returned value of this function doesn't depend on the head node IP at all, and so even if the head node were to change and we use the cached return val of this function, we'd still have the correct behavior.

Let me know if this makes sense or if there's something I can do to make this more robust.

ludwig/hyperopt/execution.py
ShreyaR: Thanks, fixed.

ludwig/hyperopt/execution.py
tgaddair: I think that sounds reasonable.

tgaddair (approved):

PR #2118, opened 2022-06-08
Comprehensive configs for trainer and combiner.

justinxzhao:
This changes `defaults.py` to source defaults directly from the `TrainerConfig`, `OptimizerConfig`, and `BaseCombinerConfig` marshmallow schemas.

Going forward, we'll have a single source of defaults (the schemas). Merging with the full schema also enables Ludwig to print and save comprehensive configs, which contain an exhaustive list of parameters, for Trainer and Combiner, insulating older models from subsequent changes to module constructor values, like #2050, which we reverted in #2069.

[Config diff](https://editor.mergely.com/fPRQo41M/). You'll notice that there are many more parameters in the config from trainer and combiner.

### DIFF
--- a/ludwig/schema/utils.py
+++ b/ludwig/schema/utils.py
@@ -1,6 +1,6 @@
 from dataclasses import field
 from typing import Dict as tDict
-from typing import List, Tuple, Union
+from typing import List, Tuple, Type, Union
 
 from marshmallow import EXCLUDE, fields, schema, validate, ValidationError
 from marshmallow_jsonschema import JSONSchema as js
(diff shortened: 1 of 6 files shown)


ludwig/utils/defaults.py
```diff
@@ -260,29 +187,25 @@ def merge_with_defaults(config):
...
-
     # ===== Input Features =====
     for input_feature in config["input_features"]:
         get_from_registry(input_feature[TYPE], input_type_registry).populate_defaults(input_feature)
 
     # ===== Combiner =====
-    set_default_value(config, "combiner", {TYPE: default_combiner_type})
+    set_default_value(config, COMBINER, {TYPE: default_combiner_type})
+    full_combiner_config, _ = load_config_with_kwargs(
+        combiner_registry[config[COMBINER][TYPE]].get_schema_cls(), config[COMBINER]
```
ksbrar, line 206: If this is working correctly locally, this is a moot point but because of the way the schemas and `combiners.py` are currently structured it is possible to import `combiner_registry` before any combiners have been registered. [See this thread](https://predibase.slack.com/archives/C03F25Q9LTB/p1654635842932799)

Might be safer to do an indirect import from `combiners.py` for now? (Though this may also cause a circular import...)

tests/ludwig/utils/test_defaults.py
```diff
@@ -85,15 +85,15 @@ def test_merge_with_defaults_early_stop(use_train, use_hyperopt_scheduler):
...
+        config[TRAINER] = {"batch_size": 42}
 
     if use_hyperopt_scheduler:
         # hyperopt scheduler cannot be used with early stopping
         config[HYPEROPT]["executor"][SCHEDULER] = SCHEDULER_DICT
 
     merged_config = merge_with_defaults(config)
 
-    expected = -1 if use_hyperopt_scheduler else default_early_stop
+    expected = -1 if use_hyperopt_scheduler else asdict(TrainerConfig())["early_stop"]
```
ksbrar, line 96: Can just do `TrainerConfig().early_stop`

ludwig/schema/utils.py
```diff
@@ -9,20 +9,22 @@
...
+def load_config_with_kwargs(
+    cls: Type["BaseMarshmallowConfig"], kwargs_overrides
+) -> "BaseMarshmallowConfig":  # noqa 0821
+    """Instatiates an instance of the marshmallow class and kwargs overrides instantiantes the schema."""
     assert_is_a_marshmallow_class(cls)
     schema = cls.Schema()
     fields = schema.fields.keys()
-    return load_config(cls, **{k: v for k, v in kwargs.items() if k in fields}), {
-        k: v for k, v in kwargs.items() if k not in fields
+    return load_config(cls, **{k: v for k, v in kwargs_overrides.items() if k in fields}), {
```
ksbrar, line 26: I like the new name.

tests/ludwig/utils/test_defaults.py
justinxzhao: Done.

ludwig/utils/defaults.py
justinxzhao: This seems to be working locally, both in unit tests as well as a real training run.

connor-mccorm (approved):

ksbrar (approved):

ludwig/utils/defaults.py
qiagu: @justinxzhao Hello, could you please have a look at https://github.com/ludwig-ai/ludwig/issues/2448.

PR #2123, opened 2022-06-09
Changed CheckpointManager to write the latest checkpoint to a consistent filename

tgaddair:
This changes the checkpoint format from encoding the global step in the filename to having a single `latest.ckpt` file.

The previous approach required having a background process that would cleanup old checkpoints as training progressed. However, since we only retain the latest checkpoint, this added extra overhead for no benefit, and also caused synchronization issues when background processes would attempt to copy checkpoint files while they were being deleted by the training process.

In this implementation, the global step is instead put into the state dict along with model and optimizer weights. Additionally, checkpoints are first written to a temp directory to guarantee atomicity and prevent background processes from syncing tempfiles that will later be deleted.

### DIFF
--- a/ludwig/utils/checkpoint_utils.py
+++ b/ludwig/utils/checkpoint_utils.py
@@ -13,6 +13,8 @@ from typing import Any, Dict, Optional
 
 import torch
 
+LATEST_FNAME = "latest.ckpt"
+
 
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):
This looks like a nice simplification.

Issue #2124, opened 2022-06-09
Tabnet combiner does not compile down to Torchscript


geoffreyangus:
When attempting to convert an ECD model with a Tabnet combiner to torchscript, we get some form of the following error:

```
def forward(self, input: Tensor) -> Tensor:
>       return F.linear(input, self.weight, self.bias)
E       RuntimeError: mat1 and mat2 shapes cannot be multiplied (2x15 and 160x320)

/usr/local/lib/python3.8/site-packages/torch/nn/modules/linear.py:103: RuntimeError
```

To reproduce this error, run `tests/integration_tests/test_torchscript.py::test_torchscript_e2e_tabular`, modifying the inline config with the following: 

```
config = {
    "input_features": input_features,
    "output_features": output_features,
    COMBINER: {"type": "tabnet"},
    TRAINER: {"epochs": 2},
}
```

geoffreyangus:
Closed with #2126.

PR #2126, opened 2022-06-09
Torchscript-compatible TabNet

geoffreyangus:
This PR implements a Torchscript-compatible TabNet (addressing #2124). The issues in the original implementation were twofold:

1. Some weirdness in the interaction between the Torchscript compiler and the weights shared between `torch.nn.Module` objects. This is a somewhat known issue (HuggingFace explicitly accounts for it through a `torchscript` flag: [link](https://huggingface.co/transformers/v1.0.0/torchscript.html#torchscript-flag-and-tied-weights)). This PR works around it by first registering a module with the exact same properties as the shared module, then doing an overwriting assignment to the shared module immediately following.
2. Inability to script custom `autograd.Function` subclasses. This is also a known issue (https://github.com/pytorch/pytorch/issues/22329). This PR works around it by decomposing the `forward` functions of the custom classes into standalone functions, which are scriptable.

The following validation was run in order to test the new changes: 
1. A new test was added in `tests/integration_tests/test_torchscript.py` which tests that a trained LudwigModel (with a TabNet combiner) has the same outputs as its torchscript equivalent.
2. Existing tests were modified in order to ensure that the custom autograd classes remained unchanged.
3. Two Ludwig models were trained on the Titanic dataset with a TabNet combiner. One was on the master branch (commit: `dd026ca2fb9e7e9dc0ef7fb9bcc73ccaae01b8a7`) and one was on this branch (commit: `34fe6da455d44edb8b6cb4947f4595303cf3511d`). The models were more or less the same: 
tabnet-performance-comparison 
Full config here: 
```
input_features:
  - name: Pclass
    type: category
  - name: Sex
    type: category
  - name: Age
    type: number
    preprocessing:
      missing_value_strategy: fill_with_mean
  - name: SibSp
    type: number
  - name: Parch
    type: number
  - name: Fare
    type: number
    preprocessing:
      missing_value_strategy: fill_with_mean
  - name: Embarked
    type: category

output_features:
  - name: Survived
    type: binary

combiner:
  type: tabnet
```

### DIFF
--- a/ludwig/modules/tabnet_modules.py
+++ b/ludwig/modules/tabnet_modules.py
@@ -182,11 +182,11 @@ class FeatureBlock(LudwigModule):
         self.size = size
         units = size * 2 if apply_glu else size
 
+        # Initialize fc_layer before assigning to shared layer for torchscript compatibilty
         self.fc_layer = nn.Linear(input_size, units, bias=False)
         if shared_fc_layer is not None:
             assert shared_fc_layer.weight.shape == self.fc_layer.weight.shape
-            self.fc_layer.weight = shared_fc_layer.weight
-            self.fc_layer.bias = shared_fc_layer.bias
+            self.fc_layer = shared_fc_layer
 
         self.batch_norm = GhostBatchNormalization(
             units, virtual_batch_size=bn_virtual_bs, momentum=bn_momentum, epsilon=bn_epsilon
--- a/ludwig/modules/tabnet_modules.py
+++ b/ludwig/modules/tabnet_modules.py
@@ -4,7 +4,7 @@ import torch
 import torch.nn as nn
 
 from ludwig.modules.normalization_modules import GhostBatchNormalization
-from ludwig.utils.entmax import entmax15, entmax_bisect, sparsemax
+from ludwig.utils.entmax import Entmax15, EntmaxBisect, Sparsemax
 from ludwig.utils.torch_utils import LudwigModule
 
 
@@ -182,10 +182,11 @@ class FeatureBlock(LudwigModule):
         self.size = size
         units = size * 2 if apply_glu else size
 
-        if shared_fc_layer:
-            self.fc_layer = shared_fc_layer
-        else:
-            self.fc_layer = nn.Linear(input_size, units, bias=False)
+        self.fc_layer = nn.Linear(input_size, units, bias=False)
+        if shared_fc_layer is not None:
+            assert shared_fc_layer.weight.shape == self.fc_layer.weight.shape
+            self.fc_layer.weight = shared_fc_layer.weight
+            self.fc_layer.bias = shared_fc_layer.bias
 
         self.batch_norm = GhostBatchNormalization(
             units, virtual_batch_size=bn_virtual_bs, momentum=bn_momentum, epsilon=bn_epsilon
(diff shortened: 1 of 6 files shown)


brightsparc (approved):
LGTM

ludwig/modules/tabnet_modules.py
```diff
@@ -182,10 +182,11 @@ def __init__(
         self.size = size
         units = size * 2 if apply_glu else size
 
-        if shared_fc_layer:
+        # Initialize fc_layer before assigning to shared layer for torchscript compatibilty
+        self.fc_layer = nn.Linear(input_size, units, bias=False)
+        if shared_fc_layer is not None:
+            assert shared_fc_layer.weight.shape == self.fc_layer.weight.shape
```
tgaddair, line 188: Is this a programming error if it happens, or a user error in the config? Want to make sure we aren't asserting on user errors, as those cannot be recovered from (and are generally disabled at runtime).

ludwig/utils/entmax/activations.py
```diff
@@ -217,11 +226,15 @@ def sparsemax(X, dim=-1, k=None):
     P : torch tensor, same shape as X
         The projection result, such that P.sum(dim=dim) == 1 elementwise.
     """
-
+    # Avoids call to custom autograd.Function during eval to ensure torchscript compatibility
+    # custom autograd.Function is not scriptable: https://github.com/pytorch/pytorch/issues/22329#issuecomment-506608053
+    if not training:
+        output, _ = _sparsemax_forward(X, dim, k)
```
tgaddair, line 232: I imagine this means we cannot use integrated gradients with torchscript, since we lose the grad info at predict time with this approach, is that right? Not the end of the world, but something we'll need to think about.

ludwig/utils/entmax/activations.py
```diff
@@ -190,7 +188,18 @@ def backward(cls, ctx, dY):
...
+    X = X - max_val  # same numerical stability trick as for softmax
+    X = X / 2  # divide by 2 to solve actual Entmax
+
+    tau_star, _ = _entmax_threshold_and_support(X, dim=dim, k=k)
+
+    Y = torch.clamp(X - tau_star, min=0) ** 2
+    return Y, {}
+
+
+def sparsemax(X, dim=-1, k=None, training=True):
```
tgaddair, line 202: Couldn't find it here, but from where does this `training` param get plumbed through to this sparsemax call?

tgaddair (approved):
Amazing! Thanks for doing the regression analysis.

ludwig/modules/tabnet_modules.py
geoffreyangus: IIUC is a programming error– this should essentially never happen unless something was implemented incorrectly

ludwig/utils/entmax/activations.py
geoffreyangus: This is coming from the `nn.Module` wrapper for each of the entmax modes: https://github.com/ludwig-ai/ludwig/blob/9a2efa6865a26bc1b7e95487d910c8723c06be6c/ludwig/utils/entmax/activations.py#L274

tgaddair: Ah, then this must be set when we call `model.train()` vs `model.eval()`. Got it, thanks!

w4nderlust (approved):

Issue #1823, opened 2022-03-16
Enable specifying custom columns and values for splitting.


justinxzhao:
Currently, Ludwig's dataset splitting requires either:
- Splitting is disabled completely
- Dataset is split randomly into train, validation, and test, according to `preprocessing.split_probabilities`
- Dataset is split according to a special metadata column in the data, `split`, with a fixed set of special values to do the association: `0`: train, `1`: validation, `2`: test.

We should extend this API to enable users to customize splitting by other columns and values.

```
preprocessing:
    split:
        (moved) force_split: false
        (moved) split_probabilities: [0.7, 0.1, 0.2]
        (new) split_column: split # Name of column that should be used for splitting
        (new) train_values: [0] # Values in the split_column that should be associated with the training split
        (new) validation_values: [1] # Values in the split_column that should be associated with the validation split
        (new) test_values: [2] # Values in the split_column that should be associated with the test split
```

Note: we may want to revisit this API if/when we support multiple test sets.

PR #2132, opened 2022-06-11
Restructured split config and added datetime splitting

### LINKED ISSUE #2129: RFC: Reorganize split configuration to support time-based split strategy
Current preprocessing configuration supports splitting based on a split column, split probabilities (train-val-test), and stratified splitting based on a category. Currently this is all flattened into the top-level of the preprocessing section:

```
preprocessing:
    force_split: false
    split_probabilities: [0.7, 0.1, 0.2]
    stratify: null
    category: {...}
    sequence: {...}
    text: {...}
```

The proposal here is to add the ability to split based on a date column such that the data is ordered by date (ascending) and then split according to the `split_probabilities`. For example, if `split_probabilities: [0.7, 0.1, 0.2]`, then the earliest 70% of the data will be used for training, the middle 10% used for validation, and the last 20% used for testing.

To achieve this, we could add a new param in this structure, but it starts to get confusing juggling mutually exclusive params. As such, the proposal would be to both add a new param and restructure this section to be more modular:

```
preprocessing:
    split:
        type: date
        column: created_ts
        probabilities: [0.7, 0.1, 0.2]
```

Existing split types would similarly be reworked:

Default:

```
preprocessing:
    split:
        type: random
        probabilities: [0.7, 0.1, 0.2]
```

Fixed:

```
preprocessing:
    split:
        type: fixed
        column: split
```

Stratified:

```
preprocessing:
    split:
        type: stratify
        column: region
        probabilities: [0.7, 0.1, 0.2]
```

One change in this approach would be that the user would need to be explicit about the split column, and splitting by fixed column would only happen if the user manually specifies it in the type.

tgaddair:
Closes #2129.


jppgks (commented):
This is a great improvement and much cleaner compared to the existing code, thanks Travis!

2 comments:

1. Assuming split_index can only be one of 3 values (0, 1 or 2), should we introduce an enum for it instead of working with integers?
2. Suggestion, potentially out of scope here: add JSON schema for split

ludwig/data/split.py
```diff
@@ -0,0 +1,222 @@
...
+    return splitter_cls(**kwargs)
+
+
+def split_dataset(
+    df: DataFrame,
+    global_preprocessing_parameters: Dict[str, Any],
+    backend: Backend,
+    random_seed: float = default_random_seed,
+) -> Tuple[DataFrame, DataFrame, DataFrame]:
+    if "split" not in global_preprocessing_parameters and SPLIT in df:
```
justinxzhao, line 209:
```suggestion
    if SPLIT not in global_preprocessing_parameters and SPLIT in df:
```

ludwig/visualize.py
```diff
@@ -234,8 +234,20 @@ def _extract_ground_truth_values(
         gt = gt_df[output_feature_name][split == ground_truth_split]
     elif split_file is not None:
         # retrieve from split file
-        split = load_array(split_file)
-        gt = gt_df[output_feature_name][split == ground_truth_split]
+        if split_file.endswith(".csv"):
+            # Legacy code path for previous split file format
```
justinxzhao, line 238: Is the plant to remove this after 0.6?

ludwig/data/dataframe/base.py
```diff
@@ -55,7 +55,11 @@ def reduce_objects(self, series, reduce_fn):
         raise NotImplementedError()
 
     @abstractmethod
-    def to_parquet(self, df, path):
+    def split(self, df, probabilities):
+        raise NotImplementedError()
+
+    @abstractmethod
+    def to_parquet(self, df, path, index=False):
```
justinxzhao, line 62: Are there type hints / docstrings we could add here?

ludwig/data/split.py
```diff
@@ -0,0 +1,222 @@
...
+        d2 = d1 + int(self.probabilities[1] * n)
+        return np.split(df.sample(frac=1, random_state=random_seed), [d1, d2])
+
+    def has_split(self, split_index: int) -> bool:
+        return self.probabilities[split_index] > 0
+
+
+@split_registry.register("fixed")
+class FixedSplitter(Splitter):
+    def __init__(self, column: str = SPLIT, **kwargs):
```
justinxzhao, line 80: At some point (perhaps in separate PR), what do you think about extending this to enable users to specify custom values for fixed splits?

For example, I can imagine users with an explicit split column, but the values in the column are like "TRAIN", "TEST" instead of 0, 1, and 2.

(default)
```yaml
split:
  type: fixed
  column: custom_split_column
  train_values: [0]
  validation_values: [1]
  test_values: [2]
```

(custom)
```yaml
split:
  type: fixed
  column: custom_split_column
  train_values: [train]
  validation_values: [test]
  test_values: [heldout]
```

Related: #1823

ludwig/data/dataframe/dask.py
```diff
@@ -88,12 +90,24 @@ def apply_objects(self, df, apply_fn, meta=None):
     def reduce_objects(self, series, reduce_fn):
         return series.reduction(reduce_fn, aggregate=reduce_fn, meta=("data", "object")).compute()[0]
 
-    def to_parquet(self, df, path):
+    def split(self, df, probabilities):
+        # First ensure that every split receives at least one partition.
```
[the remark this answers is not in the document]
tgaddair: No, so this method is inexact. We should eventually explore a better solution, but likely it would be more expensive in some way.

ludwig/data/split.py
tgaddair: Yes, I think we should definitely do this. Can be in a follow-up PR right after.

ludwig/visualize.py
tgaddair: Added deprecation warning.

ludwig/data/dataframe/dask.py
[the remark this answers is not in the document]
justinxzhao: Thanks for the clarification, and going with a cheaper approximate workaround sounds good for now. 
Could you extend the comment to include this additional detail/context?

ludwig/utils/backward_compatibility.py
```diff
@@ -153,6 +154,49 @@ def _upgrade_trainer(trainer: Dict[str, Any]):
         trainer[EVAL_BATCH_SIZE] = None
 
 
+def _upgrade_preprocessing(preprocessing: Dict[str, Any]):
```
justinxzhao, line 157: Could we add at least one unit test for this file/function?

Usefully, this would also serve as a form of documentation we can link to that shows off what the deprecated format is and what the new format should be.

justinxzhao (approved):
LGTM! Just a couple of minor comments.

justinxzhao:
Latest commits hopefully address #2174, CC: @jppgks

PR #2133, opened 2022-06-13
feat: Adding feature type shared parameter capability for hyperopt

arnavgarg1:
This change enables using the `defaults` keyword within hyperopt to set default parameters for feature groups. This will help add search spaces more concisely for datasets with a large number of features while also reducing the search space to allow for a deeper search during hyperopt. 

For e.g., a user can now add 

``` 
"defaults.text.num_filters": {"space": "choice", "categories": [64, 128, 256, 512]}
```

in their hyperopt search space so that all text features (inputs and outputs) use the sampled `cell_type` for that particular trial. 

The API for this change is as follows: `defaults.<feature_type>.<parameter>`, where:

- `feature_type`: a valid input or output feature type, for e.g., text/category/sequence etc.
- `parameter`: the parameter to search over for that particular feature type, for e.g., `embedding_size` for category features

Example Configuration:

```yaml
hyperopt:
    goal: minimize
    parameters:
        trainer.learning_rate:
            lower: 0.0001
            upper: 0.01
            space: loguniform
        defaults.text.num_filters:
            space: choice
            categories: [64, 128, 256, 512]
        defaults.category.embedding_size:
            space: choice
            categories: [64, 128, 256, 512]
        ....
    ....
```

_Note: The output `status` that gets logged every few seconds will show the parameter as `defaults.text.cell_type` in the printed table with each trial status, but underneath the surface, this is getting updated correctly. This is unfortunately not something we can control, but I think it is okay to leave it this way anyway because the table width would become too large if we were to add a column for each feature whose default parameter was updated in the trial status table that gets printed.

Screen Shot 2022-06-13 at 1 12 33 PM_


arnavgarg1:
Working on writing some tests, but wanted to make this PR so that we can try with the Uber datasets

arnavgarg1:
Working on writing some tests so will update this PR, but wanted to create the PR so we can test if it works.

arnavgarg1:
Hmm, weird that the test are failing in some cases. It seems like the actual key in the dictionary of `trial_row` is actually `config.defaults.text.cell_type` but some reason in both the `pytest/Unit Test Results` and `pytest/py3.9 torch-nightly`, I see the only valid key in the object to be `config/defaults.text.cell_type`. Not sure why this is happening, but even more so I'm not sure how to debug this @tgaddair @ShreyaR

arnavgarg1:
Found out that the reason for the key error with my test is because of this warning

`UserWarning: Dataframes will use '/' instead of '.' to delimit nested result keys in future versions of Ray. For forward compatibility, set the environment variable TUNE_RESULT_DELIM='/'` 

Not sure if this would impact other tests or anything other parsing we do in other parts of Ludwig?

tests/integration_tests/test_hyperopt.py
```diff
@@ -314,3 +315,96 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
 
     if os.path.isfile(os.path.join(tmpdir, "test_hyperopt", "hyperopt_statistics.json")):
         os.remove(os.path.join(tmpdir, "test_hyperopt", "hyperopt_statistics.json"))
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("search_space", ["random"])
+def test_hyperopt_run_hyperopt_with_shared_params(csv_filename, search_space, tmpdir, ray_cluster):
```
ShreyaR, line 322: It looks like you aren't using the `ray_cluster` argument?

Also, instead of passing in `tmpdir`, can you use the [tempfile](tmpdir) package? You should be able to remove the cleanup at the end of the tests.

tests/integration_tests/test_hyperopt.py
```diff
@@ -314,3 +315,96 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+                    cell_type = input_feature["cell_type"]
+                    # Check that cell type got updated from the sampler
+                    assert cell_type in cell_types
+                    text_input_cell_types.add(cell_type)
+            # All text features with defaults should have the same cell type for this trial
+            assert len(text_input_cell_types) == 1
+            params_fd.close()
+        # Likely unable to open trial dir so fail this test
+        except OSError:
+            assert 1 == 0
```
ShreyaR, line 407: Can you raise a more descriptive error (like a RuntimeError that states which directory was unable to be opened) instead of the assert?

tests/integration_tests/test_hyperopt.py
```diff
@@ -314,3 +315,96 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
 
     if os.path.isfile(os.path.join(tmpdir, "test_hyperopt", "hyperopt_statistics.json")):
         os.remove(os.path.join(tmpdir, "test_hyperopt", "hyperopt_statistics.json"))
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("search_space", ["random"])
+def test_hyperopt_run_hyperopt_with_shared_params(csv_filename, search_space, tmpdir, ray_cluster):
+    input_features = [
```
ShreyaR, line 323: You can abstract the initialization of the Ludwig config into a separate function, see `test_hyperopt_search_alg` above. This was you can write follow up tests on the same config if needed.

You can either edit the existing `_setup_ludwig_config` function or add a new ludwig config setup method that is specific for your test.

tests/integration_tests/test_hyperopt.py
```diff
@@ -314,3 +315,96 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+        category_feature(vocab_size=2, reduce_input="sum"),
+        category_feature(vocab_size=6),
+    ]
+
+    output_features = [category_feature(vocab_size=2, reduce_input="sum")]
+
+    rel_path = generate_data(input_features, output_features, csv_filename)
+
+    config = {
+        "input_features": input_features,
```
ShreyaR, line 335: nit: You can import defaults for `INPUT_FEATURES`, etc. here as well

ludwig/features/sequence_feature.py
```diff
@@ -226,7 +226,7 @@ def add_feature_data(
 
 
 class SequenceInputFeature(SequenceFeatureMixin, InputFeature):
-    encoder = "embed"
+    encoder = "parallel_cnn"
```
arnavgarg1, line 229: I felt like this was a typo so added this change into this PR. Official [ludwig documentation](https://ludwig-ai.github.io/ludwig-docs/0.5/configuration/features/sequence_features/#sequence-input-features-and-encoders) says that the default encoder for sequence features is a parallel_cnn

ludwig/hyperopt/execution.py
```diff
@@ -844,15 +898,33 @@ def get_parameters_dict(parameters):
     return parameters_dict
 
 
-def substitute_parameters(config, parameters):
+def substitute_parameters(config, parameters, shared_params_features_dict):
```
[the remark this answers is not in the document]
arnavgarg1: @justinxzhao will add type hints! 

Theoretically speaking, `shared_params_features_dict` can never be of type `None` with the implementation I have, but the reason I have allowed it to be None is that there are quite a few hyperopt tests that directly use the `HyperoptExecutor.execute()` function instead of calling the top level `hyperopt()` function, which overrides the creation of `shared_params_features_dict` that happens with the top level `hyperopt()` function. 

One option is to modify all the tests that display this behavior, or to just state in the type hints that it can be None. 

Should I just modify all of the tests so that it's done correctly via the top level function?

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,34 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
         hyperopt_params[EXECUTOR],
         executor_defaults,
     )
+
+
+def get_shared_params_dict(config_feature_type: str, features: Dict[str, Any]) -> Dict[str, Set]:
+    """Generates a mapping of feature type to the corresponding set of features without an encoder or one using the
+    default encoder for that feature type.
```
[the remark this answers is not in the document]
ShreyaR: @justinxzhao yes, we currently don't support hyperopt on preprocessing parameters as it's too expensive. It would be good to clarify this in the docstring as a comment.

@arnavgarg1 do we now support hyperopt search over output features? If we do, it would be good to update that in the documentation.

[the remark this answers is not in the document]
arnavgarg1: @justinxzhao The rationale is primarily that using default encoders makes it easier to ensure that the shared parameters being defined are actually applicable to all features of that type because they all share the same parameters. Different encoders/decoders are likely to have different parameters so this was just a simplification. Theoretically, we can add another layer of complexity to do defaults per feature type per encoder, but I don't think we need to provide this level of granularity at this time. Let me know if you feel differently!

@ShreyaR will add both things to the documentation/doc string.

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
justinxzhao: I would probably not modify all of the tests since it's easy enough to handle the `None` case.

One additional suggestion is to pass `shared_params_features_dict` as is, and then do shared_params_features_dict[shared_params_type] within `substitute_parameters()`.

```
shared_params_type=INPUT_FEATURES,
shared_params_features_dict=shared_params_features_dict),
```

tests/integration_tests/test_hyperopt.py
```diff
@@ -301,21 +338,122 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+        DEFAULTS
+        + "."
+        + INPUT_FEATURES
+        + "."
+        + TEXT
+        + ".cell_type": {"space": "choice", "categories": cell_types_search_space},
+        DEFAULTS + "." + OUTPUT_FEATURES + "." + CATEGORY + ".vocab_size": {"space": "randint", "lower": 4, "upper": 8},
+    }
+
+    # add hyperopt parameter space to the config
```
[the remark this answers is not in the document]
arnavgarg1: @justinxzhao I had done that initially, but in a prior code review for this PR, there was a suggestion to make this part out into a separate function. Any strong reasons beyond having all the code in one place to make it easier to read to move this all into one function? 

I think the main reason to have this in a separate function is that in case additional tests need to be added in the future using the same config, it makes sense to have it be returned programmatically rather than repeating code.

tests/integration_tests/test_hyperopt.py
```diff
@@ -301,21 +338,122 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
-    hyperopt_results = hyperopt(config, dataset=rel_path, output_directory=tmpdir, experiment_name="test_hyperopt")
-    if search_space == "random":
-        assert hyperopt_results.experiment_analysis.results_df.shape[0] == RANDOM_SEARCH_SIZE
-    else:
-        # compute size of search space for grid search
-        grid_search_size = 1
-        for k, v in search_parameters.items():
-            grid_search_size *= len(v["values"])
-        assert hyperopt_results.experiment_analysis.results_df.shape[0] == grid_search_size
+    with tempfile.TemporaryDirectory() as tmpdir:
```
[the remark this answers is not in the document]
arnavgarg1: @justinxzhao I had done that initially, but in a prior code review for this PR, there was a suggestion to use the `tempfile` package instead of the tmpdir fixture. Any strong reasons to use the fixture instead of `tempfile`?

I think the biggest advantage of `tempfile` is that the cleanup code is no longer required which is nice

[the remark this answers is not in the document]
justinxzhao: I believe the `tmpdir` pytext fixture also takes care of deleting any files that were written under `tmpdir`. Perhaps the old code manually deleting files in the `tmpdir` was redundant.

tests/integration_tests/test_hyperopt.py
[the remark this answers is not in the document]
justinxzhao: Here, my suggestion is just to make the config["hyperopt"] expressed as one block instead of multiple blocks, unless we're doing heavy parameterization, which I don't think this test necessarily needs.

```python
@pytest.mark.distributed
def test_hyperopt_run_hyperopt_with_shared_params(csv_filename, tmpdir):
    config, rel_path = _setup_ludwig_config_with_shared_params(csv_filename)

    categorical_feature_name = config[INPUT_FEATURES][2][NAME]
    output_feature_name = config[OUTPUT_FEATURES][0][NAME]

    cell_types_search_space = ["lstm", "gru"]
    vocab_size_search_space = list(range(4, 9))

    config["hyperopt"] = {
        "parameters": {
            "trainer.learning_rate": {"lower": 0.0001, "upper": 0.01, "space": "loguniform"},
            f"{categorical_feature_name}.vocab_size": {"space": "randint", "lower": 1, "upper": 3},
            "defaults.input_features.text.cell_type": {"space": "choice", "categories": ["lstm", "gru"]},
            "defaults.output_features.category.vocab_size": {"space": "randint", "lower": 4, "upper": 8},
        },
        "goal": "minimize",
        "validation_metrics": "loss",
        "executor": {"type": "ray", "num_samples": 4},
        "search_alg": {"type": "variant_generator"},
    }

    hyperopt_results = hyperopt(config, dataset=rel_path, output_directory=tmpdir, experiment_name="test_hyperopt")

    # Check that the trials did sample from defaults in the search space
    ...
```

ludwig/hyperopt/execution.py
```diff
@@ -818,7 +843,36 @@ def get_build_hyperopt_executor(executor_type):
 executor_registry = {"ray": RayTuneExecutor}
 
 
-def set_values(model_dict, name, parameters_dict):
+def set_values(
```
justinxzhao, line 846: While this is technically covered by the larger end to end integration test, it might be good to add unit tests for this function explicitly since what this function is doing is already rather non-trivial.

arnavgarg1: Agreed! I've decided to split this into two functions - one that just sets value for normal parameters from the sampler, and another that sets values for shared parameters. This will also let me get rid of the deeply nested conditionals. Will update :) 

Will try and add unit tests for my new simplified functions

tests/integration_tests/test_hyperopt.py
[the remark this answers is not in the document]
arnavgarg1: @justinxzhao Just checked and you're right, the old code was redundant since the tmpdir  pytext fixture uses tempfile internally :) Great catch!

tests/integration_tests/test_hyperopt.py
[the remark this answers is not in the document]
w4nderlust: I see an issue here. vocab size is an internal parameter, not an externally visible one (we should find a better way to make the distinction clear).
What you actually want to use .preprocessing.most_common (also not a great name, we may want to change it in the future to max_vocab_size)

tests/integration_tests/test_hyperopt.py
```diff
@@ -301,21 +338,122 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+                        # Category input features have vocab_size in range [1,3] inclusive
+                        assert vocab_size not in vocab_size_search_space
+                # All text features with defaults should have the same cell_type for this trial
+                assert len(text_input_cell_types) == 1
+                for output_feature in output_features:
+                    if output_feature[TYPE] == CATEGORY:
+                        vocab_size = output_feature["vocab_size"]
+                        # Check that vocab_size got updated from the sampler
+                        assert vocab_size in vocab_size_search_space
+                params_fd.close()
```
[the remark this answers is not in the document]
arnavgarg1: @justinxzhao I'm thinking that it makes sense to break these up into two different tests. 

1. To test whether default values are getting sampled and being returned in the `experiment_analysis_df` returned by `tune.run()` for each trial [trial table]
2. To test if the underlying config for the features gets updated correctly in the scenario that we're using default parameters in the search space [for written config]

What do you think?

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
         executor_defaults,
     )
+
+
+def get_shared_params_dict(config_feature_type: str, features: Dict[str, Any]) -> Dict[str, Dict[str, Set]]:
+    """Generates a mapping of feature type to the corresponding set of features without an encoder or one using the
+    default encoder for that feature type.They may be considered for potential shared parameter search spaces
+    depending on the parameter space defined later within the hyperopt config.
+
+    This applies to both config_feature_types - input_features and output_features. The shared parameters for both
```
justinxzhao, line 414: Can we add somewhere:

```
Note that shared default parameter search spaces are not applied to features with non-default encoders or non-default decoders, since shared default parameter values should only apply to default modules.
```

ludwig/hyperopt/execution.py
```diff
@@ -819,15 +844,92 @@ def get_build_hyperopt_executor(executor_type):
...
+        if isinstance(value, dict):
+            for sub_key, sub_value in value.items():
+                if key not in model_dict:
+                    model_dict[key] = dict()
+                model_dict[key][sub_key] = sub_value
+        else:
+            model_dict[key] = value
+
+
+def update_model_dict_with_shared_params(
```
justinxzhao, line 858: nit: Rename to `update_features_with_shared_params()`?

ludwig/hyperopt/execution.py
```diff
@@ -819,15 +844,92 @@ def get_build_hyperopt_executor(executor_type):
...
+    model_dict: Dict[str, Any],
+    trial_parameters_dict: Dict[str, Dict[str, Any]],
+    config_feature_group: str = None,
+    shared_params_features_dict: Dict[str, Dict[str, Set]] = None,
+):
+    """Updates the parameters of feature_name in model_dict based on hyperopt parameters sampled.
+
+    :param model_dict: Underlying config for the specific input/output feature populated with defaults and feature
+            specific parameters. This may be updated with values from the hyperopt search space.
```
justinxzhao, line 867:
```suggestion
    :param model_dict: Underlying config for the specific input/output feature populated with potentially a mix of default and feature-specific parameters. This may be updated with values from the hyperopt search space.
```

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
         hyperopt_params[EXECUTOR],
         executor_defaults,
     )
+
+
+def get_shared_params_dict(config_feature_type: str, features: Dict[str, Any]) -> Dict[str, Dict[str, Set]]:
```
justinxzhao, line 409: nit: `get_features_eligible_for_shared_params()`?

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
+    """Generates a mapping of feature type to the corresponding set of features without an encoder or one using the
+    default encoder for that feature type.They may be considered for potential shared parameter search spaces
+    depending on the parameter space defined later within the hyperopt config.
+
+    This applies to both config_feature_types - input_features and output_features. The shared parameters for both
+    config_feature_types must be specified separately.
+
+    TODO(#2167): Make sure each feature has a type defined in the JSONSchema for Hyperopt
```
justinxzhao, line 417: Add:

```
Returns:
  Dict of feature type -> set of feature names with that type that are eligible for shared parameters (they use the default encoder or default decoder).
```

ludwig/hyperopt/run.py
```diff
@@ -169,6 +186,12 @@ def hyperopt(
     else:
         config_dict = config
 
+    # Get mapping of input/output features that don't have an encoder for shared parameters
+    shared_params_features_dict = {
```
justinxzhao, line 190: nit: Rename to `features_eligible_for_shared_params`

ludwig/hyperopt/execution.py
```diff
@@ -819,15 +844,92 @@ def get_build_hyperopt_executor(executor_type):
...
+    :param trial_parameters_dict: Config produced by the hyperopt sampler based on the parameter search space. It maps
+            the name of the feature to the sampled parameters for that feature. For default parameters, it creates
+            nested dictionaries for each feature type.
+    :type trial_parameters_dict: dict[str, dict[str, any]]
+    :param config_feature_group: Indicates whether the feature is an input feature or output feature (can be either of
+            `input_features` or `output_features`).
+    :type config_feature_group: str
+    :param shared_params_features_dict: Mapping that stores all input and output feature names that use default
+            encoders or decoders based on their feature type. At the top level, these are separated into two groups -
+            input features and output features.
```
justinxzhao, line 878:
```suggestion
    :param shared_params_features_dict: Collection of names of features that are eligible for using shared parameters, keyed by `input_features` or `output_features` and then by feature type.
```

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
+        feature_type = feature[TYPE]
+        feature_group_to_features_map[feature_type].add(feature_name)
+
+    for feature in features:
+        if TYPE not in feature:
+            raise ValueError("Ludwig expects feature types to be defined for each feature within the config")
+        if config_feature_type == INPUT_FEATURES:
+            default_encoder = get_from_registry(feature[TYPE], input_type_registry).encoder
+            if not feature.get(ENCODER, 0) or feature.get(ENCODER) == default_encoder:
+                update_feature_group_mapping(feature)
```
justinxzhao, line 433:
```suggestion
                features_eligible_for_shared_params[feature[TYPE]].add(feature[NAME])
```

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
+
+    TODO(#2167): Make sure each feature has a type defined in the JSONSchema for Hyperopt
+    """
+
+    feature_group_to_features_map = defaultdict(set)
+
+    def update_feature_group_mapping(feature):
+        feature_name = feature[NAME]
+        feature_type = feature[TYPE]
+        feature_group_to_features_map[feature_type].add(feature_name)
```
justinxzhao, line 425:
```suggestion
```

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
+        if TYPE not in feature:
+            raise ValueError("Ludwig expects feature types to be defined for each feature within the config")
+        if config_feature_type == INPUT_FEATURES:
+            default_encoder = get_from_registry(feature[TYPE], input_type_registry).encoder
+            if not feature.get(ENCODER, 0) or feature.get(ENCODER) == default_encoder:
+                update_feature_group_mapping(feature)
+        else:
+            default_decoder = get_from_registry(feature[TYPE], output_type_registry).decoder
+            if not feature.get(DECODER, 0) or feature.get(DECODER) == default_decoder:
+                update_feature_group_mapping(feature)
```
justinxzhao, line 437:
```suggestion
                features_eligible_for_shared_params[feature[TYPE]].add(feature[NAME])
```

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
+    default encoder for that feature type.They may be considered for potential shared parameter search spaces
+    depending on the parameter space defined later within the hyperopt config.
+
+    This applies to both config_feature_types - input_features and output_features. The shared parameters for both
+    config_feature_types must be specified separately.
+
+    TODO(#2167): Make sure each feature has a type defined in the JSONSchema for Hyperopt
+    """
+
+    feature_group_to_features_map = defaultdict(set)
```
justinxzhao, line 420:
```suggestion
    features_eligible_for_shared_params = defaultdict(set)
```

ludwig/hyperopt/run.py
```diff
@@ -373,3 +404,36 @@ def update_hyperopt_params_with_defaults(hyperopt_params):
...
+        if config_feature_type == INPUT_FEATURES:
+            default_encoder = get_from_registry(feature[TYPE], input_type_registry).encoder
+            if not feature.get(ENCODER, 0) or feature.get(ENCODER) == default_encoder:
+                update_feature_group_mapping(feature)
+        else:
+            default_decoder = get_from_registry(feature[TYPE], output_type_registry).decoder
+            if not feature.get(DECODER, 0) or feature.get(DECODER) == default_decoder:
+                update_feature_group_mapping(feature)
+
+    return feature_group_to_features_map
```
justinxzhao, line 439:
```suggestion
    return features_eligible_for_shared_params
```

tests/integration_tests/test_hyperopt.py
```diff
@@ -106,6 +120,62 @@ def _setup_ludwig_config(dataset_fp: str) -> Tuple[Dict, str]:
...
+    the version of Ray. Returns None if the parameter key is not found.
+
+    TODO(#2176): There are different key name delimiters depending on Ray version. Simplify this as Ray is upgraded.
+    The delimiter in future versions of Ray will be '/' instead of '.'
+    """
+    if f"config.{parameter_key}" in trial_row:
+        return trial_row[f"config.{parameter_key}"]
+    elif f"config/{parameter_key}" in trial_row:
+        return trial_row[f"config/{parameter_key}"]
+    return None
```
justinxzhao, line 176: nit: Remove.

tests/integration_tests/test_hyperopt.py
```diff
@@ -106,6 +120,62 @@ def _setup_ludwig_config(dataset_fp: str) -> Tuple[Dict, str]:
...
+            "validation_metrics": "loss",
+            "executor": {"type": "ray", "num_samples": RANDOM_SEARCH_SIZE},
+            "search_alg": {"type": "variant_generator"},
+        },
+    }
+
+    return config, rel_path, cell_types_search_space, vocab_size_search_space
+
+
+def _get_trial_parameter_value(parameter_key: str, trial_row: str) -> str:
```
justinxzhao, line 165:
```suggestion
def _get_trial_parameter_value(parameter_key: str, trial_row: str) -> Union[str, None]:
```

tests/integration_tests/test_hyperopt.py
```diff
@@ -106,6 +120,62 @@ def _setup_ludwig_config(dataset_fp: str) -> Tuple[Dict, str]:
...
+    """Returns the parameter value from the Ray trial row, which has slightly different column names depending on
+    the version of Ray. Returns None if the parameter key is not found.
+
+    TODO(#2176): There are different key name delimiters depending on Ray version. Simplify this as Ray is upgraded.
+    The delimiter in future versions of Ray will be '/' instead of '.'
+    """
+    if f"config.{parameter_key}" in trial_row:
+        return trial_row[f"config.{parameter_key}"]
+    elif f"config/{parameter_key}" in trial_row:
```
justinxzhao, line 174:
```suggestion
    if f"config/{parameter_key}" in trial_row:
```

tests/integration_tests/test_hyperopt.py
```diff
@@ -317,5 +387,53 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+    hyperopt_results = hyperopt(config, dataset=rel_path, output_directory=tmpdir, experiment_name="test_hyperopt")
+    hyperopt_results_df = hyperopt_results.experiment_analysis.results_df
+
+    # Check that the trials did sample from defaults in the search space
+    for _, trial_row in hyperopt_results_df.iterrows():
+        vocab_size = _get_trial_parameter_value("defaults.output_features.category.vocab_size", trial_row)
+
+        cell_type = _get_trial_parameter_value("defaults.input_features.text.cell_type", trial_row)
+        if cell_type is not None:
+            cell_type = cell_type.replace('"', "")
```
justinxzhao, line 406: The .replace could probably just go in `_get_trial_parameter_value()`

tests/integration_tests/test_hyperopt.py
```diff
@@ -317,5 +387,53 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+
+    hyperopt_results = hyperopt(config, dataset=rel_path, output_directory=tmpdir, experiment_name="test_hyperopt")
+    hyperopt_results_df = hyperopt_results.experiment_analysis.results_df
+
+    # Check that the trials did sample from defaults in the search space
+    for _, trial_row in hyperopt_results_df.iterrows():
+        vocab_size = _get_trial_parameter_value("defaults.output_features.category.vocab_size", trial_row)
+
+        cell_type = _get_trial_parameter_value("defaults.input_features.text.cell_type", trial_row)
+        if cell_type is not None:
```
justinxzhao, line 405: The cell type should always be non-None in this test, right?

justinxzhao (commented):
Nice! Much cleaner. Tests look a lot cleaner too.

New comments and suggestions mainly around renaming `shared_params_features_dict` to `features_eligible_for_shared_params`, which I feel is a better name.

tests/integration_tests/test_hyperopt.py
arnavgarg1: @justinxzhao did we update to Ray 1.13?

tests/integration_tests/test_hyperopt.py
arnavgarg1: Yup, agreed

justinxzhao (approved):
Thanks for the changes! LGTM

w4nderlust:
Not sure where my comment went, but i have an issue with vocab_size. it is an internal only parameter, that is derived by most_common and preprocessing, it's not something to hyperopt on, but the pr searches over it, we should choose another parameter to search over

arnavgarg1:
> Not sure where my comment went, but i have an issue with vocab_size. it is an internal only parameter, that is derived by most_common and preprocessing, it's not something to hyperopt on, but the pr searches over it, we should choose another parameter to search over

This should be fixed now!

w4nderlust:
> > Not sure where my comment went, but i have an issue with vocab_size. it is an internal only parameter, that is derived by most_common and preprocessing, it's not something to hyperopt on, but the pr searches over it, we should choose another parameter to search over
> 
> This should be fixed now!

In test_hyperopt I can see it though

arnavgarg1:
> > > Not sure where my comment went, but i have an issue with vocab_size. it is an internal only parameter, that is derived by most_common and preprocessing, it's not something to hyperopt on, but the pr searches over it, we should choose another parameter to search over
> > 
> > 
> > This should be fixed now!
> 
> In test_hyperopt I can see it though

Good catch, I've updated all hyperopt tests in test_hyperopt to use parameters that are user-facing now :)

w4nderlust:
> > > > Not sure where my comment went, but i have an issue with vocab_size. it is an internal only parameter, that is derived by most_common and preprocessing, it's not something to hyperopt on, but the pr searches over it, we should choose another parameter to search over
> > > 
> > > 
> > > This should be fixed now!
> > 
> > 
> > In test_hyperopt I can see it though
> 
> Good catch, I've updated all hyperopt tests in test_hyperopt to use parameters that are user-facing now :)

Just to provide additional clarification: in the tests, when we define features like `category_feature()`, we provide a `vocab_size`. The reason there is that that function has two purposes: it creates a dictionary that can be both used by `synthesize_data` and as part of a ludwig config. `vocab_size` is a parameter for `synthesize_data` (it determines how many categories will there be in the generated data) and not a parameter for Ludwig config (because it is internally determined by ludwig).

So, its use in, say [tests/integration_tests/test_hyperopt.py](https://github.com/ludwig-ai/ludwig/pull/2133/files#diff-53bd1d32793cc8ee9b0a79ef7829dccfdf59d54a7d9f6f7138a6d4dd6502c3b1) line 89 and 92 is completely fine, and indeed helps generating data with a small vocabulary, which in turn helps making those tests faster (although I think for category features, it should be minimum 3). So that should be kept. So I suggest reintroducing those uses of `vocab_size`.

What I was arguing for removal, and sorry if I wasn't more precise about it before, is the presence of `vocab_size` as a parameter in the hyperopt section of the test config, as there it is basically a config override of an internal only parameter. `embedding_size` on the other hand, is a good replacement parameter you are using instead of `vocab_size`, because it is actually a parameter we expect users of hyperopt to be able to change.

Another unrelated comment:
```
num_filters_search_space = [128, 512]
embedding_size_search_space = [128, 512]
```
Those are reasonable values user may set, although in these tests we want to make them as small as possible to make tests quick. Here i suggest something like 4 and 8 to be the two options.

arnavgarg1:
> tests/integration_tests/test_hyperopt.py

That makes a lot of sense @w4nderlust , thanks for the in-depth clarification! I've added `vocab_size` back to any instance where a categorical feature is being created, but removed all instances where it was in the hyperopt search space. I've also reduced the sample space to be [4, 8] to speed up the tests.

arnavgarg1:
@justinxzhao @ShreyaR @w4nderlust Does this PR look good to go? Happy to make any additional changes if needed!

tests/integration_tests/test_hyperopt.py
```diff
@@ -313,5 +383,57 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
...
+    for _, trial_row in hyperopt_results_df.iterrows():
+        embedding_size = _get_trial_parameter_value("defaults.output_features.category.embedding_size", trial_row)
+        num_filters = _get_trial_parameter_value("defaults.input_features.text.num_filters", trial_row)
+        assert embedding_size in embedding_size_search_space
+        assert num_filters in num_filters_search_space
+
+
+@pytest.mark.distributed
+def test_hyperopt_with_shared_params_written_config(csv_filename, tmpdir):
+    config, rel_path, num_filters_search_space, embedding_size_search_space = _setup_ludwig_config_with_shared_params(
```
ShreyaR, line 406: Can you consolidate this test and `test_hyperopt_run_shared_params_trial_table` into a single test?

It seems like they run the same hyperopt experiment (which is kind of slow to run). If you run the experiment only once and test everything on the same experiment, the tests will be faster overall.

I would recommend keeping the abstraction you've used here, but merging the tests. So something like:

```
def test_shared_parameters_trial_table(hyperopt_results_df, shared_params_dict):
    # Checks that the trials did sample from defaults in the search space
    # Shared params dict: {shared_param_name: shared_param_search_space}. E.g.: {'defaults.output_features.category.embedding_size': embedding_size_search_space, ...}
    for _, trial_row in hyperopt_results_df.iterrows():
        for shared_param_name, shared_param_search_space in shared_params_dict.items():
            param_val = _get_trial_parameter_value(shared_param_name, trial_row)
            assert param_val in shared_param_search_space


def test_shared_params_written_config(hyperopt_results_df, ...):
    # Abstracts out everything that happens from lines 414 - 439


@pytest.mark.distributed
def test_hyperopt_with_shared_params(csv_filename, tmpdir):
    config, rel_path, num_filters_search_space, embedding_size_search_space = _setup_ludwig_config_with_shared_params(
        csv_filename
    )

    hyperopt_results = hyperopt(config, dataset=rel_path, output_directory=tmpdir, experiment_name="test_hyperopt")
    hyperopt_results_df = hyperopt_results.experiment_analysis.results_df

    test_shared_parameters_trial_table(...)
    test_shared_params_written_config(...)

```

tests/integration_tests/test_hyperopt.py
```diff
@@ -105,6 +119,60 @@ def _setup_ludwig_config(dataset_fp: str) -> Tuple[Dict, str]:
...
+
+
+def _get_trial_parameter_value(parameter_key: str, trial_row: str) -> Union[str, None]:
+    """Returns the parameter value from the Ray trial row, which has slightly different column names depending on
+    the version of Ray. Returns None if the parameter key is not found.
+
+    TODO(#2176): There are different key name delimiters depending on Ray version. Simplify this as Ray is upgraded.
+    The delimiter in future versions of Ray will be '/' instead of '.'
+    """
+    if f"config.{parameter_key}" in trial_row:
```
ShreyaR, line 170: Instead of checking the format of `config` + `parameter key`, can you make a check based on the ray version? [Here's an example](https://github.com/ludwig-ai/ludwig/blob/a5871819bc430b79ec97ed12724dab157874fb78/tests/integration_tests/test_hyperopt_ray_horovod.py#L61). Also specifying which ray version we need to update in the comment would be helpful.

tgaddair (commented):
Open question in my mind is whether it's necessary to specify the `input_features` section, or if we want to just specify the feature type?

arnavgarg1:
> Open question in my mind is whether it's necessary to specify the `input_features` section, or if we want to just specify the feature type?

@tgaddair I think that's a good question. 

Ideally we should allow users to do both so they can control the level of granularity. If they want to only set defaults for a specific feature group (say `input_feature_type` vs `output_feature_type`) then they can add this, otherwise they could just do something like `defaults.text.cell_type`.

For now, doing it in the way I've implemented it technically gives users both levels of control because the users can just copy and paste the same thing in their configs but just replace `input_features` with `output_features`, which is only a little bit of additional work. If we see requests or demands to not use this distinction and allow setting defaults for inputs + outputs together in one go, it can be added in a v2.

Another reason to consider this distinction is that the parameters for input features and output features of the same type (e.g. text) can be different, so it's just neater to specify it with this distinction. We could theoretically infer whether the parameter is for the encoder or the decoder based on the parameter names if we want to, but I think it might be more error-prone in terms of usability. It could be confusing for an end user to specify these things in one place without any distinction of whether the parameter works for the encoder or the decoder or both, or only have it apply to say the output feature of that type.

Happy to discuss this further if you feel differently!

tests/integration_tests/test_hyperopt.py
arnavgarg1: Makes sense to me, will update!

tests/integration_tests/test_hyperopt.py
arnavgarg1: Good suggestion, will update!

tgaddair:
@arnavgarg1, a few things I would say as a counter-point to the above:

1. In most cases the user will not have more than one output feature (in the extreme, a handful), so I wouldn't over-optimize for the case where the user has many input and output features that need their own distinct defaults. The 99% scenario for this would be the user having a lot of input features, and then any output feature specific logic would make sense to belong to the specific output feature.
2. Once we implement the hierarchical encoder and decoder sections, it should be easy for users to separate the input and output feature specific sections.
3. This would be consistent with the existing behavior of the global `preprocessing` section (which we wish to replace with this `defaults` section).

arnavgarg1:
@tgaddair Makes sense, I'll update this to not be feature group (input or output) specific for now and make sure both input and output features use the same shared parameters.

ShreyaR (approved):

PR #2139, opened 2022-06-14
rmspe: add epsilon to avoid division by zero

### LINKED ISSUE #1281: Address issue that Ludwig often reports root_mean_squared_error_percentage being infinity
This seems to be happening when target value is 0, due to divide by 0.  One example is trial_9b499_00018
in Sarcos/Hband/Tabconcat.

Piero and I discussed this; his comments:
>Regarding RMSPE. It's a nasty issue and there's no consesus, see here:
>https://stats.stackexchange.com/questions/86708/how-to-calculate-relative-error-when-the-true-value-is-zero
>That said, I belive hasing a max(epsilon, abs(y)) as the denominator instead of y is a very reasonable choice.

Also, AFAICT we aren't multiplying the overall result by 100.0 to convert to percent FWIW.

jppgks:
Fixes #1281

### DIFF
--- a/ludwig/utils/loss_utils.py
+++ b/ludwig/utils/loss_utils.py
@@ -2,7 +2,13 @@ import torch
 
 
 def rmspe_loss(targets: torch.Tensor, predictions: torch.Tensor) -> torch.Tensor:
-    """Root mean square percentage error."""
+    """
+    Root mean square percentage error.
+    
+    Bad predictions can lead to arbitrarily large RMSPE values, 
+    especially if some values of targets are very close to zero. 
+    We return a large value instead of inf when (some) targets are zero.
+    """
     epsilon = 1e-4
     # add epsilon if targets are zero to avoid division by zero
     denominator = targets + epsilon * (targets == 0).float()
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):
Thanks for the quick fix! Tested this out locally on titanic and it works reasonably well.

ludwig/utils/loss_utils.py
```diff
@@ -3,7 +3,10 @@
 
 def rmspe_loss(targets: torch.Tensor, predictions: torch.Tensor) -> torch.Tensor:
     """Root mean square percentage error."""
```
justinxzhao, line 5: Let's update this comment to explain the expected behavior when there are targets that contain 0, perhaps adapted from [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_percentage_error.html)


> MAPE output is non-negative floating point. The best value is 0.0. But note that bad predictions can lead to arbitrarily large MAPE values, especially if some y_true values are very close to zero. Note that we return a large value instead of inf when y_true is zero.

jppgks, line 5:
```suggestion
    """
    Root mean square percentage error.
    
    Bad predictions can lead to arbitrarily large RMSPE values, 
    especially if some values of targets are very close to zero. 
    We return a large value instead of inf when (some) targets are zero.
    """
```

PR #2142, opened 2022-06-14
Try installing different torch nightly.

justinxzhao:
https://github.com/pytorch/vision/issues/6103#issuecomment-1140903691


tgaddair:
@justinxzhao looks like there may be real incompatibilities with the nightly torch version. Can you try installing at the last known sucessful nightly?

tgaddair:
Ah, this is telling:

```
torch                     1.13.0.dev20220614+rocm5.0
torchaudio                0.13.0.dev20220614+rocm5.1.1
torchtext                 0.14.0.dev20220614
torchvision               0.14.0.dev20220614+rocm5.0
```

Somehow you installed a ROCM enabled version of torch, which is not what we want (we want cpu). Will raise this with torchvision folks.

PR #2147, opened 2022-06-14
Input/Output Feature Schema Refactor

connor-mccorm:
Adds Marshmallow schema support for all input and output features and their corresponding preprocessing, encoders, and decoders.


connor-mccorm:
Most tests are passing, there is one case where I need to add a regex pattern optional argument for the string input type to handle custom fill value restrictions, but once I get that done, we should be good to go.

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+
+    norm: Optional[str] = schema_utils.StringOptions(
+        ["per_file"],
+        default=None,
+        allow_none=True,
+        description="Normalization strategy for the audio files. If None, no normalization is performed. If "
+        "per_file, z-norm is applied on a 'per file' level",
+    )
+
+    audio_feature: Optional[dict] = schema_utils.Dict(
```
martindavis, line 501: Hey @connor-mccorm , just a callout here that this is a bit of a beast on the Client. All 4 of these settings have to be set together right? If so, I'm thinking a grouped set of inputs that are all required make the most sense. However, given the way this is defined, we don't have an enum to pull from for `type` and we have to infer that `num_filter_bands` is an integer while `window_*_in_s` are floats.

I bring that all up to say: we have a similar situation with the `missing_value_strategy` of `fill_with_const` and `fill_value`. In those scenarios, we chose to select a reasonable default and have the user specify each setting separately.

I'm curious, what makes `audio_feature` different that we force the customer to specify a Dict instead of individual fields with reasonable defaults?

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+
+    audio_file_length_limit_in_s: Optional[int] = schema_utils.NonNegativeFloat(
+        default=7.5,
+        allow_none=False,
+        description="Float value that defines the maximum limit of the audio file in seconds. All files longer than "
+        "this limit are cut off. All files shorter than this limit are padded with padding_value",
+    )
+
+    missing_value_strategy: Optional[str] = schema_utils.StringOptions(
+        ["fill_with_const", "fill_with_mode", "fill_with_mean", "backfill", "bfill", "pad", "ffill", "drop_row"],
```
martindavis, line 476: Since we're including `fill_with_const` in the enum, do we need to include a `fill_value` option as well?

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+        description="The computed fill value determined by the user or inferred from the data.",
+    )
+
+
+@dataclass
+@register_preprocessor("image")
+class ImagePreprocessingConfig(schema_utils.BaseMarshmallowConfig):
+
+    missing_value_strategy: Optional[str] = schema_utils.StringOptions(
+        ["fill_with_const", "fill_with_mode", "fill_with_mean", "backfill", "bfill", "pad", "ffill", "drop_row"],
```
martindavis, line 372: Same comment as the audio field: since we're including `fill_with_const`, do we need to include a `fill_value` options as well?

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+        default="fill_with_const",
+        allow_none=False,
+        description="What strategy to follow when there's a missing value in a text column",
+    )
+
+    fill_value: Optional[str] = schema_utils.String(
+        default="",
```
martindavis, line 551: Editing comment: I'm not sure what a reasonable default is for Time Series. Glad to pair up on possible solutions!

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+    )
+
+    computed_fill_value: Optional[str] = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description="The computed fill value determined by the user or inferred from the data.",
+    )
+
+    datetime_format: Optional[str] = schema_utils.String(
+        default=None,
```
martindavis, line 656: Hey @connor-mccorm  , this caused a bit of weirdness on the Client. The solution that @hungcs and I came up with for now is to make this field not required. Since Date Time has a standard ISO format, I'm curious if we could come up with a reasonable default that both Python/Ludwig and JavaScript's `Date()` library can interpret. Should we try to pair on that?

ludwig/schema/features/text_feature.py
```diff
@@ -0,0 +1,65 @@
...
+from ludwig.schema.preprocessing import BasePreprocessingConfig, PreprocessingDataclassField
+
+
+@dataclass
+class TextInputFeatureConfig(schema_utils.BaseMarshmallowConfig):
+    """TextInputFeatureConfig is a dataclass that configures the parameters used for a text input feature."""
+
+    preprocessing: BasePreprocessingConfig = PreprocessingDataclassField(feature_type="text")
+
+    encoder: Optional[str] = schema_utils.StringOptions(
```
martindavis, line 15: Seeing a list this long made me think: is there a way to pull this list from code? I think one thing I get a little concerned about is that it will be hard to maintain this list and keep it in sync with the code.

martindavis (approved):
Hey @connor-mccorm , overall excellent job making the Schema docs very clean and clear in Ludwig! I've got some questions, but I don't see anything blocking this from my perspective.

ludwig/schema/preprocessing.py
connor-mccorm: See above:)

ludwig/schema/preprocessing.py
connor-mccorm: I think this is a tricky one to set since a timeseries column could come in many forms. I will defer to the #ludwig channel to see if people have ideas for a better default.

ludwig/schema/features/text_feature.py
ksbrar: @connor-mccorm I wonder if you can use `encoder_registry` here? (tho this is just from a quick glance)

ludwig/schema/preprocessing.py
ksbrar: Not sure about what the default here should be, but to have something better than an unvalidated string...

@connor-mccorm what we could do here is define a custom marshmallow field specifically for datetime. And we would override `_json_type_mapping` as we do in the other complex fields, and in particular set the [`pattern`](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html) attribute to this regex pattern: `^([\+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([\.,]\d+(?!:))?)?(\17[0-5]\d([\.,]\d+)?)?([zZ]|([\+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)?$` (ref: https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/). Don't necessarily have to do this in this PR though.

@martindavis both RJSF and `Date` should interpret this pattern correctly.

connor-mccorm: Ya I think that a pattern for datetime format is a good call. I will add this in.

ludwig/schema/features/text_feature.py
justinxzhao: You can use `list(encoder_registry["text"].keys())`

ludwig/schema/utils.py
```diff
@@ -87,20 +88,24 @@ def RegularizerOptions(default: Union[None, str] = None, allow_none: bool = True
     return StringOptions(["l1", "l2", "l1_l2"], default=default, allow_none=allow_none, description=description)
 
 
-def String(default: Union[None, str] = None, allow_none: bool = True, description=""):
+def String(default: Union[None, str] = None, allow_none: bool = True, pattern: str = ".*", description=""):
```
connor-mccorm, line 91: @ksbrar for datetime format validation, what do you think about just using the regex pattern validation argument in the `String()` marshmallow field that I added here instead of a new custom marshmallow field?

ksbrar: good call! yes this is better, I just double checked that the pattern is plumbed through to the JSON schema automatically so nothing else is needed

ludwig/schema/preprocessing.py
connor-mccorm: Talked with @martindavis and decided that a pattern validation for another pattern validation isn't the best idea. Additionally, since the there's not an undeniably most common datetime format, without format inference, it is best to keep the default as None and just require the field when a datetime feature is selected. A future fix for this would be implementing datetime format inference for datetime format features.

ludwig/schema/utils.py
```diff
@@ -1,6 +1,7 @@
 from dataclasses import field
 from typing import Dict as tDict
-from typing import List, Tuple, Type, Union
+from typing import List as tList
```
tgaddair, line 3: Nit: python style would be `TList` instead of `tList`.

tgaddair (commented):
Is the plan to wire this into the schema validation logic in a follow-up?

tgaddair:
Sounds good @connor-mccorm !

ludwig/features/audio_feature.py
```diff
@@ -463,6 +433,7 @@ def _get_max_length_feature(audio_feature_dict, sampling_rate_in_hz, audio_lengt
             raise ValueError(f"{feature_type} is not recognized.")
 
 
+@register_input_feature("audio")
```
justinxzhao, line 436: nit: These could be constants.

ludwig/schema/encoders.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+    use_bias: Optional[bool] = schema_utils.Boolean(
+        default=True,
+        description="Whether the layer uses a bias vector.",
+    )
+
+    weights_initializer: Optional[Union[str, dict]] = schema_utils.StringOptions(  # TODO: Add support for String/Dict
+        [
+            "constant",
```
justinxzhao, line 57: Could these options be replaced with `initializer_registry.keys()`

ludwig/schema/encoders.py
```diff
@@ -0,0 +1,134 @@
...
+        description="Normalization to use in the dense layer.",
+    )
+
+    norm_params: Optional[dict] = schema_utils.Dict(
+        default=None,
+        description="Parameters for normalization if norm is either batch or layer.",
+    )
+
+    activation: Optional[str] = schema_utils.StringOptions(
+        ["elu", "leakyRelu", "logSigmoid", "relu", "sigmoid", "tanh", "softmax"],
```
justinxzhao, line 116: Could this be replaced with `activations.keys()` (perhaps renamed to `activations_registry`) from ludwig.utils.torch_utils?

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+    """TextPreprocessingConfig is a dataclass that configures the parameters used for a text input feature."""
+
+    pretrained_model_name_or_path: Optional[str] = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description="This can be either the name of a pretrained HuggingFace model or a path where it was downloaded",
+    )
+
+    tokenizer: Optional[str] = schema_utils.String(
+        default="space_punct",
```
justinxzhao, line 51: Could we restrict the options to `tokenizer_registry.keys()`?

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+        description="What strategy to follow when there's a missing value in a text column",
+    )
+
+    fill_value: Optional[str] = schema_utils.String(
+        default="<UNK>",
+        allow_none=False,
+        description="The value to replace the missing values with in case the missing_value_strategy is fill_value",
+    )
+
+    computed_fill_value: Optional[str] = schema_utils.String(
```
justinxzhao, line 116: What is the difference between `fill_value` and `computed_fill_value`?

It looks like `computed_fill_value` might be an internal-only parameter that perhaps we want to leave out of the schema.

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+        allow_none=False,
+        description="The maximum number of most common tokens in the vocabulary. If the data contains more than this "
+        "amount, the most infrequent symbols will be treated as unknown.",
+    )
+
+    padding_symbol: Optional[str] = schema_utils.String(
+        default="<PAD>",
+        allow_none=False,
+        description="The string used as a padding symbol. This special token is mapped to the integer ID 0 in the "
+        "vocabulary.",
```
justinxzhao, line 81:
```suggestion
        description="The string used as the padding symbol for sequence features. Ignored for features using huggingface encoders, which have their own vocabulary.",
```

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+    )
+
+    unknown_symbol: Optional[str] = schema_utils.String(
+        default="<UNK>",
+        allow_none=False,
+        description="The string used as an unknown placeholder. This special token is mapped to the integer ID 1 in "
+        "the vocabulary.",
```
justinxzhao, line 88:
```suggestion
        description="The string used as the unknown symbol for sequence features. Ignored for features using huggingface encoders, which have their own vocabulary.",
```

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+        description="the direction of the padding. right and left are available options.",
+    )
+
+    lowercase: Optional[bool] = schema_utils.Boolean(
+        default=False,
+        description="If true, converts the string to lowercase before tokenizing.",
+    )
+
+    missing_value_strategy: Optional[str] = schema_utils.StringOptions(
+        ["fill_with_const", "fill_with_mode", "fill_with_mean", "backfill", "bfill", "pad", "ffill", "drop_row"],
```
justinxzhao, line 348: Consider factoring out a `UNIVERSAL_MISSING_VALUE_STRATEGIES` which has the missing value strategies that are shared across all features.

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,751 @@
...
+        description="The computed fill value determined by the user or inferred from the data.",
+    )
+
+
+@dataclass
+@register_preprocessor("bag")
+class BagPreprocessingConfig(schema_utils.BaseMarshmallowConfig):
+
+    tokenizer: Optional[str] = schema_utils.String(
+        default="space",
```
justinxzhao, line 568: Restrict this to `tokenizer_registry.keys()`

tests/ludwig/marshmallow/test_fields_preprocessing.py
```diff
@@ -0,0 +1,21 @@
...
+
+import ludwig.schema.preprocessing as lsp
+
+
+def get_marshmallow_from_dataclass_field(dfield):
+    """Helper method for checking marshmallow metadata succinctly."""
+    return dfield.metadata["marshmallow_field"]
+
+
+def test_PreprocessingDataclassField():
```
justinxzhao, line 12: Let's stick with snake case that we've been using for all of tests so far.

ludwig/schema/encoders.py
connor-mccorm: Yes good call. These files will actually be used in a followup PR for the encoder schema refactor since the current implementation is not actually using this file. I will include the registry in the next PR however.

ludwig/schema/encoders.py
connor-mccorm: Same as above, will include in the encoder schema refactor.

ludwig/schema/preprocessing.py
connor-mccorm: Same as above, will include in the encoder schema refactor.

ludwig/schema/preprocessing.py
connor-mccorm: I believe it is just an internal parameter. Specifically, I think it is the fill_value that we calculate based on the data internally. However, throughout the original schema implementation and all of the testing utils, they have `computed_fill_value` as a member of the schema so I figured I would stay consistent and keep it in. I think the best thing to do hear would be to ask #ludwig channel and see why it's in the schema, and if theres not a strong reason to keep in the schema, I can remove it throughout the entire project.

ludwig/schema/preprocessing.py
connor-mccorm: Will do!

ludwig/schema/preprocessing.py
connor-mccorm: Will do!

tests/ludwig/marshmallow/test_fields_preprocessing.py
connor-mccorm: Good catch, will make the change!

ludwig/schema/encoders.py
connor-mccorm: Going to remove this file from this PR since it is not relevant until the encoder schema PR.

ludwig/schema/encoders.py
connor-mccorm: Going to remove this file from this PR since it is not relevant until the encoder schema PR.

ludwig/schema/preprocessing.py
connor-mccorm: Done!

ludwig/schema/preprocessing.py
connor-mccorm: Done!

tests/ludwig/marshmallow/test_fields_preprocessing.py
connor-mccorm: Done!

ludwig/schema/preprocessing.py
connor-mccorm: Done!

ludwig/schema/preprocessing.py
connor-mccorm: Done

ludwig/schema/preprocessing.py
connor-mccorm: Done

ludwig/schema/preprocessing.py
connor-mccorm: Checked with Piero and confirmed it's internal only, so removing from schema.

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,699 @@
...
+    from the preprocessing of each feature.
+    """
+
+    feature_type: ClassVar[Optional[str]] = None
+    "Class variable pointing to the corresponding preprocessor."
+
+    type: str
+
+
+@dataclass
```
ksbrar, line 55: If this isn't causing bugs then its a moot point, but it looks like you are registering the preprocessing configs before they are marshmallow-fied by the dataclass decorator.

ludwig/schema/optimizers.py
```diff
@@ -325,7 +325,8 @@ def _jsonschema_type_mapping(self):
         raise ValidationError(f"Invalid default: `{default}`")
     try:
         opt = optimizer_registry[default["type"].lower()][1]
-        load_default = opt.Schema().load(default)
+        load_default = opt.Schema()
```
ksbrar, line 328: Changed for readability?

ludwig/schema/features/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+def register_output_feature(name: str):
+    def wrap(cls):
+        output_type_registry[name] = cls
+        return cls
+
+    return wrap
+
+
+def update_encoders(feature_props, feature_type):
```
ksbrar, line 26: A bit confused why these two update functions are necessary. Aren't the `encoder` props on the feature schemas declared with calls to `get_encoder_classes`/`get_decoder_classes` already?

ludwig/schema/__init__.py
```diff
@@ -20,127 +20,30 @@
...
-                        "column": {"type": "string"},
-                        "decoder": {"type": "string"},
-                    },
-                    "allOf": get_output_decoder_conds(output_feature_types)
-                    + get_output_preproc_conds(output_feature_types),
-                    "required": ["name", "type"],
-                },
-            },
-            COMBINER: get_combiner_schema(),
+            INPUT_FEATURES: get_input_feature_jsonschema(),
```
ksbrar, line 34: SO clean 🤤

ludwig/features/number_feature.py
```diff
@@ -230,21 +231,6 @@ def preprocessing_defaults():
...
-        return {
-            "missing_value_strategy": {
-                "type": "string",
-                "enum": MISSING_VALUE_STRATEGY_OPTIONS,
-            },
-            "fill_value": {"type": "number"},
-            "computed_fill_value": {"type": "number"},
-            "normalization": {
-                "type": ["string", "null"],
-                "enum": list(numeric_transformation_registry.keys()),
```
ksbrar, line 244: I believe this list gets duplicated (once here, once in the preprocessing schema), which is unfortunate. Unrelated to this PR but another indication we should do a wider registry refactor at some point.

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,699 @@
...
+def PreprocessingDataclassField(feature_type: str):
+    """Custom dataclass field that when used inside a dataclass will allow the user to specify a preprocessing
+    config.
+
+    Returns: Initialized dataclass field that converts an untyped dict with params to a preprocessing config.
+    """
+
+    class PreprocessingMarshmallowField(fields.Field):
+        """Custom marshmallow field that deserializes a dict for a valid preprocessing config from the
+        preprocessing_registry and creates a corresponding `oneOf` JSON schema for external usage."""
```
ksbrar, line 652: Nit: update this docstring to remove `oneOf` ref.

ksbrar (approved):
Awesome work!!

Some questions and nits but LGTM.

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,699 @@
...
+
+    missing_value_strategy: Optional[str] = schema_utils.StringOptions(
+        MISSING_VALUE_STRATEGY_OPTIONS,
+        default="fill_with_const",
+        allow_none=False,
+        description="What strategy to follow when there's a missing value in a text column",
+    )
+
+    fill_value: Optional[str] = schema_utils.String(
+        default="<UNK>",
```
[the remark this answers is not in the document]
justinxzhao: This is already a constant: `strings_utils.UNKNOWN_SYMBOL`

ludwig/schema/utils.py
```diff
@@ -1,6 +1,7 @@
 from dataclasses import field
 from typing import Dict as tDict
```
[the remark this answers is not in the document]
connor-mccorm: Done!

ludwig/schema/preprocessing.py
```diff
@@ -0,0 +1,699 @@
...
+    )
+
+
+@dataclass
+@register_preprocessor(CATEGORY)
+class CategoryPreprocessingConfig(schema_utils.BaseMarshmallowConfig):
+    """CategoryPreprocessingConfig is a dataclass that configures the parameters used for a category input
+    feature."""
+
+    missing_value_strategy: Optional[str] = schema_utils.StringOptions(
```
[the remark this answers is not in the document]
connor-mccorm: The rationale here is that this is a required field in the config, but it can't be `None`. If the user doesn't touch it, then it populates with the default, which is why it's typed as `Optional[str]` but if they do decide to touch it, they can't set the value to None, otherwise we'll get an "Invalid missing_value_strategy error". So the solution was to set it to `None` to avoid the default.

ludwig/schema/preprocessing.py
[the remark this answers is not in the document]
connor-mccorm: Will switch this out!

ludwig/schema/preprocessing.py
connor-mccorm: Done!

ludwig/schema/__init__.py
connor-mccorm: Iknorite?? It was really satisfying replacing with these functions🤠

ludwig/schema/features/utils.py
ksbrar: gotcha! please add docstring or comment where it's called with this note

connor-mccorm: Will do!

ludwig/schema/preprocessing.py
connor-mccorm: Now moving to `parameter_metadata` schema validation method with `internal_only` flag.

ludwig/schema/preprocessing.py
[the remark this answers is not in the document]
ksbrar: Might be misunderstanding you, but the default choice is `fill_with_const`, right? Imagining I delete `Optional` for a moment and then the functionality is: if a user types nothing, `fill_with_const` is what gets chosen; at the same time, setting `allow_none` to `False` prevents the user from typing `missing_value_strategy: null` in the config. So `missing_value_strategy` can never be `None`. That seems like what we want, right?

ludwig/schema/features/text_feature.py
```diff
@@ -0,0 +1,41 @@
...
+    preprocessing: BasePreprocessingConfig = PreprocessingDataclassField(feature_type=TEXT)
+
+    encoder: Optional[str] = schema_utils.StringOptions(
+        list(get_encoder_classes(TEXT).keys()),
+        default="parallel_cnn",
+        description="Encoder to use for this text feature.",
+    )
+
+    # TODO(#1673): Need some more logic here for validating against input features
+    tied: Optional[str] = schema_utils.String(
```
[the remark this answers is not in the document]
connor-mccorm: Will move to the schema generation logic since you are right, we don't really need to define this in the schema if it's consistent across all features.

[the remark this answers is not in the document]
connor-mccorm: Done!

connor-mccorm:
Update on this: waiting for https://github.com/ludwig-ai/ludwig/pull/2224 to merge this so that I can use the `internal_only` parameter metadata flag for `computed_fill_value` in the preprocessing schemas.

ludwig/schema/features/text_feature.py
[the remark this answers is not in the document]
connor-mccorm: So after making this change, it looks like a bunch of the tests are failing now because they want to set tied to `None` which is not of type string. Was looking for a way to do that, however it got me thinking that since we want to create this schema object for the SDK that basically uses the schema definitions for each input feature type in order to create input/output features via the SDK using dot notation, it may be better to actually have these constants defined in the schema classes themselves like `name`, `type` and `tied`. So that the user can modify all these attributes when creating these objects via the SDK object. WDYT, @justinxzhao?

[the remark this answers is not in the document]
ksbrar: @connor-mccorm missing some context, but would defining some tertiary base/common class for fields like this and deriving it work, e.g. "`CommonFeatureOptions`"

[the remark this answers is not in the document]
connor-mccorm: Ya that's probably the move actually. I had that before but ran into an inheritance issue that I resolved by just removing the class entirely which was dumb of me. Didn't have the foresight to see that we would probably need that at the time. Will add now.

justinxzhao:
@connor-mccorm FYI #2224 has now merged.

connor-mccorm:
Thank you @justinxzhao, will work on merging this today!

PR #2151, opened 2022-06-15
Accept kwargs in read_xsv

jeffreyftang:
Since `df_lib` can vary, adding support for kwargs allows passing non-universal params to the underlying library.


jeffreyftang:
Test failure looks unrelated: `ImportError: ray module is not installed. To install it, try running pip install ray`

PR #2160, opened 2022-06-17
[Torchscript] Add Vector input and output feature

geoffreyangus:
This PR adds Vector input and output features. There is a small workaround in the Vector postprocessing step to convert `preds` from `Dict[str, torch.Tensor]` to `Dict[str, Any]`. The alternative is to set the return type to `Union[Dict[str, torch.Tensor], Dict[str, Any]]`. We prefer the former over the latter because Torchscript will complain downstream if we try to do dict operations on variables with type `Union[Dict[str, torch.Tensor], Dict[str, Any]]` (i.e. if we try to do `.items()` iteration).


ludwig/features/vector_feature.py
```diff
@@ -41,10 +41,32 @@
 from ludwig.utils import output_feature_utils
 from ludwig.utils.misc_utils import set_default_value
 from ludwig.utils.torch_utils import LudwigModule
+from ludwig.utils.types import TorchscriptPreprocessingInput
 
 logger = logging.getLogger(__name__)
 
 
+class _VectorPreprocessing(torch.nn.Module):
+    def forward(self, v: TorchscriptPreprocessingInput):
```
justinxzhao, line 50: nit: add return type? `torch.Tensor`?

justinxzhao (approved):

ludwig/features/vector_feature.py
geoffreyangus: Added to this function as well as the other torchscript preprocessing modules 🙂

PR #2162, opened 2022-06-17
Cleaning hyperopt logging

arnavgarg1:
This PR aims to resolve two different logging messages seen generally during training and also during hyperopt. (See issue [here](https://github.com/ludwig-ai/ludwig/issues/2156))

- The `scope global has keys` logging messages during hyperopt show up because the verbosity for logging is set to 1 by default for `HorovodConfig` when using Ray 11.2. The easiest fix is to disable verbosity for the HorovodConfig in our 11.2 compatibility module.
- Instead of creating a Tensor from the input/target feature `ndarrays` directly, pytorch wants us to create Tensors from a copy of the numpy array during training.

### DIFF
--- a/ludwig/backend/_ray112_compat.py
+++ b/ludwig/backend/_ray112_compat.py
@@ -41,7 +41,7 @@ class HorovodConfig(BackendConfig):
     """
 
     nics: Optional[Set[str]] = None
-    verbose: int = 1
+    verbose: int = 0  # Turn off verbosity by default
     key: Optional[str] = None
     ssh_port: Optional[int] = None
     ssh_identity_file: Optional[str] = None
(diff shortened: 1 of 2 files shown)


ShreyaR (approved):
Awesome, thanks for adding these!

justinxzhao:
> * Instead of creating a Tensor from the input/target feature `ndarrays` directly, pytorch wants us to create Tensors from a copy of the numpy array during training.

Curious what the warning you are seeing from torch is. @geoffreyangus FYI

justinxzhao (approved):

justinxzhao:
Nice! I've seen this one all over the place, thanks for getting rid of it!

PR #2173, opened 2022-06-21
Update to packing version instead of LooseVersion

arnavgarg1:
Moving from `distutils` to `packaging` to remove some deprecation warnings, for example:

```
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    _ray113 = LooseVersion(ray.__version__) == LooseVersion("1.13.0")
```


justinxzhao (approved):
Thanks!

PR #2183, opened 2022-06-22
Replace ray.state.nodes() with ray.nodes().

justinxzhao:


tgaddair:
https://github.com/ludwig-ai/ludwig/runs/6998208225?check_suite_focus=true#step:10:2220

```
ValueError: Metadata mismatch found in `from_delayed`.
```

Maybe @clarkzinzow knows why this is suddenly popping up?

justinxzhao:
@tgaddair @clarkzinzow There's some flakiness with the hyperopt integration tests.

After re-running the CI, the test with the `Metadata mismatch` error went away.

However, we're still running into an error with `tests/integration_tests/test_hyperopt_ray.py::test_hyperopt_ray_mlflow`

Here's the error
```
>   from ray.util.ml_utils.mlflow import MLflowLoggerUtil as _MLflowLoggerUtil
E   ImportError: cannot import name 'MLflowLoggerUtil' from 'ray.util.ml_utils.mlflow' (/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/ray/util/ml_utils/mlflow.py)
```

justinxzhao:
The MLflowLoggerUtil issue was due to a temporary workaround that was implemented in Ludwig that is no longer needed with newer versions of Ray.

PR #2185, opened 2022-06-22
enh: Updated trainer schema

ksbrar:
Updates `TrainerConfig` to use latest metadata from parameter sprint.

Not quite finished, but looking for early feedback.

#### Current limitations
* Script is focused on `TrainerConfig` rather than generalized
* Still unclear what the correct format for certain metadata fields should be (suggested_values, type_annotation, default vs default_value)
* Unclear to me whether `BaseMetadataConfig` is much more useful over just dicts, and if it is then it might be preferable to change all of our marshmallow field methods to accept that as a typed argument rather than an open kwargs
  * Also occurs to me this could be a normal dataclass rather than a marshmallow dataclass (which in our case just gives us nice extra attributes and methods like the `Schema.load` and `Schema.dump`)

### DIFF
--- a/ludwig/schema/metadata/trainer_alt.py
+++ b/ludwig/schema/metadata/trainer_alt.py
@@ -1,0 +1,662 @@
+from ludwig.schema.metadata.common import BaseMetadataConfig
+
+batch_size_metadata = BaseMetadataConfig.Schema().load(
+    {
+        "allow_none": False,
+        "default_value": 128.0,
+        "ui_display_name": "Batch Size",
+        "type_annotation": "int",
+        "description": "The number of training examples utilized in one training step of the model. If ’auto’, we will use the biggest batch size (power of 2) that can fit in memory.",  # noqa: E501
+        "default_value_reasoning": "Not too big, not too small.",
+        "schema": 'min: 1, or "auto"',
+        "example_value": None,
+        "related_parameters": "eval_batch_size",
+        "other_information": None,
+        "description_implications": "There's conflicting advice about what batch size to use. Using a higher batch size will achieve the highest throughput and training efficiency. However, there's also evidence that depending on other hyperparameters, a smaller batch size may produce a higher quality model.",  # noqa: E501
+        "suggested_values": "auto",
+        "suggested_values_reasoning": "Try at least a few different batch sizes to get a sense of whether batch size affects model performance",  # noqa: E501
+        "commonly_used": True,
+        "expected_impact": "HIGH",
+        "literature_references": None,
+    }
+)
+checkpoints_per_epoch_metadata = BaseMetadataConfig.Schema().load(
+    {
+        "allow_none": False,
+        "default_value": False,
+        "ui_display_name": "Checkpoints per epoch",
+        "type_annotation": "int",
+        "description": "Number of checkpoints per epoch. For example, 2 -> checkpoints are written every half of an epoch. Note that it is invalid to specify both non-zero `steps_per_checkpoint` and non-zero `checkpoints_per_epoch`.",  # noqa: E501
+        "default_value_reasoning": "Per-epoch behavior, which scales according to the dataset size.",
+        "schema": "min: 0",
+        "example_value": None,
+        "related_parameters": "train_steps, steps_per_checkpoint",
+        "other_information": None,
+        "description_implications": "Epoch-based evaluation (using the default: 0) is an appropriate fit for tabular datasets, which are small, fit in memory, and train quickly.\n\nHowever, this is a poor fit for unstructured datasets, which tend to be much larger, and train more slowly due to larger models.\n\nIt's important to setup evaluation such that you do not wait several hours before getting a single evaluation result. In general, it is not necessary for models to train over the entirety of a dataset, nor evaluate over the entirety of a test set, to produce useful monitoring metrics and signals to indicate model health.\n\nIt is also more engaging and more valuable to ensure a frequent pulse of evaluation metrics, even if they are partial.\n",  # noqa: E501
+        "suggested_values": "2 - 10, for larger datasets",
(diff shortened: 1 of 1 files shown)


ludwig/schema/utils.py
```diff
@@ -170,14 +186,14 @@ def PositiveInteger(default: Union[None, int] = None, allow_none=False, descript
...
-                metadata={"description": description},
+                metadata={"description": description, **kwargs},
             )
         },
         default=default,
     )
 
 
-def NonNegativeInteger(default: Union[None, int] = None, allow_none=False, description=""):
+def NonNegativeInteger(default: Union[None, int] = None, allow_none=False, description="", **kwargs):
```
justinxzhao, line 196: Could we make some of these parameters required?

`def NonNegativeInteger(default: Union[None, int], allow_none: bool, description: str, **kwargs):`

ludwig/schema/utils.py
```diff
@@ -65,40 +67,54 @@ def unload_jsonschema_from_marshmallow_class(mclass) -> tDict:
     return schema
 
 
-def InitializerOptions(default: str = "xavier_uniform", description=""):
+# Fields with open signatures (accept any kwargs):
+
+
+def InitializerOptions(default: str = "xavier_uniform", description="", **kwargs):
```
justinxzhao, line 73: Instead of kwargs, could we use `BaseMetadataConfig` directly?

ludwig/schema/trainer.py
```diff
@@ -16,175 +17,105 @@
...
+        default=LOSS, **lsmt.increase_batch_size_eval_metric_metadata
+    )
 
-    increase_batch_size_eval_split: str = schema_utils.String(default=TRAINING, description="")
+    increase_batch_size_eval_split: str = schema_utils.String(
+        default=TRAINING, **lsmt.increase_batch_size_eval_split_metadata
+    )
 
-    decay: bool = schema_utils.Boolean(default=False, description="Turn on exponential decay of the learning rate.")
+    decay: bool = schema_utils.Boolean(default=False, **lsmt.decay_metadata)
```
justinxzhao, line 98: I think it's important that we keep the descriptions and defaults visible and readable in this file. Putting everything else in the "metadata" block SGTM.

ludwig/schema/optimizers.py
```diff
@@ -354,13 +358,16 @@ class GradientClippingConfig(BaseMarshmallowConfig):
...
-def GradientClippingDataclassField(default={}, allow_none=True, description="TODO"):
+def GradientClippingDataclassField(default={}, allow_none=True, **kwargs):
     """Returns custom dataclass field for `ludwig.modules.optimization_modules.GradientClippingConfig`. Allows
     `None` by default.
 
     :param default: dict that specifies clipping param values that will be loaded by its schema class (default: {}).
     :param allow_none: Whether this field can accept `None` as a value. (default: True)
     """
+    # Prevent accidental override:
+    kwargs["allow_none"] = allow_none
```
justinxzhao, line 369: Seems odd. I can ask you more about this when we meet later today?

ludwig/schema/metadata/trainer.py
```diff
@@ -0,0 +1,594 @@
...
+    "default_value_reasoning": "Not too big, not too small.",
+    "schema": 'min: 1, or "auto"',
+    "example_value": None,
+    "related_parameters": "eval_batch_size",
+    "other_information": None,
+    "description_implications": "There's conflicting advice about what batch size to use. Using a higher batch size will achieve the highest throughput and training efficiency. However, there's also evidence that depending on other hyperparameters, a smaller batch size may produce a higher quality model.",  # noqa: E501
+    "suggested_values": "auto",
+    "suggested_values_reasoning": "Try at least a few different batch sizes to get a sense of whether batch size affects model performance",  # noqa: E501
+    "commonly_used": True,
+    "expected_impact": "HIGH",
```
justinxzhao, line 16: `ExpectedImpact.HIGH` right?

ludwig/schema/metadata/trainer.py
```diff
@@ -0,0 +1,594 @@
+batch_size_metadata = {
+    "allow_none": False,
+    "default_value": 128.0,
+    "ui_display_name": "Batch Size",
+    "type_annotation": "int",
+    "description": "The number of training examples utilized in one training step of the model. If ’auto’, we will use the biggest batch size (power of 2) that can fit in memory.",  # noqa: E501
+    "default_value_reasoning": "Not too big, not too small.",
+    "schema": 'min: 1, or "auto"',
+    "example_value": None,
+    "related_parameters": "eval_batch_size",
```
justinxzhao, line 10: `related_parameters` should be a list.

ludwig/schema/metadata/trainer.py
```diff
@@ -0,0 +1,594 @@
+batch_size_metadata = {
+    "allow_none": False,
+    "default_value": 128.0,
+    "ui_display_name": "Batch Size",
+    "type_annotation": "int",
+    "description": "The number of training examples utilized in one training step of the model. If ’auto’, we will use the biggest batch size (power of 2) that can fit in memory.",  # noqa: E501
+    "default_value_reasoning": "Not too big, not too small.",
+    "schema": 'min: 1, or "auto"',
```
justinxzhao, line 8: To avoid duplication and unintentional override conflicts, it would probably make sense to exclude the fields that are used to select the schema in the first place (`type_annotation`, `schema`), and the fields that are used to initialize the schema (`allow_none`, `default_value`, `description`, `schema`). WDYT?

ludwig/schema/utils.py
ksbrar: yup, this is what I was thinking

ludwig/schema/trainer.py
ksbrar: descriptions in particular though can get pretty long, which is my only concern

ksbrar:
Closed in favor of #2224, #2218,

PR #2187, opened 2022-06-23
fix: Set divisions for proc_cols directly from original dataset

jeffreyftang:
Depending on how the dataframe is originally read, it may or may not have known divisions. In the former case, attempting to perform the "manual join" in `DaskEngine` with certain feature types (image, audio) will fail due to `proc_cols` not having known divisions to match. 

This PR simply sets the divisions of each `proc_col` directly from the originating dataset.


geoffreyangus (approved):

tgaddair (changes requested):
Please add a test that repros the issue without this fix.

tgaddair (approved):

PR #2189, opened 2022-06-23
Fix postprocessing on binary feature columns with number dtype

geoffreyangus:
This PR ensures that the postprocessing behavior before #2058 is preserved when dealing with binary feature columns of dtype float/int. This is done primarily by reverting the changes introduced by #2058 in `ludwig/features/binary_feature.py` and `ludwig/visualize.py`. You can see the full changes in `binary_feature.py` and `visualize.py` relative to 0.5.2 [here](https://github.com/ludwig-ai/ludwig/compare/dd026ca..f7dbb45#diff-1bcb267f1a0a83aea88c3f15b64f346e2c42ea81e99400ea1504744e662ddb5f) and [here](https://github.com/ludwig-ai/ludwig/compare/dd026ca..f7dbb45#diff-a6e3d9afdd95decfc7a71b4140a9d7eb6f38dc3de3c0eb8491bf581dfa6ca234), respectively.

That said, in order to achieve this, we add several checks upstream in `cast_column` to ensure that the old behavior when determining dtypes is preserved.  The one caveat is that this requires calling `compute()`, which could be expensive in large datasets.

### DIFF
--- a/ludwig/features/binary_feature.py
+++ b/ludwig/features/binary_feature.py
@@ -40,7 +40,7 @@ from ludwig.constants import (
     TYPE,
 )
 from ludwig.features.base_feature import BaseFeatureMixin, InputFeature, OutputFeature, PredictModule
-from ludwig.utils import output_feature_utils, strings_utils
+from ludwig.utils import calibration, output_feature_utils, strings_utils
 from ludwig.utils.eval_utils import (
     average_precision_score,
     ConfusionMatrix,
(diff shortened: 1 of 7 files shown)


justinxzhao (approved):
Code LGTM. Not totally sure why this change would causes tests to be flaky.

tgaddair (approved):

brightsparc (approved):

PR #2192, opened 2022-06-24
Add attention scores of the vit encoder as an additional return value

Dennis-Rall:
This adds the vit's attention scores as return values.


justinxzhao (commented):
@Dennis-Rall thanks for putting this change together.

Are there any speed implications with outputting attentions?

If so (or if you aren't sure), I suggest that we add this as a parameter in the constructor of the encoder such that this parameter would be configurable.

Please let me know if what I'm requesting is not clear.

Dennis-Rall:
@justinxzhao I don't know how this affects speed, but I think it's definitely a good idea to add this as a configurable parameter.

Dennis-Rall:
Are the failing hyperopt tests a result of my changes?

justinxzhao:
> Are the failing hyperopt tests a result of my changes?

Nope! Tests look good, thanks for the PR -- merging.

PR #2193, opened 2022-06-25
Add boolean postprocessing to dataset type inference for automl

magdyksaleh:
# Code Pull Requests

When a csv file is read in via pandas, if it contains a column of boolean values with some missing values, it is misinterpreted as an object dtype instead of a bool dtype. This corrects that.

### DIFF
--- a/ludwig/automl/base_config.py
+++ b/ludwig/automl/base_config.py
@@ -201,6 +201,11 @@ def get_dataset_info_from_source(source: DataSource) -> DatasetInfo:
         image_values = source.get_image_values(field)
         audio_values = source.get_audio_values(field)
         avg_words = None
+        if dtype == "object":
+            # Check if it is a nullboolean field
+            is_boolean = source.check_if_boolean(field)
+            if is_boolean:
+                dtype = "bool"
         if source.is_string_type(dtype):
             avg_words = source.get_avg_num_tokens(field)
         fields.append(
--- a/ludwig/automl/base_config.py
+++ b/ludwig/automl/base_config.py
@@ -202,7 +202,9 @@ def get_dataset_info_from_source(source: DataSource) -> DatasetInfo:
         audio_values = source.get_audio_values(field)
         avg_words = None
         if dtype == "object":
-            # Check if it is a nullboolean field
+            # Check if it is a nullboolean field. We do this since if you read a csv with 
+            # pandas that has a column of booleans and some missing values, the column is 
+            # interpreted as object dtype instead of bool
             is_boolean = source.check_if_boolean(field)
             if is_boolean:
                 dtype = "bool"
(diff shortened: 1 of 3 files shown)


ludwig/automl/utils.py
```diff
@@ -78,6 +78,23 @@ def avg_num_tokens(field: Series) -> int:
     return avg_words
 
 
+def check_if_boolean(field: Series) -> bool:
+    if len(field) > 5000:
+        field = field.sample(n=5000, random_state=40)
```
tgaddair, line 83: I believe this function could be given a Dask Series, which does not support `sample(n)`. Dask only supports sampling a fraction of the data. Typically we would just use `field.head(5000)` here, and avoid calling `len`, which is also expensive in Dask.

More general question: do we need to compute this directly from the series when we already have a function called `get_distinct_values` that returns the top k distinct values? We should be able to derive this info from the distinct values instead of recomputing it.

tgaddair (changes requested):
We also need a test since presumably this is fixing a bug.

ludwig/automl/utils.py
magdyksaleh: Makes sense. Changed it accordingly

ludwig/automl/data_source.py
```diff
@@ -69,6 +75,20 @@ def get_audio_values(self, column: str, sample_size: int = 10) -> int:
...
+            for entry in unique_values:
+                try:
+                    if np.isnan(entry):
+                        continue
+                except TypeError:
+                    return False
+                if isinstance(entry, bool):
+                    continue
+                return False
+        return True
```
justinxzhao, line 90:
```suggestion
        if num_unique_values > 3:
          return False
        for entry in unique_values:
            try:
                if np.isnan(entry):
                    continue
            except TypeError:
                # When does this happen?
                return False
            if isinstance(entry, bool):
                continue
            # Encountered non-boolean type.
            return False
        return True
```

ludwig/automl/base_config.py
```diff
@@ -201,6 +201,13 @@ def get_dataset_info_from_source(source: DataSource) -> DatasetInfo:
         image_values = source.get_image_values(field)
         audio_values = source.get_audio_values(field)
         avg_words = None
+        if dtype == "object":
+            # Check if it is a nullboolean field. We do this since if you read a csv with
+            # pandas that has a column of booleans and some missing values, the column is
+            # interpreted as object dtype instead of bool
+            is_boolean = source.check_if_boolean(field)
+            if is_boolean:
```
justinxzhao, line 209: nit: Consolidate into 1 line.

justinxzhao (commented):
Agreed with @tgaddair that it would be great to add a test for this new check, perhaps in a new file like `test_data_source.py`.

magdyksaleh:
Added a test and made some changes so that it is just a function in ludwig that checks if the field is boolean. That way we don't need to propagate the change through the platform as well and instead leverage the Datasource API

tgaddair (approved):

PR #2195, opened 2022-06-27
adding hardware usage and software packages tracker

abidwael:
This provides a new `Tracker` class to track hardware usage and software packages while a block of code is being executed.

Usage:
```
with Tracker(tag='train', output_dir=model.config['backend']['cache_dir'], num_batches=model.config[TRAINER]["batch_size"], num_examples=len(training_set)) as tracker:
    # code block
    .
    .
```
Will save hardware and software usage metrics under `f"{output_dir}/{tag}_metrics.json"`


ludwig/utils/tracker.py
```diff
@@ -0,0 +1,169 @@
...
+        gpu_infos = get_gpu_info()
+        for i, gpu_info in enumerate(gpu_infos):
+            gpu_key = f"gpu_{i}"
+            self.info["system"][gpu_key] = {}
+            self.info["system"][gpu_key]["name"] = gpu_info["name"]
+            self.info["system"][gpu_key]["total_memory"] = gpu_info["total_memory"]
+            self.info["system"][gpu_key]["driver_version"] = gpu_info["driver_version"]
+            self.info["system"][gpu_key]["cuda_version"] = gpu_info["cuda_version"]
+
+        torch.cuda.synchronize()
```
ShreyaR, line 113: Do we need this here?

ludwig/utils/tracker.py
```diff
@@ -0,0 +1,169 @@
...
+        self.p.join()
+
+        self.info = load_json(os.path.join(self.output_dir, self.info["tag"] + "_temp.json"))
+        os.remove(os.path.join(self.output_dir, self.info["tag"] + "_temp.json"))
+
+        self.info["end_time"] = time.time()
+        self.info[f"{self.tag}_total_duration"] = self.info["end_time"] - self.info["start_time"]
+
+        # if self.num_batches:
```
ShreyaR, line 154: Did you mean to leave this in?

ludwig/utils/tracker.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # GPU information
+        gpu_infos = get_gpu_info()
+        for i, gpu_info in enumerate(gpu_infos):
+            gpu_key = f"gpu_{i}"
+            self.info["system"][gpu_key] = {}
+            self.info["system"][gpu_key]["name"] = gpu_info["name"]
+            self.info["system"][gpu_key]["total_memory"] = gpu_info["total_memory"]
+            self.info["system"][gpu_key]["driver_version"] = gpu_info["driver_version"]
+            self.info["system"][gpu_key]["cuda_version"] = gpu_info["cuda_version"]
```
ShreyaR, line 111: Is it possible to get GPU utilization here? That's a key metric I look for while benchmarking.

ludwig/utils/misc_utils.py
```diff
@@ -140,3 +144,39 @@ def set_saved_weights_in_checkpoint_flag(config):
     """
     for input_feature in config.get("input_features", []):
         input_feature["saved_weights_in_checkpoint"] = True
+
+
+def processify(func):
```
ShreyaR, line 149: My concern with adding a functionality like this is that it will likely only be used for `monitor`, as in the rest of the codebase we use multiprocessing directly.

Is it possible to refactor `Tracker` so that `monitor` is launched in a separate process without needing processify?

ShreyaR (commented):
Thanks for putting up this PR! Left some comments.

One general question I have is what the overhead of using `Tracker` is. If it isn't too expensive or slow to run `Tracker`, then it may make sense to use it by default for any Ludwig training process. I can see it being quite useful if we have a json with all the benchmarking stats generated in the model artifacts folder anytime we do training/evaluation.

requirements_tracker.txt
```diff
@@ -0,0 +1,3 @@
+experiment_impact_tracker
+gpustat
```
abidwael, line 2: Should this be a separate `requirements_tracker.txt` file or do I need to add it to the main `requirements.txt` file?

ludwig/utils/tracker.py
abidwael: added out of safety but I don't think it's necessary

ludwig/utils/tracker.py
abidwael: added in `monitor()`
`info["system"][gpu_key]["memory_used"].append(gpu_info.memory_used)`

abidwael:
> Thanks for putting up this PR! Left some comments.
> 
> One general question I have is what the overhead of using `Tracker` is. If it isn't too expensive or slow to run `Tracker`, then it may make sense to use it by default for any Ludwig training process. I can see it being quite useful if we have a json with all the benchmarking stats generated in the model artifacts folder anytime we do training/evaluation.

@ShreyaR I ran `model.experiment` with and without `Tracker` for `ames_housing` and `mercedes_benz_greener` and collected the total cpu and ram usage for all processes running on the machine at the time of execution. Here are the results:

Screen Shot 2022-07-12 at 3 32 20 AM

CPU seems to be more or less unaffected, but there's some RAM overhead. In my opinion, it's worth adding an optional `Tracker` ctx per @justinxzhao 's suggestion.

requirements_tracker.txt
justinxzhao: I'd be ok with adding this to the main requirements.txt file, especially if hardware resource usage tracking adds marginal overhead.

Curious about other people's opinions on this: @dantreiman @w4nderlust @tgaddair

ludwig/benchmarking/tracker.py
```diff
@@ -0,0 +1,188 @@
...
+
+class Tracker:
+    """Track system resource (hardware and software) usage by a chunk of code.
+
+    Attributes:
+        tag: a string tag about the process that we're tracking. Examples: train, evaluate, preprocess, etc.
+        output_dir: path where metrics are saved.
+        logging_interval: time interval in seconds at which system is polled for resource usage.
+        num_batches: number of batches of training or evaluation process.
+        num_batches: number of examples of training or evaluation process.
```
justinxzhao, line 81: nit: `num_examples`?

ludwig/benchmarking/tracker.py
```diff
@@ -0,0 +1,188 @@
...
+            for i, gpu_info in enumerate(gpu_infos):
+                gpu_key = f"gpu_{i}"
+                info["system"][gpu_key]["memory_used"].append(gpu_info.memory_used)
+        info["system"]["cpu_utilization"].append(psutil.cpu_percent())
+        info["system"]["ram_utilization"].append(psutil.virtual_memory().percent)
+        time.sleep(logging_interval)
+
+
+class Tracker:
+    """Track system resource (hardware and software) usage by a chunk of code.
```
justinxzhao, line 74: nit: remove "by a chunk of code".

ludwig/benchmarking/tracker.py
```diff
@@ -0,0 +1,188 @@
...
+            gpu_infos = GPUStatCollection.new_query()
+            for i, gpu_info in enumerate(gpu_infos):
+                gpu_key = f"gpu_{i}"
+                info["system"][gpu_key]["memory_used"].append(gpu_info.memory_used)
+        info["system"]["cpu_utilization"].append(psutil.cpu_percent())
+        info["system"]["ram_utilization"].append(psutil.virtual_memory().percent)
+        time.sleep(logging_interval)
+
+
+class Tracker:
```
justinxzhao, line 73: Maybe rename this `ResourceUsageTracker` to make it more clear that this is only tracking resource usage, not evaluation metrics, model health, or other things.

ludwig/benchmarking/tracker.py
```diff
@@ -0,0 +1,185 @@
...
+            gpu_infos = GPUStatCollection.new_query()
+            for i, gpu_info in enumerate(gpu_infos):
+                gpu_key = f"gpu_{i}"
+                info["system"][gpu_key]["memory_used"].append(gpu_info.memory_used)
+        info["system"]["cpu_utilization"].append(psutil.cpu_percent())
+        info["system"]["ram_utilization"].append(psutil.virtual_memory().percent)
+        time.sleep(logging_interval)
+
+
+class ResourceUsageTracker:
```
justinxzhao, line 73: Discussed offline: Add a basic unit test that shows how this class can/should be used.

abidwael:
The pre-commit.ci check will not pass because of the following block
```
# disabling print because the following imports are verbose
f = open(os.devnull, "w")
sys.stdout = f
from experiment_impact_tracker.cpu.common import get_my_cpu_info
from experiment_impact_tracker.gpu.nvidia import get_gpu_info
from experiment_impact_tracker.py_environment.common import get_python_packages_and_versions

f.close()
sys.stdout = sys.__stdout__
```
I'm temporarily redirecting stdout because the import statement is verbose. 

Made a PR in the original repo: https://github.com/Breakend/experiment-impact-tracker/pull/74
Will follow up with the maintainer.

justinxzhao (commented):
Changes LGTM, looks like there's a few last errors to resolve:

From pre-commit:
```
ludwig/benchmarking/resource_usage_tracker.py:24: [E402] module level import not at top of file
ludwig/benchmarking/resource_usage_tracker.py:25: [E402] module level import not at top of file
ludwig/benchmarking/resource_usage_tracker.py:26: [E402] module level import not at top of file
```

Finally, could you check the unit test you added? It looks like it's failing on one of the builds.

abidwael:
> Changes LGTM, looks like there's a few last errors to resolve:
> 
> From pre-commit:
> 
> ```
> ludwig/benchmarking/resource_usage_tracker.py:24: [E402] module level import not at top of file
> ludwig/benchmarking/resource_usage_tracker.py:25: [E402] module level import not at top of file
> ludwig/benchmarking/resource_usage_tracker.py:26: [E402] module level import not at top of file
> ```
> 
> Finally, could you check the unit test you added? It looks like it's failing on one of the builds.

This one is due to my previous comment [here](https://github.com/ludwig-ai/ludwig/pull/2195#issuecomment-1185028923).

justinxzhao (approved):
LGTM after the pre-commit error is fixed, and the tests are all green.

PR #2210, opened 2022-06-29
Add explicit handling for zero-length image byte buffers to avoid cryptic errors

jeffreyftang:
In practice these cryptic errors may look like `Existing exports of data: object cannot be re-sized`.


geoffreyangus (approved):
Thank you!

Issue #2199, opened 2022-06-28
model.predict() fails on goemotions v0.5.3


abidwael:
**Edit**: commit that led to the bug: https://github.com/ludwig-ai/ludwig/commit/fe77793288869a651f3b845aa7626fb5a0552ba7

**Describe the bug**
model.predict() fails on goemotions. Generates the following error:

**To Reproduce**

Code to execute
```
import logging
from ludwig.api import LudwigModel
from ludwig.datasets import goemotions
from ludwig.utils.data_utils import load_yaml

training_set, test_set, val_set = goemotions.load(split=True)
model_config = load_yaml("goemotions.yaml") # copied below
model_config['backend'] = {}
model_config['backend']['type'] = 'local'
model = LudwigModel(config=model_config, logging_level=logging.INFO)
train_stats, preprocessed_data, output_directory = model.train(dataset=training_set, experiment_name="simple_experiment", model_name="simple_model")
model.predict(test_set, skip_save_predictions=False)
```
`goemotions.yaml`
```
output_features:
  - name: emotion_ids
    type: set
input_features:
  - name: text
    type: text
    encoder: bert
preprocessing:
  text:
    max_sequence_length: 40
trainer:
  batch_size: 64
  learning_rate: 0.00002
  epochs: 5
  learning_rate_warmup_epochs: 0
  optimizer:
    type: adamw
  validation_field: emotion_ids
  validation_metric: jaccard
```
Error message
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.7/site-packages/ludwig/api.py", line 810, in predict
    postproc_predictions, self.model.output_features, output_directory, self.backend
  File "/opt/conda/lib/python3.7/site-packages/ludwig/models/predictor.py", line 320, in save_prediction_outputs
    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME))
  File "/opt/conda/lib/python3.7/site-packages/pandas/util/_decorators.py", line 207, in wrapper
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/pandas/core/frame.py", line 2685, in to_parquet
    **kwargs,
  File "/opt/conda/lib/python3.7/site-packages/pandas/io/parquet.py", line 423, in to_parquet
    **kwargs,
[...]

PR #2211, opened 2022-06-29
Fixes prediction saving for models with Set output

geoffreyangus:
This PR fixes prediction saving for Set features (#2199). 

Prior to this change, the list of probabilities crossing `threshold` was converted to `np.array` with no dtype specified. If any of the probabilities crossed the threshold, the `np.array` generated would be the dtype of the probability, typically the dtype `float32` commonly used by PyTorch. However, if NO probabilities crossed the threshold, the `np.array` generated would be dtype `float64`, which is the default for `np.array`. This PR ensures that the dtype of the generated numpy array is `float32`.

### DIFF
--- a/tests/integration_tests/test_postprocessing.py
+++ b/tests/integration_tests/test_postprocessing.py
@@ -31,6 +31,7 @@ from tests.integration_tests.utils import (
     init_backend,
     RAY_BACKEND_CONFIG,
     set_feature,
+    text_feature,
 )
 
 
@@ -163,7 +164,7 @@ def test_binary_predictions_with_number_dtype(tmpdir, backend, distinct_values):
 def test_set_feature_saving(tmpdir, pct_positive):
     backend = "local"
     input_features = [
-        category_feature(vocab_size=3),
+        text_feature(vocab_size=3),
     ]
 
     feature = set_feature()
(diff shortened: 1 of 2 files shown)


tests/integration_tests/test_postprocessing.py
```diff
@@ -131,7 +159,48 @@ def test_binary_predictions_with_number_dtype(tmpdir, backend, distinct_values):
         assert np.allclose(prob_0, 1 - prob_1)
 
 
-def predict_with_backend(tmpdir, config, data_csv_path, backend, num_predict_samples=None):
+@pytest.mark.parametrize("pct_positive", [1.0, 0.5, 0.0])
+def test_set_feature_saving(tmpdir, pct_positive):
+    backend = "local"
+    input_features = [
+        category_feature(vocab_size=3),
```
abidwael, line 166: Since `goemotions` is a text datasets, would it be possible to use `text_feature` in [tests.integration_tests.utils](https://github.com/ludwig-ai/ludwig/blob/3785ed5b53c83ea5e3e4c2df33ffd5677758896c/tests/integration_tests/utils.py#L234)?

abidwael (commented):
Thanks for putting this together Geoffrey! Left a comment about using `text_feature`, but otherwise should be good.

tests/integration_tests/test_postprocessing.py
geoffreyangus: Done!

PR #2213, opened 2022-06-29
[WIP] Adds `inference_utils.py`

geoffreyangus:


brightsparc:
So if I look at `inference_utils.py` entry point, it looks like at a minimum I would need the following files (does this sound right @geoffreyangus ?)

```
ludwig
│   ├── constants.py
│   ├── utils
│   │   ├── audio_utils.py
│   │   ├── date_utils.py
│   │   ├── fs_utils.py (imported from audio_utils.py)
│   │   ├── image_utils.py
│   │   ├── inference_utils.py (entry point)
│   │   ├── string_utils.py (imported from torch_utils.py)
│   │   ├── torch_utils.py
│   │   ├── misc_utils.py
│   │   ├── nlp_utils.py (imported from tokenizers.py)
│   │   ├── types.py
│   │   ├── tokenizers.py (imported from torch_utils.py)
```

Which has the following list of deps:

```
import torch
import pandas
import torchaudio # not yet supported in snowflake
import torchvision
import fsspec
import h5py
import requests # not supported in snowflake
from filelock import FileLock
```

The latter of these are imported by `fs_utils.py` used by `audio_utils.py`  for which  `get_bytes_obj_from_path` has supported for  reading from http.  

Given the conda distribution doesn't have `requests` package, consider replacing with `urllib3`, eg the following function:

```
def get_bytes_obj_from_http_path(path: str) -> bytes:
    data = requests.get(path, stream=True)
    if data.status_code == 404:
        upgraded = upgrade_http(path)
        if upgraded:
            logging.info(f"reading url {path} failed. upgrading to https and retrying")
            return get_bytes_obj_from_http_path(upgraded)
        else:
            raise requests.exceptions.HTTPError(f"reading url {path} failed and cannot be upgraded to https")
    return data.raw.read()
```

Could have internals replaced with something like :

```
import urllib3
http = urllib3.PoolManager()
r = http.request('GET', 'http://httpbin.org/bytes/8')
r.data
```

Or if you wanted to stream to file like this:

```
import urllib3
http = urllib3.PoolManager()
r = http.request('GET', url, preload_content=False)

with open(path, 'wb') as out:
    while True:
        data = r.read(chunk_size)
        if not data:
            break
        out.write(data)

r.release_conn()
```

brightsparc (approved):
LGTM, the latest python libraries in snowflake are 
* `certifi==2022.6.15` and 
* `urllib3==1.26.9`

PR #2218, opened 2022-06-30
Define a dataclass for parameter metadata.

justinxzhao:


ludwig/schema/parameter_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+    # Other information that is relevant for this parameter.
+    other_information: Union[str, None] = None
+
+    # If we change/increase/decrease this parameter, what's the intuition for how model performance would change, i.e.:
+    # learning curves, model speed, memory usage, etc.
+    description_implications: Union[str, None] = None
+
+    # What values would a machine learning expert suggest users try to help improve their model? Ideally, covers 95%
+    # (~2 sigma) of use cases.
+    suggested_values: Any = None
```
ksbrar, line 45: Should this be a list?

ludwig/schema/parameter_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+
+class ExpectedImpact(Enum):
+    """The expected impact of determining a "good" value for a specific parameter.
+
+    - HIGH: this parameter should almost always be included in a hyperopt run and can make or break a good model.
+    - MEDIUM: this parameter can sometimes make or break a good model.
+    - LOW: this parameter usually does not have a significant impact on model performance.
+    """
+
+    UNKNOWN = 0
```
ksbrar, line 14: What is the rationale for having an `UNKNOWN` flag? I would assume default as `LOW`

ludwig/schema/parameter_metadata.py
justinxzhao: I'm inclined to leave this as Any for now, since it could be a list, set, range, or plain string (a verbal description)

ludwig/schema/parameter_metadata.py
justinxzhao: `UNKNOWN` allows us to distinguish parameters that haven't been annotated vs. parameters that have been explicitly marked as having `LOW` impact.

PR #2224, opened 2022-06-30
Add parameter metadata to the trainer schema.

justinxzhao:


connor-mccorm (approved):

ludwig/schema/optimizers.py
```diff
@@ -354,7 +355,12 @@ class GradientClippingConfig(BaseMarshmallowConfig):
     clipvalue: Optional[float] = FloatRange(default=None, allow_none=True, description="")
 
 
-def GradientClippingDataclassField(default={}, allow_none=True, description="TODO"):
+def GradientClippingDataclassField(
+    default={},
+    allow_none=True,
+    description="Parameters for gradient clipping.",
+    parameter_metadata: ParameterMetadata = None,
```
ksbrar, line 362: I would not let `parameter_metadata` be set here, and just have it set to `TRAINER_METADATA['gradient_clipping']`. Actually I'd remove all the parameters here, being only used in the trainer it's something that should not have open arguments.

ludwig/schema/trainer.py
```diff
@@ -229,7 +310,10 @@ class GBMTrainerConfig(BaseTrainerConfig):
     eval_batch_size: int = schema_utils.PositiveInteger(
```
ksbrar, line 310: Metadata omitted on purpose? Is the `eval_batch_size` metadata necessarily different for GBM?

ludwig/schema/utils.py
```diff
@@ -687,13 +738,20 @@ def _jsonschema_type_mapping(self):
                 else []
             )
 
-            return {"oneOf": oneof_list, "title": self.name, "description": description, "default": default}
+            return {
```
ksbrar, line 741: Not part of the PR, but would you mind adding this line above after `numeric_option` is declared?

```
if not is_integer: numeric_option["format"] = "float"
```

ksbrar (changes requested):

ludwig/schema/optimizers.py
justinxzhao: Done, though I left `description` and `default` since I think it's still visually useful for those fields to appear in the Trainer schema.

ludwig/schema/trainer.py
justinxzhao: I don't think so, though this probably means we can just use the `BaseTrainerConfig`'s `eval_batch_size` definition.

ludwig/schema/utils.py
justinxzhao: Done.

ludwig/schema/metadata/trainer_metadata.py
```diff
@@ -0,0 +1,685 @@
...
+# fmt: off
+from ludwig.schema.metadata.parameter_metadata import ParameterMetadata
+
+TRAINER_METADATA = (
+{'batch_size': ParameterMetadata(ui_display_name='Batch Size',
+                                 default_value_reasoning='Not too big, not too small.',
+                                 example_value=None,
+                                 related_parameters='eval_batch_size',
+                                 other_information=None,
+                                 description_implications="There's conflicting advice about what batch size to use. "
```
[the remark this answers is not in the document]
justinxzhao: Yes! pretty printing for dataclasses is supported in python 3.10.

ludwig/schema/metadata/trainer_metadata.py
```diff
@@ -0,0 +1,685 @@
...
+                                                          'Using a higher batch size will achieve the highest '
+                                                          "throughput and training efficiency. However, there's also "
+                                                          'evidence that depending on other hyperparameters, a smaller '
+                                                          'batch size may produce a higher quality model.',
+                                 suggested_values='auto',
+                                 suggested_values_reasoning='Try at least a few different batch sizes to get a sense '
+                                                            'of whether batch size affects model performance',
+                                 commonly_used=True,
+                                 expected_impact='HIGH',
```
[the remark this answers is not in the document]
justinxzhao: Done.

ludwig/schema/optimizers.py
```diff
@@ -354,13 +355,14 @@ class GradientClippingConfig(BaseMarshmallowConfig):
     clipvalue: Optional[float] = FloatRange(default=None, allow_none=True, description="")
 
 
-def GradientClippingDataclassField(default={}, allow_none=True, description="TODO"):
+def GradientClippingDataclassField(description: str, default: Dict):
```
ksbrar, line 358: I would still keep `{}` as the default value for `default` so it's clear that in the case of `GradientClippingDataclassField` an empty dict is valid (whereas e.g. for `OptimizerDataclassField` you have to at least provide a `type` inside the dict)

ludwig/schema/trainer.py
```diff
@@ -226,12 +307,6 @@ class GBMTrainerConfig(BaseTrainerConfig):
         allow_none=False,
     )
 
-    eval_batch_size: int = schema_utils.PositiveInteger(
```
ksbrar, line 229: Just want to double check, `eval_batch_size` here had the `default` set to `128` whereas the `default` on `BaseTrainerConfig` is `None`

ksbrar (changes requested):

ludwig/schema/optimizers.py
justinxzhao: Added back the default value for `default` for now.

Outside of this PR, we should consider removing the default values for marshmallow field function arguments. This will make some of the marshmallow unit tests more verbose, but it ensures that the field's parameters (default, description, etc.) are defined in one location, at the point of instantiation. For example, right now for `OptimizerDataclassField`, we declare the `default` in two places -- as a function parameter default via the marshmallow field definition, as well as at the point of instantiation in `ECDTrainerConfig`.

Here's what I'm thinking:
- Always include `description` and `default` as required arguments so that it's impossible to instantiate a marshmallow field without explicitly specifying the `description` and `default`. This setup ensures that the defaults and descriptions are concisely listed where parameters are declared, i.e. `ECDTrainerConfig`.
- Include other arguments like `allow_none` in the marshmallow field only if we want this to be able to be configured differently across different instances of the field. For example, we should remove `allow_none` from the list of parameters for `GradientClippingDataclassField` since it's always `True`.
- Use `ParameterMetadata` to contain all other "optional" information pertaining to the field. Eventually, this will also be a required argument.

See also, [why default arguments are bad](https://quuxplusone.github.io/blog/2020/04/18/default-function-arguments-are-the-devil/).

ksbrar: Agreed on all counts!

ksbrar (approved):

PR #2226, opened 2022-06-30
Add an explicit call to merge_wtih_defaults() when loading a config from a model directory.

justinxzhao:


jppgks (approved):

tgaddair (changes requested):
Please add a test that loads an old Ludwig model and runs `predict()`. This can be stored in s3 at a public URL.

tgaddair (approved):

PR #2235, opened 2022-07-06
Update R2 score to handle single sample computation

arnavgarg1:
One of the users in Ludwig's OSS reported an issue with R2 computation in v0.5 where R2 scores can't be computed if the sample size during prediction is just 1 row. The error `torchmetrics` returns is logically correct because R2 scores can't be calculated for just one sample.

This PR updates the implementation of R2 to return `NaN` when one sample is used for R2 computation and logs a user-facing warning to use two or more samples for non NaN R2 score calculation.

### DIFF
--- a/ludwig/modules/metric_modules.py
+++ b/ludwig/modules/metric_modules.py
@@ -17,7 +17,6 @@ from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from typing import Any, Callable, Generator, Optional
 
-import numpy as np
 import torch
 import torchmetrics.functional as metrics_F
 from torch import Tensor, tensor
@@ -299,7 +298,7 @@ class R2Score(LudwigMetric):
             logger.warning(
                 """R-squared (r2) is not defined for one sample. It needs at least two samples. Returning NaN."""
             )
-            return np.nan
+            return torch.tensor(float("nan"))
 
         return _r2_score_compute(
             self.sum_squared_error, self.sum_error, self.residual, self.total, self.adjusted, self.multioutput
--- a/ludwig/modules/metric_modules.py
+++ b/ludwig/modules/metric_modules.py
@@ -12,18 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import logging
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from typing import Any, Callable, Generator, Optional
 
+import numpy as np
 import torch
 import torchmetrics.functional as metrics_F
-from torch import Tensor
+from torch import Tensor, tensor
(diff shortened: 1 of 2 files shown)


ludwig/modules/metric_modules.py
```diff
@@ -241,13 +245,65 @@ def get_inputs(cls):
...
+
+    def compute(self) -> Tensor:
+        """Computes r2 score over the metric states."""
+
+        # self.total maps to the number of observations in preds/target computed during update()
+        if self.total <= 1:
+            logger.warning(
+                """R-squared (r2) is not defined for one sample. It needs at least two samples. Returning NaN."""
+            )
+            return np.nan
```
justinxzhao, line 302: The function's return type is `Tensor`, but this looks like a numpy type. Should this return `torch.tensor[float('nan')]` instead?

tests/ludwig/modules/test_metric_modules.py
```diff
@@ -36,12 +37,17 @@ def test_rmspe_metric(preds: torch.Tensor, target: torch.Tensor, output: torch.T
...
         (torch.arange(6).reshape(3, 2), torch.arange(6, 12).reshape(3, 2), 2, torch.tensor(-12.5)),
+        (torch.tensor([0.8]), torch.arange(1), 1, np.nan),
     ],
 )
 def test_r2_score(preds: torch.Tensor, target: torch.Tensor, num_outputs: int, output: torch.Tensor):
     metric = metric_modules.R2Score(num_outputs=num_outputs)
     metric.update(preds, target)
-    assert metric.compute() == output
+    computed_score = metric.compute()
+    if np.isnan(computed_score):
```
justinxzhao, line 47: Avoid logic in tests:

https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests

tests/ludwig/modules/test_metric_modules.py
```diff
@@ -36,12 +37,17 @@ def test_rmspe_metric(preds: torch.Tensor, target: torch.Tensor, output: torch.T
     [
         (torch.arange(3), torch.arange(3, 6), 1, torch.tensor(-12.5)),
         (torch.arange(6).reshape(3, 2), torch.arange(6, 12).reshape(3, 2), 2, torch.tensor(-12.5)),
+        (torch.tensor([0.8]), torch.arange(1), 1, np.nan),
```
justinxzhao, line 40: Instead of overloading this test, and adding additional conditional logic in the test to handle this test case, I would create a separate test (unparameterized) for this special sample=1 case, i.e. `test_r2_score_single_prediction`.

tgaddair (approved):

justinxzhao (approved):

PR #2237, opened 2022-07-06
Update get_repeatable_train_val_test_split to handle non-stratified split w/ no existing split

amholler:
As titled.

Tested using new unit test added as well as on experiments repo regression datasets with no existing split column.

### DIFF
--- a/ludwig/utils/dataset_utils.py
+++ b/ludwig/utils/dataset_utils.py
@@ -5,14 +5,15 @@ from ludwig.constants import TEST_SPLIT, TRAIN_SPLIT, VALIDATION_SPLIT
 
 
 def get_repeatable_train_val_test_split(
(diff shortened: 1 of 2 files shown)


ludwig/utils/dataset_utils.py
```diff
@@ -5,14 +5,15 @@
 
 
 def get_repeatable_train_val_test_split(
-    df_input, stratify_colname, random_seed, frac_train=0.7, frac_val=0.1, frac_test=0.2
+    df_input, stratify_colname="", random_seed=42, frac_train=0.7, frac_val=0.1, frac_test=0.2
```
justinxzhao, line 8: Can we use `from ludwig.utils.defaults import default_random_seed`?

justinxzhao (approved):

ludwig/utils/dataset_utils.py
amholler: good idea, will do

PR #2238, opened 2022-07-06
Fix nan in entmax loss and flaky sparsemax/entmax loss tests

dantreiman:
Should fix https://github.com/ludwig-ai/ludwig/issues/2228

I observed two situations where these tests fail:
1.  The original issue, which happens only with `elementwise_mean` reduction and `ignore_index=True`.  With `ignore_index=True` we choose `iix = y[0]`.  Since the batch size is 5, there is a 1/1000 chance that all y[i] == y[0].  This results in a divide by zero in the loss reduction.  Modified loss function to return zero loss if size is 0.
2.  In `test_index_ignored`, it is possible (since this is a sparse loss function) that the ignored index makes exactly 0 contribution to the total loss.  In that case, the strict less-than `assert loss_ignore(x, y) < loss_noignore(x, y)` fails.  Changed to less than or equal to (<=)

### DIFF
--- a/ludwig/utils/entmax/losses.py
+++ b/ludwig/utils/entmax/losses.py
@@ -17,14 +17,14 @@ class _GenericLoss(nn.Module):
         loss = self.loss(X, target)
         if self.ignore_index >= 0:
             ignored_positions = target == self.ignore_index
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):
Nice find!

PR #2239, opened 2022-07-06
Fix preprocessing dataset split API backwards compatibility upgrade bug.

justinxzhao:


ludwig/utils/backward_compatibility.py
```diff
@@ -187,7 +187,7 @@ def _upgrade_preprocessing(preprocessing: Dict[str, Any]):
         )
 
         if "type" not in split_params:
-            split_params["type"] = "random" if force_split else "fixed"
+            split_params["type"] = "fixed" if force_split else "random"
```
tgaddair, line 190: `force_split` is intended to ignore the split column, so if the user had `force_split=True` they would for sure want to use random splitting.

Where things get interesting is if the user had `force_split=False`. In that case they would want to use fixed iff the dataset contains a split column. However, the way this logic was working before, it would use fixed even if no split column was present, hence the error.

As such, it may be that the safest thing to do here is to set the split type to `random` regardless of the `force_split` param value. This has the effect of changing the behavior and forcing users to set the split type to fixed themselves, but should be the safest.

justinxzhao: The latest commit I pushed goes with the latter option, which I agree is the safest.

Users would have used `force_split=False` because it was the default (random) or if they had a feature named `split`, and didn't want to use that for fixed splitting (and thus they probably just want to use random splitting).

PR #2241, opened 2022-07-07
Use Ray Datasets to read binary files in parallel

tgaddair:


ludwig/data/dataframe/dask.py
```diff
@@ -47,13 +66,45 @@ def set_parallelism(self, parallelism):
...
+            if v.npartitions == dataset.npartitions:
+                # Outer join cols with equal partitions
+                v.divisions = dataset.divisions
+                dataset[k] = v
+            else:
+                # If partitions have changed (e.g. due to conversion from Ray dataset), we handle separately
+                repartitioned_cols[k] = v
+
+        if repartitioned_cols:
+            if not dataset.known_divisions:
```
tgaddair, line 90: Is this primarily to handle the image columns, which lack a global index? I just want to make sure we're handling things like numerical columns correctly, where we rely on the index in the case where we drop missing values from the column.

geoffreyangus: Great question– I left a hopefully descriptive (if long) comment in `preprocessing.build_dataset` (at time of writing this comment, line 1087). Essentially, we only need a globally unique index if we know there will be a repartition step (i.e. with image and audio features). If there will NOT be a repartition step (i.e. purely tabular data), partition-level unique indices are sufficient because we don't need to do a global re-alignment of partitions– the join op in `df_like` can happen at the partition-level as per usual.

geoffreyangus:
This PR improves the way path-specified Image/Audio features are loaded during preprocessing. Prior to this change, Image/Audio paths were loaded and placed directly into their source partitions, allowing partitions to balloon in size and causing undue memory pressure. The change implemented here allows the Ray backend to create new partitions when reading paths. The results of such a change are promising across a variety of benchmarking datasets:

| Dataset                              | Branch duration (secs)     |              |
|--------------------------------------|----------------------------|--------------|
|                                      | master                     | fast-im-read |
| Tabular: Criteo (100MB)              | 56.88                      | 56.37        |
| Tabular: Criteo (1GB)                | 289.9                      | 292.55       |
| Image: iSpy2 (~5k rows)              | 450.62                     | 95.22        |
| Image+Text: Twitter Bots (~43k rows) | 1568.34                    | 291.38       |
| Audio: respiratory (~7k rows)        | N/A (crashed after ~1,980) | 308.75       |
| Image+Text: H&M Shopping (~95k rows) | N/A                        | 4503.92      |

A few callouts:
- If training on Image/Audio features on a Ray/Dask backend, a globally unique index is now explicitly required in order to re-align the partitions across various features. Under certain conditions, `preprocessing.build_dataset` will automatically reset the indices to ensure they are globally unique.
- The number of new partitions created during Image/Audio path preprocessing will always be at least `min(len(dataset), 200)`, where 200 is the Ray default number of parallel readers. This number will increase to ensure that the estimated size of the resulting partitions is less than 50MB.
- An [issue](https://github.com/modin-project/modin/issues/4686) with Modin was discovered during implementation. A small workaround on the Ray backend will be reverted once the issue is resolved.

geoffreyangus:
> This PR improves the way path-specified Image/Audio features are loaded during preprocessing. Prior to this change, Image/Audio paths were loaded and placed directly into their source partitions, allowing partitions to balloon in size and causing undue memory pressure. The change implemented here allows the Ray backend to create new partitions when reading paths. The results of such a change are promising across a variety of benchmarking datasets:
> 
> Dataset	Branch duration (secs)	
> master	fast-im-read
> Tabular: Criteo (100MB)	56.88	56.37
> Tabular: Criteo (1GB)	289.9	292.55
> Image: iSpy2 (~5k rows)	450.62	95.22
> Image+Text: Twitter Bots (~43k rows)	1568.34	291.38
> Audio: respiratory (~7k rows)	N/A (crashed after ~1,980)	308.75
> Image+Text: H&M Shopping (~95k rows)	N/A	4503.92
> A few callouts:
> 
> * If training on Image/Audio features on a Ray/Dask backend, a globally unique index is now explicitly required in order to re-align the partitions across various features. Under certain conditions, `preprocessing.build_dataset` will automatically reset the indices to ensure they are globally unique.
> * The number of new partitions created during Image/Audio path preprocessing will always be at least `min(len(dataset), 200)`, where 200 is the Ray default number of parallel readers. This number will increase to ensure that the estimated size of the resulting partitions is less than 50MB.
> * An [issue](https://github.com/modin-project/modin/issues/4686) with Modin was discovered during implementation. A small workaround on the Ray backend will be reverted once the issue is resolved.

^^resurfacing this

ShreyaR (approved):
Wow, this was a pretty substantial change with a bunch of optimizations throughout image reading! Thanks a lot for getting this in 🎉

arnavgarg1 (approved):
Amazing work @geoffreyangus! LGTM 🚢

PR #2242, opened 2022-07-07
Unnest Audio Feature Preprocessing Config

connor-mccorm:
Un-nesting the Audio Feature Preprocessing Parameter config so that it is more consistent and is easier to deal with for the frontend. Currently, the audio feature config looks like this:
```
            "audio": {
                "audio_file_length_limit_in_s": 7.5,
                "missing_value_strategy": "backfill",
                "in_memory": True,
                "padding_value": 0,
                "norm": None,
                "audio_feature": {
                    "type": "fbank",
                    "window_length_in_s": 0.04,
                    "window_shift_in_s": 0.02,
                    "num_filter_bands": 80,
                },
```

With the un-nested audio feature, it will look like this:
```
"audio": {
                "audio_file_length_limit_in_s": 7.5,
                "missing_value_strategy": "backfill",
                "in_memory": True,
                "padding_value": 0,
                "norm": None,
                "type": "fbank",
                "window_length_in_s": 0.04,
                "window_shift_in_s": 0.02,
                "num_fft_points": None,
                "window_type": "hamming",
                "num_filter_bands": 80,
            },
```


ksbrar:
@connor-mccorm for completeness do you mind adding an example of what the unnesting looks like to the description? (e.g. copy paste from the tests)

ludwig/features/audio_feature.py
```diff
@@ -89,12 +104,12 @@ def preprocessing_defaults():
             "in_memory": True,
             "padding_value": 0,
             "norm": None,
```
ksbrar, line 106: Hmm, is it possible to use the schema you defined here, or would that cause a circular import? Seems like the parameter values are duplicated.

ludwig/schema/preprocessing.py
```diff
@@ -557,6 +557,40 @@ class AudioPreprocessingConfig(schema_utils.BaseMarshmallowConfig):
         "!= raw. The following parameters can/should be defined in the dictionary ",
     )
 
+    type: str = schema_utils.StringOptions(
```
ksbrar, line 560: Review won't let me tag the correct line, but above in the definition for `missing_value_strategy` you could use the `BACKFILL` constant

ludwig/constants.py
```diff
@@ -183,6 +183,14 @@
 CONTINUE_PROMPT = "Do you want to continue? "
 
 DEFAULT_AUDIO_TENSOR_LENGTH = 70000
+AUDIO_FEATURE_KEYS = [
```
ksbrar, line 186: idk if this would be too complicated, but an alternate pattern you could use is to make an AudioFeatureKeys schema that is additionally derived by your audio feature schema. Then instead of having this list of keys, it would be a matter of checking if a feature is in that list of attributes. But this is clean, just throwing out an idea if it would be easier to maintain :P

ksbrar (approved):
Nice! LGTM, just a few Qs.

connor-mccorm:
Added description @ksbrar

ludwig/features/audio_feature.py
connor-mccorm: So for preprocessing defaults, that is actually going to be an entirely separate PR after the encoder nesting refactor. I wanted to separate it out into it's own PR since it is a slightly complicated refactor in itself so for the time being we can just let the defaults populate with these methods. But you're absolutely right. The end goal is to have the schema populate the defaults and get rid of these static methods for each feature type.

ludwig/schema/preprocessing.py
connor-mccorm: oh good call, will change!

ludwig/constants.py
connor-mccorm: So I think this would work pretty well and would have the added benefit of having things more schema centralized. On the counter argument, it would probably be like 5x the amount of code since I would have to build out this sub schema whereas here it's pretty small and I don't think there are any plans to get rid of constants. My final vote would be to just keep this implementation, but if you/Justin feel that the other way offers more benefits, I am down to change it.

ksbrar: totally up to you, ya if this list isn't going to change then current approach is much simpler

ludwig/features/audio_feature.py
ksbrar: gotcha

justinxzhao (approved):

PR #2245, opened 2022-07-08
Support saving numpy predictions to remote FS

hungcs:
can potentially use `open_file` for local saving too?

### DIFF
--- a/ludwig/data/postprocessing.py
+++ b/ludwig/data/postprocessing.py
@@ -24,17 +24,18 @@ from ludwig.backend import LOCAL_BACKEND
 from ludwig.data.utils import convert_to_dict
 from ludwig.utils.data_utils import DATAFRAME_FORMATS, DICT_FORMATS
 from ludwig.utils.dataframe_utils import to_numpy_dataset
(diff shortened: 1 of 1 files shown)


ludwig/data/postprocessing.py
```diff
@@ -66,8 +67,12 @@ def _save_as_numpy(predictions, output_directory, saved_keys, backend):
...
         k = k.replace("<", "[").replace(">", "]")  # Replace <UNK> and <PAD> with [UNK], [PAD]
         if k not in saved_keys:
-            np.save(npy_filename.format(make_safe_filename(k)), v)
-            saved_keys.add(k)
+            if has_remote_protocol(output_directory):
+                with open_file(npy_filename.format(make_safe_filename(k)), mode="wb") as f:
+                    np.save(f, v)
+            else:
+                np.save(npy_filename.format(make_safe_filename(k)), v)
+                saved_keys.add(k)
```
tgaddair, line 75: Why is the `saved_keys` stuff only in the else?

tgaddair (approved):

PR #2247, opened 2022-07-09
Fixed handling of invalud number values to treat as missing values

tgaddair:
In #2058, the second call to `handle_missing_values` was removed, but this was necessary to ensure that invalid values (which are cast to NaN) during `cast_columns`) are properly considered as missing values.

It's an open question as to whether this is the desired behavior, or if we should instead treat this as a user error (the user needs to clean the data to conform to the expected dtype). The correct answer here may be to make this configurable per feature.

However, because this changes the behavior from v0.5.2 to v0.5.3, this should be considered a regression and reverted to the prior behavior for now. This change will require a patch release.

### DIFF
--- a/tests/integration_tests/test_preprocessing.py
+++ b/tests/integration_tests/test_preprocessing.py
@@ -1,4 +1,6 @@
 import os
+import random
+import string
 
 import numpy as np
 import pandas as pd
@@ -6,6 +8,7 @@ import pytest
 
 from ludwig.api import LudwigModel
 from ludwig.constants import COLUMN, PROC_COLUMN
+from ludwig.data.concatenate_datasets import concatenate_df
 from tests.integration_tests.utils import (
     audio_feature,
     binary_feature,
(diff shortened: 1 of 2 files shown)


tests/integration_tests/test_preprocessing.py
```diff
@@ -139,3 +143,34 @@ def test_dask_known_divisions(feature_fn, csv_filename, tmpdir):
...
+
+    num_feat = number_feature()
+    input_features = [num_feat]
+    output_features = [binary_feature()]
+    config = {"input_features": input_features, "output_features": output_features}
+
+    training_data_csv_path = generate_data(input_features, output_features, data_csv_path)
+    df = pd.read_csv(training_data_csv_path)
+
+    # convert numbers to random strings
```
justinxzhao, line 160: nit: optional, instead of defining a new random string function, it seems cleaner to use a synthetic text input feature and then only change the input feature type just in the config, like config["input_features"][0]["type"] = "number"

justinxzhao (approved):
Thanks!

PR #2248, opened 2022-07-10
Fixed splitting when providing pre-split inputs

tgaddair:
Prior behavior was not overriding the preprocessing split type to `fixed` when providing a pre-split dataframe. This was resulting in input dataframes being concatenated and then randomly split into a different train-test-val split.


justinxzhao (approved):
LGTM.

PR #2251, opened 2022-07-11
Triton ensemble export

abidwael:
Exports Triton configs and scripted models as well an an ensemble config.

1. train a ludwig model.
```

training_set, test_set, _ = titanic.load(split=True)
model = LudwigModel(config="./titanic.yaml")

train_stats, preprocessed_data, output_directory = model.train(training_set=training_set,
                                                               test_set=test_set,
                                                               experiment_name="simple_experiment",
                                                               model_name="simple_model",
                                                               skip_save_processed_input=True)
```
2. To export to models and configs to a Triton-compliant structure 
```
export_triton(model, data_example, output_path, model_name, model_version, device, device_count)
```
3. Find the exported models under `model_repository/`


ludwig/utils/triton_utils.py
```diff
@@ -49,142 +67,434 @@ def forward(self, {input_signature}):
...
-    for feature_name, feature in model.model.input_features.items():
-        metadata = model.training_set_metadata[feature_name]
-        spec.append(
-            TRITON_SPEC.format(
-                prefix="INPUT",
-                index=len(spec),
-                name=feature.feature_name,
-                data_type=_get_type_map(feature.get_preproc_input_dtype(metadata)),
-                data_dims=1,  # hard code for now
+def raw_feature_to_inference_input(s: pd.Series, feature) -> Union[List[str], torch.Tensor]:
```
brightsparc, line 138: Can we use the `to_inference_model_input_from_series` in the `ludwig.utils.inference_utils.py` here?

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +67,434 @@ def forward(self, {input_signature}):
...
+    """Transform input for a feature to be compatible with what's required by TorchScript."""
+    if s.dtype == "object" or type(feature) is CategoryInputFeature:
+        return s.astype("str").to_list()
+    return torch.from_numpy(s.to_numpy())
+
+
+def raw_to_inference_input(df: pd.DataFrame, model: LudwigModel):
```
brightsparc, line 145: Likewise can we use the `to_inference_module_input_from_dataframe` method here?

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +67,434 @@ def forward(self, {input_signature}):
...
+        for name, feature in model.model.input_features.items()
+    }
+
+
+def to_triton_dimension(content: Union[List[str], List[torch.Tensor], List[TorchAudioTuple], torch.Tensor]):
+    if isinstance(content, list) and content:
+        if isinstance(content[0], str):
+            return [len(content)]
+        """
+        todo (Wael): check these and add other types
```
brightsparc, line 158: Will require special checks for image here which can have 1-4 channels see `read_image_from_path` and `get_image_read_mode_from_num_channels` in `image_utils.py`

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +67,434 @@ def forward(self, {input_signature}):
...
-    model_path = os.path.join(model_dir, "model.pt")
-    model_ts.save(model_path)
-    # Save the default onfig to <model_repository>/<model_name>/config.pbtxt
-    config_path = os.path.join(output_path, model_name, "config.pbtxt")
-    with open(config_path, "w") as f:
-        f.write(_get_model_config(model))
-    return model_path, config_path
+
+    inference_module = InferenceModule.from_ludwig_model(model.model, model.config, model.training_set_metadata, DEVICE)
+    split_modules = [inference_module.preprocessor, inference_module.predictor, inference_module.postprocessor]
```
brightsparc, line 479: Nice, this is a good way to get modules.

brightsparc (changes requested):
@abidwael This is looking good.  I suggest you look to leverage the inference_utils which is being split out in this draft pr #2213 (perhaps you could land this first).  So that you are not repeating the pre/post processing functions.

There will need to be some additional work to support images [see example](https://github.com/predibase/predibase-infra/blob/main/notebooks/tooso-torchscript.ipynb) for model that includes text and image columns.

Would also be nice if we could have a way of providing triton [InstanceGroup](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#instance-groups) settings to be passed in, at a minimum `cpu_count` and `gpu_count` could be useful properties which we could then use herufistic 
```
pre/post processing == cpu_count
predict = gpu_count if gpu_count > 0 else cpu_count
```

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +89,488 @@ def forward(self, {input_signature}):
...
+
+    triton_master_preprocessor: TritonMaster
+    triton_master_predictor: TritonMaster
+    triton_master_postprocessor: TritonMaster
+    model_name: str
+    output_path: str
+    model_version: int
+
+    def __post_init__(self):
+        self.ensemble_model_name = self.model_name + "_" + ENSEMBLE
```
brightsparc, line 338: Given this is the `entrypoint` for the model, and we will need a way to reference this upstream, I suggest not having a suffix for the ensemble file such that if the `model_name` was `titanic_ensemble` then the directory structure would be the following.  

```
model_repository
├── titanic_ensemble
│   ├── 1
│   │   └── model.txt
│   ├── 2
│   │   └── model.txt
│   └── config.pbtxt
├── titanic_ensemble_postprocess
│   ├── 1
│   │   └── model.pt
│   ├── 2
│   │   └── model.pt
│   └── config.pbtxt
├── titanic_ensemble_predict
│   ├── 1
│   │   └── model.pt
│   ├── 2
│   │   └── model.pt
│   └── config.pbtxt
└── titanic_ensemble_preprocess
    ├── 1
    │   └── model.pt
    ├── 2
    │   └── model.pt
    └── config.pbtxt
```

Also, for your constants I notice these strings are `UPPERCASE` perhaps these should be lower?

abidwael:
- The preprocessor and postprocessor are always exported with `max_batch_size = 0`. This is because Triton cannot batch certain input types like strings. No specific reason behind setting `max_batch_size = 0` for the postprocessor.
- The default for predictor is `max_batch_size = 1` with dynamic batching enabled.

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +93,522 @@ def forward(self, {input_signature}):
...
+    def _get_dynamic_batching_spec(self):
+        if self.inference_stage == PREDICTOR:
+            return DYNAMIC_BATCHING_TEMPLATE
+        return ""
+
+    def get_model_config(self) -> str:
+        """Generate a Triton config for a model from the input and output features."""
+        max_batch_size = 0
+        if self.inference_stage == PREDICTOR:
+            max_batch_size = 1
```
brightsparc, line 461: Should we add an argument to allow passing this in, given that we would probably want this to be more than 1 in most cases (perhaps we could default to 64 if we don't get a value?)

ludwig/utils/triton_utils.py
```diff
@@ -19,28 +55,36 @@ def __init__(self, inference_module):
...
 INSTANCE_SPEC = """
     {{
         count: {count}
         kind: {kind}
     }}"""
 
-TRITON_CONFIG_TEMPLATE = """
-name: "pytorch_raw"
+DYNAMIC_BATCHING_TEMPLATE = """dynamic_batching {
+    max_queue_delay_microseconds: 100
```
brightsparc, line 81: Can define an export parameters to pass for `max_queue_delay_microseconds`, with default of 100

ludwig/utils/triton_utils.py
```diff
@@ -19,28 +55,36 @@ def __init__(self, inference_module):
...
 """
 
 TRITON_SPEC = """
     {{
-        name: "{prefix}__{index}" # "{name}"
+        name: "{key}" # "{name}"
         data_type: {data_type}
         dims: [ {data_dims} ]
-    }}"""
+        {reshape_spec}}}"""
```
brightsparc, line 72: `nit`: Fix formatting to be consistent with open bracket.

brightsparc (changes requested):
Request to add more args / defaults for dynamic batching.

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +93,522 @@ def forward(self, {input_signature}):
...
+            gen_module = gen_ts.GeneratedInferenceModule(self.module)
+            scripted_module = torch.jit.script(gen_module)
+        return scripted_module
 
 
 def export_triton(
-    model: LudwigModel, output_path: str, model_name: str = "ludwig_model", model_version: Union[int, str] = 1
-) -> Tuple[str, str]:
+    model: LudwigModel,
+    data_example: pd.DataFrame,
```
brightsparc, line 538: If possible would be nice to make `data_example` optional, or not required given we can generate a dataset from the config eg:

```
        from ludwig.data.dataset_synthesizer import build_synthetic_dataset
        import pandas as pd
        import shutil
        import tarfile
        def generate_df(config, num_examples=2):
            features = config['input_features'] + config['output_features']
            df = build_synthetic_dataset(num_examples, features)
            data = [next(df) for _ in range(num_examples)]
            return pd.DataFrame(data[1:], columns=data[0])
      try:
          example_df = generate_df(model.config)
          output_paths: Dict[str, Tuple[str, str]] = export_triton(model, example_df, tmpdir,
              model_name, model_version, device, device_count)
          logger.info("Got paths", output_paths)
      except Exception as e:
          msg = f"error in triton export for model {model_name}"
          logger.error(msg, e)
          raise Exception(msg)
      finally:
          # Clean up any temporay files that might have been created
          shutil.rmtree('audio_files', ignore_errors=True)
          shutil.rmtree('image_files', ignore_errors=True)

```

ludwig/utils/triton_utils.py
```diff
@@ -539,13 +545,15 @@ def generate_scripted_module(self):
...
-    model: LudwigModel,
-    data_example: pd.DataFrame,
-    output_path: str = "model_repository",
-    model_name: str = "ludwig_model",
-    model_version: Union[int, str] = 1,
-    device: str = "cpu",
-    device_count: int = 1,
+        model: LudwigModel,
+        data_example: pd.DataFrame,
+        predictor_max_batch_size: int = 1,
```
brightsparc: `nit` can we move these to the end, given they are probably the most "optional" args, and if folks want to include all inputs without naming them, they could omit these.

abidwael: Which ones should be moved to the end/what order you think would make most sense?

brightsparc: I think just the two new properties you added for `max_batch_size` and `max_queue_delay_microseconds` could go after device count.

ludwig/utils/triton_utils.py
```diff
@@ -49,142 +95,538 @@ def forward(self, {input_signature}):
...
-) -> Tuple[str, str]:
+    model: LudwigModel,
+    data_example: pd.DataFrame,
+    predictor_max_batch_size: int = 64,
+    max_queue_delay_microseconds: int = 100,
+    output_path: str = "model_repository",
+    model_name: str = "ludwig_model",
+    model_version: Union[int, str] = 1,
+    device: str = "cpu",
+    device_count: int = 1,
```
brightsparc, line 559: In order to better reflect the ability to target a different number of cpu and gpu instances, I suggest we change the configuration so its close to ray [init](https://docs.ray.io/en/latest/ray-core/package-ref.html#ray-init) eg:
* `num_cpus: Optional[int]=1`,
* `num_gpus: Optional[int]=0`,
* `max_batch_size: Optional[int]=64`,
* `max_queue_delay_microseconds: Optional[int]=100`
From here we can use the `num_gpus>0` to determine if we target the gpu device for the predictor

ludwig/utils/triton_utils.py
```diff
@@ -50,142 +98,620 @@ def forward(self, {input_signature}):
...
+            example_input,
+            INFERENCE_STAGES[i],
+            predictor_max_batch_size,
+            max_queue_delay_microseconds,
+            model_name,
+            output_path,
+            model_version,
+            model.config,
+            device_types[i],
+            model_instance_count=instance_counts[i],
```
brightsparc, line 694: Should we use named parameters here for all args given the number of parameters, and the fact that order might change in future?

ludwig/utils/triton_utils.py
```diff
@@ -50,142 +98,620 @@ def forward(self, {input_signature}):
...
+        with open(config_path, "w") as f:
+            formatted_config = remove_empty_lines(self.get_config())
+            f.write(formatted_config)
+        return config_path
+
+    def save_ensemble_dummy_model(self) -> str:
+        """Scripts the model and saves it."""
+        if not isinstance(self.model_version, int) or self.model_version < 1:
+            raise ValueError("Model version has to be a non-zero positive integer")
+        pass
```
brightsparc, line 445: Should there bee a `pass` here?

ludwig/utils/triton_utils.py
```diff
@@ -50,142 +98,620 @@ def forward(self, {input_signature}):
...
+            output_path,
+            model_version,
+            model.config,
+            device_types[i],
+            model_instance_count=instance_counts[i],
+        )
+        example_input = triton_master.output_data_example
+        config_path = triton_master.save_config()
+        model_path, size = triton_master.save_model()
+        paths[INFERENCE_STAGES[i]] = (config_path, model_path, size)
```
brightsparc, line 699: Instead of this being a dictionary of tuples, could we simply return a list of list which include the following properties on a dictionary (or TritonArtifact class if you wanted to make it strongly typed):
* `model_name`
* `model_version`
* `platform`
* `path`
* `content_type`
* `content_length`

The rational being that in future we may need to include some additional python files or other assets which don't fall neatly into `config` or `model` categorisation.  We can also uses the `content_type` and `content_length` properties when upload to the s3 object store.

So an example output would be:
```
[
  { 
  "model_name": "titanic",
  "model_version": 1"
  "platform": "ensemble",
  "path": "titanic/config.pbtxt"
  "content_type": "text/x-protobuf",
  "content_length": 10000
  },
  { 
  "model_name": "titanic",
  "model_version": 1"
  "platform": "ensemble",
  "path": "titanic/1/placeholder.txt",
  "content_type": "text/plain",
  "content_length": 10
  },
  { 
  "model_name": "titanic_predictor",
  "model_version": 1"
  "platform": "pytorch_libtorch",
  "path": "titanic_predictor/config.pbtxt"
  "content_type": "text/x-protobuf",
  "content_length": 10
  },
  { 
  "model_name": "titanic_predictor",
  "model_version": 1"
  "platform": "pytorch_libtorch",
  "path": "titanic_predictor/1/model.pt"
  "content_type": "application/octet-stream",
  "content_length": 10
  }
]
```

brightsparc (changes requested):
This is getting there, I've added some comments.

I'd like to see a bit more flexibility added around the list of paths that we return so that we can support different file types in future, and pass along the content type when uploading to s3.  This could be done in another PR if necessary.

We do need this PR to update [integration tests](https://github.com/ludwig-ai/ludwig/blob/triton-ensemble-export/tests/integration_tests/test_triton.py#L93) and the [cli export
](https://github.com/ludwig-ai/ludwig/blob/triton-ensemble-export/ludwig/export.py#L81) as well as fix any remaining tests.

ludwig/utils/triton_utils.py
abidwael: If it's a `List[List[TritonArtifact]]`, would the outer list (currently) be of size 4 (preprocessor, predictor, postprocessor, ensemble)? And the inner list would be of any size that could contain any number of `TritonArtifacts`?

In this case, if someone wanted to get a Triton config or model, would they have to iterate over the list and check the fields?

brightsparc: I'm proposing it would be a flat List[TritonArtifact] where each of the stages would emit two items to the list.   The downstream implementation is just to copy these files along with their content type into a cloud bucket, to it doesn't really care that there is a series of integration stages for each model.

Yes, we would filter on the model name / version / content type if we wanted to get specific config.

ludwig/utils/triton_utils.py
```diff
@@ -351,11 +375,22 @@ def save_config(self) -> str:
...
             f.write(formatted_config)
-        return config_path
+
+        config_artifact = TritonArtifact(
+            model_name=self.full_model_name,
+            model_version=self.model_version,
+            platform="pytorch_libtorch",
+            path=config_path,
+            content_type="text/x-protobuf",
+            content_length=0,
```
brightsparc, line 390: Add file size.

ludwig/utils/triton_utils.py
```diff
@@ -431,14 +466,24 @@ def get_config(self):
...
-        return config_path
 
-    def save_ensemble_dummy_model(self) -> str:
+        config_artifact = TritonArtifact(
+            model_name=self.ensemble_model_name,
+            model_version=self.model_version,
+            platform="ensemble",
+            path=config_path,
+            content_type="text/x-protobuf",
+            content_length=0,
```
brightsparc, line 481: Add file size.

ludwig/utils/triton_utils.py
```diff
@@ -449,7 +494,16 @@ def save_ensemble_dummy_model(self) -> str:
...
             f.write("no model for the ensemble")
 
-        return model_path
+        model_artifact = TritonArtifact(
+            model_name=self.ensemble_model_name,
+            model_version=self.model_version,
+            platform="ensemble",
+            path=model_path,
+            content_type="text/plain",
+            content_length=0,
```
brightsparc, line 503: Add file size.

brightsparc (approved):
LGTM; just add file size.

PR #2254, opened 2022-07-11
Added Backwards Compatibility for Audio Feature Preprocessing

connor-mccorm:


ludwig/utils/backward_compatibility.py
```diff
@@ -70,6 +70,17 @@ def _upgrade_feature(feature: Dict[str, Any]):
     if feature.get(TYPE) == "numerical":
         warnings.warn('Feature type "numerical" renamed to "number" and will be removed in v0.6', DeprecationWarning)
         feature[TYPE] = NUMBER
+    if feature.get(TYPE) == "audio":
```
hungcs, line 73: can we use constants for all the below?
```
from ludwig.constants import COMBINER, INPUT_FEATURES, OUTPUT_FEATURES, TYPE, AUDIO, PREPROCESSING, SPLIT, etc...
```

connor-mccorm: Yup, will add now

hungcs (approved):

hungcs:
thanks for the quick turnaround!

connor-mccorm:
Absolutely! So I added the existing constants, but there are a bunch of other small things in there that don't have constants yet (i.e. "stratify" , "force_split", "split_probabilities"). Do you think it's ok to just leave these as strings, or should I add constants for all the string indexers used here.

justinxzhao (approved):

hungcs:
> Absolutely! So I added the existing constants, but there are a bunch of other small things in there that don't have constants yet (i.e. "stratify" , "force_split", "split_probabilities"). Do you think it's ok to just leave these as strings, or should I add constants for all the string indexers used here.

would be nice to add those too!

hungcs:
does `merge_with_defaults` need to be updated? when calling this, i get the old preprocessing > audio_features

ludwig/utils/backward_compatibility.py
```diff
@@ -186,24 +216,35 @@ def _upgrade_preprocessing(preprocessing: Dict[str, Any]):
...
+        preprocessing[SPLIT] = split_params
+
+    if AUDIO in preprocessing:
+        if "audio_feature" in preprocessing[AUDIO]:
+            for k, v in preprocessing[AUDIO]["audio_feature"].items():
+                preprocessing[AUDIO][k] = v
+            del preprocessing[AUDIO]["audio_feature"]
+        warnings.warn(
+            "Parameters specified at the `audio_feature` parameter level have been unnested and should now "
+            "be specified at the preprocessing level",
```
tgaddair, line 232: Please mention the removal version to be v0.7.

connor-mccorm: Will do!

PR #2255, opened 2022-07-11
Optional artifact logging for MLFlow

ShreyaR:
Skip artifact logging for MLFlowCallback with an initialization parameter

### DIFF
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -2,7 +2,7 @@ import logging
 import os
 import queue
 import threading
-from typing import Any, Callable, Dict
+from typing import Any, Dict
 
 from ludwig.callbacks import Callback
 from ludwig.constants import TRAINER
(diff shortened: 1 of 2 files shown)


jeffreyftang (approved):

ludwig/contribs/mlflow/__init__.py
```diff
@@ -28,7 +28,7 @@ def _get_or_create_experiment_id(experiment_name):
 
 
 class MlflowCallback(Callback):
-    def __init__(self, tracking_uri=None):
+    def __init__(self, tracking_uri=None, skip_artifact_logging: bool = False):
```
jeffreyftang, line 31: I know there's precedent for naming like this, but typically it's easier to read a variable name like `log_artifacts = True` rather than `if not skip_artifact_logging`

tgaddair: I agree.

tgaddair (changes requested):
Please see @jeffreyftang 's comment.

We also need to handle the model logging setup in `on_trainer_train_setup`.

Please also add a test to verify correctness.

ludwig/contribs/mlflow/__init__.py
```diff
@@ -80,14 +84,16 @@ def on_train_init(self, base_config, experiment_name, output_directory, resume_d
                 run_name = os.path.basename(output_directory)
                 self.run = mlflow.start_run(experiment_id=self.experiment_id, run_name=run_name)
 
-        mlflow.log_dict(to_json_dict(base_config), "config.yaml")
+        if self.log_artifacts:
```
tgaddair, line 87: I don't think this is an artifact. Can you clarify this one?

ShreyaR: This logs the base config as an artifact, discussed more in the docs for [mlflow.log_dict](https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.log_dict). The dictionary gets logged as a yaml file like in the screenshot below:
Screen Shot 2022-07-27 at 3 36 14 PM

We could still want to log this dictionary as an artifact even when artifact logging for other files is turned off. In that case however, the artifact locations, etc. will all be managed by mlflow.

jeffreyftang: For something like the config it seems reasonable to log regardless.

tgaddair: Ah, I see. I agree with disabling it, then, because otherwise mlflow will just write it to the local (ephemeral) disk.

tgaddair (approved):

jeffreyftang (approved):

PR #2260, opened 2022-07-12
benchmarking utils

abidwael:
A collection of utility functions for the benchmarking tool.

### DIFF
--- a/ludwig/benchmarking/utils.py
+++ b/ludwig/benchmarking/utils.py
@@ -1,9 +1,7 @@
-import copy
 import logging
 import os
 import shutil
-from datetime import datetime
-from typing import Any, Dict, Tuple, Union
+from typing import Any, Dict
 
 import fsspec
 import pandas as pd
@@ -23,7 +21,7 @@ from ludwig.utils.fs_utils import get_fs_and_path
 
 def load_from_module(
     dataset_module: BaseDataset, output_feature: Dict[str, str], subsample_frac: float = 1
-) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+) -> pd.DataFrame:
     """Load the ludwig dataset, optionally subsamples it, and returns a repeatable split. A stratified split is
     used for classification datasets.
 
@@ -43,10 +41,7 @@ def load_from_module(
     else:
         dataset = get_repeatable_train_val_test_split(dataset, random_seed=default_random_seed)
 
-    train_df = dataset[dataset["split"] == 0]
-    val_df = dataset[dataset["split"] == 1]
-    test_df = dataset[dataset["split"] == 2]
-    return train_df, val_df, test_df, dataset
+    return dataset
 
 
 def flatten_dict(d: Dict[str, Any], sep: str = ".") -> Dict[str, Any]:
--- a/ludwig/benchmarking/utils.py
+++ b/ludwig/benchmarking/utils.py
@@ -72,12 +72,7 @@ def export_artifacts(
     protocol, _ = fsspec.core.split_protocol(export_base_path)
     fs, _ = get_fs_and_path(export_base_path)
     try:
-        experiment_name = (
-            experiment["experiment_name"]
-            if experiment["experiment_name"]
-            else datetime.now().strftime("%Y.%m.%d.%H:%M:%S")
-        )
-        export_full_path = os.path.join(export_base_path, experiment["dataset_name"], experiment_name)
+        export_full_path = os.path.join(export_base_path, experiment["dataset_name"], experiment["experiment_name"])
         fs.put(report_path, os.path.join(export_full_path, REPORT_JSON), recursive=True)
         fs.put(
             os.path.join("configs", experiment["config_path"]),
--- a/ludwig/benchmarking/utils.py
+++ b/ludwig/benchmarking/utils.py
@@ -3,72 +3,46 @@ import logging
 import os
 import shutil
 from datetime import datetime
-from types import ModuleType
 from typing import Any, Dict, Tuple, Union
 
 import fsspec
 import pandas as pd
 from botocore.exceptions import ClientError
 from globals import CONFIG_YAML, EXPERIMENT_RUN, REPORT_JSON
+# todo (Wael): add to ludwig.globals
 from s3fs.errors import translate_boto_error
 
 from ludwig.constants import CATEGORY
 from ludwig.globals import MODEL_HYPERPARAMETERS_FILE_NAME
+from ludwig.datasets.base_dataset import BaseDataset
 from ludwig.utils.dataset_utils import get_repeatable_train_val_test_split
-from ludwig.utils.defaults import default_random_seed, merge_with_defaults
+from ludwig.utils.defaults import default_random_seed
 from ludwig.utils.fs_utils import get_fs_and_path
 
 
-def get_full_config(config: Dict[str, Any]) -> Dict[str, Any]:
-    """Return a full Ludwig config given an experiment config.
-
-    config: ludwig config.
-    """
-    config_dict = copy.deepcopy(config)
-    config = merge_with_defaults(config_dict)
-    return config
-
-
 def load_from_module(
-    dataset_module: ModuleType, output_feature: Dict[str, str], subsample_frac: float = 1
+    dataset_module: BaseDataset, output_feature: Dict[str, str], subsample_frac: float = 1
 ) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame, pd.DataFrame]:
-    """Take in Ludwig Python module of the dataset, load it, and split it if it isn't already (by adding "split"
-    column).
+    """Load the ludwig dataset, optionally subsamples it, and returns a repeatable split.
+    A stratified split is used for classification datasets.
 
-    dataset_module: ludwig dataset module (e.g. ludwig.datasets.sst2, ludwig.datasets.ames_housing, etc.)
-    subsample_frac: percentage of the total datasets to load.
+    dataset_module: ludwig datasets module (e.g. ludwig.datasets.sst2, ludwig.datasets.ames_housing, etc.)
+    subsample_frac: percentage of the total dataset to load.
     """
     dataset = dataset_module.load(split=False)
     if subsample_frac < 1:
         dataset = dataset.sample(frac=subsample_frac, replace=False, random_state=default_random_seed)
 
     if output_feature["type"] == CATEGORY:
-        dataset = get_repeatable_train_val_test_split(
-            dataset,
-            stratify_colname=output_feature["name"],
-            random_seed=default_random_seed,
-            frac_train=0.75,
-            frac_val=0.1,
-            frac_test=0.15,
-        )
+        dataset = get_repeatable_train_val_test_split(dataset, stratify_colname=output_feature["name"],
+                                                      random_seed=default_random_seed,)
     else:
-        dataset = get_repeatable_train_val_test_split(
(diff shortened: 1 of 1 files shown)


ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+    config: ludwig config.
+    """
+    config_dict = copy.deepcopy(config)
+    config = merge_with_defaults(config_dict)
+    return config
+
+
+def load_from_module(
+    dataset_module: ModuleType, output_feature: Dict[str, str], subsample_frac: float = 1
```
justinxzhao, line 33: `BaseDataset` seems more specific than `ModuleType`.

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+            frac_test=0.15,
+        )
+    else:
+        dataset = get_repeatable_train_val_test_split(
+            dataset, random_seed=default_random_seed, frac_train=0.75, frac_val=0.1, frac_test=0.15
+        )
+
+    train_df = dataset[dataset["split"] == 0]
+    val_df = dataset[dataset["split"] == 1]
+    test_df = dataset[dataset["split"] == 2]
```
justinxzhao, line 61: I believe `get_repeatable_train_val_test_split` already adds the `split` column to each of the dataframes. Could you check?

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def load_from_module(
+    dataset_module: ModuleType, output_feature: Dict[str, str], subsample_frac: float = 1
+) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+    """Take in Ludwig Python module of the dataset, load it, and split it if it isn't already (by adding "split"
+    column).
+
+    dataset_module: ludwig dataset module (e.g. ludwig.datasets.sst2, ludwig.datasets.ames_housing, etc.)
+    subsample_frac: percentage of the total datasets to load.
```
justinxzhao, line 39: nit: datasets -> dataset

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    config_dict = copy.deepcopy(config)
+    config = merge_with_defaults(config_dict)
+    return config
+
+
+def load_from_module(
+    dataset_module: ModuleType, output_feature: Dict[str, str], subsample_frac: float = 1
+) -> Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame, pd.DataFrame]:
+    """Take in Ludwig Python module of the dataset, load it, and split it if it isn't already (by adding "split"
```
justinxzhao, line 35: nit: "Loads the ludwig dataset, optionally subsamples it, and returns a repeatable split. A stratified split is used for classification datasets."

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+        dataset = dataset.sample(frac=subsample_frac, replace=False, random_state=default_random_seed)
+
+    if output_feature["type"] == CATEGORY:
+        dataset = get_repeatable_train_val_test_split(
+            dataset,
+            stratify_colname=output_feature["name"],
+            random_seed=default_random_seed,
+            frac_train=0.75,
+            frac_val=0.1,
+            frac_test=0.15,
```
justinxzhao, line 52: Is there a strong reason to not go with the defaults of 0.7, 0.2, and 0.1 that's the ludwig and get_repeatable_train_val_test_split default?

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    return f"{num:.1f}Yi{suffix}"
+
+
+def flatten_dict(d: Dict[str, Any], sep: str = ".") -> Dict[str, Any]:
+    [flat_dict] = pd.json_normalize(d, sep=sep).to_dict(orient="records")
+    return flat_dict
+
+
+def export_artifacts(
+    experiment: Dict[str, str], report_path: str, experiment_output_directory: str, export_base_path: str
```
justinxzhao, line 80: What are the two strings in the `experiment` argument?

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    export_base_path: path (directory) where artifacts are
+        exported. (e.g. s3://benchmarking.us-west-2.ludwig.com/bench/)
+    """
+    protocol, _ = fsspec.core.split_protocol(export_base_path)
+    fs, _ = get_fs_and_path(export_base_path)
+    try:
+        experiment_name = (
+            experiment["experiment_name"]
+            if experiment["experiment_name"]
+            else datetime.now().strftime("%Y.%m.%d.%H:%M:%S")
```
justinxzhao, line 99: What's the reason for not appending a timestamp to the experiment name always?

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+    """Save the experiment artifacts to the `bench_export_directory`.
+
+    experiment: experiment dict that contains dataset name,
+        experiment name, and config path.
+    report_path: path where the experiment metrics report is
+        saved.
+    experiment_output_directory: path where the model, data,
+        and logs of the experiment are saved.
+    export_base_path: path (directory) where artifacts are
+        exported. (e.g. s3://benchmarking.us-west-2.ludwig.com/bench/)
```
justinxzhao, line 91: Maybe clarify that this is an s3 path, or rename the variable to `s3_base_path`.

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+def export_artifacts(
+    experiment: Dict[str, str], report_path: str, experiment_output_directory: str, export_base_path: str
+) -> None:
+    """Save the experiment artifacts to the `bench_export_directory`.
+
+    experiment: experiment dict that contains dataset name,
+        experiment name, and config path.
+    report_path: path where the experiment metrics report is
+        saved.
+    experiment_output_directory: path where the model, data,
```
justinxzhao, line 88: Are these expected to be separate directories? I'm wondering if it makes sense to consolidate.

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+            dataset, random_seed=default_random_seed, frac_train=0.75, frac_val=0.1, frac_test=0.15
+        )
+
+    train_df = dataset[dataset["split"] == 0]
+    val_df = dataset[dataset["split"] == 1]
+    test_df = dataset[dataset["split"] == 2]
+    return train_df, val_df, test_df, dataset
+
+
+def bytes_to_human_readable(num: Union[float, int], suffix: str = "B") -> str:
```
justinxzhao, line 65: It looks like this code is doing something non-trivial. Please add a unit test for it under `tests/ludwig/benchmarking/test_utils.py`

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,125 @@
...
+from ludwig.utils.fs_utils import get_fs_and_path
+
+
+def get_full_config(config: Dict[str, Any]) -> Dict[str, Any]:
+    """Return a full Ludwig config given an experiment config.
+
+    config: ludwig config.
+    """
+    config_dict = copy.deepcopy(config)
+    config = merge_with_defaults(config_dict)
```
justinxzhao, line 28: Is there a reason why the caller of this utility couldn't just call `merge_with_defaults()` directly?

justinxzhao (commented):
Thank you for breaking up the changes into these smaller PRs!

ludwig/benchmarking/utils.py
abidwael: It does, but it returns one instead of three separate dataframes.

ludwig/benchmarking/utils.py
abidwael: No, will switch to defaults.

ludwig/benchmarking/utils.py
abidwael: Added to the docstring:
```
experiment: experiment dict that contains "dataset_name" (e.g. ames_housing),
        "experiment_name" (specified by user), and "config_path" (path to experiment config.
        Relative to ludwig/benchmarks/configs).
```

ludwig/benchmarking/utils.py
abidwael: One potential reason is that it would be easier to retrieve the experiment report (or filter by experiment name) without having the date there. 
Example: we want to benchmark a new version (e.g. `0.5.4`) of ludwig on a number of datasets. If we later want to compare all `0.5.3` vs. `0.5.4` benchmarks, it would be easier to filter by this name without the date. The dated names will be `0.5.4-2022.07.12.11:23:44`, `0.5.4-2022.07.12.11:31:19` and it might be more of a challenge to use these names. However, we can make the timestamp available in the exported report.

What do you think?

ludwig/benchmarking/utils.py
abidwael: It can be any local or remote path since we're using `fsspec`. Added clarification to docstring.

ludwig/benchmarking/utils.py
abidwael: `report_path` will be something like `titanic/reports/full_report.json`
`experiment_output_directory` can be passed as the `output_directory` argument in `api.experiment`. This whole directory is being zipped as `artifacts.zip` and pushed under `export_base_path`

We can go with a solution where we just export `artifacts.zip` but I think there are times where we just need the `full_report.json`, so I'm in favor of keeping them separate.

ludwig/benchmarking/utils.py
abidwael: Function was being used in the tracker, but no longer.

ludwig/benchmarking/utils.py
abidwael: Agreed, but removing function as it's not used anymore.

ludwig/benchmarking/utils.py
justinxzhao: Is there more that you are doing with the datasets such that you need to manually split them up yourself?

You should be able to use the combined dataframe returned by `get_repeatable_train_val_test_split` with ludwig directly, with `split: {type: fixed}`.

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    if output_feature["type"] == CATEGORY:
+        dataset = get_repeatable_train_val_test_split(
+            dataset,
+            stratify_colname=output_feature["name"],
+            random_seed=default_random_seed,
+        )
+    else:
+        dataset = get_repeatable_train_val_test_split(dataset, random_seed=default_random_seed)
+
+    return dataset
```
justinxzhao, line 44: nit: 

```suggestion
    return get_repeatable_train_val_test_split(dataset, random_seed=default_random_seed)
```

ludwig/benchmarking/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    dataset = dataset_module.load(split=False)
+    if subsample_frac < 1:
+        dataset = dataset.sample(frac=subsample_frac, replace=False, random_state=default_random_seed)
+
+    if output_feature["type"] == CATEGORY:
+        dataset = get_repeatable_train_val_test_split(
+            dataset,
+            stratify_colname=output_feature["name"],
+            random_seed=default_random_seed,
+        )
```
justinxzhao, line 40:
```suggestion
        return get_repeatable_train_val_test_split(
            dataset,
            stratify_colname=output_feature["name"],
            random_seed=default_random_seed,
        )
```

justinxzhao (approved):

PR #2027, opened 2022-05-12
feat: Added model type GBM (LightGBM tree learner), as an alternative to ECD

jppgks:
This PR introduces additional model type GBM (tree learner), as an alternative to ECD.

Limitations:
- supports only single output feature of type binary, number or category

In scope:
- adding `model_type` to the Ludwig config
- adding `type` to the trainer section of the Ludwig config
- introducing an `BaseModel` class, implemented by both ECD and GBM
- moving trainers into a separate folder
- implementing GBM training functionality in `LightGBMTrainer` and `LightGBMRayTrainer`

Out of scope:
- support for non-trainable feature encoders (e.g. pretrained text embeddings) => follow-up PR

---

Todo:
- [x] update/add tests reflecting new `model_type` functionality
- [x] replace occurences of ECD with `AbstractModel`
- [x] add tests for new GBM functionality
- [x] finish GBM implementation

Help needed:
- [x] ~Alternative to having separate trainer registry per backend (local, ray)~ OK for now
- [x] How to correctly update the schema `model_type` and trainer `type` (Thanks @ksbrar!)

PR #2264, opened 2022-07-12
Fix: Don't skip saving the model if the save path already exists.

### LINKED ISSUE #2259: Incorrect predictions are produced
**Describe the bug**
After merge of PR #2027, incorrect predictions are generated.

**To Reproduce**

1. Checkout repo to commit aa0c63b
1. Run this simple linear regression model

```
import logging
from typing import Callable, Union, Dict, Any, Optional
import shutil

import pandas as pd

from sklearn.datasets import make_regression
import matplotlib.pyplot as plt

import torch
from torch import Tensor


# self test
if __name__ == "__main__":
    import numpy as np

    config = {
        "input_features": [
            {"type": "number", "name": "x00"},
        ],
        "output_features": [
            {
                "name": "y",
                "type": "number",
            },
        ],
        "trainer":{
            "epochs": 50,
            "learning_rate": 0.001,
            "optimizer": {"type": "sgd", "momentum": 0.9},
        },
    }

    # Generate some non-linear noisy training data
    n_points = 500
    np.random.seed(1919)
    x_train = np.linspace(-1, 1, n_points)[:, np.newaxis]
    y_train = x_train + 0.4 * x_train * np.random.randn(n_points)[:, np.newaxis]
    # plt.figure(1)
    # plt.scatter(x_train, y_train, alpha=0.2);
    # plt.xlabel('x')
    # plt.ylabel('y')
    # plt.savefig("raw2_data.png")
    df = pd.DataFrame(
        np.hstack([x_train.reshape(-1,1), y_train.reshape(-1,1)]),
        columns=["x00", "y"]
    )
    df.to_csv('data/train.csv', index=False)
    print(df.head(10))

    model = LudwigModel(
        config=config,
        logging_level=logging.WARN
    )

    shutil.rmtree("/tmp/results", ignore_errors=True)
    training_stats, preprocessed_data, results_dir = model.train(
        training_set="data/train.csv", output_directory="/tmp/results", random_seed=1919
    )

    for n, w in model.model.named_parameters():
        print(n, w)

    predictions = model.predict("data/train.csv")
    print(f"test data predictions head:\n{predictions[0].head(10)}")
    print(f"test data predictions head:\n{predictions[0].tail(10)}")

    plt.figure(4)
    plt.scatter(x_train, y_train, marker='.', alpha=0.2, label='data')
    y_preds = model.predict(df)
    y_hat = y_preds[0]["y_predictions"]
    plt.plot(x_train, y_hat, color='black', label='model $\mu$')
    plt.xlabel('x')
    plt.ylabel('y')
    plt.legend()
    plt.savefig("linear_regression_plot.png")


    print(model.model)

    from io import StringIO
    my_data = """x00
    -0.99
    0
    0.99
    """

    preds = model.predict(pd.read_csv(StringIO(my_data)))
    print(preds)


    print("all done")
```
This will produce this plot, which  is incorrect.
image

**Expected behavior**
Running the same program at the prior commit d0e8439  the plot looks like this, which is the expected result
image

**Screenshots**

- OS: MacOS 12.4 with Docker Desktop 4.6.0 
- Python version 3.8
- Ludwig version. 0.6.dev

justinxzhao:
Without this change, the model is only saved the first time `save()` is called (at the end of the first round of evaluation).

This is a bug that was introduced in #2027. This was not caught by any model loading unit tests like `tests/integration_tests/test_model_save_and_load.py` because the best saved weights are reloaded automatically at the end of `train()`, so there's never a discrepancy between the model returned from `train()` and models loaded from disk.

@jimthompson5802 will send a follow up PR that adds a test that catches this regression specifically.

Closes #2259

### DIFF
--- a/ludwig/models/ecd.py
+++ b/ludwig/models/ecd.py
@@ -15,7 +15,6 @@ from ludwig.models.base import BaseModel
 from ludwig.schema.utils import load_config_with_kwargs
 from ludwig.utils import output_feature_utils
 from ludwig.utils.data_utils import clear_data_cache
-from ludwig.utils.fs_utils import open_file, path_exists
 from ludwig.utils.torch_utils import get_torch_device
 
 
@@ -152,9 +151,7 @@ class ECD(BaseModel):
     def save(self, save_path):
         """Saves the model to the given path."""
         weights_save_path = os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME)
-        if not path_exists(weights_save_path):
-            with open_file(weights_save_path, "wb") as f:
-                torch.save(self.state_dict(), f)
+        torch.save(self.state_dict(), weights_save_path)
 
     def load(self, save_path):
         """Loads the model from the given path."""
--- a/tests/integration_tests/test_model_save_and_load.py
+++ b/tests/integration_tests/test_model_save_and_load.py
@@ -1,13 +1,16 @@
+import logging
 import os.path
 import random
 import tempfile
 
 import numpy as np
+import pandas as pd
 import torch
 
 from ludwig.api import LudwigModel
-from ludwig.constants import TRAINER
+from ludwig.constants import LOSS, NAME, TRAINER, TRAINING
 from ludwig.data.split import get_splitter
+from ludwig.modules.loss_modules import MSELoss
 from ludwig.utils.data_utils import read_csv
 from tests.integration_tests.utils import (
     audio_feature,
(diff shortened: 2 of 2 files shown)


tgaddair (approved):

jimthompson5802 (approved):
I confirmed that this fix addresses the issue I found.

jimthompson5802:
@justinxzhao   I added the test we talked about into `tests/integration_tests/test_model_save_and_load.py::test_model_weights_match_training`

I confirmed that this test will pick up the situation when the wrong weights are loaded from checkpoint.

Assuming we get a clean run in the CI test, this PR is good to go.

dantreiman (approved):

PR #2266, opened 2022-07-12
Run github checks for pushes and merges to *-stable.

justinxzhao:


dantreiman (approved):

dantreiman:
We should document that we must use dashes in branch names "-stable" (or support "_stable" also).

justinxzhao:
> We should document that we must use dashes in branch names "-stable" (or support "_stable" also).

I can include this detail when documenting the ludwig release process.

PR #2269, opened 2022-07-13
Encoder refactor

connor-mccorm:
PR to refactor Encoder/Decoder section of input/output features so that it is nested one level deeper. For example, an input feature that originally looked like this:
```
input_features:
      - name: image_feature_1
         type: image
         preprocessing:
              height: 7.5
              width: 7.5
              num_channels": 4
         encoder: resnet
         num_channels: 4
         dropout: 0.1
         resnet_size: 100
```
should now be defined like this:
```
input_features:
      - name: image_feature_1
         type: image
         preprocessing:
              height: 7.5
              width: 7.5
              num_channels": 4
         encoder: 
              type: resnet
              num_channels: 4
              dropout: 0.1
              resnet_size: 100
````

I have added some backwards compatibility logic and a test for it as well. Also since it is such a large PR, if you are an expert in a specific area of the codebase and don't have time to do the whole thing, feel free to just inject your expertise where it works best for you!

### CODE REVIEW

**Review by brightsparc — APPROVED** [review]
Huge effort in getting this done.  I'm not seeing any jumping out as inconsistent.  It seems we have type of `embed` being reused across various feature types, but assume this is correct as it's not a change form the previous code.  Hopefully the unit tests will be able to confirm the functionality is consistent once you get.a clear run.

**Comment by brightsparc** [comment]
Are these nesting changes being flowed down to the [hyperopt](https://ludwig-ai.github.io/ludwig-docs/0.5/getting_started/hyperopt/) settings when applied at the feature level?

Eg if we were to add `sparse` and `dense` choice for  hyper opt for [category features](https://ludwig-ai.github.io/ludwig-docs/0.4/configuration/features/category_features/) would it be like this:

```
input_features:
    - name: genres
      type: set
    - name: content_rating
      type: category
      encoder:
        type: dense
    - name: top_critic
      type: binary
    - name: runtime
      type: number
    - name: review_content
      type: text
      encoder: embed
output_features:
    - name: recommended
      type: binary
hyperopt:
  goal: maximize
  output_feature: recommended
  metric: accuracy
  split: validation
  parameters:
    review_content.embedding_size:
      space: choice
      categories: [128, 256]
    review_content.encoder: 
       type: 
         space: choice
         categories:[sparse, dense]
```

Or like this:

```
hyperopt:
  goal: maximize
  output_feature: recommended
  metric: accuracy
  split: validation
  parameters:
    review_content.embedding_size:
      space: choice
      categories: [128, 256]
    review_content.encoder.type: 
         space: choice
         categories:[sparse, dense]
```

**Comment by connor-mccorm** [comment]
It would be the latter actually. Was actually testing this out today specifically in `test_hyperopt_with_shared_params`.

**Comment by connor-mccorm** [comment]
Apologies for the delay, found an issue with backwards compatibility so went back to fix that. Should be good to review now.

File: ludwig/constants.py [hunk 1]
```diff
@@ -205,6 +206,34 @@
     "num_filter_bands",
 ]
 
+INPUT_FEATURE_KEYS = [
```
dantreiman, line 209: AFAICT these constants are only used in `backward_compatibility.py`, maybe it would be simpler to define them there.

ludwig/decoders/generic_decoders.py
```diff
@@ -37,6 +38,10 @@ def __init__(self, input_size: int = 1, num_classes: int = None, **kwargs):
     def forward(self, inputs, **kwargs):
         return inputs
 
+    @staticmethod
```
dantreiman, line 41: Just brainstorming here: maybe `@register_encoder` and `@register_decoder` could take the schema class as an optional argument, and the registry could provide an accessor to get the schema for specified encoder/decoder type.

Or alternatively we could register the schema class `@register_encoder_schema`? 

To me, something like:
```
get_encoder_schema(feature_type, encoder)
```

is more readable than
```
get_encoder_cls(feature_type, encoder).get_schema_cls()
```

ludwig/utils/misc_utils.py
```diff
@@ -100,7 +100,14 @@ def set_default_value(dictionary, key, value):
 def set_default_values(dictionary, default_value_dictionary):
     # Set multiple default values
     for key, value in default_value_dictionary.items():
-        set_default_value(dictionary, key, value)
+        if key not in dictionary:  # Event where the key is not in the dictionary yet
+            dictionary[key] = value
+        elif isinstance(value, dict) and not value:  # Event where dict is empty
```
dantreiman, line 105:
```suggestion
        elif value == {}:  # Case where value is an empty dict
```

ludwig/utils/misc_utils.py
```diff
@@ -100,7 +100,14 @@ def set_default_value(dictionary, key, value):
...
+            set_default_value(dictionary, key, value)
+        elif isinstance(value, dict) and bool(value):  # Event where dictionary is nested - recursive call
```
dantreiman, line 107: I think this works without the explicit cast to bool... though if I'm missing something pls leave a code comment explaining this.
```suggestion
        elif isinstance(value, dict) and value:  # Event where dictionary is nested - recursive call
```

dantreiman (approved):
Overall this is great!  LGTM, Added a few minor comments

ludwig/constants.py
connor-mccorm: Ya that's probably a good call. Will move them now.

ludwig/decoders/generic_decoders.py
connor-mccorm: This is a fair point. I would personally go for the first option in an effort to avoid more siloed registries. Will make the change now.

ludwig/utils/misc_utils.py
connor-mccorm: Changed!

ludwig/utils/misc_utils.py
connor-mccorm: I believe you are correct. Making the change now.

ludwig/decoders/generic_decoders.py
connor-mccorm: After thinking a bit more, one thing to consider is that the method `get_schema_cls()` is used in input_features, output_features, combiners, encoders, and decoders. If I change it for the encoders, I would want to change it for the rest of the modules to remain consistent. This won't be really tricky but before I do this, just want to make sure that this is the way we want to do this. Any thoughts @dantreiman? If anyone else has thoughts on this as well, I'd love to hear them.

connor-mccorm: Thanks for the comprehensive review. I appreciate the diligence:)

ludwig/utils/defaults.py
```diff
@@ -223,7 +223,7 @@ def update_feature_from_defaults(config: Dict[str, Any], feature_dict: Dict[str,
         default_encoder_or_decoder = get_default_encoder_or_decoder(feature_dict, config_feature_group)
         if default_params_for_feature_type[TYPE] != default_encoder_or_decoder:
             # Update type and populate defaults for the encoder or decoder type
-            feature_dict[parameter] = default_params_for_feature_type[TYPE]
```
arnavgarg1, line 226: @connor-mccorm Are you able to try the test in this PR? https://github.com/ludwig-ai/ludwig/pull/2342

The new thing this is testing is if the encoder's additional dependencies (pretrained path, for e.g.) get injected into the input feature config. I have a feeling this will fail with the current implementation of this code block because with the new refactor, the `TYPE` check on line 221 will not be hit and it'll directly try line 233. The issue with that is the line 223 - 227 actually get the encoder or decoder class and then pull it any additional default parameters that are needed for that encoder or decoder class. The fix would just involve doing this step incase there's an encoder type specified in `default_params_for_feature_type`. 

Let me know if this makes sense!

connor-mccorm: Will merge and fix this up. Thanks for the additional test!

ludwig/hyperopt/execution.py
```diff
@@ -964,6 +966,18 @@ def update_features_with_shared_params(
...
+    # Remove encoder/decoder from output/input features
+    if config_feature_group == INPUT_FEATURES:
+        if DECODER in sampled_default_shared_params:
+            del shared_params_copy[DECODER]
+            sampled_default_shared_params = shared_params_copy
+    else:
+        if ENCODER in sampled_default_shared_params:
+            del shared_params_copy[ENCODER]
+            sampled_default_shared_params = shared_params_copy
+
```
arnavgarg1, line 980: nit - can refactor this to:

```
    if config_feature_group == INPUT_FEATURES:
        if DECODER in sampled_default_shared_params:
            del shared_params_copy[DECODER]
    else:
        if ENCODER in sampled_default_shared_params:
            del shared_params_copy[ENCODER]
    sampled_default_shared_params = shared_params_copy
```

ludwig/hyperopt/execution.py
```diff
@@ -1011,6 +1025,7 @@ def substitute_parameters(
         # Update or overwrite any feature specific hyperopt params
         update_section_dict(input_feature, input_feature[COLUMN], parameters_dict)
     for output_feature in config[OUTPUT_FEATURES]:
+        print("OUTPUT_FEATURE: \n", output_feature)
```
arnavgarg1, line 1028: This might be here by accident?

ludwig/hyperopt/run.py
```diff
@@ -441,11 +441,11 @@ def get_features_eligible_for_shared_params(
             raise ValueError("Ludwig expects feature types to be defined for each feature within the config.")
         if config_feature_type == INPUT_FEATURES:
             default_encoder = get_default_encoder_or_decoder(feature, INPUT_FEATURES)
```
arnavgarg1, line 443: @connor-mccorm I have a feeling this function might need modifications since it infers type directly from `feature.get('type')` instead of `feature.get('encoder').get('type')` - might just want to double check

ludwig/schema/decoders/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    class DecoderMarshmallowField(fields.Field):
+        """Custom marshmallow field that deserializes a dict for a valid decoder config from the decoder_registry
+        and creates a corresponding `oneOf` JSON schema for external usage."""
+
+        def _deserialize(self, value, attr, data, **kwargs):
+            if value is None:
+                return None
+            if isinstance(value, dict):
+                if TYPE in value and value[TYPE] in get_decoder_classes(feature_type):
+                    enc = get_decoder_cls(feature_type, default).get_schema_cls()
```
arnavgarg1, line 40: nit: should this be `dec` for decoder instead of `enc` for encoder?

ludwig/schema/preprocessing.py
```diff
@@ -177,6 +177,7 @@ class NumberPreprocessingConfig(schema_utils.BaseMarshmallowConfig):
 
 @register_preprocessor(BINARY)
 @dataclass
+@register_preprocessor(BINARY)
```
arnavgarg1, line 180: Not sure if `@register_preprocessor(BINARY)` is accidentally duplicated twice or supposed to be like this intentionally?

connor-mccorm: Ah yes this was a mistake, just found this in my current branch, will remove.

ludwig/utils/backward_compatibility.py
```diff
@@ -106,6 +135,67 @@ def _upgrade_feature(feature: Dict[str, Any]):
...
+        else:
+            feature[ENCODER] = encoder
+
+        for k in nested_params:
+            del feature[k]
+        warnings.warn(
+            "Encoder specific parameters should now be nested within a dictionary under the 'encoder' parameter. "
+            "Support for un-nested encoder specific parameters will be removed in v0.7",
+            DeprecationWarning,
+        )
```
arnavgarg1, line 171: @connor-mccorm Here's one way around the code duplication:

```
def _upgrade_nested_params(feature: Dict[str, Any], type: str) -> None:
    enc_or_dec = feature.get(type, {})
    if isinstance(enc_or_dec, str):
            encoder = {TYPE: enc_or_dec}
            feature[ENCODER] = encoder

        nested_params = []
        for k, v in feature.items():
            if k not in INPUT_FEATURE_KEYS:
                encoder[k] = v
                nested_params.append(k)

        if ENCODER in feature:
            feature[ENCODER].update(encoder)
        else:
            feature[ENCODER] = encoder

        for k in nested_params:
            del feature[k]
        warnings.warn(
            f"{type} specific parameters should now be nested within a dictionary under the '{type}' parameter. "
            f"Support for un-nested {type} specific parameters will be removed in v0.7",
            DeprecationWarning,
        )

def _upgrade_encoder_decoder_params(feature: Dict[str, Any], input_feature: bool):
    if input_feature:
        _upgrade_nested_params(feature, ENCODER)
   else:
       _upgrade_nested_params(feature, DECODER)
```

This might not be exactly right but something along these lines.

ludwig/utils/defaults.py
connor-mccorm: Just pulled, looks like it's passing with a couple of changes to the indexing logic.

ludwig/hyperopt/execution.py
connor-mccorm: Done!

ludwig/hyperopt/execution.py
connor-mccorm: Good catch, removed it!

ludwig/utils/defaults.py
arnavgarg1: @connor-mccorm that's awesome!

arnavgarg1:
Overall this is great - awesome work 👏 ! I've left a few minor comments on some potential refactoring but those are mostly nits.

ludwig/schema/decoders/utils.py
connor-mccorm: Yes good catch, copy and pasted the same logic so forgot to change it. Changing now!

ludwig/utils/backward_compatibility.py
connor-mccorm: Refactored this to remove the duplication. Could you possibly take a look and ok it?

ludwig/utils/backward_compatibility.py
```diff
@@ -106,6 +135,54 @@ def _upgrade_feature(feature: Dict[str, Any]):
...
+    if module_type in feature:
+        feature[module_type].update(module)
+    else:
+        feature[module_type] = module
+
+    for k in nested_params:
+        del feature[k]
+
+    if len(nested_params) > 0:
+        warn = True
```
arnavgarg1, line 176: nit - you can set warn=True right after line 165 (within the if statement) because you know you've added at least one item there already. That way you don't need this explicit check here

connor-mccorm: Right right, I thought of that but felt weird potentially setting it multiple times, I guess it's probably cleaner that way since it's less lines and we don't care if it sets it multiple times.

arnavgarg1 (approved):

ludwig/utils/backward_compatibility.py
arnavgarg1: @connor-mccorm Yeah I wasn't sure either, I was thinking the same thing too, but either way is fine really. Nbd at all

ludwig/utils/backward_compatibility.py
```diff
@@ -106,6 +135,52 @@ def _upgrade_feature(feature: Dict[str, Any]):
...
 
+def _upgrade_encoder_decoder_params(feature: Dict[str, Any], input_feature: bool) -> None:
+    """
+    This function nests un-nested encoder/decoder parameters to conform with the new config structure for 0.6
+    Args:
+        feature (Dict): Feature to nest encoder/decoder params for.
+        input_feature (Bool): Whether this feature is an input feature or not.
+
+    Returns:
+
```
justinxzhao, line 146: nit: Remove empty Return docs.

ludwig/utils/backward_compatibility.py
```diff
@@ -291,6 +366,12 @@ def upgrade_deprecated_fields(config: Dict[str, Any]):
     for feature in config.get(INPUT_FEATURES, []) + config.get(OUTPUT_FEATURES, []):
         _upgrade_feature(feature)
 
+    for feature in config.get(INPUT_FEATURES, []):
+        _upgrade_encoder_decoder_params(feature, True)
```
justinxzhao, line 370: nit: Use `_upgrade_encoder_decoder_params(feature, input_feature=True)` and `_upgrade_encoder_decoder_params(feature, input_feature=False)` so that readers don't have to look up `_upgrade_encoder_decoder_params` to figure out what that extra boolean is for.

ludwig/utils/misc_utils.py
```diff
@@ -100,7 +100,14 @@ def set_default_value(dictionary, key, value):
 def set_default_values(dictionary, default_value_dictionary):
```
justinxzhao, line 100: nit: Either in this PR (or add a TODO) to clarify what `set_default_values` does now, with a better docstring and additional tests that demonstrates it handling recursive/empty dictionaries.

justinxzhao (approved):
Thanks for the latest changes @connor-mccorm! LGTM except for minor nits.

jppgks:
Will we be replacing dictionary accesses with the schema class in a follow up PR?

jppgks (approved):
Great step towards a cleaner and more robust codebase. Thanks a lot @connor-mccorm!

ludwig/schema/utils.py
```diff
@@ -385,6 +386,116 @@ def FloatRange(
...
+            assert isinstance(default, int) or isinstance(default, tuple) or isinstance(default, list)
+            val(type(default))
+            if isinstance(default, tuple) or isinstance(default, list):
+                for i in default:
+                    assert isinstance(i, int)
+        except Exception:
+            raise ValidationError(f"Invalid default: `{default}`")
+    return field(
+        metadata={
+            "marshmallow_field": fields.Integer(
```
ksbrar, line 408: Does this work as intended? Prima facie it would seem this only works for the `int` case but not `tuple` or `list` (

ludwig/schema/utils.py
```diff
@@ -385,6 +386,116 @@ def FloatRange(
...
+                load_default=default,
+                dump_default=default,
+                metadata={"description": description},
+            )
+        },
+        default=default,
+    )
+
+
+def PositiveIntegerOrTupleOrStringOptions(
```
ksbrar, line 420: Nice!

ludwig/schema/utils.py
```diff
@@ -407,6 +518,40 @@ def Dict(default: Union[None, TDict] = None, description: str = "", parameter_me
     )
 
 
+def List(
```
ksbrar, line 521: Nice!

ksbrar (approved):
🚀🚀

Approved with a few nits. Awesome work!!

connor-mccorm:
@jppgks Yes, I am currently writing the config object which joins all the schemas together into the internal replacement for the config in a separate PR.

ludwig/utils/backward_compatibility.py
connor-mccorm: This is a very good callout, thanks Justin!

ludwig/schema/decoders/sequence_decoders.py
```diff
@@ -0,0 +1,92 @@
+from marshmallow_dataclass import dataclass
+
+from ludwig.schema import utils as schema_utils
+from ludwig.schema.decoders.base import BaseDecoderConfig
+
+
+@dataclass
+class SequenceGeneratorDecoderConfig(BaseDecoderConfig):
+
+    type: str = schema_utils.StringOptions(
```
ksbrar, line 10: One more nit, set `allow_none` to `False` for these fields

ksbrar (changes requested):

ludwig/schema/decoders/sequence_decoders.py
connor-mccorm: Ah true, will do that now.

ksbrar (approved):

ludwig/data/dataset_synthesizer.py
```diff
@@ -111,7 +113,12 @@ def return_none(feature):
 
 
 def assign_vocab(feature):
-    feature["idx2str"] = build_vocab(feature.get("vocab_size", 10))
+    if DECODER in feature:
+        feature["idx2str"] = build_vocab(feature[DECODER].get("vocab_size", 10))
+    elif ENCODER in feature:
+        feature["idx2str"] = build_vocab(feature[ENCODER].get("vocab_size", 10))
+    else:
+        feature["idx2str"] = build_vocab(10)
```
w4nderlust, line 121: this parameter is a parameter of the config for synthesizer. synthesizer configs shouldn't need encoder and decoder i believe

connor-mccorm: So in this instance, the `vocab_size` parameter has moved into the encoder or the decoder depending on if the feature is an input or output feature. So this logic is basically allowing the `assign_vocab` function to grab it from the respective nested dictionary since some features have a `decoder` key but not an `encoder` key and vice versa.

ludwig/features/bag_feature.py
```diff
@@ -94,16 +94,15 @@ def add_feature_data(
...
+    encoder = {TYPE: "embed", "vocab": []}
 
     def __init__(self, feature, encoder_obj=None):
         super().__init__(feature)
         self.overwrite_defaults(feature)
         if encoder_obj:
             self.encoder_obj = encoder_obj
         else:
-            self.encoder_obj = self.initialize_encoder(feature)
+            self.encoder_obj = self.initialize_encoder()
```
w4nderlust, line 105: how are parameters passed to the necoder initilization if not through feature?

connor-mccorm: So they are all saved on the object now which is fully populated with the user specified parameters via line 101 in `self.overwrite_defaults(feature)`

w4nderlust: I believe they are accessed by self.encoder, which is not actually an encoder but an encoder config. In general not a huge fan of functions that refer to internal class state instead of passing the parameters explicitly, as those functions may fail then that internal state is not set

w4nderlust: I'm seeing the same pattern is true for decoder, so we would want to consider both if we want to make changes to this

w4nderlust: also do we actually need to keep araound the encoder end decoder configs in self after the encoder and decoder objects are initialized?

ludwig/features/audio_feature.py
```diff
@@ -452,15 +453,13 @@ def _get_max_length_feature(preprocessing_parameters, sampling_rate_in_hz, audio
 
 @register_input_feature(AUDIO)
 class AudioInputFeature(AudioFeatureMixin, SequenceInputFeature):
-    encoder = "parallel_cnn"
-    max_sequence_length = None
-    embedding_size = None
+    encoder = {TYPE: "parallel_cnn", "max_sequence_length": None, "embedding_size": None}
```
tgaddair, line 456: In a follow-up PR, we'll want to change the encoder here from a dict to something using the schema.

tgaddair (changes requested):
Lots of good work here! The biggest issue I see, however, is that we're not taking advantage of the schema in this refactor. Instead, there are many occurrences of code like:

```
self.encoder["weight"]
```

When if we were using the schema objects, we could instead write:

```
self.encoder.weight
```

This would be more type safe and easier to debug. I think it is worth making this change (converting the user config dict to schema objects BEFORE using them in the Feature classes), given how prominent this pattern of using encoders / decoders as dicts is here.

ludwig/features/bag_feature.py
tgaddair: I agree, and would rather us init the encoder / decoder from the config before we even get to the input / output features. This is one of the major reasons for the move to schema objects.

ludwig/features/bag_feature.py
```diff
@@ -94,16 +94,15 @@ def add_feature_data(
 
 @register_input_feature(BAG)
 class BagInputFeature(BagFeatureMixin, InputFeature):
-    encoder = "embed"
-    vocab = []
+    encoder = {TYPE: "embed", "vocab": []}
```
tgaddair, line 97: I don't like the presence of this encoder dict in general, in part for reasons described elsewhere, but also because it mixes config with state. `vocab` is state that needs to be tracked by the feature (possibly within the encoder object). The encoder type chosen by the user is config, and should ideally live as part of the schema as well (we need to make sure changing this default doesn't break old models).

justinxzhao: With the full encoder defaults defined in the feature schemas, are these class-global features still needed?

connor-mccorm: No I am going to remove most of them, the ones that seemingly need to stay for now are `metric_functions` and `default_validation_metric` since the parent class OutputFeature has them as abstract methods.

ludwig/features/bag_feature.py
dantreiman: We don't need to retain the encoder configs in the feature after instantiating the encoder, but there are enough places in the code which reference them I'd recommend removing them in a follow-up.

jppgks:
Can this be closed now that #2370 is merged? @connor-mccorm @dantreiman

connor-mccorm:
Indeed it can! @jppgks

PR #2273, opened 2022-07-13
fix: Make `BaseTrainerConfig` an abstract class

ksbrar:
A user should not be able to directly instantiate this class.

### DIFF
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -1,5 +1,6 @@
 from typing import List, Optional, Union
 
+from abc import ABCMeta
 from marshmallow_dataclass import dataclass
 
 from ludwig.constants import COMBINED, LOSS, MODEL_ECD, MODEL_GBM, TRAINING, TYPE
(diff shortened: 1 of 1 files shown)


ludwig/schema/trainer.py
```diff
@@ -25,7 +26,7 @@ def wrap(trainer_config: BaseTrainerConfig):
 
 
 @dataclass
-class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig):
+class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig, metaclass=ABCMeta):
```
justinxzhao, line 29: Why not just inherit from ABC?

```suggestion
class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig, ABC):
```

https://docs.python.org/3/library/abc.html#abc.ABC

ludwig/schema/trainer.py
```diff
@@ -1,3 +1,4 @@
+from abc import ABCMeta
```
justinxzhao, line 1:
```suggestion
from abc import ABC
```

ludwig/schema/trainer.py
ksbrar: good point!

connor-mccorm (approved):

justinxzhao (approved):

PR #2274, opened 2022-07-13
Use pytest tmpdir fixture wherever temporary directories are used in tests.

justinxzhao:


ludwig/schema/trainer.py
```diff
@@ -64,7 +64,11 @@ class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig):
         default=None,
         allow_none=True,
         min_exclusive=0,
-        description="Size of batch to pass to the model for evaluation.",
+        description=(
```
tgaddair, line 67: We should also mention `auto`, which will set the eval batch size to maximize GPU bandwidth.

justinxzhao: Done.

dantreiman (approved):
LGTM

PR #2280, opened 2022-07-14
Save and load pre-trained encoders (transfer learning support)

dantreiman:
- Refactors LudwigModule into its own file `ludwig_module.py` (from `torch_utils.py`)
    - Adds `LudwigModuleState` dataclass, which is a POD representation any ludwig module.
    - Adds `get_state()` and `@classmethod restore_from_state()` methods forto`LudwigModule`
- Adds `serialization.py`, which reads and writes Ludwig modules to .h5 files.
- Implements serialization for text encoders
- Modifies input feature schema to allow encoder name **or** URL (`schema_utils.StringOptionsOrURL`)
- Load pretrained encoders before preprocessing, re-use preprocessing data (vocab) from pretrained
- [Temporarily] passes all encoder parameters to superclass constructor to save in a dictionary.  This code will be unnecessary once https://github.com/ludwig-ai/ludwig/pull/2269 is in.
- Adds `serialization_test.py` with transfer learning example in `test_transfer_learning`

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -55,6 +55,7 @@ from ludwig.data.split import get_splitter, split_dataset
 from ludwig.encoders.registry import get_encoder_cls
 from ludwig.features.feature_registries import base_type_registry
 from ludwig.features.feature_utils import compute_feature_hash
+from ludwig.modules import serialization
 from ludwig.utils import data_utils, strings_utils
 from ludwig.utils.backward_compatibility import upgrade_metadata
 from ludwig.utils.config_utils import merge_config_preprocessing_with_feature_specific_defaults
@@ -101,7 +102,7 @@ from ludwig.utils.data_utils import (
     use_credentials,
 )
 from ludwig.utils.defaults import default_preprocessing_parameters, default_random_seed
-from ludwig.utils.fs_utils import file_lock, path_exists
+from ludwig.utils.fs_utils import file_lock, is_url, path_exists
 from ludwig.utils.misc_utils import get_from_registry, merge_dict, resolve_pointers
 from ludwig.utils.types import DataFrame, Series
 
@@ -1274,14 +1275,18 @@ def build_preprocessing_parameters(
 
         # deal with encoders that have fixed preprocessing
         if ENCODER in feature_config:
-            if TYPE in feature_config[ENCODER]:
+            encoder_fixed_parameters = {}
+            if URL in feature_config[ENCODER]:
+                encoder_state = serialization.load_state_from_file(feature_config["encoder"])
+                encoder_fixed_parameters = encoder_state.metadata
+            elif TYPE in feature_config[ENCODER]:
                 encoder_class = get_encoder_cls(feature_config[TYPE], feature_config[ENCODER][TYPE])
                 if hasattr(encoder_class, "fixed_preprocessing_parameters"):
-                    encoder_fpp = encoder_class.fixed_preprocessing_parameters
+                    encoder_fixed_parameters = encoder_class.fixed_preprocessing_parameters
 
-                    preprocessing_parameters = merge_dict(
-                        preprocessing_parameters, resolve_pointers(encoder_fpp, feature_config, "feature.")
-                    )
+            preprocessing_parameters = merge_dict(
+                preprocessing_parameters, resolve_pointers(encoder_fixed_parameters, feature_config, "feature.")
+            )
 
         fill_value = precompute_fill_value(dataset_cols, feature_config, preprocessing_parameters, backend)
 
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1330,6 +1330,7 @@ def build_data(
 ) -> Dict[str, DataFrame]:
     """Preprocesses the input dataframe columns, handles missing values, and potentially adds metadata to
     training_set_metadata.
+
     Args:
         input_cols: Input dataframe to be processed.
         feature_configs: List of feature configs.
@@ -1362,6 +1363,7 @@ def build_data(
 def balance_data(dataset_df: DataFrame, output_features: List[Dict], preprocessing_parameters: Dict, backend: Backend):
     """The purpose of this function is to balance the training dataset using either over-sampling or under-
     sampling.
+
     Args:
         dataset_df: Input dataframe to be over-sampled or under-sampled.
         output_features: List of feature configs.
@@ -1406,6 +1408,7 @@ def balance_data(dataset_df: DataFrame, output_features: List[Dict], preprocessi
 
 def precompute_fill_value(dataset_cols, feature, preprocessing_parameters, backend):
     """Precomputes the fill value for a feature.
+
     NOTE: this is called before NaNs are removed from the dataset. Modifications here must handle NaNs gracefully.
     NOTE: this is called before columns are cast. Modifications here must handle dtype conversion gracefully.
     """
@@ -1681,6 +1684,7 @@ def _preprocess_file_for_training(
     callbacks=None,
 ):
     """Method to pre-process csv data.
+
     :param features: list of all features (input + output)
     :param dataset: path to the data
     :param training_set:  training data
@@ -1783,6 +1787,7 @@ def _preprocess_df_for_training(
     callbacks=None,
 ):
     """Method to pre-process dataframes.
+
     This doesn't have the option to save the processed data as hdf5 as we don't expect users to do this as the data can
     be processed in memory
     """
@@ -1843,6 +1848,7 @@ def preprocess_for_prediction(
     callbacks=None,
 ):
     """Preprocesses the dataset to parse it into a format that is usable by the Ludwig core.
+
     :param model_path: The input data that is joined with the model
            hyperparameter file to create the config file
     :param data_csv: The CSV input data file
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1278,14 +1278,21 @@ def build_preprocessing_parameters(
 
         # deal with encoders that have fixed preprocessing
         if ENCODER in feature_config:
-            if TYPE in feature_config[ENCODER]:
-                encoder_class = get_encoder_cls(feature_config[TYPE], feature_config[ENCODER][TYPE])
+            encoder_config = feature_config[ENCODER]
+            encoder_fixed_parameters = {}
+            if "pretrained_encoder" in encoder_config and is_url(encoder_config["pretrained_encoder"]):
+                # Since we are loading a pre-trained encoder, fix its preprocessing parameters so that we don't infer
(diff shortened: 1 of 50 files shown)


tests/ludwig/modules/test_serialization.py
```diff
@@ -0,0 +1,164 @@
...
+        "output_features": output_features,
+        "combiner": {"type": "concat", "output_size": 14},
+    }
+    model1 = LudwigModel(model1_config)
+    train_stats1, _, _ = model1.train(dataset=data_csv, output_directory=os.path.join(tmpdir, "model1"))
+    # Saves pre-trained encoder to file.
+    trained_input_feature = model1.model.input_features[text_input_name]
+    input_feature_encoder = trained_input_feature.encoder_obj
+    saved_path = os.path.join(tmpdir, "text_encoder.h5")
+    # Note: I'd like to make the preprocessing metadata a property of the encoder, instead of having to get it from
```
justinxzhao, line 128: I'm in favor of this change. The encoder config should have everything that is needed for the encoder to be properly initialized, which may include computed preprocessing data.

tests/ludwig/modules/test_serialization.py
```diff
@@ -0,0 +1,164 @@
...
+    output_features = [category_feature(vocab_size=5, reduce_input="sum")]
+    data_csv = generate_data(input_features, output_features, os.path.join(tmpdir, "dataset.csv"))
+    config = {
+        "input_features": input_features,
+        "output_features": output_features,
+        "combiner": {"type": "concat", "output_size": 14},
+    }
+    model_1 = LudwigModel(config)
+    model_1.train(dataset=data_csv, output_directory=tmpdir)
+    # Gets pre-trained encoder.
```
justinxzhao, line 56: nits: Here and in other tests, consider adding whitespace to break up the tests by [Arrange-Act-Assert](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/)

ludwig/modules/fully_connected_modules.py
```diff
@@ -18,11 +18,39 @@
...
 from torch.nn import BatchNorm1d, BatchNorm2d, Dropout, LayerNorm, Linear, ModuleList
 
-from ludwig.utils.torch_utils import activations, initializer_registry, LudwigModule
+from ludwig.modules.ludwig_module import LudwigModule
+from ludwig.utils.torch_utils import activations, initializer_registry
 
 logger = logging.getLogger(__name__)
 
 
+class Dense(LudwigModule):
```
justinxzhao, line 27: nit: Add a docstring.

ludwig/modules/ludwig_module.py
```diff
@@ -0,0 +1,123 @@
...
+    type: str  # Module Type
+    ludwig_version: str  # Version of ludwig which saved this object
+    config: Dict  # Module Config
+    metadata: Dict  # Preprocessing data (same a straining_set_metadata.json)
+    saved_weights: Dict[str, np.ndarray]  # Saved weights of this module
+    children: Dict[str, "LudwigModuleState"]  # Child modules
+
+
+class LudwigModule(Module):
+    def __init__(self):
```
justinxzhao, line 54: nit: Add a docstring.

ludwig/schema/utils.py
```diff
@@ -51,6 +67,15 @@ def load_config_with_kwargs(
     }
 
 
+def create_anyof_cond(if_pred: TDict, then_preds: TList[TDict]):
```
justinxzhao, line 70: Is this used anywhere? If not, remove.

ludwig/schema/features/utils.py
```diff
@@ -35,8 +35,18 @@ def update_encoders(feature_props, feature_type):
         None
     """
     for key in get_encoder_classes(feature_type):
-        if key not in feature_props["encoder"]["enum"]:
-            feature_props["encoder"]["enum"].append(key)
+        encoder_props = feature_props["encoder"]
```
justinxzhao, line 38: This looks non-trivial. Can we add a unit test for `update_encoders`?

ludwig/modules/serialization.py
```diff
@@ -0,0 +1,149 @@
...
+from ludwig.modules.ludwig_module import LudwigModule, LudwigModuleState, module_registry
+from ludwig.utils.fs_utils import download_h5, upload_h5
+
+logger = logging.getLogger(__name__)
+
+
+class NumpyEncoder(json.JSONEncoder):
+    """Python json library does not support serialization of numpy types.
+
+    This custom encoder converts numpy scalars to python types
```
justinxzhao, line 33: nit: .

ludwig/modules/serialization.py
```diff
@@ -0,0 +1,149 @@
...
+        type=group.attrs["type"],
+        ludwig_version=group.attrs["ludwig_version"],
+        config=json.loads(group.attrs["config"]),
+        metadata=json.loads(group.attrs["metadata"]),
+        saved_weights={k: v[()] for k, v in group.items() if isinstance(v, h5py.Dataset)},
+        children={k: _load_state_from_group(v) for k, v in group.items() if isinstance(v, h5py.Group)},
+    )
+
+
+@lru_cache(maxsize=3)
```
justinxzhao, line 99: Curious -- how did you determine 3 as the maxsize to use?

ludwig/modules/serialization.py
```diff
@@ -0,0 +1,149 @@
...
+    state = load_state_from_file(path_or_uri)
+    return instantiate_module_from_state(state, device=device)
+
+
+def update_state_for_current_version(state: LudwigModuleState) -> LudwigModuleState:
+    """Update saved Ludwig object from previous version."""
+    if state == LUDWIG_VERSION:
+        return state
+    else:
+        # TODO: check version, apply updates if needed.
```
justinxzhao, line 147: We certainly don't need to solve this now, though one proposal for how this could be implemented like for if/when parameters are renamed:

In ParameterMetadata, we can add a field i.e, `backwards_compatibility_source`, that contains tuple of older Ludwig version and which field that a module parameter should try to source its value from, if it's not explicitly in the config, as well as whether loading should crash if loading is not successful.

Hypothetical example: `output_size`'s metadata: `backwards_compatibility_source: [("<0.4", "fc_size", True)]`

justinxzhao (commented):
Thanks for the PR! Overall, looks solid to me. I've left a few comments, mostly nits and discussion.

ludwig/schema/utils.py
dantreiman: Not used anymore, good call.

ludwig/modules/serialization.py
dantreiman: No particular reason.

ludwig/schema/features/utils.py
dantreiman: Will do in a separate PR: https://linear.app/predibase/issue/MLQ-38

tests/ludwig/modules/test_serialization.py
dantreiman: We're agreed on that.  Will do metadata refactor in a follow-up PR - I have an open linear issue for this.

ludwig/modules/serialization.py
dantreiman: To add on to that, I'd like to add an option for a general hook to register a transformation (for a specific version and section of the config).  My plan is to implement that for regular configs (refactoring `backwards_compatibility.py`), then re-using that same system here in `serialization.py`.

ludwig/schema/features/image_feature.py
```diff
@@ -15,7 +15,7 @@ class ImageInputFeatureConfig(BaseInputFeatureConfig):
 
     preprocessing: BasePreprocessingConfig = PreprocessingDataclassField(feature_type=IMAGE)
 
-    encoder: Optional[str] = schema_utils.StringOptions(
+    encoder: Optional[str] = schema_utils.StringOptionsOrURL(
```
connor-mccorm, line 18: I believe this will also need to be refactored once encoder refactor lands since there is a nested structure now. Will probably be as simple as just moving this option to the corresponding parameter in the encoders, however this may have downstream effects in the rest of the PR.

ludwig/schema/features/text_feature.py
```diff
@@ -16,7 +16,7 @@ class TextInputFeatureConfig(BaseInputFeatureConfig):
 
     preprocessing: BasePreprocessingConfig = PreprocessingDataclassField(feature_type=TEXT)
 
-    encoder: Optional[str] = schema_utils.StringOptions(
+    encoder: Optional[str] = schema_utils.StringOptionsOrURL(
```
connor-mccorm, line 19: Same comment as above, going to need to be refactored after encoder PR lands.

ludwig/schema/features/utils.py
```diff
@@ -35,8 +35,18 @@ def update_encoders(feature_props, feature_type):
         None
     """
     for key in get_encoder_classes(feature_type):
```
connor-mccorm, line 37: So I actually removed this in the Encoder Refactor PR since I was able to change the import structure to support all registrations. I can see an argument for keeping this just in case as a fail safe so I'm ok keeping this in though I believe it may be unnecessary.

connor-mccorm (commented):
Great work on this, code looks real good. There are a few things that need to be changed when my encoder refactor PR goes in this week, but shouldn't be too tricky to make the changes. Let me know if you have any questions when doing these refactors and I'll be happy to help!

ludwig/schema/features/utils.py
dantreiman: SGTM.

ludwig/schema/features/image_feature.py
dantreiman: Good point.  I'm going to get rid of StringOptionsOrURL and just use a String field, similar to `pretrained_model: str = schema_utils.String(`

ludwig/data/preprocessing.py
```diff
@@ -1238,13 +1239,20 @@ def build_preprocessing_parameters(
...
         if ENCODER in feature_config:
-            encoder_class = get_encoder_cls(feature_config[TYPE], feature_config["encoder"])
-            if hasattr(encoder_class, "fixed_preprocessing_parameters"):
-                encoder_fpp = encoder_class.fixed_preprocessing_parameters
-
-                preprocessing_parameters = merge_dict(
-                    preprocessing_parameters, resolve_pointers(encoder_fpp, feature_config, "feature.")
-                )
+            encoder_fixed_parameters = {}
+            if is_url(feature_config["encoder"]):
```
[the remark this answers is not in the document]
dantreiman: Hmmmm, I'd like to keep the semantics of encoder type the same, so I think I need a new field.  How about `pretrained_encoder:` or `pretrained_encoder_url:`, like:
```
input_features:
  - name: image
    type: image
    encoder:
      pretrained_encoder_url: file://pretrained_encoder.h5
```

PR #2284, opened 2022-07-15
Fixes #2279

### LINKED ISSUE #2279: "Error parsing date..."
**Describe the bug**

I'm attempting to train a model from data in a large JSON array.  An example of the relevant slice looks like:

```json
[
  {
    "date": "1941-03-01",
    "foo": "foo"
    "bar": "bar"
  },
  {
    "date": "2001-05-15",
    "foo": "foo"
    "bar": "bar"
  }
]
```

When I run the following command:

`ludwig train --dataset input.json --data_format json -c config.yaml`

After a few moments, I get the following output for every element of the array:

```
Error parsing date: 1941-03-01 00:00:00 with error strptime() argument 1 must be str, not Timestamp Please provide a datetime format that parses it in the preprocessing section of the date feature in the config. The preprocessing fill in value will be used.For more details: https://ludwig-ai.github.io/ludwig-docs/0.5/configuration/features/date_features/

Error parsing date: 2001-05-15 00:00:00 with error strptime() argument 1 must be str, not Timestamp Please provide a datetime format that parses it in the preprocessing section of the date feature in the config. The preprocessing fill in value will be used.For more details: https://ludwig-ai.github.io/ludwig-docs/0.5/configuration/features/date_features/
```

I've tried this with several config options in `config.yaml`, including each of the following:

```yaml
input_features:
  - name: date
    type: date
    preprocessing:
      datetime_format: "%Y-%m-%d %H:%M:%S"
```

```yaml
input_features:
  - name: date
    type: date
    preprocessing:
      datetime_format: "%Y-%m-%d"
```

```yaml
input_features:
  - name: date
    type: date
```

All 3 result in the same error message.

It seems to me what's happening is that the date string is being parsed too soon in the process -- it's being turned into a Timestamp object BEFORE being fed into strptime(), hence the error.

I'm not familiar enough with the codebase to know if this is a bug, an issue with JSON as the input source, or something else, but some guidance would be appreciated!

**Expected behavior**
The date should parse according to the provided format.

- OS: MacOS
- Version 12.3.1
- Python version 3.10.5
- Ludwig version 0.5.4

noahlh:
This PR (temporarily) fixes a case where `pandas.read_json` parses a date feature as a `datetime` object prior to feature data building.  Test case included and passing.  As discussed in #2279.

### DIFF
--- a/ludwig/features/date_feature.py
+++ b/ludwig/features/date_feature.py
@@ -68,7 +68,9 @@ class DateFeatureMixin(BaseFeatureMixin):
     @staticmethod
     def date_to_list(date_str, datetime_format, preprocessing_parameters):
         try:
-            if datetime_format is not None:
+            if isinstance(date_str, datetime): 
+                datetime_obj = date_str
+            elif datetime_format is not None:
                 datetime_obj = datetime.strptime(date_str, datetime_format)
             else:
                 datetime_obj = parse(date_str)
--- a/ludwig/features/date_feature.py
+++ b/ludwig/features/date_feature.py
@@ -68,7 +68,7 @@ class DateFeatureMixin(BaseFeatureMixin):
     @staticmethod
     def date_to_list(date_str, datetime_format, preprocessing_parameters):
         try:
-            if isinstance(date_str, datetime): 
+            if isinstance(date_str, datetime):
                 datetime_obj = date_str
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):
Thanks for the change!

PR #2286, opened 2022-07-15
Fixed default eval_batch_size when setting batch_size=auto

tgaddair:


ludwig/api.py
```diff
@@ -524,7 +525,7 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                         self.config[TRAINER][BATCH_SIZE] = tuned_batch_size
                         trainer.batch_size = tuned_batch_size
 
-                    if self.config[TRAINER][EVAL_BATCH_SIZE] == AUTO:
+                    if self.config[TRAINER][EVAL_BATCH_SIZE] in {AUTO, None}:
```
justinxzhao, line 528: Does this mean we can get rid of [this extra condition](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/trainers/trainer.py#L144)?

justinxzhao (approved):

ludwig/api.py
tgaddair: Not quite, as the above code path is only executed when `batch_size=auto`

PR #2293, opened 2022-07-20
Use ray dataset and drop type casting in binary_feature prediction post processing for speedup

magdyksaleh:
# Code Pull Requests

Update the binary feature post-processing to use ray datasets to speed up overall time 

# Documentation Pull Requests


tgaddair (commented):
@magdyksaleh I'm going to suggest a different approach here, since I suspect the issue we're seeing here with `binary_feature` will be an issue for other output feature types as well.

The key idea is that all of the `postprocess_predictions` functions for all output features are only applying per-row transformations. There are not transformations being applied that require computing column-level metadata in this step. As such, instead of passing Dask DFs to the postprocess functions, we can run one top-level `map_batches` call that applies the transformations to batches as pandas dataframes.

Concretely:

1. Change the `postprocess` function in `postprocessing.py` to something like this:

```
def posproc_preds(df):
            return output_feature.postprocess_predictions(
                df,
                training_set_metadata[of_name],
            )
    
    predictions = backend.df_engine.map_batches(predictions, posproc_preds)
```

2. Change every `postprocess_predictions` function by removing `output_directory` and `backend` from the signature.

3. Change all the calls to `backend.df_engine.map_objects` in `postprocess_predictions` to simply call `df.map`, since it's just a Pandas DF.

ludwig/data/dataframe/base.py
```diff
@@ -46,6 +46,10 @@ def map_objects(self, series, map_fn, meta=None):
     def map_partitions(self, series, map_fn, meta=None):
         raise NotImplementedError()
 
+    @abstractmethod
+    def try_map_batches(self, series, map_fn, batch_format="pandas", meta=None):
```
tgaddair, line 50: I would rename to just `map_batches`, and maybe drop the `batch_format` param, since it would be difficult to make it work with other backends.

magdyksaleh:
> @magdyksaleh I'm going to suggest a different approach here, since I suspect the issue we're seeing here with `binary_feature` will be an issue for other output feature types as well.
> 
> The key idea is that all of the `postprocess_predictions` functions for all output features are only applying per-row transformations. There are not transformations being applied that require computing column-level metadata in this step. As such, instead of passing Dask DFs to the postprocess functions, we can run one top-level `map_batches` call that applies the transformations to batches as pandas dataframes.
> 
> Concretely:
> 
>     1. Change the `postprocess` function in `postprocessing.py` to something like this:
> 
> 
> ```
> def posproc_preds(df):
>         for of_name, output_feature in output_features.items():
>             return output_feature.postprocess_predictions(
>                 df,
>                 training_set_metadata[of_name],
>             )
>     
>     predictions = backend.df_engine.map_batches(predictions, posproc_preds)
> ```
> 
>     2. Change every `postprocess_predictions` function by removing `output_directory` and `backend` from the signature.
> 
>     3. Change all the calls to `backend.df_engine.map_objects` in `postprocess_predictions` to simply call `df.map`, since it's just a Pandas DF.

I really like this suggestion! I'll try it out now. Thanks travis

ludwig/data/postprocessing.py
```diff
@@ -45,13 +45,17 @@ def postprocess(
...
-            training_set_metadata[of_name],
-            output_directory=output_directory,
-            backend=backend,
-        )
+    def postprocess_batch(df):
+        for of_name, output_feature in output_features.items():
+            df = output_feature.postprocess_predictions(
+                df,
+                training_set_metadata[of_name],
+                output_directory=output_directory,
```
tgaddair, line 53: Remove `output_directory` and `backend`

ludwig/data/dataframe/base.py
```diff
@@ -46,6 +46,10 @@ def map_objects(self, series, map_fn, meta=None):
     def map_partitions(self, series, map_fn, meta=None):
         raise NotImplementedError()
 
+    @abstractmethod
+    def map_batches(self, series, map_fn):
```
tgaddair, line 50: First element should be a dataframe, not a series.

tgaddair (approved):
LGTM! Please also double-check performance on higgs with this new approach before landing.

magdyksaleh:
This is ready to merge pending CI passing

PR #2294, opened 2022-07-20
catching urllib3 exception

abidwael:
Changed exception catching in `ludwig.utils.fs_utils.get_bytes_obj_from_path` from `except requests.exceptions.RequestException as e` to `except Exception as e`.

Used `Exception` instead of a `urllib3` specific base exception because `urllib3` doesn't have one.


geoffreyangus:
LGTM. Is there a test we can add to ensure that this behaves as expected?

justinxzhao (approved):

PR #2310, opened 2022-07-25
Ensure unit tests are using `tmpdir`

geoffreyangus:
Updated several unit tests to ensure that Ludwig models save their artifacts into the pytest fixture `tmpdir`, in order to ensure cleanup and relieve memory pressure overall during tests.


justinxzhao (approved):
Thanks for the change!

PR #2318, opened 2022-07-27
Custom to_dask() implementation to include meta during Dask DataFrame creation

arnavgarg1:
Ray's [`to_dask()`](https://docs.ray.io/en/latest/_modules/ray/data/dataset.html#Dataset.to_dask) function is called when predictions are being generated, but the Ray implementation doesn't pass in any meta related information into `dd.from_delayed`. I believe that in some instances (small dataset with a large number of partitions, empty partitions, etc.), there isn't enough data in the first Dask partition for Dask to infer the meta correctly (or at all). 

This PR updates `from_ray_dataset` within our DaskEngine to use a custom implementation that creates meta information from the first 100 rows of the ray dataset and uses it during the creation of the Dask DataFrame from the Ray Dataset. Dask won't need to infer the meta information anymore for the dataframe, guaranteeing fewer problems in downstream transformations. This gives us more downstream control.

The hack has been added because in Ray 1.12 and 1.13, the resulting column from `read_binary_files` was cast into type object for being passed into Dask, but in Ray nightly, they've started casting the column to type TensorDtype (thanks for the catch @geoffreyangus). This change is also needed for https://github.com/ludwig-ai/ludwig/pull/2241 to pass the tests for Ray nightly.

### DIFF
--- a/ludwig/data/dataframe/dask.py
+++ b/ludwig/data/dataframe/dask.py
@@ -24,6 +24,7 @@ import ray
 import ray.data
 from dask.diagnostics import ProgressBar
 from ray.data.block import Block, BlockAccessor
+from ray.data.extensions import TensorDtype
 from ray.util.client.common import ClientObjectRef
 from ray.util.dask import ray_dask_get
 
@@ -31,7 +32,7 @@ from ludwig.data.dataframe.base import DataFrameEngine
 from ludwig.utils.data_utils import split_by_slices
 
 TMP_COLUMN = "__TMP_COLUMN__"
-
+META_SAMPLE_SIZE = 100
 
 logger = logging.getLogger(__name__)
 
@@ -145,9 +146,10 @@ class DaskEngine(DataFrameEngine):
                 )
             return block.to_pandas()
 
-        # Use first few row from ray dataset to generate meta to infer types even if there are NaNs
-        meta = dataset.limit(100).to_pandas()
-        # meta = meta.dtypes.apply(lambda x: x.name).to_dict()
+        # Use a sample of the ray dataset to generate meta to infer types even if there are NaNs
+        meta = dataset.limit(META_SAMPLE_SIZE).to_pandas()
+        # HACK: Dask cannot handle TensorDtype yet, so we ensure those are of dtype object
+        meta = {k: object if isinstance(dtype, TensorDtype) else dtype for k, dtype in meta.dtypes.items()}
         ddf = dd.from_delayed([block_to_df(block) for block in dataset.get_internal_block_refs()], meta=meta)
         return ddf
 
--- a/ludwig/data/dataframe/dask.py
+++ b/ludwig/data/dataframe/dask.py
@@ -145,8 +145,9 @@ class DaskEngine(DataFrameEngine):
                 )
             return block.to_pandas()
 
-        # Use first row from ray dataset to generate meta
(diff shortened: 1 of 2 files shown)


ludwig/data/dataframe/dask.py
```diff
@@ -129,7 +133,25 @@ def to_ray_dataset(self, df):
...
+                raise ValueError(
+                    "Dataset.to_dask() must be used with Dask-on-Ray, please "
+                    "set the Dask scheduler to ray_dask_get (located in "
+                    "ray.util.dask)."
+                )
+            return block.to_pandas()
+
+        # Use a sample of the ray dataset to generate meta to infer types even if there are NaNs
+        meta = dataset.limit(META_SAMPLE_SIZE).to_pandas()
+        # HACK: Dask cannot handle TensorDtype yet, so we ensure those are of dtype object
```
justinxzhao, line 151: Is there a TODO/issue we can point to track when Dask will be able to handle TensorDtype?

ludwig/data/dataframe/dask.py
```diff
@@ -129,7 +133,25 @@ def to_ray_dataset(self, df):
...
+            if isinstance(block, (ray.ObjectRef, ClientObjectRef)):
+                raise ValueError(
+                    "Dataset.to_dask() must be used with Dask-on-Ray, please "
+                    "set the Dask scheduler to ray_dask_get (located in "
+                    "ray.util.dask)."
+                )
+            return block.to_pandas()
+
+        # Use a sample of the ray dataset to generate meta to infer types even if there are NaNs
+        meta = dataset.limit(META_SAMPLE_SIZE).to_pandas()
```
justinxzhao, line 150: This is also a significant change. Could you add a note about this in the docstring? Does this change any inferred types for existing datasets?

ludwig/data/dataframe/dask.py
arnavgarg1: @justinxzhao I just checked Dask's GitHub and I don't see any existing reported issues related to handling TensorDType. This check is only relevant for Ray 2.0 and the latest nightly releases, so essentially Dask on Ray. Would it make sense to flag it on both Ray and Dask's ends? Or just Dask?

arnavgarg1: Ideally, I'm hoping that the Ray team updates their `to_dask()` function in the future to handle this neatly.

ludwig/data/dataframe/dask.py
arnavgarg1: @justinxzhao I will update this PR with a docstring to make it clear why we're making this change. 
As far as I know, this doesn't break any existing tests or types for existing datasets. I haven't come across any issues during benchmarking, and I believe @geoffreyangus also ran benchmarks with this change across a variety of datasets and no issues popped up. 

@geoffreyangus are you able to confirm?

arnavgarg1:
Making this a draft PR for now since we haven't run into this issue yet.

ludwig/data/dataframe/dask.py
clarkzinzow: FYI, Dask does not have support for Pandas extension types in general, it looks like. https://github.com/dask/dask/issues/5720

arnavgarg1:
Closing this PR in favor of https://github.com/ludwig-ai/ludwig/pull/2728

PR #2327, opened 2022-07-28
Pin nightly versions to fix CI

geoffreyangus:
This PR pins Ray nightly to a working commit to ensure tests pass.

In addition, we pin torchvision and subsequently torch, torchtext, and torchaudio due to a similar bug to that found here: https://github.com/ludwig-ai/ludwig/pull/2072.


tgaddair (commented):
I think the torchvision issue was resolved, worth trying again without the pin for that dep.

geoffreyangus:
@tgaddair seems like tests pass with unpinned torch– feel free to review 🙂

tgaddair (approved):

PR #2328, opened 2022-07-28
Removes empty partitions after dropping rows and splitting datasets

geoffreyangus:
This PR addresses two separate issues: https://github.com/ludwig-ai/ludwig/issues/2324 and https://github.com/ludwig-ai/ludwig/issues/2308.

The issues are addressed by culling empty partitions from the Dask DataFrame at two points: (1) after dropping rows with NaNs (part of the DROP_ROWS missing value strategy) and (2) after splitting the dataset into train/val/test.

In order to maintain/increase performance, we add a persist call at the end of `build_dataset`, which makes it relatively inexpensive to compute the length of partitions repeatedly downstream.


arnavgarg1 (approved):
Thanks for making this change - looks good to me!

ShreyaR (approved):

ludwig/data/dataframe/dask.py
```diff
@@ -129,7 +149,24 @@ def to_ray_dataset(self, df):
         return from_dask(df)
 
     def from_ray_dataset(self, dataset) -> dd.DataFrame:
-        return dataset.to_dask()
+        """Custom Ray to Dask conversion implementation to pass in meta during dd.DataFrame creation."""
```
arnavgarg1, line 152: @geoffreyangus should we revert back to `dataset.to_dask()` if the empty partitions issue doesn't stem from a need to pass in meta?

geoffreyangus: Yup, reverted!

arnavgarg1:
LGTM!

PR #2329, opened 2022-07-28
Added feature specific parameter test for hyperopt

arnavgarg1:
We don't have any feature specific parameter tests for hyperopt - all the existing tests hyperopt over trainer/combiner parameters, or over shared/default parameters. We should have a test for feature-specific parameters to catch changes to the schema, such as those coming to Ludwig 0.6


justinxzhao (commented):
Thanks for adding this test! Let's hold off merging until #2269 merges.

justinxzhao (approved):

PR #2332, opened 2022-07-30
Test for parameter updates in Ludwig Combiners - Part 3

jimthompson5802:
# Code Pull Requests

Adds parameter update testing into Ludwig combiner unit tests

Adds parameter update testing to Attention unit test 

Fix error in TabTransformerCombiner constructor when setting up layers for embeddable and unembeddable features.


tests/ludwig/combiners/test_combiners.py
```diff
@@ -414,6 +475,10 @@ def test_transformer_combiner(encoder_outputs: tuple, transformer_output_size: i
             ("number", [BATCH_SIZE, 1]),  # passthrough encoder
             ("category", [BATCH_SIZE, 64]),
         ],
+        [  # single binary and single categorical
+            ("binary", [BATCH_SIZE, 1]),  # passthrough encoder
+            ("category", [BATCH_SIZE, 64]),  # passthrough encoder
+        ],
```
justinxzhao, line 481: I synced with @ShreyaR and @dantreiman about these tests, and we agreed that this is very much worth checking in, if we can clean up the tests for `TabTransformer` by splitting/un-parameterizing this big test into several smaller ones, with the main goal of removing the chain of conditionals.

```python
def test_tabtransformer_combiner_number_and_binary_without_category():
    ...

def test_tabtransformer_combiner_number_and_binary_with_category():
    ...

def test_tabtransformer_combiner_number_or_binary_without_category():
    ...

def test_tabtransformer_combiner_number_or_binary_with_category():
    ...
```

There might be some additional setup duplication, but the resulting tests would be more encapsulated and readable -- does that seem feasible/reasonable to do?

jimthompson5802: Got it...not a problem.  I should be able get this done in a day or two.

PR #2333, opened 2022-08-01
Add wine quality notebook to demonstrate using config defaults

arnavgarg1:
This notebook uses the `wine_quality` dataset from Kaggle to demonstrate how to create and use a Ludwig config with the new defaults section. It defines all 4 subsections - preprocessing, encoder, decoder, and loss.


examples/wine_quality/model_defaults_example.ipynb
```diff
@@ -0,0 +1,299 @@
+{
```
justinxzhao, line 1: This seems potentially more useful/readable as a live colab link under docs, like the example notebook for [text classification](https://ludwig-ai.github.io/ludwig-docs/0.5/examples/text_classification/). WDYT?

arnavgarg1: @justinxzhao Definitely agree. Would it make sense to have it in both places? I saw that we did something similar for hyperopt so I figured I'd replicate it. Happy to do whatever you think makes the most sense!

justinxzhao: Having this available in multiple places SGTM, though I'm not sure about the best way to keep the copies in sync. Maybe we can think about that later when we consolidate all of the "examples" we have across docs and the ludwig repo.

justinxzhao (approved):

examples/wine_quality/model_defaults_example.ipynb
arnavgarg1: Awesome, that sounds good. Happy to revisit when we're thinking of consolidating :)

PR #2335, opened 2022-08-02
Adds registry to organize backward compatibility updates around versions and config sections

dantreiman:
[DRAFT PR]: feature complete but needs documentation.

- Adds `version_transformation.py`, which defines a registry of transformation functions tagged by version.

- `version_transformation.py` provides `upgrade_to_latest_version()`
- Removes compatibility updating code from`defaults.py`'s `merge_with_defaults()`.  Moves update tests to `test_backward_compatibility.py`

Breaks up functions in `backward_compatibility.py` into smaller units, tags each one with a version, i.e.
```
@register_config_transformation("0.5")
def rename_training_to_trainer(config: Dict[str, Any]):
    if TRAINING in config:
        warnings.warn('Config section "training" renamed to "trainer" and will be removed in v0.6', DeprecationWarning)
        config[TRAINER] = config[TRAINING]
        del config[TRAINING]
    return config
```

### DIFF
--- a/ludwig/data/dataset_synthesizer.py
+++ b/ludwig/data/dataset_synthesizer.py
@@ -94,6 +94,20 @@ DATETIME_FORMATS = {
 }
 
 
+def _get_feature_encoder_or_decoder(feature):
+    """Returns the nested decoder or encoder dictionary for a feature.
+
+    If neither encoder nor decoder is present, creates an empty encoder dict and returns it.
+    """
+    if DECODER in feature:
+        return feature[DECODER]
+    elif ENCODER in feature:
+        return feature[ENCODER]
+    else:
+        feature[ENCODER] = {}
+        return feature[ENCODER]
+
+
 def generate_string(length):
     sequence = []
     for _ in range(length):
@@ -113,12 +127,9 @@ def return_none(feature):
 
 
 def assign_vocab(feature):
-    if DECODER in feature:
-        feature["idx2str"] = build_vocab(feature[DECODER].get("vocab_size", 10))
-    elif ENCODER in feature:
-        feature["idx2str"] = build_vocab(feature[ENCODER].get("vocab_size", 10))
-    else:
-        feature["idx2str"] = build_vocab(10)
+    encoder_or_decoder = _get_feature_encoder_or_decoder(feature)
+    encoder_or_decoder["idx2str"] = build_vocab(encoder_or_decoder.get("vocab_size", 10))
+    encoder_or_decoder["vocab_size"] = len(encoder_or_decoder["idx2str"])
 
 
 def build_feature_parameters(features):
@@ -206,7 +217,12 @@ def generate_datapoint(features):
 
 
 def generate_category(feature):
-    return random.choice(feature["idx2str"])
+    if DECODER in feature:
+        return random.choice(feature[DECODER]["idx2str"])
+    elif ENCODER in feature:
+        return random.choice(feature[ENCODER]["idx2str"])
+    else:
+        logger.error(f"Feature {str(feature)} should have either an encoder or decoder.")
 
 
 def generate_number(feature):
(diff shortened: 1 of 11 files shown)


connor-mccorm (commented):
This is a solid step to making our backwards compatibility more robust, very great job here. One question, were you planning on landing this before or after the Encoder Refactor PR? If before, I need to make the necessary changes to conform to this. If after, I imagine you'll have to make a couple changes for the backwards compatibility logic I added.

dantreiman:
> This is a solid step to making our backwards compatibility more robust, very great job here. One question, were you planning on landing this before or after the Encoder Refactor PR? If before, I need to make the necessary changes to conform to this. If after, I imagine you'll have to make a couple changes for the backwards compatibility logic I added.

Either ordering is fine with me, I don't think the merge will be too hard.

ludwig/utils/backward_compatibility.py
```diff
@@ -51,11 +48,55 @@
...
+        return config_transformation_registry.update_config(
+            config, from_version=config["ludwig_version"], to_version=LUDWIG_VERSION
+        )
+    else:
+        return config
 
 
 def _traverse_dicts(config: Any, f: Callable[[Dict], None]):
-    """Applies function f to every dictionary contained in config.
+    """Recursively Applies function f to every dictionary contained in config.
```
justinxzhao, line 99: nit: applies

tests/ludwig/utils/test_backward_compatibility.py
```diff
@@ -93,3 +110,82 @@ def test_audio_feature_backward_compatibility():
...
+
+    hparams = updated_config[HYPEROPT]["parameters"]
+    assert "training.learning_rate" not in hparams
+    assert "trainer.learning_rate" in hparams
+
+    assert "sampler" not in updated_config[HYPEROPT]
+
+    assert updated_config[HYPEROPT]["executor"]["type"] == "ray"
+    assert "num_samples" in updated_config[HYPEROPT]["executor"]
+    assert "scheduler" in updated_config[HYPEROPT]["executor"]
```
justinxzhao, line 158: Instead of these explicit asserts, could we simply assert on the full config, or at least in addition? I find these asserts to be less readable for groc-ing the overall config change.

ludwig/utils/backward_compatibility.py
```diff
@@ -68,25 +109,40 @@ def _traverse_dicts(config: Any, f: Callable[[Dict], None]):
...
-        )
-        config[CONV_USE_BIAS] = config[CONV_BIAS]
-        del config[CONV_BIAS]
-    if DEFAULT_BIAS in config:
-        warnings.warn(
-            'Parameter "default_bias" renamed to "default_use_bias" and will be removed in v0.6', DeprecationWarning
-        )
-        config[DEFAULT_USE_BIAS] = config[DEFAULT_BIAS]
-        del config[DEFAULT_BIAS]
+@register_config_transformation("0.5")
```
justinxzhao, line 112: What do you think about a slightly modified design where:
- config transformations are registered with the `warning` message, e.g.

    ```python
    @register_config_transformation("0.5", deprecation_warning='Config section "training" renamed to "trainer" and will be removed in v0.6')
    ```

- config transformations return a tuple `(config, bool)`, with the `bool` indicating whether the transformation was actually applied.
- `VersionTransformation.transform_config()` logs all of the deprecation warnings in one place, anytime `t.transform_config()` returns `(config, True)`.

A couple of benefits of this design is that it would encourage better encapsulated transformations, and centralize deprecation logging. WDYT?

justinxzhao (commented):
Nice PR! I'm a fan of these version-registered config transformations.

tests/ludwig/utils/test_backward_compatibility.py
dantreiman: I agree, but I just copied these tests over from test_defaults.py.  I didn't want to make this refactoring and rewrite the tests in the same PR, so I kept them as is.

ludwig/utils/backward_compatibility.py
dantreiman: I like the idea of requiring a deprecation message, thats a great idea!  However, this would force us to up to break up `_upgrade_feature`, `_upgrade_hyperopt`, `_upgrade_preprocessing_defaults`, and any other function which potentially makes multiple edits to the config.

It also would make it harder to use format strings in logs messages, i.e.
```
    f'executor type "{executor_type}" not supported, converted to "ray" will be flagged as error '
    "in v0.6",
```

We could enforce a convention "one config edit per function", but I'd rather have a looser guideline like "one conceptual config change per function" and its up to the author and reviewer to make that decision.

justinxzhao (approved):

connor-mccorm:
@dantreiman Did you want me to review this again after you've made the refactors from encoder PR?

dantreiman:
> 

Thanks, would appreciate a re-review.  As soon as all the tests are green I'm done.

ludwig/data/dataset_synthesizer.py
```diff
@@ -206,7 +217,12 @@ def generate_datapoint(features):
 
 
 def generate_category(feature):
-    return random.choice(feature["idx2str"])
+    if DECODER in feature:
```
connor-mccorm, line 220: nit: should we be using the helper function that you defined above here: `_get_feature_encoder_or_decoder`

tests/integration_tests/utils.py
```diff
@@ -236,64 +236,120 @@ def number_feature(normalization=None, **kwargs):
...
+            }
+        )
     recursive_update(feature, kwargs)
     feature[COLUMN] = feature[NAME]
     feature[PROC_COLUMN] = compute_feature_hash(feature)
     return feature
 
 
-def sequence_feature(**kwargs):
+def sequence_feature(output_feature=False, **kwargs):
```
connor-mccorm, line 321: Just a couple comments here. I thought of this option for refactoring these feature utils during my encoder/decoder PR however realized that this would require going back through all the Ludwig tests and making sure all output features have this flag which takes forever. Is there a plan in a followup PR to go through and refactor all these tests? Also since you've added this output feature flag to some of the feature utils, should we add them to all that have an encoder in the util config definition (i.e. image_feature)?

connor-mccorm (approved):
Looks great! This is actually a very helpful PR for my config object PR since it decouples backwards compatibility from merge with defaults which I want to get rid of now that the schema houses all the defaults. A couple nits/comments/questions regarding integration test utils among other things but overall, looks good to go IMO.

tests/integration_tests/utils.py
dantreiman: I added the `output_feature` kwarg, but ended up not really needing to use it anywhere.  

The conditional:
```
    if DECODER in kwargs:
        output_feature = True
```
was sufficient to handle all of our tests.  I thought about removing the `output_feature` kwarg and simplifying to `output_feature = DECODER in kwargs`, but decided not to because:
1.  Specifying output_feature=True might be useful for future test code.
2. `def sequence_feature(output_feature=False` makes it obvious that this could generate either an input or output feature.  I prefer having the explicit option, though that is just a style preference.

connor-mccorm: Gotcha, ya I'm ok leaving it in and just having tests be refactored as needed going forward / new tests using it if necessary. Do you think we should add that kw arg in for the other feature gen utils then?

dantreiman: The simple ones like binary don't really need it, nor image/audio/timeseries which don't have output features yet

connor-mccorm: Makes sense. LGTM then!

PR #2342, opened 2022-08-02
Add test with encoder dependencies for global defaults

arnavgarg1:
This PR adds a test for global defaults that uses an encoder with additional dependencies. Specifically, when the encoder is set to something like `bert`, there are additional fields from the encoder class that need to be injected into the input feature for things to work correctly. This PR makes sure that the fields are being populated as expected.


connor-mccorm (approved):
Good check. LGTM

PR #2345, opened 2022-08-03
fix: restore overwrite of eval_batch_size on GBM schema

jppgks:
Overwritten here to provide a default value. In some places, we fall back to `eval_batch_size` if `batch_size` is not specified. GBM does not have a value for `batch_size`, so we need to specify `eval_batch_size` here.

### DIFF
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -312,6 +312,19 @@ class GBMTrainerConfig(BaseTrainerConfig):
         allow_none=False,
     )
 
+    # NOTE: Overwritten here to provide a default value. In many places, we fall back to eval_batch_size if batch_size
+    # is not specified. GBM does not have a value for batch_size, so we need to specify eval_batch_size here.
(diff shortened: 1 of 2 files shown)


ludwig/schema/trainer.py
```diff
@@ -312,6 +312,19 @@ class GBMTrainerConfig(BaseTrainerConfig):
...
 
+    # NOTE: Overwritten here to provide a default value. In many places, we fall back to eval_batch_size if batch_size
+    # is not specified. GBM does not have a value for batch_size, so we need to specify eval_batch_size here.
+    eval_batch_size: Union[None, int, str] = schema_utils.IntegerOrAutoField(
+        default=128,
+        allow_none=False,
+        min_exclusive=0,
+        description=(
+            "Size of batch to pass to the model for evaluation. "
+            "If ’auto’, the biggest batch size (power of 2) that can fit in memory will be used."
```
justinxzhao, line 323: Is `auto` still work as expected? If so, then we should just default to `auto`. If not, then we should change this to be a non-None Integer field.

justinxzhao (approved):

PR #2346, opened 2022-08-03
feat: Add basic hyperopt schema

ksbrar:
After discussion with @justinxzhao I will be breaking this PR up into 6 smaller ones. I'll primarily be pulling from this branch, so I'll leave this PR up as reference for how it all should come together.

Smaller PRs in order:
- [x] 1. `get_output_metric_functions` - https://github.com/ludwig-ai/ludwig/pull/2759
- [ ] 1.5. [Optional] Improvements to metric access - ref: https://github.com/ludwig-ai/ludwig/pull/2759#discussion_r1023369967
- [x] 2. `BaseMarshmallowConfig` changes - https://github.com/ludwig-ai/ludwig/pull/2779
- [x] 3. `FloatTupleMarshmallowField` changes - https://github.com/ludwig-ai/ludwig/pull/2780
- [x] 4. `instantiate_search_algorithm` + `register_search_algorithm` + hyperopt registry + RayTuneExecutor using registry - https://github.com/ludwig-ai/ludwig/pull/2781
- [ ] 5. `ExecutorConfig` + `HyperoptConfig` (config additions only, no functional changes) - https://github.com/ludwig-ai/ludwig/pull/2784
- [ ] 6.  Finally using the new `HyperoptConfig` + backwards compatibility changes - https://github.com/ludwig-ai/ludwig/pull/2785
- [ ] 6.5. [Optional] Basic config object integration - may leave this to you @connor-mccorm .
----------------------------------------------------------
[Pastebin showing updated JSON schema of the hyperopt section of the Ludwig config](https://pastebin.com/D65YACJ2)
----------------------------------------------------------

### Summary of changes:

This PR adds a basic marshmallow-backed JSON schema for the hyperopt section of the config. Below is a summary of key changes:

#### A new registry replaces the manually defined `ludwig.hyperopt.search_algos.search_algo_registry`.
* Added `ludwig.constants.TRAIN` constant
* Added `ludwig.hyperopt.registry`, which is a top level file that contains this new `search_algorithm_registry` and its associated getter/register methods. This registry is then used for populating the search algorithm section of the hyperopt config.
* In the `ludwig.hyperopt.execution` module, the `self.search_algorithm` instance variable on `RayTuneExecutor` is now set using this registry as well.

#### The `search_algorithm` variable is instantiated slightly differently on new `RayTuneExecutor` instances.
Currently, `search_alg` is an optional argument on the constructor on the `RayTuneExecutor` class - if it's none then by default it is set to `HyperoptSA`. But with the new registry and schemas, this argument should never equal `None` when it is passed to the parameter (either the user sets it or it will be filled in by the schema) - so it is now a required constructor argument. Furthermore, `RayTuneExecutor.search_algorithm` gets instantiated using a new `instantiate_search_algorithm` function which takes the `search_alg` argument, uses the `type` inside of it to pick the right algorithm class, and then passes the whole dict to that class's constructor. The net result is the same as in the current flow, but the logic is slightly different overall.

#### Metric functions are now defined on an output feature's schema rather than on the output feature class itself.
* Currently, the supported metric functions are defined directly on output feature classes themselves using a `metric_functions` class variable. With this PR, these definitions are now pulled from a function on the config class for each respective output feature class called `get_output_metric_functions`.
* Relevant files: anything under `ludwig/features/` or `ludwig/schema/features` (basically, the same small refactor change is replicated across each class and schema).

#### `FloatTupleMarshmallowField` now have optional `min` and `max` kwargs.
* Small upgrade necessary for a few hyperopt params.

#### `BaseMarshmallowConfig` is now an abstract class.
* Small enhancement.

#### There is a new `ludwig.schema.hyperopt` module.
There are four parts to this module. 
*`ludwig.schema.hyperopt.__init__` assembles the full hyperopt schema, in accordance with 0.6/0.7 conventions. The top level parameters are (with varying levels of validation): `output_feature` (basic string validation), `goal` (minimize or maximize), `metric` (enum validation but not fully restricted based on output feature type), `split` (train, validation, test), `search_alg` (full, separate schema validation), `executor` (full, separate schema validation), and `parameters` (simple type validation to check if it is a dict or not).
* `ludwig.schema.hyperopt.executor` - basic Ray (ONLY) executor settings
* `ludwig.schema.hyperopt.scheduler` - full schema validation with separate configs for each Ray scheduler supported in Ludwig (organized in a new `scheduler_config_registry`). Certain constants were copied over from Ray to prevent a direct Ray import here, see comments for further details. Certain parameters are marked with `TODO`(s) because they cannot be truly represented by JSON schemas (e.g. `Callable` parameters), so they may require further testing.
* `ludwig.schema.hyperopt.search_alg`. - a very basic schema that forces `search_alg` to be a dict with a `type` key (that must be set to a value registered in the `search_algorithm_registry` from above).

##### 1. Upgrade script failures/changes
I had to make two changes in the upgrade script that slightly change expectations. In particular...
- [ ] `search_alg` was previously not consistently being upgraded to a dict (e.g. it was possible to set `search_alg: 'bohb'` rather than the now schema-enforced `search_alg: {'type': 'bohb'}`). Just to be clear, is this what we want?
- [ ] it had been possible to manually set `scheduler` to be an empty dict. I don't think this should be allowed, so the upgrade script will now delete this key from the hyperopt dict if it is empty. Again, is this acceptable?

##### 2. Should we support separate `search_alg` configs in Ludwig?
If you look at the new `BaseSearchAlgorithmConfig`, it has only one parameter - the type. The implication is that there should be other child config classes since Ray supports different sets of parameters for its various search space constructors (e.g. https://docs.ray.io/en/master/tune/api_docs/suggestion.html#ax-tune-search-ax-axsearch). I didn't implement these yet in the spirit of keeping the PR smaller and because I haven't seen examples in our docs of hyperopt with type-specific search space attributes set like this. But in `ludwig.hyperopt.execution` it looks like (if I'm reading it correctly) when a search algorithm is instantiated we *do* support an arbitrary dict of kwargs to pass to the ray constructor (see https://github.com/ludwig-ai/ludwig/blob/d864fbd976abdb2fec0f608ca02fdf6b32b02711/ludwig/hyperopt/execution.py#L725). Is that the implication?
- [ ] In summary, should we support type-specific search algorithm configs? If so, they can be added in a fast follow-up PR.
- [ ] Each search algorithm class has an instance variable called `search_alg_dict` which currently holds an entire dict of values to represent that class - including it's own `type`. Should this be broken up into 2+ variables?

##### 3. Avoiding a circular import error in `BaseSearchAlgorithmConfig`
- [ ] In order to avoid a circular import, I force populate the `search_algorithm_registry` by importing `ludwig.hyperopt.search_algos` inside of the class declaration for `BaseSearchAlgorithmConfig` (just ctrl-f "import ludwig.hyperopt.search_algos" in the diff view) rather than at the top of the file. I hope this is valid Python style, but I would like if someone could confirm this. If not, then I have to rearrange the registry declarations a bit.

##### 4. Config object integration?
- [ ] Mainly a question for @connor-mccorm : do I need to do anything else to integrate this schema properly with the config object? Do I need to add the `repr=False` tag to these dataclasses? Do I need to make edits in `model_config` at all?

### Potential follow-up PRs.
1. Add separate search algorithm configs as mentioned above.
2. Use implications/ref-definitions to support simultaneous restrictions on `output_feature` and `metric` (CC @connor-mccorm I think there might be a simple way to do this, something we can discuss later).


ludwig/schema/hyperopt/executor.py
```diff
@@ -0,0 +1,63 @@
...
+
+    type: str = schema_utils.StringOptions(options=["ray"], default="ray", allow_none=False)
+
+    num_samples: int = schema_utils.PositiveInteger(default=10, description="")
+
+    time_budget_s: int = schema_utils.PositiveInteger(default=3600, description="")
+
+    cpu_resources_per_trial: int = schema_utils.PositiveInteger(default=1, description="")
+
+    gpu_resources_per_trial: int = schema_utils.PositiveInteger(default=1, description="")
```
arnavgarg1, line 23: The default for this actually 0

ludwig/schema/hyperopt/hyperopt.py
```diff
@@ -0,0 +1,97 @@
...
+        ofcls = output_type_registry[oftype]
+        all_metrics += ofcls.metric_functions.keys()
+    return all_metrics
+
+
+@dataclass
+class HyperoptConfig(schema_utils.BaseMarshmallowConfig, ABC):
+    """Basic hyperopt settings."""
+
+    output_feature: str = "combined"  # TODO: make more restrictive
```
arnavgarg1, line 52: Not sure if there's a way to specify that this can only be combined, or one of the output feature types in the schema/config. But this is the correct default 👍

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,72 @@
...
+    "mean_loss",
+)
+TIME_TOTAL_S = "time_total_s"
+
+
+@dataclass
+class SchedulerConfig(schema_utils.BaseMarshmallowConfig):
+    """Basic scheduler settings."""
+
+    type: str = schema_utils.StringOptions(options=["async_hyperband"], default="async_hyperband", allow_none=False)
```
arnavgarg1, line 24: @ksbrar So as far as I know, if no scheduler is provided, Ray Tune actually defaults to the FIFO scheduler type (https://docs.ray.io/en/releases-1.12.1/tune/api_docs/schedulers.html#fifoscheduler), and there are many other options as well that we support. Essentially anything that can be created by `tune.create_scheduler` (https://docs.ray.io/en/releases-1.12.1/tune/api_docs/schedulers.html#shim-instantiation-tune-create-scheduler), so I would think almost all the algorithms on the page can be supported 

cc: @ShreyaR can you confirm?

ludwig/schema/hyperopt/executor.py
```diff
@@ -0,0 +1,63 @@
...
+
+    scheduler: SchedulerConfig = SchedulerDataclassField(description="")
+
```
arnavgarg1, line 26: I'd also add `kubernetes_namespace` (https://ludwig-ai.github.io/ludwig-docs/0.5/configuration/hyperparameter_optimization/#scheduler)

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -0,0 +1,54 @@
...
+from marshmallow_dataclass import dataclass
+
+from ludwig.schema import utils as schema_utils
+
+
+@dataclass
+class SearchAlgorithmConfig(schema_utils.BaseMarshmallowConfig):
+    """Basic search algorithm settings."""
+
+    type: str = schema_utils.StringOptions(options=["hyperopt"], default="hyperopt", allow_none=False)
```
arnavgarg1, line 14: Same here - there's definitely more options that we theoretically support in Ludwig: https://docs.ray.io/en/releases-1.12.1/tune/api_docs/suggestion.html#search-algorithms-tune-suggest, so just allowing hyperopt as the type might be very restrictive for Ludwig users if we merge this in.

ludwig/schema/hyperopt/hyperopt.py
```diff
@@ -0,0 +1,97 @@
...
+
+from marshmallow_dataclass import dataclass
+
+from ludwig.constants import LOSS
+from ludwig.features.feature_registries import output_type_registry
+from ludwig.schema import utils as schema_utils
+from ludwig.schema.hyperopt.executor import ExecutorConfig, ExecutorDataclassField
+from ludwig.schema.hyperopt.search_algorithm import SearchAlgorithmConfig, SearchAlgorithmDataclassField
+
+# hyperopt:
```
connor-mccorm, line 12: Do we need this commented out code here still?

ludwig/schema/hyperopt/hyperopt.py
```diff
@@ -0,0 +1,97 @@
...
+
+    executor: ExecutorConfig = ExecutorDataclassField(description="")
+
+    parameters: Any = None
+
+
+def get_hyperopt_jsonschema():
+    return schema_utils.unload_jsonschema_from_marshmallow_class(HyperoptConfig)
+
+    # def allowed_types_for_trainer_schema(cls) -> List[str]:
```
connor-mccorm, line 68: Is this code for a more rich hyperopt schema down the line?

connor-mccorm (commented):
Schema wise looks good. Just had a couple of questions regarding the code that's commented out. Once you address Arnav's feedback and get a green run I think we should be ready to go.

ludwig/schema/hyperopt/hyperopt.py
ksbrar: @arnavgarg1 yup that is the intention with the TODO. It's possible but it'll require some refactoring worth a different PR

ludwig/schema/hyperopt/scheduler.py
ksbrar: gotcha, will add those in

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,72 @@
...
+
+    time_attr: str = schema_utils.StringOptions(
+        options=list(DEFAULT_RESULT_KEYS), default=TIME_TOTAL_S, allow_none=False
+    )
+
+    max_t: int = schema_utils.PositiveInteger(default=3600, description="")
+
+    grace_period: int = schema_utils.PositiveInteger(default=72, description="")
+
+    reduction_factor: int = schema_utils.FloatRange(default=5, min=0, min_inclusive=False)
```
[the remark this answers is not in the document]
ksbrar: sure, will do

ludwig/schema/hyperopt/hyperopt.py
ksbrar: nope!

ludwig/schema/hyperopt/hyperopt.py
ksbrar: yea basically just some leftover experimentation but too complicated for one PR ;) - will delete

ludwig/schema/hyperopt/search_algorithm.py
ksbrar: will add in

ludwig/schema/hyperopt/executor.py
ksbrar: fixed

ludwig/schema/hyperopt/executor.py
ksbrar: added - is `None` the correct default?

arnavgarg1: Yup

ludwig/schema/hyperopt/scheduler.py
ShreyaR: @arnavgarg1 yep, can confirm that any scheduler that tune supports should work and that the default is `FIFO`. Here's the list:
```
    "fifo": FIFOScheduler,
    "async_hyperband": AsyncHyperBandScheduler,
    "asynchyperband": AsyncHyperBandScheduler,
    "median_stopping_rule": MedianStoppingRule,
    "medianstopping": MedianStoppingRule,
    "hyperband": HyperBandScheduler,
    "hb_bohb": HyperBandForBOHB,
    "pbt": PopulationBasedTraining,
    "pbt_replay": PopulationBasedTrainingReplay,
    "pb2": _pb2_importer,
    "resource_changing": ResourceChangingScheduler,
```

ludwig/hyperopt/registry.py
```diff
@@ -0,0 +1,13 @@
+from ludwig.utils.registry import Registry
+
+search_algorithm_registry = Registry()
+
+
+def register_search_algorithm(name: str):
+    print("test")
```
connor-mccorm, line 7: Should we remove this?

ludwig/hyperopt/search_algos.py
```diff
@@ -30,19 +29,27 @@ def check_for_random_seed(self, ludwig_random_seed: int) -> None:
             self.search_alg_dict[self.random_seed_attribute_name] = ludwig_random_seed
 
 
+def get_search_algorithm_cls(name: str) -> Type[SearchAlgorithm]:
```
connor-mccorm, line 32: What do you think about defining this along with the registry itself in the registry file and then importing it into this file. I feel like that might make a bit more organizational sense but I don't know. WDYT?

ludwig/schema/__init__.py
```diff
@@ -45,7 +46,7 @@ def get_schema():
             COMBINER: get_combiner_jsonschema(),
             TRAINER: get_trainer_jsonschema(),
             PREPROCESSING: {},
-            HYPEROPT: {},
+            # HYPEROPT: get_hyperopt_jsonschema(),
```
connor-mccorm, line 49: Do we want to enable this?

ludwig/schema/hyperopt/schedulers/__init__.py
```diff
@@ -0,0 +1,78 @@
+from dataclasses import field
```
connor-mccorm, line 1: Why make schedulers a directory instead of just a file?

ludwig/schema/hyperopt/hyperopt.py
```diff
@@ -0,0 +1,39 @@
...
+
+    search_alg: SearchAlgorithmConfig = SearchAlgorithmDataclassField(description="")
+
+    executor: ExecutorConfig = ExecutorDataclassField(description="")
+
+    parameters: Any = None
+
+
+def get_hyperopt_jsonschema():
+    return schema_utils.unload_jsonschema_from_marshmallow_class(HyperoptConfig)
```
connor-mccorm, line 39: Do we want to change this to a json structure to add a title? Maybe something like this:
```
def get_hyperopt_jsonschema():
    props = schema_utils.unload_jsonschema_from_marshmallow_class(HyperoptConfig)["properties"]

    return {
        "type": "object",
        "properties": props,
        "title": "hyperopt_options",
        "description": "Schema for hyperopt",
    }
```

connor-mccorm (approved):
Overall looks pretty good. Just a couple organizational comments and nits. Thanks for putting this together!

ludwig/schema/__init__.py
```diff
@@ -20,10 +20,9 @@
 from jsonschema import Draft7Validator, validate
 from jsonschema.validators import extend
 
-from ludwig.constants import (
+from ludwig.constants import (  # HYPEROPT,
```
arnavgarg1, line 23: nit: Remove this comment?

ludwig/schema/hyperopt/scheduler.py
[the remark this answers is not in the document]
ksbrar: done.

ludwig/schema/hyperopt/scheduler.py
ksbrar: done.

ludwig/schema/hyperopt/search_algorithm.py
ksbrar: done

ludwig/schema/hyperopt/executor.py
```diff
@@ -0,0 +1,65 @@
...
+from marshmallow import fields, ValidationError
+from marshmallow_dataclass import dataclass
+
+from ludwig.schema import utils as schema_utils
+from ludwig.schema.hyperopt.scheduler import SchedulerConfig, SchedulerDataclassField
+
+
+@dataclass
+class ExecutorConfig(schema_utils.BaseMarshmallowConfig):
```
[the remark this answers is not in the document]
ksbrar: addedd

ludwig/schema/hyperopt/hyperopt.py
ksbrar: fixed

ludwig/schema/__init__.py
ksbrar: fixed

ludwig/schema/__init__.py
ksbrar: fixed

ludwig/hyperopt/registry.py
ksbrar: removed

ludwig/hyperopt/search_algos.py
ksbrar: done

ludwig/schema/hyperopt/schedulers/__init__.py
ksbrar: might be worth making it a directory as its quite long, but fixed for now.

ludwig/schema/__init__.py
ksbrar: done

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+        )
+        conds.append(preproc_cond)
+    return conds
+
+
+def SchedulerDataclassField(default={"type": "async_hyperband"}, description="Hyperopt scheduler settings."):
+    """Custom dataclass field that when used inside of a dataclass will allow any scheduler in
+    `ludwig.schema.hyperopt.scheduler.scheduler_registry`.
+
+    Sets default optimizer to 'async_hyperband'.
```
connor-mccorm, line 572: Looks like there are a handful of copy/paste remainders from the optimizer here, may want to `ctrl+f` and address all of them so the docstring makes sense.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+        """Custom marshmallow field that deserializes a dict to a valid optimizer from
+        `ludwig.modules.optimization_modules.optimizer_registry` and creates a corresponding `oneOf` JSON schema
+        for external usage."""
+
+        def _deserialize(self, value, attr, data, **kwargs):
+            if value is None:
+                return None
+            if isinstance(value, dict):
+                if "type" in value and value["type"] in scheduler_config_registry:
+                    opt = scheduler_config_registry[value["type"].lower()][1]
```
connor-mccorm, line 589: probably better to call this `scheduler_config` or `scheduler` or something like that. WDYT?

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -0,0 +1,57 @@
...
+
+    """Basic search algorithm settings."""
+
+    type: str = schema_utils.StringOptions(
+        options=list(search_algorithm_registry.keys()), default="hyperopt", allow_none=False
+    )
+
+
+def SearchAlgorithmDataclassField(description: str = "", default: Dict = {"type": "variant_generator"}):
+    allow_none = False
```
connor-mccorm, line 23: May be cleaner to put this as an optional argument with a default if we want to allow the option for None, otherwise, I think it may be better to just put `None` in the return below. WDYT?

ludwig/schema/hyperopt/executor.py
```diff
@@ -0,0 +1,100 @@
...
+def ExecutorDataclassField(description: str, default: Dict = {}):
+    allow_none = False
+
+    class ExecutorMarshmallowField(fields.Field):
+        def _deserialize(self, value, attr, data, **kwargs):
+            if isinstance(value, dict):
+                try:
+                    return ExecutorConfig.Schema().load(value)
+                except (TypeError, ValidationError):
+                    raise ValidationError(f"Invalid params for scheduler: {value}, see ExecutorConfig class.")
```
connor-mccorm, line 74: Shouldn't this read "Invalid params for executor"?

ludwig/schema/hyperopt/executor.py
```diff
@@ -0,0 +1,100 @@
...
+                description="Automatically set number of concurrent trials.",
+            ),
+        ],
+    )
+
+    scheduler: BaseSchedulerConfig = SchedulerDataclassField(description="")
+
+
+def ExecutorDataclassField(description: str, default: Dict = {}):
+    allow_none = False
```
connor-mccorm, line 66: May be cleaner to put this as an optional argument with a default if we want to allow the option for None, otherwise, I think it may be better to just put None in the return below. WDYT?

ludwig/schema/__init__.py
```diff
@@ -56,7 +57,7 @@ def get_schema(model_type: str = MODEL_ECD):
             COMBINER: get_combiner_jsonschema(),
             TRAINER: get_trainer_jsonschema(model_type),
             PREPROCESSING: get_preprocessing_jsonschema(),
-            HYPEROPT: {},
+            HYPEROPT: get_hyperopt_jsonschema(),
```
connor-mccorm, line 60: 🤤

ludwig/hyperopt/search_algos.py
```diff
@@ -145,27 +158,3 @@ def __init__(self, search_alg_dict: Dict) -> None:
...
     def check_for_random_seed(self, ludwig_random_seed: int) -> None:
         pass
-
-
-def get_search_algorithm(search_algo):
-    search_algo_name = search_algo.get(TYPE, None)
-    return get_from_registry(search_algo_name, search_algo_registry)(search_algo)
-
-
-search_algo_registry = {
```
connor-mccorm, line 155: Feels good eh!

connor-mccorm (commented):
Great job on this. I know these changes take a ton of time and thought so well done. Will wait to approve the individual factored out PRs, but left a handful of feedback since I imagine you're just factoring out and there's not gonna be a ton of new code past this PR.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+        default=TRAINING_ITERATION,
+        allow_none=False,
+        description=(
+            "A training result attr to use for comparing time. Note that you can pass in something non-temporal such as"
+            " training_iteration as a measure of progress, the only requirement is that the attribute should increase "
+            "monotonically."
+        ),
+    )
+
+    metric: Optional[str] = schema_utils.StringOptions(
```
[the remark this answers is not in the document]
ksbrar: Definitely agree with the logic, but I want to clarify whether `metric` supports `None` or not. If so, then I would want to keep the `Optional` or perhaps replace it with `Union[str, None]` to be super specific.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+        options=list(DEFAULT_RESULT_KEYS) + [RAY_TUNE_DESULT_DEFAULT_METRIC],
+        default=None,
+        description=(
+            "The training result objective value attribute. Stopping procedures will use this attribute. If None but a "
+            "mode was passed, the ray.tune.result.DEFAULT_METRIC will be used per default."
+        ),
+    )
+
+    mode: Optional[str] = schema_utils.StringOptions(
```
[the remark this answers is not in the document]
ksbrar: Same answer as above ;)

ludwig/schema/hyperopt/scheduler.py
ksbrar: will do!

ludwig/schema/hyperopt/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    """Basic hyperopt settings."""
+
+    def get_hyperopt_metric_options():
+        all_metrics = []
+        for oftype in output_config_registry:
+            ofcls = output_config_registry[oftype]
+            all_metrics += ofcls.get_output_metric_functions().keys()
+        return all_metrics
+
+    output_feature: str = "combined"  # TODO: make more restrictive
```
arnavgarg1, line 24: @ksbrar Is the plan to make this so that it's either combined or one of the output feature names?

ludwig/schema/hyperopt/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+@dataclass
+class HyperoptConfig(schema_utils.BaseMarshmallowConfig, ABC):
+    """Basic hyperopt settings."""
+
+    def get_hyperopt_metric_options():
+        all_metrics = []
+        for oftype in output_config_registry:
+            ofcls = output_config_registry[oftype]
+            all_metrics += ofcls.get_output_metric_functions().keys()
+        return all_metrics
```
arnavgarg1, line 22: Would it make sense to make this a set instead of a list since many output features have overlapping metrics?

ludwig/schema/hyperopt/executor.py
```diff
@@ -0,0 +1,100 @@
...
+                default=1, allow_none=False, description="Manually set a number of concurrent trials."
+            ),
+            schema_utils.StringOptions(
+                options=["auto"],
+                default="auto",
+                allow_none=False,
+                description="Automatically set number of concurrent trials.",
+            ),
+        ],
+    )
```
arnavgarg1, line 60: 🥇

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+        ),
+    )
+
+    max_t: int = schema_utils.PositiveInteger(
+        default=81,
+        description=(
+            "max time units per trial. Trials will be stopped after max_t time units (determined by time_attr) have "
+            "passed."
+        ),
+    )
```
arnavgarg1, line 507: @ksbrar Since dataclass attributes like `mode` and `max_t` are repeated in quite a few of the schedulers, would it by any chance make sense to create methods outside of the dataclasses to generate these fields and then just call those functions from within the Dataclass? Although the description and options are unlikely to change for these fields, I'm thinking that it might be a lot easier to maintain / change in the future that way if needed.

ludwig/schema/hyperopt/scheduler.py
arnavgarg1: +1

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+                    opt = scheduler_config_registry[value["type"].lower()][1]
+                    try:
+                        return opt.Schema().load(value)
+                    except (TypeError, ValidationError) as e:
+                        raise ValidationError(
+                            f"Invalid params for optimizer: {value}, see `{opt}` definition. Error: {e}"
+                        )
+                raise ValidationError(
+                    f"Invalid params for optimizer: {value}, expect dict with at least a valid `type` attribute."
+                )
```
arnavgarg1, line 598: @ksbrar Does this ValidationError indicate that scheduler can no longer be None, and that at the very least, the hyperopt section of the Ludwig config needs to look something like this to not have validation errors?

```yaml
hyperopt:
...
...
executor:
    type: fifo
```

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,641 @@
...
+            if value is None:
+                return None
+            if isinstance(value, dict):
+                if "type" in value and value["type"] in scheduler_config_registry:
+                    opt = scheduler_config_registry[value["type"].lower()][1]
+                    try:
+                        return opt.Schema().load(value)
+                    except (TypeError, ValidationError) as e:
+                        raise ValidationError(
+                            f"Invalid params for optimizer: {value}, see `{opt}` definition. Error: {e}"
```
arnavgarg1, line 594: Should be scheduler instead of optimizer

ludwig/schema/hyperopt/scheduler.py
arnavgarg1: Hmm, actually reading through, it seems like it'll default to `async_hyperband` if it isn't specified. Can we default to `fifo` instead? I belive if no scheduler is passed into Ray.tune, then it uses `fifo` by default, but I guess we could be opinionated about it: cc: @ShreyaR

arnavgarg1 (approved):
@ksbrar Awesome work! Super excited to land this soon 🚀 Left a few small comments!

ludwig/schema/hyperopt/__init__.py
ksbrar: Eventually!

ludwig/schema/hyperopt/search_algorithm.py
ksbrar: Good catch, will move `allow_none` into the return. I don't think it should be configurable to a user of the field, though.

ludwig/schema/hyperopt/executor.py
ksbrar: will fix

ludwig/schema/hyperopt/executor.py
ksbrar: Same as above

ludwig/schema/__init__.py
ksbrar: ⚡

ludwig/hyperopt/search_algos.py
ksbrar: 💯

ludwig/schema/hyperopt/scheduler.py
ksbrar: Sure, I can move them into a common parent class of some sort.

ludwig/schema/hyperopt/scheduler.py
ksbrar: @arnavgarg1 exactly, this was one of the questions I called out in the description as well. I feel that this should be the behavior going forward, but happy to allow `None` if y'all say so.

Yes, can also set the default to `fifo`.

ludwig/schema/hyperopt/scheduler.py
ksbrar: will fix.

ksbrar:
Closed in favor of separate, smaller PRs.

PR #2347, opened 2022-08-03
fix: GBM tests failing after new release from upstream dependency

jppgks:
hummingbird-ml depends on onnxconverter-common. onnxconverter-common 1.12 introduced a dependency on onnxruntime. Since we do not use the onnx features from hummingbird, we can avoid the addition of onnxruntime dependency.


tgaddair (approved):
Nice! We should ask hummingbird team to make onnx deps optional.

PR #2348, opened 2022-08-03
fix: Properly serialize `ParameterMetadata` to JSON

ksbrar:
Currently, because `ParameterMetadata` is just a dataclass when we add it to any JSON payload it is never serialized.


justinxzhao (commented):
It looks like this PR will need to be adjusted to account for the case when parameter_metadata is `None`, as not all parameters have real parameter_metadata set yet.

Maybe in all of the `.to_json()` calling sites, this should be changed to:

```python
"parameter_metadata": parameter_metadata.to_json() if parameter_metadata else None
```

justinxzhao (approved):

PR #2359, opened 2022-08-08
Respect the resource requests in RayPredictor

magdyksaleh:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -626,7 +626,7 @@ class RayPredictor(BasePredictor):
         return {**get_trainer_kwargs(), **self.trainer_kwargs}
 
     def get_resources_per_worker(self) -> Tuple[int, int]:
-        trainer_kwargs = {**get_trainer_kwargs(), **self.trainer_kwargs}
+        trainer_kwargs = self.get_trainer_kwargs()
         resources_per_worker = trainer_kwargs.get("resources_per_worker", {})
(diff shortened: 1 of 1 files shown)


ludwig/backend/ray.py
```diff
@@ -623,10 +623,10 @@ def __init__(
         self.df_engine = df_engine
 
     def get_trainer_kwargs(self) -> Dict[str, Any]:
-        return {**self.trainer_kwargs, **get_trainer_kwargs()}
+        return {**get_trainer_kwargs(), **self.trainer_kwargs}
 
     def get_resources_per_worker(self) -> Tuple[int, int]:
-        trainer_kwargs = {**self.trainer_kwargs, **get_trainer_kwargs()}
+        trainer_kwargs = {**get_trainer_kwargs(), **self.trainer_kwargs}
```
arnavgarg1, line 629: @magdyksaleh Should this line just be a call to `self.get_trainer_kwargs()` and we just make this change/fix within `get_trainer_kwargs` function defined for `RayPredictor`? I think we don't actually need to do this twice and this was a typo/oversight from past work on the `RayPredictor` class.

magdyksaleh: agree that's cleaner

arnavgarg1 (approved):
Nice catch on the ordering of arguments from `get_trainer_kwargs`! LGTM 🚢 
Feel free to merge if all the tests pass :)

PR #2363, opened 2022-08-09
More precise resource usage tracking

abidwael:
Main features
- Using `torch.profiler` to track resource usage more precisely.
- Getting process-specifc CPU and CPU memory metrics from `psutils` instead of metrics for the whole machine.
- Directory re-organization.

How it works
```
with LudwigProfiler(tag="trainer.train", output_dir="outdir", use_torch_profiler=True):
    # code line 1
    # ...
    # code line n
```
or
```
@LudwigProfiler(tag="LudwigModel.evaluate", output_dir="outdir", use_torch_profiler=True):
def evaluate(args, ...):
    # code line 1
    # ...
    # code line n
```
This will profile the code wrapped with the context manager/decorator and write the resource usage metrics to a number of files following this example direcotry:
```
outdir
└── resource_usage_metrics
    ├── system_resource_usage
    │   ├── LudwigModel.evaluate
    │   │   └── run_0.json
    │   └── trainer.train
    │       └── run_0.json
    └── torch_ops_resource_usage
        ├── LudwigModel.evaluate
        │   └── run_0.json
        └── trainer.train
            └── run_0.json
```
Example output for `resource_usage_metrics/system_resource_usage/trainer.train/run_0.json`
```
{
    "average_cpu_memory_usage": 116578474.66666667,
    "average_cpu_utilization": 0.69375,
    "average_global_cpu_memory_available": 2158860800,
    "code_block_tag": "trainer.train",
    "cpu_architecture": "X86_64",
    "cpu_name": "Apple M1 Pro",
    "disk_footprint": 2899968,
    "ludwig_version": "0.6.dev",
    "max_cpu_memory_usage": 120967168,
    "max_cpu_utilization": 1.21,
    "min_global_cpu_memory_available": 2153541632,
    "num_cpu": 10,
    "total_cpu_memory_size": 34359738368,
    "total_execution_time": 2596286.124999999
}
```
Example output for `resource_usage_metrics/torch_ops_resource_usage/trainer.train/run_0.json`
```
{
    "num_oom_events": 0,
    "torch_average_cpu_memory_usage": 922671.6896565272,
    "torch_cpu_time": 749387.0,
    "torch_cpu_time_total": 749387,
    "torch_cuda_time": 0.0,
    "torch_cuda_time_total": 0,
    "torch_max_cpu_memory_usage": 4193200,
    "torch_self_cpu_time_total": 530170,
    "torch_self_cuda_time_total": 0
}
```


ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
     merged_file_path = os.path.join(os.getcwd(), experiment_name, "metrics_report", "{}.json".format("full_report"))
     save_json(merged_file_path, full_report)
     return full_report, merged_file_path
+
+
+def initialize_stats_dict(function_events):
```
jppgks, line 44: missing type hints

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+    """Initialize dictionary which stores resource usage information per tagged code block."""
+    info = dict()
+    for event_name in [evt.name for evt in function_events if "ludwig" in evt.name]:
+        info[event_name] = {"code_block_tag": event_name}
+        info[event_name]["runs"] = []
+    return info
+
+
+def get_memory_details(event):
```
jppgks, line 53: missing type hints

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+    """Get device name and number of bytes (de)allocated during an event."""
+    if event.device_type() in [DeviceType.CPU, DeviceType.MKLDNN, DeviceType.IDEEP]:
+        return "cpu", event.nbytes()
+    elif event.device_type() in [DeviceType.CUDA, DeviceType.HIP]:
+        return f"cuda_{event.device_index()}", event.nbytes()
+
+
+def get_devices_usage(kineto_event, mem_records_acc, run_usage_info):
```
jppgks, line 61: missing type hints

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+    for device in memory_lists:
+        memory_lists[device].append(0)  # just in case we have an empty list
+        run_usage_info[f"average_{device}_memory_usage"] = mean(memory_lists[device])
+        run_usage_info[f"max_{device}_memory_usage"] = max(memory_lists[device])
+        run_usage_info[f"average_{device}_memory_usage_str"] = _format_memory(mean(memory_lists[device]))
+        run_usage_info[f"max_{device}_memory_usage_str"] = _format_memory(max(memory_lists[device]))
+    return run_usage_info
+
+
+def get_device_timing(function_event, run_usage_info):
```
jppgks, line 80: missing type hints

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+    run_usage_info["cpu_time_total"] = function_event.cpu_time_total
+    run_usage_info["cpu_time_total_str"] = function_event.cpu_time_total_str
+    run_usage_info["cpu_time"] = function_event.cpu_time
+    run_usage_info["cpu_time_str"] = function_event.cpu_time_str
+    run_usage_info["cuda_time"] = function_event.cuda_time
+    run_usage_info["cuda_time_str"] = function_event.cuda_time_str
+    return run_usage_info
+
+
+def get_resource_usage_report(kineto_events, function_events, info):
```
jppgks, line 97: missing type hints

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
     merged_file_path = os.path.join(os.getcwd(), experiment_name, "metrics_report", "{}.json".format("full_report"))
     save_json(merged_file_path, full_report)
     return full_report, merged_file_path
+
```
jppgks, line 42: Is there any logic in the ludwig/benchmarking folder that gets replaced by the logic added here?

jppgks (changes requested):

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+        info[event_name]["runs"] = []
+    return info
+
+
+def get_memory_details(event):
+    """Get device name and number of bytes (de)allocated during an event."""
+    if event.device_type() in [DeviceType.CPU, DeviceType.MKLDNN, DeviceType.IDEEP]:
+        return "cpu", event.nbytes()
+    elif event.device_type() in [DeviceType.CUDA, DeviceType.HIP]:
+        return f"cuda_{event.device_index()}", event.nbytes()
```
justinxzhao, line 58: What's the expected behavior when the `device_type` is none of these 4?

If `None` is a valid return, the function should be documented as so. If only these 4 device types are valid, then an error should be raised, or we should define a reasonable default, i.e. `(f"custom_device_type_{event.device_index()}", event.nbytes())`

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+
+
+def get_devices_usage(kineto_event, mem_records_acc, run_usage_info):
+    """Get CPU and CUDA memory usage for and event."""
```
justinxzhao, line 62: nit: an

ludwig/benchmarking/reporting.py
justinxzhao, line 61: nit: perhaps rename to `get_device_memory_usage` to be more specific.

ludwig/benchmarking/reporting.py
justinxzhao, line 80: The docstring says "run durations". Maybe `get_device_run_durations()`, as this is more clear than "timings"

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+    run_usage_info["cpu_time_total_str"] = function_event.cpu_time_total_str
+    run_usage_info["cpu_time"] = function_event.cpu_time
+    run_usage_info["cpu_time_str"] = function_event.cpu_time_str
+    run_usage_info["cuda_time"] = function_event.cuda_time
+    run_usage_info["cuda_time_str"] = function_event.cuda_time_str
+    return run_usage_info
+
+
+def get_resource_usage_report(kineto_events, function_events, info):
+    """Get relevant information from Kineto events and function events exported by the profiler."""
```
justinxzhao, line 98: What are Kineto events? Can you add a link?

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +39,92 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+    function_events = p.profiler.function_events
+    assert Counter([evt.name for evt in function_events if "ludwig" in evt.name]) == Counter(
+        [evt.name() for evt in kineto_events if "ludwig" in evt.name()]
+    )
+    info = initialize_stats_dict(function_events)
+    info = get_resource_usage_report(kineto_events, function_events, info)
+    for code_block_tag, report in info.items():
+        os.makedirs(os.path.join(os.getcwd(), experiment_name, "metrics_report"), exist_ok=True)
+        file_path = os.path.join(
+            os.getcwd(), experiment_name, "metrics_report", f"{code_block_tag}_resource_usage.json"
```
justinxzhao, line 128: I would recommend taking the `outdir` as a separate explicit argument rather than deriving it implicitly from `os.getcwd()`. Users of this function might be in for a surprise when calling this results in results written to some directory elsewhere.

ludwig/benchmarking/reporting.py
abidwael: Yes, will be added in a separate PR

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +38,152 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+        memory_so_far[device] += nbytes
+        max_so_far[device] = max(max_so_far[device], memory_so_far[device])
+        average_so_far[device] = (memory_so_far[device] + (average_so_far[device] * count_so_far[device])) / (
+            count_so_far[device] + 1
+        )
+        count_so_far[device] += 1
+    for device in count_so_far:
+        run_usage_info[f"average_{device}_memory_usage"] = average_so_far[device]
+        run_usage_info[f"max_{device}_memory_usage"] = max_so_far[device]
+    return run_usage_info
```
justinxzhao, line 96: Why are we computing the metrics statefully? Is it expensive to iterate through `records_in_interval`?

```python
devices = set([get_memory_details(mem_record[0])[0] for mem_record in records_in_interval])

memory_usage = defaultdict(list)
for mem_record in records_in_interval:
    device, nbytes = get_memory_details(mem_record[0])
    memory_usage[device].append(nbytes)

max_memory_usage = defaultdict(int)
average_memory_usage = defaultdict(int)
for device in devices:
    max_memory_usage[device] = max(memory_usage[device])
    average_memory_usage[device] = mean(memory_usage[device])
...
```

ludwig/benchmarking/reporting.py
```diff
@@ -33,3 +38,152 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
+        [event.name() for event in main_kineto_events]
+    )
+
+    info = initialize_stats_dict(main_function_events)
+    info = get_resource_usage_report(main_kineto_events, main_function_events, memory_events, info)
+    for code_block_tag, report in info.items():
+        os.makedirs(output_dir, exist_ok=True)
+        file_path = os.path.join(output_dir, f"{code_block_tag}_resource_usage.json")
+        save_json(file_path, report)
+        print("exported to", file_path)
```
justinxzhao, line 189: Use logging.info() instead of prints.

ludwig/benchmarking/reporting.py
abidwael: Yes, it is expensive. I made the change from `.append`, then `max` and `mean` after timing this function.

As an example, when we profile a `ludwig.experiment` on the `ames_housing` dataset (1,460 examples) with `train_step: 5` which will take 0.87 seconds, `len(records_in_interval) = 54372`. Doing `.append`, `max` and `mean` for longer jobs will be pretty expensive.

ludwig/benchmarking/reporting.py
```diff
@@ -141,49 +142,54 @@ def get_resource_usage_report(
...
+    main_function_events = [evt for evt in function_events if evt.name in tags]
+    main_kineto_events = [event for event in kineto_events if event.name() in tags]
+    memory_events = [[event, False] for event in kineto_events if profiler_util.MEMORY_EVENT_NAME in event.name()]
+    # profiler_util.OUT_OF_MEMORY_EVENT_NAME seems to only be in newer versions of torch.
     out_of_memory_events = [event for event in kineto_events if "[OutOfMemory]" in event.name()]
     return main_kineto_events, main_function_events, memory_events, out_of_memory_events
 
 
-def export_metrics_from_torch_profiler(profile: torch.profiler.profiler.profile, output_dir: str):
+def export_metrics_from_torch_profiler(tags: list, profile: torch.profiler.profiler.profile, output_dir: str):
```
jppgks: `tags` here is list, whereas in other places it is `set`, let's be consistent across. Additionally, can we specify what's the type of the items in the set/list? Is it strings/callables?

ludwig/benchmarking/reporting.py
```diff
@@ -41,90 +40,146 @@ def create_metrics_report(experiment_name: str) -> Tuple[Dict[str, Any], str]:
...
 
 
-def export_metrics_from_torch_profiler(p: profile, experiment_name: str):
-    """Export time and resource usage metrics (CPU and CUDA) from a PyTorch profiler."""
+def get_all_events(
+    kineto_events: List[_KinetoEvent], function_events: profiler_util.EventList
+) -> Tuple[List[_KinetoEvent], List[profiler_util.FunctionEvent], List[Any], List[_KinetoEvent]]:
+    """Return main Kineto and function events (tagged with "ludwig.*"), memory and out of memory events.
+
+    :param tags: set of code block/function tags we report resource usage metrics for.
```
jppgks, line 148: `tags` does not seem to present as parameter in the function header

ludwig/benchmarking/resource_usage_tracker.py
```diff
@@ -32,37 +37,40 @@
...
-def monitor(
-    queue: Queue, info: Dict[str, Any], output_dir: str, logging_interval: int, cuda_is_available: bool
-) -> None:
+def monitor(queue: Queue, info: Dict[str, Any], logging_interval: int, cuda_is_available: bool) -> None:
     """Monitors hardware resource use.
 
-    Populate `info` with system specific metrics (GPU, CPU, RAM) at a `logging_interval` interval and saves the output
-    in `output_dir`.
+    Populate `info` with system specific metrics (CPU/CUDA, CPU/CUDA memory) at a `logging_interval` interval
+    and saves the output in `output_dir`.
```
jppgks, line 44: `output_dir` seems to be removed from the function header, but is still referred to in the docstring here

tests/ludwig/benchmarking/test_profiler.py
```diff
@@ -0,0 +1,61 @@
...
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "evaluate", "run_0.json"))
+    assert os.path.exists(os.path.join(tmpdir, "torch_ops_resource_usage", "evaluate", "run_0.json"))
+
+    func(0.25)
+    func(0.5)
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_0.json"))
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_1.json"))
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_2.json"))
+
+    shutil.rmtree(tmpdir)
```
justinxzhao, line 61: tmpdir is cleaned up automatically at the end of the test.

tests/ludwig/benchmarking/test_profiler.py
```diff
@@ -0,0 +1,61 @@
...
+        func(0.1)
+
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "evaluate", "run_0.json"))
+    assert os.path.exists(os.path.join(tmpdir, "torch_ops_resource_usage", "evaluate", "run_0.json"))
+
+    func(0.25)
+    func(0.5)
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_0.json"))
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_1.json"))
+    assert os.path.exists(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_2.json"))
```
justinxzhao, line 59: Maybe 

```python
assert os.listdir(os.path.join(tmpdir, "system_resource_usage", "test_function", "run_0.json")) == ["run_0.json", "run_1.json", "run_2.json"]
```

tests/ludwig/benchmarking/test_profiler.py
```diff
@@ -0,0 +1,61 @@
...
+    config = {
+        "input_features": [{"name": "input_1", "type": "number"}, {"name": "input_2", "type": "number"}],
+        "output_features": [{"name": "output_1", "type": "number"}],
+        "combiner": {"type": "concat", "output_size": 14},
+        TRAINER: {"epochs": 1},
+    }
+
+    model = LudwigModel(config=config, backend="local")
+
+    with LudwigProfiler(tag="train", output_dir=tmpdir, use_torch_profiler=False, logging_interval=0.1):
```
justinxzhao, line 34: nit: Use `profile_1`, `profile_2`, etc, to make the tag->os directory relationship more obvious.

ludwig/constants.py
```diff
@@ -219,3 +219,4 @@
 EXPERIMENT_RUN = "experiment_run"
 TRAIN_TAG = "train"
 EVAL_TAG = "evaluate"
+LUDWIG_TAG = "[ludwig]"
```
justinxzhao, line 222: Add a comment for what this tag is used for.

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+    :param memory_events: list of memory events.
+    :param info: dictionary used to record resource usage metrics.
+    """
+    mem_records_acc = profiler_util.MemRecordsAcc(memory_events)
+    main_kineto_events = sorted(
+        (evt for evt in main_kineto_events if LUDWIG_TAG in evt.name()), key=lambda x: x.correlation_id()
+    )
+    main_function_events = sorted((evt for evt in main_function_events if LUDWIG_TAG in evt.name), key=lambda x: x.id)
+    assert [evt.id for evt in main_function_events] == [evt.correlation_id() for evt in main_kineto_events]
+    assert [evt.name for evt in main_function_events] == [evt.name() for evt in main_kineto_events]
```
justinxzhao, line 116: These asserts would cause a crash in production if the conditions aren't met. Are these asserts essential?

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+    run_usage_info["torch_cuda_time"] = get_torch_op_time(function_event.cpu_children, "cuda_time_total")
+    return run_usage_info
+
+
+def get_resource_usage_report(
+    main_kineto_events: List[_KinetoEvent],
+    main_function_events: List[profiler_util.FunctionEvent],
+    memory_events: List[Any],
+    info: Dict[str, Any],
+) -> Dict[str, Any]:
```
justinxzhao, line 102: If there's a finite number of keys, it would be worth defining a dataclass instead of using `Dict[str, Any]`.

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+            total += getattr(e, attr)
         else:
-            raise ValueError("Tag unrecognized. Please choose 'train' or 'evaluate'.")
+            total += get_torch_op_time(e.cpu_children, attr)
+    return total
+
+
+def get_device_run_durations(
+    function_event: profiler_util.FunctionEvent, run_usage_info: Dict[str, Any]
+) -> Dict[str, Any]:
```
justinxzhao, line 86: Define a dataclass for run usage info.

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+
+    :param events: list of events.
+    :param attr: a FunctionEvent attribute. Expecting one of "cpu_time_total", "cuda_time_total".
+    """
+    if attr not in ["cpu_time_total", "cuda_time_total"]:
+        return -1
+
+    total = 0
+    for e in events:
+        if LUDWIG_TAG not in e.trace_name:
```
justinxzhao, line 77: Which events have LUDWIG_TAG in the trace name and which do not? Probably worth a comment.

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+        return "cpu", kineto_event.nbytes()
+    elif kineto_event.device_type() in [DeviceType.CUDA, DeviceType.HIP]:
+        return f"cuda_{kineto_event.device_index()}", kineto_event.nbytes()
+    else:
+        raise ValueError(f"Device {kineto_event.device_type()} is not valid.")
+
+
+def get_device_memory_usage(
+    kineto_event: _KinetoEvent, mem_records_acc: profiler_util.MemRecordsAcc, run_usage_info: Dict[str, Any]
+) -> Dict[str, Any]:
```
justinxzhao, line 37: Dataclass?

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+from typing import Any, Dict, List, Tuple, Union
+
+import torch
+from torch._C._autograd import _KinetoEvent
+from torch.autograd import DeviceType, profiler_util
+
+from ludwig.constants import LUDWIG_TAG
+
+
+def initialize_stats_dict(main_function_events: List[profiler_util.FunctionEvent]) -> Dict[str, Any]:
```
justinxzhao, line 11: This returns Dict[str, List], no?

ludwig/benchmarking/reporting.py
```diff
@@ -1,35 +1,166 @@
...
+    """Initialize dictionary which stores resource usage information per tagged code block.
+
+    :param main_function_events: list of main function events.
+    """
+    info = dict()
```
justinxzhao, line 16: nit. Use `{}`

ludwig/benchmarking/profiler.py
```diff
@@ -0,0 +1,274 @@
+"""some parts are inspired from https://github.com/Breakend/experiment-impact-
```
justinxzhao, line 1: nit: Capitalization "Some".

ludwig/benchmarking/profiler.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            self.info.pop("cpu_utilization")
+            self.info["average_cpu_utilization"] = 0
+        if self.info["cpu_memory_usage"]:
+            self.info["average_cpu_memory_usage"] = mean(self.info.pop("cpu_memory_usage"))
+        else:
+            self.info.pop("cpu_memory_usage")
+            self.info["average_cpu_memory_usage"] = 0
+
+        temp_dir = os.path.join(self.output_dir, "system_resource_usage", self.info["code_block_tag"])
```
justinxzhao, line 250: Is this a `temp_dir` or the actual output subdirectory? If the latter, maybe use `output_subdir`

ludwig/benchmarking/reporting.py
abidwael: They aren't. Removing them

ludwig/benchmarking/profiler.py
abidwael: No longer valid. Deleting.

ludwig/benchmarking/reporting.py
abidwael: The number of keys and their names depend on how many times someone uses `LudwigProfiler`. So the size and keys of this dictionary can vary from run to another.

ludwig/benchmarking/reporting.py
abidwael: The size and keys of this dictionary will depend on the devices that the code is run on. The contents will mainly differ based on the number of GPUs.

ludwig/benchmarking/profiler.py
```diff
@@ -0,0 +1,271 @@
...
+    def populate_static_information(self) -> None:
+        """Populate the report with static software and hardware information."""
+        self.info["ludwig_version"] = LUDWIG_VERSION
+        self.info["start_disk_usage"] = shutil.disk_usage(os.path.expanduser("~")).used
+
+        # CPU information
+        cpu_info = get_my_cpu_info()
+        self.info["cpu_architecture"] = cpu_info["arch"]
+        self.info["num_cpu"] = cpu_info["count"]
+        self.info["cpu_name"] = cpu_info["brand_raw"]
```
jppgks, line 135: Fails on local when running tests with:
```
>       self.info["cpu_name"] = cpu_info["brand_raw"]
E       KeyError: 'brand_raw'
```

ludwig/benchmarking/profiler.py
```diff
@@ -0,0 +1,271 @@
...
+from ludwig.constants import LUDWIG_TAG
+from ludwig.globals import LUDWIG_VERSION
+from ludwig.utils.data_utils import save_json
+
+# disabling print because the following imports are verbose
+f = open(os.devnull, "w")
+sys.stdout = f
+from experiment_impact_tracker.cpu.common import get_my_cpu_info  # noqa E402
+from experiment_impact_tracker.gpu.nvidia import get_gpu_info  # noqa E402
+from experiment_impact_tracker.py_environment.common import get_python_packages_and_versions  # noqa E402
```
jppgks, line 29: seems unused 
```suggestion
```

jppgks (commented):
Would it be hard to add a human-readable output? If not, adding the units for each of the reported metrics would help

jppgks (dismissed):

abidwael:
> Would it be hard to add a human-readable output? If not, adding the units for each of the reported metrics would help

@jppgks in PR #2372

ludwig/benchmarking/profiler_dataclasses.py
```diff
@@ -0,0 +1,81 @@
...
+    # Avergae CPU memory (RAM) utilization of the code block.
+    average_cpu_memory_usage: float
+
+    # Per device usage. Dictionary containing max and average memory used per device.
+    device_usage: Dict[str, DeviceUsageMetrics]
+
+    def to_flat_dict(self):
+        nested_dict = dataclasses.asdict(self)
+        nested_dict[""] = nested_dict.pop("device_usage")
+        return flatten_dict(nested_dict, sep="")
```
justinxzhao, line 61: Pull this out into a separate function, also since it's duplicated across the two classes.

```python
def profiler_dataclass_to_flat_dict(data: Union[SystemResourceMetrics, TorchProfilerMetrics]) -> Dict:
    """Returns a flat dictionary representation, with the device_usage key removed."""
    nested_dict = dataclasses.asdict(data)
    nested_dict[""] = nested_dict.pop("device_usage")
    return flatten_dict(nested_dict, sep="")
```

justinxzhao (commented):
This is looking a lot cleaner, nice!

ludwig/benchmarking/profiler.py
abidwael: Wasn't able to reproduce this. Let me know if you run into this again.

justinxzhao (approved):

PR #2370, opened 2022-08-09
Encoder refactor V2

dantreiman:
This PR contains all of Connor's commits from https://github.com/ludwig-ai/ludwig/pull/2269 (`:encoder_refactor`) merged atop the master branch, plus additional refactoring and test fixes in `:daniel/encoder_refactor`
```
master -----------------+------+>
                        |      |
daniel/encoder_refactor |      +*---->
                        |       | 
encoder_refactor        +------->

+ = branch
* = merge commit
```

A few notes:
- We have a glue function `load_config_with_kwargs` which loads a config from a dictionary, and collects any key-value pairs not captured by the config in an extra `kwargs` dictionary.  I removed this from instantiation of input features, now all input feature and encoder params are captured by the config.
- `load_config_with_kwargs` is still in use for combiners to pass in `output_size`.


ludwig/constants.py
```diff
@@ -58,7 +63,9 @@
 PERPLEXITY = "perplexity"
 JACCARD = "jaccard"
 PREDICTIONS = "predictions"
+TOP_K = "top_k"
```
justinxzhao, line 66: nit: This is duplicated.

ludwig/features/audio_feature.py
```diff
@@ -469,21 +457,23 @@ def forward(self, inputs, mask=None):
...
+        input_feature[ENCODER]["max_sequence_length"] = feature_metadata["max_length"]
+        input_feature[ENCODER]["embedding_size"] = feature_metadata["feature_dim"]
+        input_feature[ENCODER]["should_embed"] = False
 
     @staticmethod
     def populate_defaults(input_feature):
-        set_default_values(input_feature, {TIED: None, "preprocessing": {}})
+        defaults = AudioInputFeatureConfig()
+        set_default_values(input_feature, {TIED: defaults.tied, PREPROCESSING: {}})
+        set_default_values(input_feature, {ENCODER: {TYPE: defaults.encoder.type}})
```
justinxzhao, line 476: nit: This could be one statement.

```python
set_default_values(
    input_feature, 
    {
        TIED: defaults.tied, 
        PREPROCESSING: {}, 
        ENCODER: {TYPE: defaults.encoder.type}
    }
)
```

ludwig/features/base_feature.py
```diff
@@ -122,39 +122,33 @@ class BaseFeature:
     classes to avoid the diamond pattern.
     """
 
-    def __init__(self, feature, *args, **kwargs):
+    def __init__(self, feature: BaseFeatureConfig):
```
justinxzhao, line 125: Nice! Glad you were able to make this change.

ludwig/features/category_feature.py
```diff
@@ -161,15 +159,16 @@ def add_feature_data(
 
 @register_input_feature(CATEGORY)
 class CategoryInputFeature(CategoryFeatureMixin, InputFeature):
-    encoder = "dense"
+    # encoder = {TYPE: "dense"}
```
justinxzhao, line 162: nit: Remove.

justinxzhao (approved):
Minor nits. LGTM.

ludwig/features/base_feature.py
```diff
@@ -184,56 +181,52 @@ def create_preproc_module(metadata: Dict[str, Any]) -> torch.nn.Module:
...
+    ):
         """Defines defaults, overwrites them based on the feature dictionary, and sets up dependencies.
 
         Any output feature can depend on one or more other output features. The `other_output_features` input dictionary
         should contain entries for any dependent output features, which is accomplished by constructing output features
         in topographically sorted order. Attributes of any dependent output features are used to properly initialize
         this feature's sizes.
         """
-        super().__init__(*args, feature=feature, **kwargs)
+        feature = self.load_config(feature)
```
connor-mccorm, line 198: Just a quick question, but why do we load the config in the specific input/output features and the base output feature, but not in the base input feature?

ludwig/constants.py
dantreiman: Pre-commit merge error, good catch!

connor-mccorm (commented):
Thank you so much for finishing this off. Just had a small question about `load_config` in features/base.py, and also wanted to discuss a piece of feedback that Piero gave in the older PR but besides that I think it should be good to merge. Happy to chat now about these points or we can talk in standup at 11.

ludwig/features/base_feature.py
dantreiman: This is vestigial code - My first instinct was to `load_config` in the base class, but then I changed my mind.  `load_config` in `OutputFeature` doesn't do anything right now, will remove.

ludwig/data/dataset_synthesizer.py
```diff
@@ -111,7 +113,12 @@ def return_none(feature):
 
 
 def assign_vocab(feature):
-    feature["idx2str"] = build_vocab(feature.get("vocab_size", 10))
+    if DECODER in feature:
```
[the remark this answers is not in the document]
dantreiman: I recommend doing that as a follow-up task after this PR is merged, because what we have here works and the follow-up task is parallelizable AFAIK nothing else we're doing depends on it.  Created a follow-up task in linear here:  https://linear.app/predibase/issue/MLQ-54/refactor-usage-of-encoder-configdecoder-config-features

dantreiman:
I recommend doing https://github.com/ludwig-ai/ludwig/pull/2370#pullrequestreview-1072904998 as a follow-up task after this PR is merged, because what we have here works and the follow-up task is parallelizable AFAIK nothing else we're doing depends on it.  Created a follow-up task in linear: here:  https://linear.app/predibase/issue/MLQ-54/refactor-usage-of-encoder-configdecoder-config-features

connor-mccorm (approved):
LGTM

ludwig/features/category_feature.py
```diff
@@ -419,26 +417,26 @@ def postprocess_predictions(
 
     @staticmethod
     def populate_defaults(output_feature):
+        defaults = CategoryOutputFeatureConfig()
```
tgaddair, line 420: Is `populate_defaults()` needed anymore? I believe all defaults should be encoded in the schema, so I don't think we need to set them explicitly, but perhaps I'm missing something.

ludwig/features/date_feature.py
```diff
@@ -114,15 +114,14 @@ def add_feature_data(
 
 @register_input_feature(DATE)
 class DateInputFeature(DateFeatureMixin, InputFeature):
-    encoder = "embed"
-
-    def __init__(self, feature, encoder_obj=None):
-        super().__init__(feature)
-        self.overwrite_defaults(feature)
+    def __init__(self, input_feature_config: DateInputFeatureConfig, encoder_obj=None, **kwargs):
+        input_feature_config = self.load_config(input_feature_config)
```
tgaddair, line 118: Is it necessary to call `load_config` here? If the `input_feature_config` is always a `DateInputFeatureConfig`, then it should be a no-op, right?

ludwig/features/feature_utils.py
```diff
@@ -116,11 +116,11 @@ def get_input_size_with_dependencies(
     """
     input_size_with_dependencies = combiner_output_size
     for feature_name in dependencies:
-        if other_output_features[feature_name].num_fc_layers:
+        if other_output_features[feature_name].decoder["num_fc_layers"]:
```
tgaddair, line 119: I thought the decoder was schema-fied as well. Is that not the case?

ludwig/features/h3_feature.py
```diff
@@ -116,15 +111,16 @@ def add_feature_data(
 
 @register_input_feature(H3)
 class H3InputFeature(H3FeatureMixin, InputFeature):
-    encoder = "embed"
+    # encoder = {TYPE: "embed"}
```
tgaddair, line 114: Remove commented code.

tgaddair (approved):
Amazing work! I had a couple comments, but no worries if the answer is "we'll do this in the follow-up". But I would like to know if we're on the same page about the ideal state, and what the blockers are to getting there.

ludwig/features/feature_utils.py
connor-mccorm: Ya you're right, this is a good catch. I think the dependencies aren't actually being acknowledged here which is why this isn't throwing an error. Looking into this.

ludwig/features/category_feature.py
connor-mccorm: So I believe there are a couple of defaults that still need to be set with `populate_defaults()` before it gets to the actual feature initialization, IIRC for the preprocessing step. I will double check on this. Either way we will be removing this once the full config object lands so you are correct that they shouldn't be needed anymore with the new schemas, just think theres a small amount of additional wiring needed to fully remove this.

ludwig/features/date_feature.py
connor-mccorm: So I believe @dantreiman implemented this section so I will defer to him for the architecture choice here. What I can say is that we should be able to remove this once the config object PR is landed since the full config object will be initialized earlier in the pipeline so we won't have to load the individual configs here in the feature initializers.

dantreiman: Woops.  I meant to declare the type of `input_feature_config` as `Union[dict, DateInputFeatureConfig]`.  I intended to keep that initialization path working until we finally finish schema-fication because many of our tests construct input features from a config dict.

ludwig/features/feature_utils.py
dantreiman: Wow thanks for catching that: this code would have failed if it had ever been called.  I'll update it and add a new test to `test_dependencies.py`

ludwig/features/category_feature.py
dantreiman: For now, `populate_defaults()` is needed for preprocessing, since we don't instantiate the feature objects until after preprocessing.  We don't need an implementation per class anymore though, because all of the defaults can be stored in the schema.  @connor-mccorm is planning to remove populate_defaults in his config object PR, which will be a follow-up to this one.

ludwig/schema/features/base.py
```diff
@@ -1,7 +1,57 @@
...
 from ludwig.schema import utils as schema_utils
+from ludwig.schema.metadata.parameter_metadata import ParameterMetadata
+
+
+@dataclass
+class BaseFeatureConfig(schema_utils.BaseMarshmallowConfig):
+    """Base class for feature configs."""
+
+    name: str = schema_utils.String(
+        allow_none=True,
```
connor-mccorm, line 29: Shouldn't allow none for `name` and `type` be `false`?

ludwig/features/bag_feature.py
```diff
@@ -94,16 +88,14 @@ def add_feature_data(
...
-        self.overwrite_defaults(feature)
+    def __init__(self, input_feature_config: Union[BagInputFeatureConfig, Dict], encoder_obj=None, **kwargs):
+        input_feature_config = self.load_config(input_feature_config)
+        super().__init__(input_feature_config, **kwargs)
+        self.encoder_config = input_feature_config.encoder
         if encoder_obj:
             self.encoder_obj = encoder_obj
         else:
-            self.encoder_obj = self.initialize_encoder(feature)
+            self.encoder_obj = self.initialize_encoder()
```
w4nderlust, line 98: I'm not a fan of setting the encoder_config to self and then inside self.initialize_encoder() referring to it through self. I would rather pass the encoder_config explicitly to the initialize_encoder function.

w4nderlust: This applies to basically every input feature type

connor-mccorm: I'm actually working on this as we speak. Changing this to a method where we pass the encoder config into the encoder initializer to create the encoder. The encoder config will then live on the encoder object. This will avoid any issues with tied weights pointing to one encoder object but an encoder config pointing to another.

ludwig/features/feature_utils.py
```diff
@@ -116,11 +116,11 @@ def get_input_size_with_dependencies(
     """
     input_size_with_dependencies = combiner_output_size
     for feature_name in dependencies:
-        if other_output_features[feature_name].num_fc_layers:
+        if other_output_features[feature_name].decoder_config.num_fc_layers:
             input_size_with_dependencies += other_output_features[feature_name].fc_stack.output_shape[-1]
```
w4nderlust, line 120: previously the fc stack of each output feature was not pert of the decoder, was like a inbetween step between the combiner output and the decoder input. For that reason, parameters like num_fc_stack where assigned to the output feature and not to the decoder. Not sure if this was changed, but here I see something weird: the num_fc_layers is a parameter of the decoder, but in the following line the fc_stack from which the output shape is obtained is not inside the decoder but it's a property of the output feature (like it was before), so maybe this needs to be reconsiliated.

ludwig/features/bag_feature.py
```diff
@@ -94,16 +88,14 @@ def add_feature_data(
...
-    encoder = "embed"
-    vocab = []
-
-    def __init__(self, feature, encoder_obj=None):
-        super().__init__(feature)
-        self.overwrite_defaults(feature)
+    def __init__(self, input_feature_config: Union[BagInputFeatureConfig, Dict], encoder_obj=None, **kwargs):
+        input_feature_config = self.load_config(input_feature_config)
+        super().__init__(input_feature_config, **kwargs)
+        self.encoder_config = input_feature_config.encoder
```
w4nderlust, line 94: just a note to remind of the inconsistency of the self.encoder_config assignment in case there is a tied encoder and so encoder_obj is not None. This applies to all input feature types

ludwig/schema/features/base.py
dantreiman: If allow_none is false, then we must provide a (non-None) default, so I left allow_none True.  What would be the default if allow_none False?

ludwig/features/bag_feature.py
connor-mccorm: Great catch, I'm actually working on changing this right now which will address this comment and your first comment!

ludwig/features/sequence_feature.py
```diff
@@ -270,19 +261,17 @@ def add_feature_data(
...
+    def __init__(self, input_feature_config: Union[SequenceInputFeatureConfig, Dict], encoder_obj=None, **kwargs):
+        input_feature_config = self.load_config(input_feature_config)
+        super().__init__(input_feature_config, **kwargs)
         # TODO: Potentially abstract this feature-specific attribute overwrite to a consolidated design.
-        if "vocab" in feature:
-            feature["vocab_size"] = len(feature["vocab"])
-        self.overwrite_defaults(feature)
+        self.encoder_config = input_feature_config.encoder
+        if getattr(self.encoder_config, "vocab", None):
+            self.encoder_config.vocab_size = len(self.encoder_config.vocab)
```
w4nderlust, line 270: we can revise this later, but I'm not a fan of having two parameters for vocab and vocab size around

ludwig/features/feature_utils.py
dantreiman: Output features still have an fc_stack pre-decoder.  This is my mistake, will fix

dantreiman: Whoa, but we did move these params to the decoder though we still instantiate the FC stack in OutputFeatureBase:

ludwig/trainers/trainer_lightgbm.py
```diff
@@ -479,7 +479,7 @@ def _construct_lgb_params(self) -> Tuple[dict, dict]:
                 output_params = {
                     "objective": "multiclass",
                     "metric": ["multi_logloss"],
-                    "num_class": feature.num_classes,
+                    "num_class": feature.decoder_config.num_classes,
```
w4nderlust, line 482: I believe num_classes should likely be a property of the output feature, not the decoder, as here for the lightgbm trainer it should not be required to provide a decoder config to begin with. @jppgks to doublecheck

ludwig/features/feature_utils.py
w4nderlust: I think either is fine, if those parameters end up being decoder ones or output feature ones, we should be clear though about which is which and to have consistency ;)

ludwig/schema/features/base.py
connor-mccorm: This is a good point. I realized that since the json schema mapping still has them as required params, this won't cause an issue with users potentially not specifying a name or feature type so we can disregard this!

ludwig/features/bag_feature.py
connor-mccorm: Just finished this switch so we no longer use self in the initialization of encoders and decoders. The config is directly piped in and then becomes an attribute of the `encoder_obj` / `decoder_obj`

ludwig/features/bag_feature.py
connor-mccorm: This should be resolved now that the encoder and decoder configs live on the `encoder_obj` / `decoder_obj` now.

ludwig/features/sequence_feature.py
connor-mccorm: Agreed it is a little clunky and should probably be set more elegantly. I will create an issue on my Linear board to refactor this in a followup PR.

ludwig/trainers/trainer_lightgbm.py
connor-mccorm: Awaiting @jppgks reply to determine next steps!

ludwig/features/feature_utils.py
dantreiman: After investigating this I made some additional changes:

- Moved fc_stack parameters from decoders back into output feature
- Made threshold a property of BinaryOutputFeatureConfig, not the decoder
- Similarly made clip a property of NumberOutputFeatureConfig
- Output feature's `input_size` should be the output of the combiner (plus any dependencies).  Decoder's `input_size` is the output size of the fc stack.

ludwig/trainers/trainer_lightgbm.py
jppgks: Agree with @w4nderlust

jppgks: by default, GBM uses the passthrough decoder https://github.com/ludwig-ai/ludwig/blob/4405ef8745aab886e973dcdf3ad77c2048bf9f74/ludwig/utils/defaults.py#L313

PR #2372, opened 2022-08-11
Summarizing performance metrics and resource usage results

abidwael:
Takes in the saved results from the PyTorch profiler and exports reports (stdout and csv) to compare across experiments.

Example usage:

```
from ludwig.benchmarking.summarize import summarize_metrics

metric_diffs, resource_usage_diffs = summarize_metrics(
    "/Users/projects/ludwig/benchmarking/configs/benchmarking_config.yaml",
    "local_test_1",
    "local_test_2",
    "s3://some_bucket/benchmarks/")

for experiment_metric_diff in metric_diffs:
    print(experiment_metric_diff.to_string())

for experiment_resource_diff in resource_usage_diffs:
    for tag_diff in experiment_resource_diff:
        print(tag_diff.to_string())
```

This will print an export to CSV the following table for any function flagged to be profiled. In the current example, we profile the function `ludwig.api.LudwigModel.evaluate`.
```
Metrics for dataset: ames_housing
Output Feature Name  Metric Name             local_test_1  local_test_2  Diff          Diff Percentage
SalePrice            mean_absolute_error     180309.547    180308.281    -1.266        -0.001
SalePrice            mean_squared_error      38554296320.0 38553845760.0 -450560.0     -0.001
SalePrice            r2                      -5.38         -5.38         0.0           -0.001
SalePrice            root_mean_squared_error 196352.484    196351.328    -1.156        -0.001
SalePrice            root_mean_squared_percentage_error 1.0           1.0           -0.0          -0.001

Resource usage for: LudwigModel.evaluate
Metric Name                    local_test_1         local_test_2         Diff                 Diff Percentage
average_cpu_memory_usage       338.08 Mb            660.02 Mb            321.94 Mb            95.226
average_cpu_utilization        90.9                 810.9732558139535    720.0732558139536    792.16
cpu_memory_available           6.36 Gb              990.04 Gb            983.68 Gb            15459.893
disk_footprint                 32768                17068032             17035264             51987.5
max_cpu_memory_usage           338.08 Mb            669.75 Mb            331.68 Mb            98.106
max_cpu_utilization            98.5                 1958.1               1859.6               1887.919
num_cpu                        8                    255                  247                  3087.5
num_runs                       1                    1                    0                    0.0  
torch_average_cpu_memory_usage 182.36 Kb            182.36 Kb            -0.03467944375006482 b -0.0 
torch_cpu_time                 231.868ms            128.488s             128.256s             55314.392
torch_cpu_time_total           231.868ms            128.488s             128.256s             55314.392
torch_cuda_time                0.000us              0.000us              0.000us              inf  
torch_cuda_time_total          0.000us              0.000us              0.000us              inf  
torch_max_cpu_memory_usage     467.41 Kb            467.41 Kb            0 b                  0.0  
torch_self_cpu_time_total      201.276ms            3.229s               3.028s               1504.485
torch_self_cuda_time_total     0.000us              0.000us              0.000us              inf  
total_execution_time           235.431ms            128.492s             128.256s             54477.191
```


ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -5,9 +5,9 @@
 
 from torch.autograd.profiler_util import _format_memory, _format_time
 
-import ludwig.modules.metric_modules
+import ludwig.modules.metric_modules # noqa: F401
 from ludwig.globals import MODEL_HYPERPARAMETERS_FILE_NAME, REPORT_JSON
-from ludwig.modules.metric_registry import get_metric_classes, metric_feature_registry
+from ludwig.modules.metric_registry import get_metric_classes, metric_feature_registry # noqa: F401
```
abidwael: Added noqa for metric registries to be populated

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -1,17 +1,21 @@
...
+import ludwig.modules.metric_modules  # noqa: F401
 from ludwig.globals import MODEL_HYPERPARAMETERS_FILE_NAME, REPORT_JSON
-from ludwig.modules.metric_registry import get_metric_classes
+from ludwig.modules.metric_registry import get_metric_classes, metric_feature_registry  # noqa: F401
 from ludwig.utils.data_utils import load_json
 
 
 @dataclass
-class ExperimentSummary:
+class MetricsSummary:
```
justinxzhao, line 15: I think I prefer `ExperimentSummary`, but curious to know why you wanted to rename?

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -31,7 +35,7 @@ class ExperimentSummary:
 
 
 @dataclass
-class MetricDiff:
+class Diff:
```
justinxzhao, line 38: `MetricDiff` is more descriptive than `Diff`

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -43,15 +47,31 @@ class MetricDiff:
...
 
     # Percentage of change the metric with respect to base_value.
-    diff_percentage: float
+    diff_percentage: Union[float, str]
+
+    def __post_init__(self):
+        """Add human-readable string representations to the field."""
+
+        if "memory" in self.name:
+            self.base_value_str = _format_memory(self.base_value)
```
justinxzhao, line 60: These methods from torch have leading underscores, which suggests that they should be interpreted and treated as private/internal methods.

If we want to use the functionality verbatim, then we should copy the implementation and attribute instead of using the internal method directly, or look to see if they have a public API.

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -66,20 +86,200 @@ class ExperimentsDiff:
...
 
-    # List of metric diffs for a metric of the two experiments.
-    metrics: List[MetricDiff]
+    # Name of the experimental experiment.
+    experimental_experiment_name: str
+
+    # `List[Diff]` containing diffs for metric of the two experiments.
+    metrics: List[Diff]
+
+    def to_csv(self, path):
```
justinxzhao, line 206: This to_csv method probably shouldn't belong to the data class. The implementation of how the data class should be materialized, like in a file system, should be independent of the data class itself.

I'd recommend defining separate methods, i.e. `write_experiment_summary_to_csv(experiment_summary: ExperimentSummary, path)`, `write_resource_usage_summary_to_csv(resource_usage_summary: ResourceUsageSummary, path)`, parallel to the `build_*` methods that you've defined for these dataclasses.

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -66,20 +86,200 @@ class ExperimentsDiff:
...
+            f.write(
+                csv_str.format(
+                    "Code Block Tag",
+                    "Metric Name",
+                    self.base_experiment_name,
+                    self.experimental_experiment_name,
+                    "Diff",
+                    "Diff Percentage",
+                )
+            )
```
justinxzhao, line 220: Python has a widely used CSV library for reading and writing CSV files. Let's use that instead of baking our own implementation.

https://docs.python.org/3/library/csv.html#csv.DictWriter

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -66,20 +86,200 @@ class ExperimentsDiff:
...
+                        metric.name,
+                        metric.base_value_str,
+                        metric.experimental_value_str,
+                        metric.diff_str,
+                        diff_percentage,
+                    )
+                )
+        print("Exported report to", full_path)
+
+    def __str__(self):
```
justinxzhao, line 237: In my experience, overriding the built-in python `__str__` method for a dataclass is unusual/uncommon. Could we define a separate `to_string()` method and use that instead? It can be a class method.

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -138,3 +325,60 @@ def build_experiments_diff(
         experimental_summary=experimental_summary,
         metrics=metrics,
     )
+
+
+def build_resource_usage_summary(path):
```
justinxzhao, line 330: nit: missing type hints. Would also be good to add a docstring with the detail that this is returning the ResourceUsageSummary with the mean across all runs, if there are multiple runs.

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -138,3 +325,60 @@ def build_experiments_diff(
...
+):
+    base_experiment_reports = set(os.listdir(base_dir))
+    experimental_experiment_reports = set(os.listdir(experimental_dir))
+    shared_reports = base_experiment_reports.intersection(experimental_experiment_reports)
+
+    diffs = []
+    for report in shared_reports:
+        base_path = os.path.join(base_dir, report)
+        experimental_path = os.path.join(experimental_dir, report)
+        base_summary = build_resource_usage_summary(base_path)
```
justinxzhao, line 369: nit: base_report_path

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -138,3 +325,60 @@ def build_experiments_diff(
...
+        experimental_summary = build_resource_usage_summary(experimental_path)
```
justinxzhao, line 370: nit: experimental_report_path

ludwig/benchmarking/summary_dataclasses.py
```diff
@@ -138,3 +325,60 @@ def build_experiments_diff(
...
+    runs = report["runs"]
+
+    def average_runs(runs):
+        average_run = {"num_runs": len(runs)}
+        for metric in runs[0].keys():
+            average_run[metric] = mean([run[metric] for run in runs])
+        return average_run
+
+    average_run = average_runs(runs)
+    metric_names = set(average_run.keys())
```
justinxzhao, line 342: I don't think we need to `set()` this. `.keys()` should already be uniqued.

justinxzhao, line 342:
```suggestion
    average_run = {metric: mean([run[metric] for run in runs]) for metric in runs[0].keys()}
    average_run["num_runs"] = len(runs)
    metric_names = set(average_run.keys())
```

ludwig/benchmarking/summarize.py
```diff
@@ -85,102 +87,15 @@ def build_summary(
     """
     config = load_yaml(bench_config_path)
     downloaded_names = set(download_artifacts(config, base_experiment, experimental_experiment, download_base_path))
+    print(downloaded_names)
```
justinxzhao, line 90: nit: is this print intentional?

ludwig/benchmarking/summarize.py
```diff
@@ -85,102 +87,15 @@ def build_summary(
...
     experiment_diffs = []
     for n in downloaded_names:
         if isinstance(n, tuple) and len(n) == 2:
             (dataset_name, local_dir) = n
-            e = ExperimentsDiff(dataset_name, base_experiment, experimental_experiment, local_dir)
-            if not e.empty:
+            try:
+                e = build_metrics_diff(dataset_name, base_experiment, experimental_experiment, local_dir)
```
justinxzhao, line 96: No single character variable names. Let's use `experiment_diff`

ludwig/benchmarking/summarize.py
```diff
@@ -85,102 +87,15 @@ def build_summary(
...
                 experiment_diffs.append(e)
+            except Exception:
+                print("Exception encountered while creating diff summary for", dataset_name)
+                print(traceback.format_exc())
```
justinxzhao, line 100: Instead of print, let's use logging.warning() or logging.info(). Or raising an error if crashing is appropriate.

ludwig/benchmarking/summary_dataclasses.py
abidwael: replaced it because now we have `metrics` (accuracy, r2, etc.) and `resource usage metrics`. Both metrics and resource usage metrics are part of an `experiment`. I changed it to prevent confusion. wdyt?

ludwig/benchmarking/summary_dataclasses.py
justinxzhao: I would rename this to `MetricDiff`.

`Diff` is rather generic, and the global namespace doesn't make this clearer, as there's other "Diff" dataclasses in the same file.

```python
from ludwig.benchmarking.summary_dataclasses import Diff  # seems ambiguous.
```

PR #2373, opened 2022-08-11
Wrap read of cached training set metadata in try/except for robustness

jeffreyftang:
In practice the read may sometimes fail even when the previous existence check passes, although whether this is due to a race condition or some quirk of the underlying filesystem isn't fully understood.

Converting an errored read into a cache miss seems reasonable and more robust behavior.


justinxzhao (approved):
Seems like a reasonable change to me!

PR #2379, opened 2022-08-12
Add default split config

hungcs:
Currently on the predibase side, we rely on `defaults` to render preprocessing forms (I assume this will be replaced later by rjsf, but for now, the split form is missing)

we can either
1) Update the defaults here (though not sure this should be the default)
2) Update the split defaults on the predibase engine side
2) Hardcode the split probabilities form on the predibase client side

edit:
```
E               AssertionError: Defaults specified for `preprocessing` but `preprocessing` is
E                               not a feature type recognised by Ludwig.
```
does this mean there is no longer a global preprocessing?

### DIFF
--- a/ludwig/utils/defaults.py
+++ b/ludwig/utils/defaults.py
@@ -255,7 +255,6 @@ def merge_with_defaults(config: dict) -> dict:  # noqa: F821
         config[DEFAULTS] = dict()
 
     # Update defaults with the default feature specific preprocessing parameters
-    config[DEFAULTS][PREPROCESSING][SPLIT] = base_preprocessing_split
     for feature_type, preprocessing_defaults in default_feature_specific_preprocessing_parameters.items():
         # Create a new key with feature type if defaults is empty
         if feature_type not in config.get(DEFAULTS):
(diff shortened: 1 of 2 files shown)


ksbrar (approved):
@hungcs I believe that is the correct default!

arnavgarg1 (approved):
LGTM!

arnavgarg1:
@hungcs

```
E               AssertionError: Defaults specified for `preprocessing` but `preprocessing` is
E                               not a feature type recognised by Ludwig.
```

This error comes from https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/defaults.py#L128, where basically it makes sure that if `defaults` is passed into the Ludwig config, that it conforms to the following structure:

```
defaults:
    text:
        preprocessing:
            parameter_1: value
            ...
        encoder:
            parameter_1: value
            ...
        decoder:
            parameter_1: value
            ...
        loss:
            parameter_1: value
            ...
    category:
        preprocessing:
            parameter_1: value
            ...
        encoder:
            parameter_1: value
            ...
        decoder:
            parameter_1: value
            ...
        loss:
            parameter_1: value
            ...
    ...
```

So I guess the Ludwig config wasn't being built correctly, because you might have passed in something along the lines of:

```
defaults:
   preprocessing:
      ....
```

which isn't the correct structure.

Hope this helps!

ludwig/utils/defaults.py
```diff
@@ -63,8 +63,9 @@
 
 default_random_seed = 42
 
+base_preprocessing_split = {"type": "random", "probabilities": list(DEFAULT_PROBABILITIES)}
```
justinxzhao, line 66: Let's make this all caps, since it's a constant.

justinxzhao (approved):

PR #2380, opened 2022-08-13
Reduce dropout prob in test_conv1d_stack

dantreiman:
For some reason this specific test case `test_convolutional_modules.py::test_conv1d_stack` fails for me on my M1 mac (though it works consistently on Linux)

I'm not sure why, maybe its a different random number generator or different version of PyTorch.  Anyway, reducing the dropout prob fixes it for me.

### DIFF
--- a/tests/ludwig/modules/test_convolutional_modules.py
+++ b/tests/ludwig/modules/test_convolutional_modules.py
@@ -123,7 +123,7 @@ def test_conv1d_layer(
     assert out_tensor.size() == (BATCH_SIZE, output_seq_size, NUM_FILTERS)
 
 
-@pytest.mark.parametrize("dropout", [0, 0.9])
+@pytest.mark.parametrize("dropout", [0, 0.5])
 @pytest.mark.parametrize(
     "layers, num_layers",
     [


jimthompson5802 (approved):
This should be OK.  This should not affect the parameter update test.   Famous last words of course.

dantreiman:
> This should be OK. This should not affect the parameter update test. Famous last words of course.

Thanks!  Do you have any suggestions on how to handle situations like this? I imagine this will always happen with some probability when testing dropout.

jimthompson5802:
Unfortunately, this is like the No-Free Lunch theorem.  :smile:  

 I have not figured out away to reason  about this to know how to set up the assert tests ahead of time. It depends on the random seed chosen, combination of config parameters and, as I learned, from the `TabTransformerCombiner` work it may also depend on the combination of input features.

 For me, running the test and seeing the counts and then following the logic in the constructor and `forward()` method.  If some parameters are not updated, then the list `not_updated` will give a hint on where to look in the constructor and `forward()` method.  From that you'll have to determine if it is  normal", where you adjust the counts with some magic numbers, or there is a subtle error to be fixed, like the one found in `TabTransformerCombiner`.

This may help.  I describe a 2 step process I go through to figure out the above. https://github.com/ludwig-ai/ludwig-docs/pull/187     If there is anything I can add or change to this write-up, feel free to add feedback to the PR.

PR #2382, opened 2022-08-15
Fix: Random dataset splitting with 0.0 probability for optional validation or test sets.

### LINKED ISSUE #2368: Setting validation split probability to 0 causes size errors
When using the following config:

```yaml
preprocessing:
    split:
        probabilities: [0.7, 0, 0.3]
```

Ludwig throws the following error: `ValueError: Dataset is empty following preprocessing`. 

This error actually surfaces when writing/reading from the cache because Ludwig instantiates a `RayDataset` object that has an explicit check for size being > 0. 

When you do 0.7, 0.3, and 0, the validation set reports N/A for metrics and train/test set had metrics. When you do 0.7, 0, 0.3, then you hit the error messages above.

justinxzhao:
We retain that validation and test sets are optional in Ludwig. Only the training dataset is required.

This PR fixes a couple of exceptions that arise when a user's configuration results in an empty validation or an empty test set when relying on Ludwig's random splitting during data preprocessing when `0` is used as one of the split probabilities, i.e.

```yaml
preprocessing:
    split:
        probabilities: [0.7, 0, 0.3]
```

### 1: `ValueError: need at least one array to stack`

This comes from `to_numpy_dataset()` when it gets an empty data frame. The fix is to call `to_list()` instead of `np.stack()` if the dataframe is empty.

### 2: `ValueError: Dataset is empty following preprocessing`

This comes from `backend.dataset_manager.create()`. The fix is to move checking for empty datasets after creating the validation and test datasets, pass if the validation or test datasets are empty, and raise a `ValueError` specifically if the training set is empty.

### Fix: Zombie test datasets.

This PR also includes a fix to `RandomSplitter` to ensure that the test set is certainly empty if the split probability for it is `0`.

Closes #2368, Co-authored-by @arnavgarg1

### DIFF
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -74,8 +74,6 @@ class RayDataset(Dataset):
     ):
         self.df_engine = backend.df_engine
         self.ds = self.df_engine.to_ray_dataset(df) if not isinstance(df, str) else read_remote_parquet(df)
-        if self.size == 0:
-            raise ValueError("Dataset is empty following preprocessing")
         self.features = features
         self.training_set_metadata = training_set_metadata
         self.data_hdf5_fp = training_set_metadata.get(DATA_TRAIN_HDF5_FP)
--- a/ludwig/data/split.py
+++ b/ludwig/data/split.py
@@ -67,7 +67,13 @@ class RandomSplitter(Splitter):
 
         n = len(df)
         d1 = int(self.probabilities[0] * n)
-        d2 = d1 + int(self.probabilities[1] * n)
+        if not self.probabilities[-1]:
+            # If the last probability is 0, then use the entire remaining dataset for validation.
+            d2 = n
+        else:
+            d2 = d1 + int(self.probabilities[1] * n)
+
+        # Note that sometimes this results in the test set with 1 example even if the last probability is 0.
         return np.split(df.sample(frac=1, random_state=random_seed), [d1, d2])
 
     def has_split(self, split_index: int) -> bool:
--- a/ludwig/utils/dataframe_utils.py
+++ b/ludwig/utils/dataframe_utils.py
@@ -61,9 +61,9 @@ def to_numpy_dataset(df: DataFrame, backend: Optional["Backend"] = None) -> Dict
         res = df[col]
         if backend and is_dask_backend(backend):
             res = res.compute()
-        try:
+        if len(df.index) != 0:
             dataset[col] = np.stack(res.to_numpy())
-        except ValueError:
+        else:
             # Dataframe is empty.
             # Use to_list() directly, as np.stack() requires at least one array to stack.
             dataset[col] = res.to_list()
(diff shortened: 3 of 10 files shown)


ludwig/utils/dataframe_utils.py
```diff
@@ -61,7 +61,12 @@ def to_numpy_dataset(df: DataFrame, backend: Optional["Backend"] = None) -> Dict
         res = df[col]
         if backend and is_dask_backend(backend):
             res = res.compute()
-        dataset[col] = np.stack(res.to_numpy())
+        try:
+            dataset[col] = np.stack(res.to_numpy())
+        except ValueError:
```
jppgks, line 66: can we check for the dataframe to be empty instead of triggering a ValueError?

tests/ludwig/data/test_split.py
```diff
@@ -63,6 +63,35 @@ def compute(dfs):
...
 
 
+@pytest.mark.parametrize(
+    ("df_engine",),
+    [
+        pytest.param(PandasEngine(), id="pandas"),
+        pytest.param(DaskEngine(_use_ray=False), id="dask", marks=pytest.mark.distributed),
+    ],
+)
+def test_random_split_zero_probability_for_test_produces_no_zombie(df_engine):
```
jppgks, line 73: love the name 😂

ludwig/utils/dataframe_utils.py
justinxzhao: Good suggestion! Done.

jppgks (approved):

ludwig/data/split.py
```diff
@@ -67,7 +67,13 @@ def split(
...
         n = len(df)
         d1 = int(self.probabilities[0] * n)
-        d2 = d1 + int(self.probabilities[1] * n)
+        if not self.probabilities[-1]:
+            # If the last probability is 0, then use the entire remaining dataset for validation.
+            d2 = n
+        else:
+            d2 = d1 + int(self.probabilities[1] * n)
+
+        # Note that sometimes this results in the test set with 1 example even if the last probability is 0.
```
arnavgarg1, line 76: Nice fix!

arnavgarg1 (approved):
LGTM! Thanks for the quick fix :)

arnavgarg1:
Seems like some of the ray GBM tests are failing

ludwig/data/dataset/ray.py
```diff
@@ -74,8 +74,6 @@ def __init__(
     ):
         self.df_engine = backend.df_engine
         self.ds = self.df_engine.to_ray_dataset(df) if not isinstance(df, str) else read_remote_parquet(df)
-        if self.size == 0:
```
tgaddair, line 77: I wonder if we should still warn the user in these cases, in case the full test dataset ended up being dropped for one reason or another when they didn't expect that.

tgaddair: If we do, it should be in the preprocessing file below, though.

justinxzhao: Logging a warning when seems reasonable. Added!

PR #2386, opened 2022-08-15
Enable saving hyperopt checkpoints with multi-node clusters

arnavgarg1:
This PR enables two things:

1. Ludwig users can now create a custom  RayTune `SyncClient` and pass it into RayTune for custom checkpointing syncing behavior.
2. Function calls to `fsspec` related functions in `fs_utils` can now take in custom storage options (credentials) for different protocol types. This means that model metadata, parameters, checkpoints etc can now be written to remote storage options such as GCS, Azure, AWS, Minio, etc - anything that `fsspec` supports.

Both of these changes together enable being able to perform multi-node checkpoint syncing for hyperopt.

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -29,13 +29,14 @@ from ludwig.api import LudwigModel
 from ludwig.backend import initialize_backend, RAY
 from ludwig.backend.ray import initialize_ray
 from ludwig.callbacks import Callback
-from ludwig.constants import MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
+from ludwig.constants import HYPEROPT_LOCAL_DIR, MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
+from ludwig.globals import MODEL_HYPERPARAMETERS_FILE_NAME, TRAIN_SET_METADATA_FILE_NAME
 from ludwig.hyperopt.results import HyperoptResults, TrialResults
 from ludwig.hyperopt.search_algos import get_search_algorithm
 from ludwig.hyperopt.utils import load_json_values, substitute_parameters
 from ludwig.modules.metric_modules import get_best_function
 from ludwig.utils import metric_utils
-from ludwig.utils.data_utils import hash_dict, NumpyEncoder
+from ludwig.utils.data_utils import hash_dict, NumpyEncoder, save_json, use_credentials
 from ludwig.utils.defaults import default_random_seed, merge_with_defaults
 from ludwig.utils.fs_utils import has_remote_protocol, safe_move_file
 from ludwig.utils.misc_utils import get_from_registry
@@ -44,7 +45,8 @@ _ray_200 = version.parse(ray.__version__) >= version.parse("2.0")
 if _ray_200:
     from ray.air import Checkpoint
     from ray.tune.search import SEARCH_ALG_IMPORT
-    from ray.tune.syncer import get_node_to_storage_syncer, SyncConfig
+
+    # from ray.tune.syncer import get_node_to_storage_syncer, SyncConfig
 else:
     from ray.ml import Checkpoint
     from ray.tune.suggest import SEARCH_ALG_IMPORT
@@ -165,6 +167,8 @@ class RayTuneExecutor:
         self.max_concurrent_trials = max_concurrent_trials
         self.sync_config = None
         self.sync_client = None
+        self.sync_function_template = kwargs.get("sync_function_template", None)
+        self.delete_function_template = kwargs.get("delete_function_template", None)
         # Head node is the node to which all checkpoints are synced if running on a K8s cluster.
         self.head_node_ip = ray.util.get_node_ip_address()
 
@@ -303,17 +307,18 @@ class RayTuneExecutor:
     def _gpu_resources_per_trial_non_none(self):
         return self.gpu_resources_per_trial if self.gpu_resources_per_trial is not None else 0
 
-    def _get_remote_checkpoint_dir(self, trial_dir: Path) -> Optional[Union[str, Tuple[str, str]]]:
+    def _get_remote_checkpoint_dir(
+        self, trial_dir: Path, use_tmp: bool = False
+    ) -> Optional[Union[str, Tuple[str, str]]]:
         """Get the path to remote checkpoint directory."""
         if self.sync_config is None:
             return None
 
         if self.sync_config.upload_dir is not None:
             # Cloud storage sync config
-            remote_checkpoint_dir = os.path.join(
-                self.sync_config.upload_dir, *_get_relative_checkpoints_dir_parts(trial_dir)
-            )
-            return remote_checkpoint_dir
+            if use_tmp:
+                return os.path.join(self.sync_config.upload_dir, "tmp", *_get_relative_checkpoints_dir_parts(trial_dir))
+            return os.path.join(self.sync_config.upload_dir, *_get_relative_checkpoints_dir_parts(trial_dir))
         elif self.kubernetes_namespace is not None:
             # Kubernetes sync config. Returns driver node name and path.
             # When running on kubernetes, each trial is rsynced to the node running the main process.
@@ -321,8 +326,8 @@ class RayTuneExecutor:
             return (node_name, trial_dir)
         else:
             logger.warning(
-                "Checkpoint syncing disabled as syncing is only supported to remote cloud storage or on Kubernetes "
-                "clusters is supported. To use syncing, set the kubernetes_namespace in the config or use a cloud URI "
+                "Checkpoint syncing disabled as it is only supported to remote cloud storage or on Kubernetes "
+                "clusters. To use syncing, set the kubernetes_namespace in the config or use a cloud URI "
                 "as the output directory."
             )
             return None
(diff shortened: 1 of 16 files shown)


tests/ludwig/utils/test_fs_utils.py
```diff
@@ -39,6 +42,14 @@ def test_get_fs_and_path_unicode():
     assert create_file("http://a/æ.jpg") == "a/æ.jpg"
 
 
+@pytest.mark.filesystem
+def test_get_fs_and_path_with_local_directory():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        fs, path = get_fs_and_path(tmpdir)
+        assert isinstance(fs, fsspec.implementations.local.LocalFileSystem)
+        assert path == tmpdir
```
justinxzhao, line 50:
```suggestion
def test_get_fs_and_path_with_local_directory(tmpdir):
    fs, path = get_fs_and_path(tmpdir)
    assert isinstance(fs, fsspec.implementations.local.LocalFileSystem)
    assert path == tmpdir
```

ludwig/utils/fs_utils.py
```diff
@@ -33,14 +33,59 @@
...
+    KEY,
+    MLFLOW_S3_ENDPOINT_URL,
+    S3,
+    SECRET,
+)
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_STORAGE_OPTIONS = {
```
justinxzhao, line 50: Where does this structure come from? Maybe add a comment about this.

ludwig/hyperopt/execution.py
```diff
@@ -178,7 +190,7 @@ def __init__(
         self.time_budget_s = time_budget_s
         self.max_concurrent_trials = max_concurrent_trials
         self.sync_config = None
-        self.sync_client = None
+        self.sync_client = kwargs.get("sync_client", None)
```
arnavgarg1, line 193: @arnavgarg1 can revert this change if things look good, don't need this anymore

ludwig/utils/fs_utils.py
```diff
@@ -152,30 +224,41 @@ def safe_move_file(src, dst):
...
-def makedirs(url, exist_ok=False):
-    fs, path = get_fs_and_path(url)
+def makedirs(url, exist_ok=False, storage_options: Optional[Dict[str, Any]] = None):
+    fs, path = get_fs_and_path(url, storage_options=storage_options)
+    print(f"[makedirs] url: {url}")
+    print(f"[makedirs] fs: {fs}")
+    print(f"[makedirs] path: {path}")
+    logger.info(f"logger [makedirs] url: {url}")
+    logger.info(f"logger [makedirs] fs: {fs}")
+    logger.info(f"logger [makedirs] path: {path}")
```
arnavgarg1, line 243: Remove logging statements once done

ludwig/hyperopt/execution.py
arnavgarg1: This is needed, won't be reverting.

ludwig/utils/fs_utils.py
```diff
@@ -152,30 +179,30 @@ def safe_move_file(src, dst):
...
     else:
         safe_move_file(src, tgt)
 
 
-def makedirs(url, exist_ok=False):
-    fs, path = get_fs_and_path(url)
-    fs.makedirs(path, exist_ok=exist_ok)
-    if not path_exists(url):
-        with fsspec.open(url, mode="wb"):
-            pass
```
arnavgarg1, line 169: This creates a file, not a directory. That's why I deleted it

ludwig/utils/fs_utils.py
```diff
@@ -33,14 +33,41 @@
...
+def create_fs(protocol: Union[str, None], storage_options: Optional[Dict[str, Any]]) -> fsspec.filesystem:
+    """Create filesystem object with correct storage options based on the protocol."""
+    if not protocol:
+        # Defaults to LocalFileSystem
+        return fsspec.filesystem(protocol)
+
+    if not storage_options:
+        logger.info(f"Using default storage options for `{protocol}` filesystem.")
+        if protocol == S3:
+            s3 = S3RemoteStorageOptions()
```
tgaddair, line 51: Why is s3 handled specially here? Do we need to do anything for other filesystems?

One way to make this clearer could be:

```
if not storage_options:
    logger.info(f"Using default storage options for `{protocol}` filesystem.")
    if protocol == S3:
        storage_options = S3RemoteStorageOptions().get_storage_options()

 try:
    return fsspec.filesystem(protocol, **storage_options)
...
```

ludwig/utils/remote_storage_options.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_storage_options(self):
+        return {KEY: self.get_key(), SECRET: self.get_secret(), CLIENT_KWARGS: self.get_client_kwargs()}
+
+
+class S3RemoteStorageOptions(BaseRemoteStorageOptions):
+    """Get credentials from environment variables."""
+
+    def __init__(self):
+        super().__init__(AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, MLFLOW_S3_ENDPOINT_URL)
```
tgaddair, line 60: Doesn't make sense to me to have mlflow coupling here. Looks like the idea here is to pull storage options from certain environment variables when they're not passed in explicitly. However, this shouldn't be necessary, as fsspec already has a mechanism for plumbing credentials through the environment, either through the `use_credentials` helper that sets the `conf` dictionary, or through the `FSSPEC_` env vars:

https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration

So I would rely on the built-in fsspec plumbing mechanisms instead of recreating it here.

ludwig/utils/fs_utils.py
tgaddair: Following up on my comment below, I now believe that this `storage_options` handling shouldn't be necessary.

Also, this approach has an additional bug, in that it will break existing `use_credentials` behavior by overriding it with credentials from the environment (when both are using s3).

tgaddair (changes requested):

ludwig/utils/fs_utils.py
tgaddair: There's actually an important reason why it was written this way (though unfortunately it was not commented, which would have made it clear).

Basically, S3-like filesystems have no concept of "directories". Directories are a higher level abstraction that some APIs provide on top of files which contain `/` characters to make them easier to organize. So when you call `makedir` on an S3-like filesystem, it can be a no-op (depending on how the client implements it). So the workaround (if downstream code assumes the existence of this "directory") is to create a file with the directory name instead.

arnavgarg1: Got it! That makes a lot of sense - will add this back with comments for future Ludwig developers.

ludwig/utils/fs_utils.py
arnavgarg1: @tgaddair S3 is handled especially here because users may want to pass in a different S3 compatible storage like Minio. The issue is that to do this, users need to pass in the `--endpoint_url` flag to AWSCLI because AWSCLI currently doesn't have an environment variable that can be configured to pick up this URL automatically (although they are currently working on a proposal to add it in: https://github.com/aws/aws-sdk/issues/229). Within fsspec, that would involve manually passing in client_kwargs with the endpoint_url defined in the object. I think this is the only reason to do something special for S3 at the moment. 

I can't entirely say what would be required for other file systems yet, but those should automatically get created via fsspec.

arnavgarg1: About the `use_credentials` method - I think that this actually won't be an issue because both `use_credentials` and this approach will ned up using the same key and secret since they're derived from the same environment variables. The only thing that changes is the endpoint URL, which in either case would need to change for S3 compatible storage since use_credentials would also require this new endpoint URL (if it is specified).

ludwig/utils/remote_storage_options.py
arnavgarg1: @tgaddair the `MLFLOW_S3_ENDPOINT_URL` is actually just a placeholder name and can be change to any other name. It doesn't actually end up involving any coupling with MLFLOW. I'll rename this environment variable to something more generic like S3_ENDPOINT_URL to make this more clear! 

I like the idea of using the FSSPEC_ env vars as a potential substitute to avoid plumbing S3 credentials in this particular case with endpoint URLs. Will update and clear this up so that we don't need to do any of this manually.

arnavgarg1: Although following from the thread here (https://github.com/fsspec/s3fs/issues/432) where you've requested this before as well, it doesn't seem like there's a neat way to specify the endpoint URL since it needs to be set inside a nested dictionary. How do you feel about just using this as an environment variable for now but with a different name?

ludwig/hyperopt/execution.py
```diff
@@ -320,8 +326,32 @@ def _get_remote_checkpoint_dir(self, trial_dir: Path) -> Optional[Union[str, Tup
...
+            return remote_checkpoint_dir
+        elif self.kubernetes_namespace is not None:
+            # Kubernetes sync config. Returns driver node name and path.
+            # When running on kubernetes, each trial is rsynced to the node running the main process.
+            node_name = self._get_kubernetes_node_address_by_ip()(self.head_node_ip)
+            return (node_name, trial_dir)
+        else:
+            logger.warning(
+                "Checkpoint syncing disabled as it is only supported to remote cloud storage or on Kubernetes "
+                "clusters. To use syncing, set the kubernetes_namespace in the config or use a cloud URI "
```
arnavgarg1, line 354: - [ ] Merge into `_get_remote_checkpoint_dir` with optional argument "tmp" to avoid code duplication

arnavgarg1:
Closing this in favor of https://github.com/ludwig-ai/ludwig/pull/2617

PR #2390, opened 2022-08-15
Updating images to Ray 2.0.0 and CUDA 11.3

abidwael:
Upgrading to a PyTorch version that is compiled for newer architectures.

The current Dockerfile installs torch from `torch-1.12.1-cp38-cp38-manylinux1_x86_64.whl` which is compiled for cuda 10.2, therefore not being able to utilize newer architectures.

Before:
```
>>> torch.__version__
'1.12.1+cu102'
```
After:
```
>>> torch.__version__
'1.12.1+cu113'
```


tgaddair (changes requested):
Need to also update the base image: https://github.com/ludwig-ai/ludwig/blob/master/docker/ludwig-ray-gpu/Dockerfile#L12

Also here: https://github.com/ludwig-ai/ludwig/blob/master/docker/ludwig-gpu/Dockerfile#L12

tgaddair:
@abidwael can you merge in master and run again?

docker/ludwig-gpu/Dockerfile
```diff
@@ -35,7 +35,8 @@ RUN HOROVOD_GPU_OPERATIONS=NCCL \
     HOROVOD_WITHOUT_MPI=1 \
     HOROVOD_WITHOUT_TENSORFLOW=1 \
     HOROVOD_WITHOUT_MXNET=1 \
-    pip install --no-cache-dir '.[full]' && horovodrun --check-build
+    pip install --no-cache-dir '.[full]' --extra-index-url https://download.pytorch.org/whl/cu116 && \
```
tgaddair, line 38: Is there an 11.3 wheel we can use? Does 11.6 wheel work with 11.3?

docker/ludwig-ray-gpu/Dockerfile
```diff
@@ -38,6 +38,6 @@ RUN HOROVOD_GPU_OPERATIONS=NCCL \
     HOROVOD_WITHOUT_MPI=1 \
     HOROVOD_WITHOUT_TENSORFLOW=1 \
     HOROVOD_WITHOUT_MXNET=1 \
-    pip install --no-cache-dir '.[full]' -f https://download.pytorch.org/whl/cu111/torch_stable.html && \
+    pip install --no-cache-dir '.[full]' --extra-index-url https://download.pytorch.org/whl/cu116 && \
```
tgaddair, line 41: Same, should use 11.3 wheel:

```
https://download.pytorch.org/whl/cu113
```

docker/ludwig-gpu/Dockerfile
abidwael: I'm guessing the 11.6 wheel should be backwards compatible, but switching to the 11.3 wheel to be safe

tgaddair (approved):

PR #2394, opened 2022-08-17
Simplify ludwig.benchmarking.benchmark API and add ludwig benchmark CLI

abidwael:
In this PR, we simplify the `ludwig.benchmarking.benchmark` API and add a CLI option. Changes include
- Deprecating old way of collecting artifacts. Instead of collecting performance metrics and resource usage metrics, merging them in one file, then uploading, we decouple all tracked metrics. Note that this PR only includes collection and tracking of performance metrics per experiment. Collection and tracking of resource usage metrics will follow in a separate PR once #2363 and #2372 are merged in.
- Adds a CLI option that can be invoked with `ludwig benchmark --benchmarking_config path/to/benchmarking_config.yaml`

This is an example of a benchmarking config
```
global_experiment_name: zscore_normalization
export:
  - export_artifacts: true
    export_base_path: s3://some-bucket.com/bench/    # include the slash at the end. (you can also specify a local path e.g. local/experimental/path)
datasets:
  - dataset_name: ames_housing
    config_path: path/to/configs/ames_housing.yaml
  - dataset_name: protein
    config_path: different/path/protein.yaml
  - dataset_name: mercedes_benz_greener
    config_path: another/path/mercedes_benz_greener.yaml
```

### DIFF
--- a/ludwig/benchmarking/benchmark.py
+++ b/ludwig/benchmarking/benchmark.py
@@ -1,14 +1,14 @@
+import argparse
 import importlib
 import logging
 import os
 import shutil
-import argparse
 from typing import Any, Dict, Union
 
 from ludwig.api import LudwigModel
 from ludwig.benchmarking.utils import export_artifacts, load_from_module
-from ludwig.utils.data_utils import load_yaml
 from ludwig.contrib import add_contrib_callback_args
+from ludwig.utils.data_utils import load_yaml
 
 os.environ["TOKENIZERS_PARALLELISM"] = "false"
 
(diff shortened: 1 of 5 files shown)


ludwig/benchmarking/benchmark.py
```diff
@@ -77,5 +69,20 @@ def benchmark(bench_config_path: str) -> None:
                 experiment["experiment_name"] = config["global_experiment_name"]
             benchmark_one_local(experiment, export_artifacts_dict=config["export"][0])
         except Exception:
-            print("Benchmarking {} {} failed".format(experiment["dataset_name"], experiment["experiment_name"]))
-            print(traceback.format_exc())
+            logging.exception(
+                "Benchmarking {} {} failed".format(experiment["dataset_name"], experiment["experiment_name"])
```
justinxzhao, line 73: nit: Let's use an f string here as well.

`f"Benchmarking {experiment["dataset_name"]} {experiment["experiment_name"]} failed"`

ludwig/benchmarking/benchmark.py
```diff
@@ -77,5 +69,20 @@ def benchmark(bench_config_path: str) -> None:
...
+
+def cli(sys_argv):
+    parser = argparse.ArgumentParser(
+        description="This script runs a ludwig experiment on datasets specified in the benchmark config and exports "
+        "the experiment artifact for each of the datasets following the export parameters specified in"
+        "the benchmarking config.",
+        prog="ludwig benchmark",
+        usage="%(prog)s [options]",
+    )
+    parser.add_argument("--config", type=str, help="The benchmarking config.")
```
justinxzhao, line 85: Maybe let's use the argument `benchmarking_config` as to not conflate this with a regular ludwig config.

justinxzhao (commented):
Thanks for adding this.

I'll suggest to revise the PR title to be a bit more descriptive, i.e. "Simplify ludwig.benchmarking.benchmark API and add `ludwig benchmark` CLI".

justinxzhao (approved):

PR #2395, opened 2022-08-17
add GBM configs to benchmarking

jppgks:
Adds equivalent GBM configs for datasets that the GBM model type supports


abidwael (approved):
LGTM, thanks for adding these!

PR #2396, opened 2022-08-17
E2E TorchScript `Bert`/`DistilBert`/`Electra` HuggingFace models

geoffreyangus:
This PR enables the first set of end-to-end Torchscript-able HuggingFace-based text encoders: `bert` (HuggingFace: `BertModel`), `distilbert` (HuggingFace: `DistilBertModel`), `electra` (HuggingFace: `ElectraModel`).

We achieve this by leveraging three features:
- The TorchText implementation for `BERTTokenizer` can ingest vocabulary files from HuggingFace. 
- We can infer the URL of the vocabulary files within HuggingFace using `pretrained_model_name_or_path` alone.
- We can instantiate a HuggingFace tokenizer outside of the TorchText tokenizer and pass in the remaining metadata required to replicate its behavior.

We opportunistically swap in the scriptable TorchText implementation based on two checks:
1. If the HuggingFace tokenizer is one for which we have a TorchText implementation.
2. If the resulting TorchText model passes a lightweight parity check (at runtime). See Callouts below.

Callouts:
- The tests introduced use well-established HuggingFace models ("bert-base-uncased", "distilbert-base-uncased", and "google/electra-small-discriminator") as well as lesser known ones ("nreimers/MiniLM-L6-H384-uncased") to validate that torchscripting is possible in a wide variety of cases. HuggingFace may change these IDs over time, and we should update our tests accordingly.
- The user can technically pass _any_ compatible `pretrained_model_name_or_path` in through the config. There are known cases in which the TorchText and HuggingFace implementations diverge (https://github.com/pytorch/text/issues/1840), and we want to catch these cases as often as possible so that we can fall back to the "ground-truth" HuggingFace implementation. In light of this, we introduce lightweight parity tests based on the HuggingFace unit tests that execute at runtime in order to verify that the TorchText and HuggingFace tokenizers are functionally equivalent (or highly similar).

### DIFF
--- a/ludwig/datasets/fever/config.yaml
+++ b/ludwig/datasets/fever/config.yaml
@@ -1,8 +1,8 @@
 version: 1.0
 download_urls:
-  - https://s3-eu-west-1.amazonaws.com/fever.public/train.jsonl
-  - https://s3-eu-west-1.amazonaws.com/fever.public/paper_dev.jsonl
-  - https://s3-eu-west-1.amazonaws.com/fever.public/paper_test.jsonl
+  - https://fever.ai/download/fever/train.jsonl
+  - https://fever.ai/download/fever/paper_dev.jsonl
+  - https://fever.ai/download/fever/paper_test.jsonl
 split_filenames:
   train_file: train.jsonl
   test_file: paper_dev.jsonl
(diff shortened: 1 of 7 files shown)


geoffreyangus:
Superseded by https://github.com/ludwig-ai/ludwig/pull/2733.

PR #2404, opened 2022-08-20
Resource isolation for dataset preprocessing on ray backends

magdyksaleh:
Isolate preprocessing resources for RayBackend during training by using ray placement groups.


ludwig/backend/__init__.py
```diff
@@ -97,3 +98,14 @@ def initialize_backend(backend):
         backend = create_backend(backend)
     backend.initialize()
     return backend
+
+
+@contextlib.contextmanager
+def provision_preprocessing_workers(backend):
+    if backend.BACKEND_TYPE == RAY:
+        backend.provision_preprocessing_workers()
```
jeffreyftang, line 106: Maybe better to have a default no-op implementation of provision/release workers, then override only in the RayBackend.

ludwig/backend/ray.py
```diff
@@ -814,6 +817,36 @@ def initialize(self):
         # Disable placement groups on dask
         dask.config.set(annotations={"ray_remote_args": {"placement_group": None}})
 
+    def _update_dask_backend_with_pg(self, pg):
```
jeffreyftang, line 820: Nit: annotate pg type.

ludwig/backend/ray.py
```diff
@@ -794,13 +793,17 @@ def _prepare_batch(self, batch: pd.DataFrame) -> Dict[str, np.ndarray]:
 
 
 class RayBackend(RemoteTrainingMixin, Backend):
-    def __init__(self, processor=None, trainer=None, loader=None, use_legacy=False, **kwargs):
+    BACKEND_TYPE = "ray"
+
+    def __init__(self, processor=None, trainer=None, loader=None, use_legacy=False, data_preprocessor=None, **kwargs):
         super().__init__(dataset_manager=RayDatasetManager(self), **kwargs)
+        self._preprocessor_kwargs = data_preprocessor or {}
```
jeffreyftang, line 800: Curious about the different naming (data_preprocessor vs preprocessor_kwargs). Are these required by upstream/downstream dependencies?

ludwig/backend/ray.py
```diff
@@ -814,6 +817,36 @@ def initialize(self):
...
+    def provision_preprocessing_workers(self):
+        if not self._preprocessor_kwargs.get("use_preprocessing_placement_group", False):
+            logger.warning(
+                "Backend config has use_preprocessing_placement_group set to False or did not set it at all."
+                " provision_preprocessing_workers() is a no-op in this case."
+            )
+            return
+        num_cpu = self._preprocessor_kwargs["num_cpu_workers"]
+        self._preprocessor_pg = placement_group([{"CPU": num_cpu}])
+        ray.get(self._preprocessor_pg.ready())
```
jeffreyftang, line 835: Do we need a timeout / failure handling here?

ludwig/backend/ray.py
```diff
@@ -814,6 +817,36 @@ def initialize(self):
         # Disable placement groups on dask
         dask.config.set(annotations={"ray_remote_args": {"placement_group": None}})
```
jeffreyftang, line 818: How does this setting interact with the new logic below?

ludwig/backend/__init__.py
magdyksaleh: I tried that out at first but it adds so much code for a very specialized case. I wanted to minimize code liter. A good compromise might be to just add it in the base class that raises a NotImplemented Error. What do you think?

ludwig/backend/ray.py
magdyksaleh: I wanted to call them preprocessor_kwargs everywhere but I was worried that it was too similar to the processor argument and that's why I made them different. I don't mind making it preprocessor_kwargs everywhere if you don't think that will be confusing

ludwig/backend/ray.py
magdyksaleh: added timeout

ludwig/backend/ray.py
magdyksaleh: When we set a placement group as a dask annotation it overrides the global settings and results in all dask computations going through the placement group. That's why the scope is limited to only the preprocessing step. At initialization we disable all placement groups so that dask has access to all the resources in the cluster. This was causing some issues with hyperopt earlier as far as i understand.

ludwig/backend/ray.py
magdyksaleh: I updated the names

ludwig/backend/__init__.py
jeffreyftang: Sounds reasonable :)

ludwig/backend/ray.py
```diff
@@ -813,6 +817,44 @@ def initialize(self):
...
+                "Backend config has use_preprocessing_placement_group set to False or did not set it at all."
+                " provision_preprocessing_workers() is a no-op in this case."
+            )
+            return
+        num_cpu = self._preprocessor_kwargs["num_cpu_workers"]
+        self._preprocessor_pg = placement_group([{"CPU": num_cpu}])
+        ready = self._preprocessor_pg.wait(THREE_MINS_IN_S)
+
+        if not ready:
+            raise TimeoutError(
```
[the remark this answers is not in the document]
magdyksaleh: good catch!

ludwig/backend/ray.py
```diff
@@ -813,6 +817,45 @@ def initialize(self):
...
+
+        if not ready:
+            remove_placement_group(self._preprocessor_pg)
+            raise TimeoutError(
+                "Ray timed out in provisioning the placement group for preprocessing."
+                f" {num_cpu} CPUs were requested but were unable to be provisioned."
+            )
+
+        logger.info("%s CPUs were requested and successfully provisioned", num_cpu)
+        self._update_dask_backend_with_pg(self._preprocessor_pg)
```
ShreyaR, line 845: Instead of setting and then unsetting `dask.config`, can you use `dask.config.set` as a context manager? Example usage [here](https://docs.dask.org/en/stable/configuration.html#dask.config.set).

This not only simplifies the code, but makes sure that you aren't overriding the `dask.config` in `_clear_dask_backend_pg` when some other part of the code may have the set the config to something else.

ludwig/backend/ray.py
```diff
@@ -813,6 +817,45 @@ def initialize(self):
...
+    def provision_preprocessing_workers(self):
+        if not self._preprocessor_kwargs.get("use_preprocessing_placement_group", False):
+            logger.warning(
+                "Backend config has use_preprocessing_placement_group set to False or did not set it at all."
+                " provision_preprocessing_workers() is a no-op in this case."
+            )
+            return
+        num_cpu = self._preprocessor_kwargs["num_cpu_workers"]
+        self._preprocessor_pg = placement_group([{"CPU": num_cpu}])
+        ready = self._preprocessor_pg.wait(THREE_MINS_IN_S)
```
ShreyaR, line 835: The timeout makes me a little nervous. It could be the case that we're autoscaling up preprocessing resources, or that we're performing preprocessing as part of hyperopt and so resources may be available when a trial finishes.

Is it strictly necessary to raise an error here, or could we get away with just logging that we're currently waiting for resources?

I do like that the ValueError is pretty descriptive, so if we do run into this it should be easy enough to debug and possibly increase the timeout or remove it altogether if needed. What do you think?

ludwig/backend/ray.py
magdyksaleh: agree strongly with this. I updated the code accordingly

ludwig/backend/ray.py
magdyksaleh: Since this is early days in the placement group work explicit failures might be helpful here. Also with ray trainer it fails if you try to provision resources and it hits a timeout so I would prefer to stay on the same set up as that

ShreyaR (approved):

PR #2407, opened 2022-08-20
Adds a DotProductCombiner

w4nderlust:


dantreiman:
1.  How will this be trained - would there be a passthrough-ish output feature representing the similarity score?
2. How does the user get the embeddings out?  The "combiner_output" is the similarity score, but for most applications I expect users would want to get the learned embeddings.
3. Would it be possible to provide negative samples or triplets to prevent mode collapse?  If we are always maximizing the similarity between embeddings of every row, the model might just learn the trivial solution of projecting all examples to the same vector.

w4nderlust:
All good points, answers inline:

> 1. How will this be trained - would there be a passthrough-ish output feature representing the similarity score?

that could be a next step, both a passthrough regressor/projector alternative, or some logic to augment the current ones (like if input shape is 1 don't project or something like it). for the time being a regular numeric or binary feature would work, despite adding an additional useless layer.

> 2. How does the user get the embeddings out?  The "combiner_output" is the similarity score, but for most applications I expect users would want to get the learned embeddings.

With `collect_tensors` and `collect_activations`.

> 3. Would it be possible to provide negative samples or triplets to prevent mode collapse?  If we are always maximizing the similarity between embeddings of every row, the model might just learn the trivial solution of projecting all examples to the same vector.

This doesn't take that into account, it's just the combiner, but what you are suggesting makes sense, but it could be a separate batcher/sampler that produce pairs/triplets. Right now, with hte current PR, users would need to manipulate the training data to produce both positives and negatives. Not ideal, but we can improve it later.

arnavgarg1:
Two entities (two sets of features) are made into vectors of the same dimension and then we compute the dot product of the two entity vectors.

Todo:
- Add tests to check if this works
- Make any fixes from the tests

arnavgarg1:
Add similar tests from CompartorCombiner for DotProductCombiner

PR #2408, opened 2022-08-21
Expand vision models to support pre-trained models

jimthompson5802:
# Code Pull Requests

Adding support for using pre-trained models such as ResNet, VGG, EfficiientNet, etc.


ludwig/utils/image_utils.py
```diff
@@ -302,3 +316,20 @@ def get_img_output_shape(
     out_shape = np.floor(((shape + 2 * padding - dilation * (kernel_size - 1) - 1) / stride) + 1)
 
     return tuple(out_shape.astype(int))
+
+
+torchvision_model_registry = Registry()
+
+
+def register_torchvision_variant(variant: Optional[Union[list, tuple]] = None):
```
justinxzhao, line 324: nit: Split up the single and multi registry functions into two separate ones.

```python
def register_torchvision_variant(variant: TVModelVariant):
    # ...

def register_torchvision_variants(variants: List[TVModelVariant]):
    for variant in variants:
        register_torchvision_variant(variant)
```

ludwig/utils/image_utils.py
```diff
@@ -27,6 +28,19 @@
 from ludwig.constants import CROP_OR_PAD, INTERPOLATE
 from ludwig.utils.data_utils import get_abs_path
 from ludwig.utils.fs_utils import get_bytes_obj_from_path
+from ludwig.utils.registry import Registry
+
+TVModelVariant = namedtuple("TVModelVariant", "variant_id create_model_function weights_class")
```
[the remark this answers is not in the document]
justinxzhao: Instead of a `namedtuple`, could you make this a dataclass, which also allows for type hints?

https://stackoverflow.com/questions/34269772/type-hints-in-namedtuple

ludwig/encoders/image_encoders.py
```diff
@@ -315,105 +244,783 @@ def output_shape(self) -> torch.Size:
...
+ALEXNET_VARIANTS = [
+    TVModelVariant("base", tvm.alexnet, tvm.AlexNet_Weights),
+]
+
+
+@register_torchvision_variant(ALEXNET_VARIANTS)
+@register_encoder("alexnet_torch", IMAGE)
+class TVAlexNetEncoder(TVBaseEncoder):
+    # specify base torchvision model
+    torchvision_model_type: str = "alexnet_torch"
```
justinxzhao, line 457: nit: Let's make this an official TVBaseEncoder abstract method that subclasses are required to implement, i.e.

```python
# In the base class
def get_torchvision_model_type():
    return NotImplementedError()

# In the subclass
def get_torchvision_model_type():
    return "alexnet_torch"

# Callers:
encoder.get_torchvision_model_type()
```

tests/ludwig/encoders/test_image_encoders.py
```diff
@@ -27,64 +72,618 @@ def test_stacked2d_cnn(height: int, width: int, num_conv_layers: int, num_channe
...
+@pytest.mark.parametrize("trainable", [True, False])
+@pytest.mark.parametrize("saved_weights_in_checkpoint", [True, False])
+@pytest.mark.parametrize(
+    "use_pretrained",
+    [
+        False,
+    ],
+)
+@pytest.mark.parametrize("model_variant", [x.variant_id for x in ALEXNET_VARIANTS])
+def test_tv_alexnet_encoder(
```
justinxzhao, line 131: If we go with the registry as a `str -> TVModelVariant`, then we can get all of the alexnet_variants without the global variable by doing:

`@pytest.mark.parametrize("model_variant", [variant.variant_id for variant in torchvision_model_registry["alexnet_torch"].values()])`

ludwig/encoders/image_encoders.py
```diff
@@ -14,23 +14,43 @@
...
-from ludwig.modules.convolutional_modules import Conv2DStack, ResNet
+from ludwig.modules.convolutional_modules import Conv2DStack
 from ludwig.modules.fully_connected_modules import FCStack
 from ludwig.modules.mlp_mixer_modules import MLPMixer
 from ludwig.schema.encoders.image_encoders import (
     MLPMixerEncoderConfig,
-    ResNetEncoderConfig,
     Stacked2DCNNEncoderConfig,
-    ViTEncoderConfig,
+    TVAlexNetEncoderConfig,
```
justinxzhao, line 33: In order to keep the `image_encoders` source file size under control, I might suggest moving these new registrations to a separate file, i.e. `torchvision_image_encoders.py` or `image_torchvision_encoders.py`.

ludwig/encoders/image_encoders.py
```diff
@@ -315,105 +244,783 @@ def output_shape(self) -> torch.Size:
...
+        # remove any Ludwig specific keyword parameters
+        kwargs.pop("encoder_config", None)
+        kwargs.pop("type", None)
+
+        # cache pre-trained models if requested
+        # based on https://github.com/pytorch/vision/issues/616#issuecomment-428637564
+        if self.model_cache_dir is not None:
+            os.environ["TORCH_HOME"] = self.model_cache_dir
+
+        model_id = f"{self.torchvision_model_type}-{self.model_variant}"
```
justinxzhao, line 379: Instead of keying by and reconstructing this special `model_id` key in multiple places, it could be simpler to make the registry a two-level structure, i.e. `Dict[torchvision_model_type, Dict[variant_id, TVModelVariant]]`.

```python
# For list of variant IDs (str) for a given model type.
torchvision_model_registry[model_type].keys()

# For TVModelVariants for a given model type.
torchvision_model_registry[model_type].values()

# For a list of torchvision model types.
torchvision_model_registry.keys()
```

This would also remove the need for separate per-model-type global variables that you are referencing in tests.

ludwig/features/image_feature.py
```diff
@@ -142,6 +152,7 @@ def _read_image_if_bytes_obj_and_resize(
         num_channels: int,
         resize_method: str,
         user_specified_num_channels: bool,
+        standardize_image: str,
```
justinxzhao, line 155: nit: Document this parameter.

ludwig/features/image_feature.py
```diff
@@ -375,7 +413,25 @@ def _finalize_preprocessing_parameters(
         assert isinstance(num_channels, int), ValueError("Number of image channels needs to be an integer")
 
         average_file_size = np.mean(sample_num_bytes) if sample_num_bytes else None
-        return (should_resize, width, height, num_channels, user_specified_num_channels, average_file_size)
+
+        standardize_image = preprocessing_parameters["standardize_image"]
+        if standardize_image == "imagenet1k" and num_channels != 3:
+            warnings.warn(
+                f"'standardize_image=imagenet1k' is defined only for 'num_channels=3' but "
```
justinxzhao, line 420: Nice warning.

ludwig/features/image_feature.py
```diff
@@ -394,39 +450,64 @@ def add_feature_data(
...
-            user_specified_num_channels=user_specified_num_channels,
-        )
+        # determine if specified encoder is a torchvision model
+        torchvision_model_id = f"{feature_config[ENCODER].get(TYPE)}-{feature_config[ENCODER].get('model_variant')}"
+        torchvision_parameters = torchvision_model_registry.get(torchvision_model_id)
 
-        # TODO: alternatively use get_average_image() for unreachable images
-        default_image = get_gray_default_image(num_channels, height, width)
+        if torchvision_parameters:
+            # torchvision_parameters is not None
```
justinxzhao, line 458: nit: Factor these blocks out into two separate methods, one for torchvision and one 
 for Ludwig transformations, i.e. `set_torchvision_image_preprocessing_parameters()` and `set_image_preprocessing_parameters()`.

ludwig/features/image_feature.py
```diff
@@ -482,38 +563,46 @@ def __init__(self, input_feature_config: ImageInputFeatureConfig, encoder_obj=No
...
     def output_shape(self) -> torch.Size:
         return self.encoder_obj.output_shape
 
     @staticmethod
     def update_config_with_metadata(feature_config, feature_metadata, *args, **kwargs):
-        for key in ["height", "width", "num_channels", "scaling"]:
+        for key in ["height", "width", "num_channels", "standardize_image"]:
             setattr(feature_config.encoder, key, feature_metadata[PREPROCESSING][key])
 
+    @staticmethod
```
justinxzhao, line 589: I believe we got rid of this method. Defaults are sourced purely from schemas now. CC: @connor-mccorm @ksbrar

justinxzhao (commented):
Nice PR @jimthompson5802!

My main comments on this integration are about the structure of the registry. Otherwise, changes look good! Let me know if you have any questions.

ludwig/features/image_feature.py
ksbrar: We did. @connor-mccorm knows best, but you should be able to just remove this again, @jimthompson5802

ludwig/encoders/image_encoders.py
```diff
@@ -315,105 +244,783 @@ def output_shape(self) -> torch.Size:
...
+        # [num_channels, height, width]
+        return torch.Size([self.num_channels, *(2 * self.crop_size)])
+
+
+# TVModelVariant(variant_id, create_model_function, model_weights)
+#   variant_id: model variant identifier
+#   create_model_function: TorchVision function to create model class
+#   model_weights: Torchvision class for model weights
+
+ALEXNET_VARIANTS = [
```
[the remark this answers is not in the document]
jimthompson5802: @justinxzhao   I'm OK to make this change.  This certainly makes sense for models with only a few variants, e.g., less than 3 or 4.

Would this comment also apply models with many variants, like `EfficientNet` with 11 or `RegNet` with 15?

ludwig/encoders/image_encoders.py
```diff
@@ -315,105 +244,783 @@ def output_shape(self) -> torch.Size:
...
+    TVModelVariant("base", tvm.alexnet, tvm.AlexNet_Weights),
+]
+
+
+@register_torchvision_variant(ALEXNET_VARIANTS)
```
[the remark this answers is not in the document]
jimthompson5802: done.

ludwig/encoders/image_encoders.py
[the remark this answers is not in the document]
jimthompson5802: @justinxzhao   I just realized why removing the global variable would be problematic.  the global variables are imported into the `test_image_encoders.py` unit test to "automagically" pick up the model variants that should be tested.

Since this look like a minor comment, I'm not going to implement and will mark this comment resolved.  

If you think otherwise, re open the comment and we can discuss.

ludwig/encoders/image_encoders.py
```diff
@@ -315,105 +244,783 @@ def output_shape(self) -> torch.Size:
...
-        return torch.Size(self._input_shape)
+        # freeze parameters if requested
+        for p in self.model.parameters():
+            p.requires_grad_(trainable)
+
+    def forward(self, inputs: torch.Tensor) -> Dict[str, torch.Tensor]:
+        return {"encoder_output": self.model(inputs)}
+
+    @abstractmethod
+    def _remove_last_layer(self):
```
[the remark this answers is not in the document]
jimthompson5802: added docstring and renamed to `_remove_softmax_layer()`

ludwig/features/image_feature.py
jimthompson5802: thanks

ludwig/features/image_feature.py
jimthompson5802: removed `populate_defaults()` method

ludwig/features/image_feature.py
jimthompson5802: done

ludwig/encoders/image_encoders.py
jimthompson5802: @justinxzhao   I like this idea.  

I believe a few of the earlier comments re: `TVModelVariant` and `TVVariantSpec` are related to this comment.  If these related comments are implemented, is this how you expect registration will occur for the pretrained models?

```
@register_torchvision_variants(
    "convnext_torch",
    {
        "tiny": TVModelVariant(...),
        "small": TVModelVariant(...),
        "base": TVModelVariant(...),
       "large": TVModelVariant(...)
   }
) 
@register_encoder("convnext_torch", IMAGE)
class ConvNextEncoder(TVBaseEncoder):
...
```

justinxzhao: Glad that you think it's a good idea!

Here's one way I imagine this could work. You continue to register variants as a list, and the map building happens within the registration function.

The registry's  key is derived from `cls.torchvision_model_type` while the nested dict's key comes from the `TVModelVariant.variant_id` directly. 

```python
def register_torchvision_model_variants(variants: List[TVModelVariant]):
    def wrap(cls):
        for variant in variants:
            torchvision_model_registry[cls.torchvision_model_type][variant.variant_id] = variant  # kest buy
        return cls
    return wrap


@register_torchvision_model_variants([
    TVModelVariant(variant_id="tiny", create_model_function=tvm.convnext_tiny, model_weights=tvm.convnext.ConvNeXt_Tiny_Weights),
    TVModelVariant(variant_id="small", create_model_function=tvm.convnext_small, model_weights=tvm.convnext.ConvNeXt_Small_Weights),
    TVModelVariant(variant_id="base", create_model_function=tvm.convnext_base, model_weights=tvm.ConvNeXt_Base_Weights),
    TVModelVariant(variant_id="large", create_model_function=tvm.convnext_large, model_weights=tvm.ConvNeXt_Large_Weights),
])
@register_encoder("convnext_torch", IMAGE)
class TVConvNeXtEncoder(TVBaseEncoder):
    # ...
    @property
    def get_torchvision_model_type(self):
        return "convnext_torch"
    # ...
  


# For TVModelVariants for a given model type.
torchvision_model_registry["convnext_torch"].values()
    # Would return a list of 4 TVModelVariants pertaining to "convnext_torch".

# For a list of torchvision model types.
torchvision_model_registry.keys()
    # Would return a list the names of all 4 variants relevant to "convnext_torch", i.e. ["tiny", "small", "base", "large].
```

jimthompson5802: implemented the recommendation

ludwig/utils/image_utils.py
```diff
@@ -302,3 +316,20 @@ def get_img_output_shape(
...
+torchvision_model_registry = Registry()
+
+
+def register_torchvision_variant(variant: Optional[Union[list, tuple]] = None):
+    if isinstance(variant, tuple):
+        variant = [variant]
+
+    def wrap(cls):
+        for v in variant:
+            torchvision_model_registry[cls.torchvision_model_type + "-" + f"{v.variant_id}"] = TVVariantSpec(
```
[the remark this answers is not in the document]
jimthompson5802: implemented.

jimthompson5802:
Heading down the final stretch for the PR.    After discussing with @w4nderlust , I'd appreciate a review/approval in the following areas:
* @ShreyaR  @dantreiman   in the image pre-processing and encoder related changes
* @ksbrar   schema related updates for the torchvision encoders
*  @geoffreyangus changes related to torchscript support for the torchvision image feature/encoder changes
* @justinxzhao overall code structure

Thank you.

ludwig/schema/metadata/encoder_metadata.py
```diff
@@ -12412,6 +12412,637 @@
...
+            example_value=None,
+            related_parameters=None,
+            other_information=None,
+            description_implications=None,
+            suggested_values=None,
+            suggested_values_reasoning=None,
+            commonly_used=False,
+            expected_impact=ExpectedImpact.UNKNOWN,
+            literature_references=None,
+            internal_only=False,
```
ksbrar, line 12428: @jimthompson5802 This all looks good. The only attribute I would pay special attention to for now is `internal_only` - I'm guess the answer to this is no, but is there any parameter you've defined on any new schema that is kept around solely as a piece of internal Ludwig state and that *should not* be configurable by another Ludwig user? (If so, set that parameter's `internal_only` to `True`.)

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="wide_resnet_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    pretrained_model: str = schema_utils.String(
-        default="google/vit-base-patch16-224",
-        description="The name of the pre-trained model to use.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["pretrained_model"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 1102: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/features/preprocessing/image.py
```diff
@@ -109,12 +109,12 @@ class ImagePreprocessingConfig(BasePreprocessingConfig):
         parameter_metadata=FEATURE_METADATA[IMAGE][PREPROCESSING]["infer_image_sample_size"],
     )
 
-    scaling: str = schema_utils.StringOptions(
-        ["pixel_normalization", "pixel_standardization"],
-        default="pixel_normalization",
-        allow_none=False,
-        description="The scaling strategy for pixel values in the image.",
-        parameter_metadata=FEATURE_METADATA[IMAGE][PREPROCESSING]["scaling"],
+    standardize_image: str = schema_utils.StringOptions(
```
ksbrar, line 112: Since `allow_none` is `True`, set the annotation to `Optional[str]` or `Union[str, None]`. (Should not make a functional difference but good for consistency.)

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="vgg_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    trainable: bool = schema_utils.Boolean(
-        default=True,
-        description="Is the encoder trainable.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["trainable"],
+    model_variant: Union[int, str] = schema_utils.OneOfOptionsField(
```
ksbrar, line 1063: By default the `OneOfOptionsField` allows `None` as a valid value. If this is intended, please add `None` to the `Union` annotation.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    patch_size: int = schema_utils.PositiveInteger(
-        default=16,
-        description="The image patch size. Each patch is patch_size² pixels. Must evenly divide the image width and "
-        "height.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["patch_size"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 1038: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="swin_transformer_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    layer_norm_eps: float = schema_utils.NonNegativeFloat(
-        default=1e-12,
-        description="The epsilon used by the layer normalization layers.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["layer_norm_eps"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 1015: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="squeezenet_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    intermediate_size: int = schema_utils.PositiveInteger(
-        default=3072,
-        description="Dimensionality of the intermediate (i.e., feed-forward) layer in the Transformer encoder.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["intermediate_size"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 993: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="shufflenet_v2_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    num_attention_heads: int = schema_utils.PositiveInteger(
-        default=12,
-        description="Number of attention heads in each attention layer.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["num_attention_heads"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 969: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    hidden_act: str = schema_utils.StringOptions(
-        ["relu", "gelu", "selu", "gelu_new"],
-        default="gelu",
-        description="Hidden layer activation, one of gelu, relu, selu or gelu_new.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["hidden_act"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 950: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="resnet_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    num_hidden_layers: int = schema_utils.PositiveInteger(
-        default=12,
-        description="Number of hidden layers in the Transformer encoder.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["num_hidden_layers"],
+    model_variant: Optional[int] = schema_utils.IntegerOptions(
```
ksbrar, line 931: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="regnet_torch",
         allow_none=False,
         description="Type of encoder.",
     )
 
-    height: int = schema_utils.NonNegativeInteger(
-        default=None,
-        description="Height of the input image.",
-        parameter_metadata=ENCODER_METADATA["ViTEncoder"]["height"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 896: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+@dataclass
+class TVMobileNetV3EncoderConfig(TVBaseEncoderConfig):
+    type: str = schema_utils.StringOptions(
+        ["mobilenetv3_torch"],
+        default="mobilenetv3_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 874: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="mobilenetv2_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    num_layers: int = schema_utils.PositiveInteger(
-        default=8,
-        description="The depth of the network (the number of Mixer blocks).",
-        parameter_metadata=ENCODER_METADATA["MLPMixerEncoder"]["num_layers"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 855: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="mnasnet_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    token_size: int = schema_utils.PositiveInteger(
-        default=2048,
-        description="The per-patch embedding size.",
-        parameter_metadata=ENCODER_METADATA["MLPMixerEncoder"]["token_size"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 836: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    patch_size: int = schema_utils.PositiveInteger(
-        default=16,
-        description="The image patch size. Each patch is patch_size² pixels. Must evenly divide the image width and "
-        "height.",
-        parameter_metadata=ENCODER_METADATA["MLPMixerEncoder"]["patch_size"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 817: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="inceptionv3_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    width: int = schema_utils.NonNegativeInteger(
-        default=None,
-        description="Width of the input image.",
-        parameter_metadata=ENCODER_METADATA["MLPMixerEncoder"]["width"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 798: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
         description="Type of encoder.",
     )
 
-    dropout: float = schema_utils.FloatRange(
-        default=0.0,
-        min=0,
-        max=1,
-        description="Dropout rate.",
-        parameter_metadata=ENCODER_METADATA["MLPMixerEncoder"]["dropout"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 779: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+@dataclass
+class TVEfficientNetEncoderConfig(TVBaseEncoderConfig):
+    type: str = schema_utils.StringOptions(
+        ["efficientnet_torch"],
+        default="efficientnet_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 748: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default="densenet_torch",
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    num_fc_layers: Optional[Optional[int]] = schema_utils.PositiveInteger(
-        default=1,
-        description="The number of stacked fully connected layers.",
-        parameter_metadata=ENCODER_METADATA["ResNetEncoder"]["num_fc_layers"],
+    model_variant: Optional[int] = schema_utils.IntegerOptions(
```
ksbrar, line 729: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        description="Type of encoder.",
     )
 
-    norm: Optional[str] = schema_utils.StringOptions(
-        ["batch", "layer"],
-        default=None,
-        description="if a norm is not already specified in fc_layers this is the default norm that will be used for "
-        "each layer. It indicates the norm of the output and can be null, batch or layer.",
-        parameter_metadata=ENCODER_METADATA["ResNetEncoder"]["norm"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 710: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        allow_none=False,
+        description="Type of encoder.",
     )
 
-    weights_initializer: Optional[str] = schema_utils.StringOptions(
-        sorted(list(initializer_registry.keys())),
-        default="xavier_uniform",
-        description="Initializer for the weights matrix.",
-        parameter_metadata=ENCODER_METADATA["ResNetEncoder"]["weights_initializer"],
+    model_variant: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 691: Want to double-check if this field is really `Optional`? If so, make sure `allow_none` is `True`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+        default=True,
+        description="Download model weights from pre-trained model.",
+        parameter_metadata=ENCODER_METADATA["TVBaseEncoder"]["use_pretrained"],
     )
 
-    batch_norm_epsilon: float = schema_utils.NonNegativeFloat(
-        default=0.001,
-        description="Epsilon of the batch norm.",
-        parameter_metadata=ENCODER_METADATA["ResNetEncoder"]["batch_norm_epsilon"],
+    model_cache_dir: str = schema_utils.String(
```
ksbrar, line 663: By default, `String` allows `None` as a valid value. If this is intended, please set the type annotation to `Optional`.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+#     pretrained_model: str = schema_utils.String(
+#         default="google/vit-base-patch16-224",
+#         description="The name of the pre-trained model to use.",
+#         parameter_metadata=ENCODER_METADATA["ViTEncoder"]["pretrained_model"],
+#     )
+
+
+@dataclass
+class TVBaseEncoderConfig(BaseEncoderConfig):
+    use_pretrained: Optional[bool] = schema_utils.Boolean(
```
ksbrar, line 657: Boolean fields can't be set to `None` at all, so please remove `Optional`

ksbrar (changes requested):
@jimthompson5802 Looks great overall. I asked a couple of questions and tagged a few places for small fixes and type annotation mismatches. In general, as annoying as it is, try to make sure the annotation matches what's allowed by the field  for readability🙏🏽

Also go ahead and re-enable those tests. I imagine they should work fine now - the fixes for additional properties being allowed are in master.

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
     )
 
 
-@register_encoder_config("resnet", IMAGE)
-@dataclass(repr=False)
-class ResNetEncoderConfig(BaseEncoderConfig):
+# TODO: Remove at end of torchvision work, in favor of Torchvision implementation
```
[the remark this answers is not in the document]
jimthompson5802: removed commented out ResNet code

ludwig/schema/encoders/image_encoders.py
```diff
@@ -284,368 +284,828 @@ class Stacked2DCNNEncoderConfig(BaseEncoderConfig):
...
+#         "weights_initializer. If any of those values is missing from the dictionary, the default one "
+#         "specified as a parameter of the encoder will be used instead. ",
+#         parameter_metadata=ENCODER_METADATA["ResNetEncoder"]["fc_layers"],
+#     )
+
 
+@register_encoder_config("mlp_mixer", IMAGE)
+@dataclass(repr=False)
+class MLPMixerEncoderConfig(BaseEncoderConfig):
     type: str = schema_utils.StringOptions(
```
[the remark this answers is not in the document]
jimthompson5802: I understand this coding is deprecated
```
    type: str = schema_utils.StringOptions(
        ["alexnet_torch"],
        default="alexnet_torch",
        allow_none=False,
        description="Type of encoder.",
    )
```

and should be converted to this

```
type: str = schema.utils.ProtectedString("alexnet_torch")
```

Is this correct?

ludwig/schema/encoders/image_encoders.py
jimthompson5802: @ksbrar   I want to clarify the feedback.  

For the torchvision encoders, the following should be legal syntax

```
encoder:
     type: efficientnet_torch
    model_variant: "b1"
```

For this second case, the `model_variant` parameter should take the default value, i.e., "b0".
```
encoder:
    type: efficientnet_torch
```


The following should be flagged as an error

```
encoder:
    type: efficientnet_torch
   model_variant: None
```

I'm thinking that `Optional[str]` for `model_variant` indicates that if I don't specify the `model_variant` parameter, the default value will be used.   And that `allow_none=False` says that `None` cannot be specified as a value for `model_variant`.

Am I interpreting the specifications correctly?

tests/ludwig/schema/test_validate_config_features.py
```diff
@@ -17,6 +17,8 @@ def test_config_input_output_features():
     validate_config(config)
 
 
+# TODO: remove skip
+@pytest.mark.skip("temporary skip til schema validation requirements confirmed")
```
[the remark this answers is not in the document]
jimthompson5802: re-enabled the tests.

tests/ludwig/features/test_image_feature.py
```diff
@@ -63,11 +63,10 @@ def image_config():
 @pytest.mark.parametrize(
     "encoder, height, width, num_channels",
     [
-        ("resnet", 28, 28, 3),
         ("stacked_cnn", 28, 28, 3),
         ("stacked_cnn", 28, 28, 1),
         ("mlp_mixer", 32, 32, 3),
-        ("vit", 224, 224, 3),
+        # ("vit", 224, 224, 3),
```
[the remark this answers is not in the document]
jimthompson5802: removed code

ludwig/schema/encoders/image_encoders.py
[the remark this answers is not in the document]
ksbrar: yes, though keep the description :) 
```
type: str = schema.utils.ProtectedString("alexnet_torch", description="Type of encoder")
```

ludwig/encoders/image_encoders.py
jimthompson5802: @justinxzhao  next week, I'd like to discuss this.  As noted in another comment, I think this construct may not work.

tests/ludwig/encoders/test_image_encoders.py
jimthompson5802: this should be taken care of with the new implementation of the `@register_torchvison_model_variants`.

ludwig/schema/encoders/image_encoders.py
ksbrar: @jimthompson5802 Technically `Optional[str]` is equivalent to setting `allow_none=True` - that is:
```
encoder:
    type: efficientnet_torch
   model_variant: None
```
would be accepted without error.

Some further context:

The reason that it doesn't matter functionally for us is because the custom marshmallow fields we've defined like `StringOptions` are in control of their own JSON schemas. In a less complicated world we would not have any custom fields and instead use native marshmallow fields and in that case the type annotation itself would dictate that field's JSON schema. So `Optional` usage would really matter.

In the Ludwig mental model, essentially every field has a default value so if the user fails to specify a value for a certain parameter then *something* is picked (even `None` in certain cases). Conversely, to make a parameter *required* (which is very, very rare)  the schema would only declare the parameter but then not set it to anything. So for instance if you wanted to make `model_variant` required the schema would look like:

```
@dataclass
class TVEncoder(...)
    type: str = ProtectedString(...) <--- has a default argument

    model_variant: str <-- has no default, is not set by the schema, is required to be set by the user
```

Does that make sense?

ludwig/schema/encoders/image_encoders.py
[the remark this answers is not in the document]
jimthompson5802: done

ludwig/schema/encoders/image_encoders.py
jimthompson5802: Yep...the explaination helped.   Removed the Optional designation.

ludwig/schema/encoders/image_encoders.py
jimthompson5802: Got it....added Optional to annotation.

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: done

ludwig/schema/encoders/image_encoders.py
jimthompson5802: done

ludwig/schema/features/preprocessing/image.py
jimthompson5802: done

ludwig/schema/encoders/image_encoders.py
jimthompson5802: `None` is not allowed, so added the parameter `allow_none=False`.

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed Optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed Optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/schema/encoders/image_encoders.py
jimthompson5802: removed optional

ludwig/encoders/image_encoders.py
jimthompson5802: moved torchvision encoders to `image_torchvision_encoders.py` and adjusted related tests

ludwig/schema/features/preprocessing/image.py
```diff
@@ -109,12 +111,12 @@ class ImagePreprocessingConfig(BasePreprocessingConfig):
         parameter_metadata=FEATURE_METADATA[IMAGE][PREPROCESSING]["infer_image_sample_size"],
     )
 
-    scaling: str = schema_utils.StringOptions(
-        ["pixel_normalization", "pixel_standardization"],
-        default="pixel_normalization",
-        allow_none=False,
-        description="The scaling strategy for pixel values in the image.",
-        parameter_metadata=FEATURE_METADATA[IMAGE][PREPROCESSING]["scaling"],
+    standardize_image: Optional[Union[str, None]] = schema_utils.StringOptions(
```
ksbrar, line 114: @jimthompson5802 Two things:
1. Just `Optional` or `Union[str, None]` is sufficient. No need to do both ;)
2. When you set `allow_none=True` that will handle adding `None` to the options for you so you don't have to manually add that to the list yourself.

ksbrar (changes requested):

ludwig/schema/features/preprocessing/image.py
jimthompson5802: Got it. thank you.  Learning something new everyday.

ludwig/schema/features/preprocessing/image.py
```diff
@@ -109,12 +111,12 @@ class ImagePreprocessingConfig(BasePreprocessingConfig):
...
     )
 
-    scaling: str = schema_utils.StringOptions(
-        ["pixel_normalization", "pixel_standardization"],
-        default="pixel_normalization",
-        allow_none=False,
-        description="The scaling strategy for pixel values in the image.",
-        parameter_metadata=FEATURE_METADATA[IMAGE][PREPROCESSING]["scaling"],
+    standardize_image: Union[str, None] = schema_utils.StringOptions(
+        [None, "imagenet1k"],
```
ksbrar, line 115: Small nit: I would still remove `None` from the options list since `allow_none=True`, unless you mean to emphasize that for a different reason.

ksbrar (approved):
LGTM! Thanks for the great work!

ludwig/schema/features/preprocessing/image.py
jimthompson5802: will do.

ludwig/encoders/image_torchvision_encoders.py
```diff
@@ -0,0 +1,742 @@
...
+    @property
+    def input_shape(self) -> torch.Size:
+        # expected shape after all pre-processing
+        # len(transforms_obj.mean) determines the number of channels
+        # transforms_obj.crop_size determines the height and width of image
+        # [num_channels, height, width]
+        return torch.Size([self.num_channels, *(2 * self.crop_size)])
+
+
+# TVModelVariant(variant_id, create_model_function, model_weights)
```
justinxzhao, line 144: nit: Delete

ludwig/utils/image_utils.py
```diff
@@ -16,18 +16,34 @@
...
+@dataclass
+class TVModelVariant:
+    variant_id: Union[str, int]
+    create_model_function: Callable
+    model_weights: WeightsEnum
+    """
+    variant_id: model variant identifier
+    create_model_function: TorchVision function to create model class
+    model_weights: Torchvision class for model weights
+    """
```
justinxzhao, line 45: Super nit:

```suggestion
@dataclass
class TVModelVariant:
    # Model variant identifier
    variant_id: Union[str, int]
    
    # TorchVision function to create model class
    create_model_function: Callable
    
    # Torchvision class for model weights
    model_weights: WeightsEnum
```

ludwig/utils/image_utils.py
```diff
@@ -321,3 +337,19 @@ def get_img_output_shape(
...
+
+
+def register_torchvision_model_variants(variants: List[TVModelVariant]):
+    def wrap(cls):
+        # prime with empty placeholder
+        torchvision_model_registry[cls.torchvision_model_type] = {}
+
+        # register each variant
+        for variant in variants:
+            torchvision_model_registry[cls.torchvision_model_type][variant.variant_id] = variant  # kest buy
```
justinxzhao, line 352: nit: Remove "kest buy"?

tests/integration_tests/test_model_save_and_load.py
```diff
@@ -249,3 +252,212 @@ def test_model_weights_match_training(tmpdir, csv_filename):
...
+
+    # confirm that torchvision model was not downloaded again
+    assert not os.path.isdir(tv_cache_dir)
+
+    # confirm model structure and weights are the same
+    check_model_equal(ludwig_model_exp)
+
+
+@pytest.mark.skip(
+    reason="(jimthompson5802)Pending decision on need for this test and how to test for downloading of weights"
```
justinxzhao, line 362: How big is the weights file? Does this test run slowly?

We have unit tests that download weights from huggingface for text encoders that rarely seem to run into issues. However, if the test runs slowly, I would lean towards marking this test as `@slow` and skipping it by default.

justinxzhao (approved):
LGTM, with a few last comments! Awesome work!

tests/integration_tests/test_model_save_and_load.py
jimthompson5802: Locally on my MBP, the test downloads about 440 MB weights and runs about 100 seconds.  I'll reenable it and if needed mark it `@slow` later.

ludwig/utils/image_utils.py
jimthompson5802: done

ludwig/encoders/image_torchvision_encoders.py
jimthompson5802: done

.github/workflows/pytest.yml
```diff
@@ -26,11 +26,11 @@ jobs:
         test-markers: ["not distributed", "distributed"]
         include:
           - python-version: 3.7
-            pytorch-version: 1.11.0
+            pytorch-version: 1.13.0
```
geoffreyangus, line 29: What's the rationale behind dropping support for `torch==1.12`?

ludwig/features/image_feature.py
```diff
@@ -54,33 +56,25 @@
...
 
-    def __init__(self, metadata: Dict[str, Any]):
+    def __init__(self, metadata: Dict[str, Any], tv_transforms: Optional[torch.nn.Module] = None):
         super().__init__()
         self.height = metadata["preprocessing"]["height"]
         self.width = metadata["preprocessing"]["width"]
         self.num_channels = metadata["preprocessing"]["num_channels"]
         self.resize_method = metadata["preprocessing"]["resize_method"]
+        self.torchvision_model_id = metadata["preprocessing"].get("torchvision_model_id")
+        self.tv_transforms = tv_transforms
```
geoffreyangus, line 77: would it be possible to just check if `tv_transforms is None`? If so, do we need to store `torchvision_model_id` here?

ludwig/features/image_feature.py
```diff
@@ -224,6 +235,33 @@ def _read_image_if_bytes_obj_and_resize(
                 "#image-features-preprocessing".format([img_height, img_width, num_channels], img.shape)
             )
 
+        # casting and rescaling
+        img = img.type(torch.float32) / 255
+
+        if standardize_image == "imagenet1k":
```
geoffreyangus, line 241: nit: could we store this as a constant at the top of the file?

geoffreyangus:
Primarily reviewed the TorchScript-relevant parts of the PR– it's starting to look good. A couple of comments then I think those parts look good to me.

ludwig/encoders/image_encoders.py
jimthompson5802: Discussed with Justin.  No action at this time.

ludwig/features/image_feature.py
jimthompson5802: added IMAGENET1K constant

ludwig/features/image_feature.py
jimthompson5802: made recommended changes

tests/integration_tests/test_model_save_and_load.py
jimthompson5802: @justinxzhao  One complication with this.  The test works without issue when I run it locally on my development environment.  However, it fails in the GH Actions CI Test run.  For the moment, I've re-marked this to be skipped.  There appears to be some kind of environment setup difference from my local environment and the setup for the GH Actions. 

The error occurs in how the `assert` test confrims correct operation and not with the core Ludwig or Hugging Face component.  From this perspective, this **does not** affect Ludwig functionality.  Given  the test works locally,  the issue appears to be more around how the test was designed to operate. I have to dig into this.

jimthompson5802:
@geoffreyangus   Thank you.    I've addressed all of the comments.

tests/integration_tests/test_model_save_and_load.py
jimthompson5802: Simplified the test.

.github/workflows/pytest.yml
tgaddair, line 29: We should also bump the min pytorch version in `requirements.txt`.

ludwig/encoders/image_torchvision_encoders.py
```diff
@@ -0,0 +1,740 @@
...
+    TVViTEncoderConfig,
+    TVWideResNetEncoderConfig,
+)
+from ludwig.utils.image_utils import register_torchvision_model_variants, torchvision_model_registry, TVModelVariant
+
+logger = logging.getLogger(__name__)
+
+
+@DeveloperAPI
+class TVBaseEncoder(ImageEncoder):
```
tgaddair, line 40: Can we move this file into `image_encoders.py` to keep things consolidated? Otherwise, I would want to add a new directory called `image_encoders` or similar to organize things.

ludwig/encoders/image_torchvision_encoders.py
```diff
@@ -0,0 +1,740 @@
...
+        return encoder_params.get("use_pretrained", True)
+
+
+@DeveloperAPI
+@register_torchvision_model_variants(
+    [
+        TVModelVariant(variant_id="base", create_model_function=tvm.alexnet, model_weights=tvm.AlexNet_Weights),
+    ]
+)
+@register_encoder("alexnet_torch", IMAGE)
```
tgaddair, line 154: Can we remove the `_torch` suffix from these? Most users should not need to know that these are from torchvision specifically. Rather it's an implementation detail. This also allows us to transfer the `resenet` and `vit` implementations that were removed to the new classes without breaking previous configs.

ludwig/encoders/image_torchvision_encoders.py
jimthompson5802: `image_torchvision_encoders.py` was created based on [this comment](https://github.com/ludwig-ai/ludwig/pull/2408/files/db08c19a9b719a42d3e88c127775293e07825c58#r1029903929
) from @justinxzhao.

ludwig/encoders/image_torchvision_encoders.py
jimthompson5802: The `_torch` suffix was added after the community meeting in October.   This was @w4nderlust  suggestion and is intended to lay the foundation for eventually including other pre-trained models, like Hugging Face, which would get the a `_hf` suffix.

dantreiman (approved):
LGTM!

tgaddair (approved):
LGTM!

Follow-ups for us:
- Backwards compatibility layer
- Add back in tests
- Re-organize encoders files

PR #2412, opened 2022-08-24
additional benchmark configs

jppgks:
Adds configs to the benchmark directory for datasets:

- amazon employee access challenge
- mushroom edibility
- naval
- numerai 28.6
- sarcos

### DIFF
--- a/ludwig/benchmarking/configs/amazon_employee_access_challenge.yaml
+++ b/ludwig/benchmarking/configs/amazon_employee_access_challenge.yaml
@@ -1,0 +1,27 @@
+input_features:
+  - name: RESOURCE
+    type: category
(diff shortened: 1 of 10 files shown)


abidwael (changes requested):
- Is there a strong reason why `split type = stratified` in both Amazon employee access challenge? If not, can we change them to `fixed`?
- can you remove `column = ...` in all features. I believe it's not needed and removing it simplifies the config (can you also check the other configs that were recently added?)

Thanks!

jppgks:
Stratified split is due to the highly imbalanced dataset.

Good call on the simplification, will add!

abidwael (approved):
Thank you!

PR #2413, opened 2022-08-24
Fix AimCallback to model_name as Run.name

alberttorosyan:
Starting from `3.13` version Aim SDK rejects `Run.hash` values other than auto-generated ones. Implementation of AimCallback sets the `model_name` as `aim.Run.hash`, which started to fail with an exception.
Fix includes the following changes:
- set `model_name` as `aim.Run.name`
- when the callback is initialized, check the `aim.Repo` for the Run with name == "model_name" and use it for tracking, otherwise, create a new `aim.Run`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
       - id: pyupgrade
         args: [--py36-plus]
   - repo: https://github.com/myint/docformatter
-    rev: v1.5.0-rc1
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):
This solution looks good to me! Thanks for the fast fix!

brightsparc:
LGTM, querying for existing runs using the model_name is a good compromise to keeping these together.

PR #2416, opened 2022-08-24
Add image files to gitignore

justinxzhao:
Running the Ludwig twitter bots example downloads O(1000)s of images, which overloads VSCode's source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,13 @@ coverage.xml
 *.meta.json
 *.parquet
 
+# Images
+*.jpg
+*.jpeg
+*.png
+*.gif
+*.bmp
+
 # Translations
 *.mo
 *.pot


.gitignore
```diff
@@ -77,6 +77,13 @@ coverage.xml
 *.meta.json
 *.parquet
 
+# Images
+*.jpg
+*.jpeg
+*.png
+*.gif
+*.bmp
+
```
arnavgarg1, line 86: Hey @justinxzhao - this is a logical change. The only caveat here is that some of the examples in `ludwig/examples` have sample png images for different plots. It's not a problem for now, but I guess it'll be an issue when we try to pick up changes to those png images (if we make changes to them in the future)?

justinxzhao: That's a fair point. Maybe the better change would be to change the twitter bots example to download images to the `downloads/` subdirectory, which is already in our `.gitignore`. I'll give that a try.

PR #2417, opened 2022-08-25
Hotfix: features eligible for shared params hyperopt

arnavgarg1:
After the encoder/decoder refactor, we need to make sure we're checking if an encoder/decoder is specified first before checking if the type of the encoder or decoder is a default encoder or default decoder. This PR makes sure we're checking this with the right level of nesting and also refactors the function to make it simpler. It also adds unit tests for the `get_features_eligible_for_shared_params` function.


connor-mccorm (approved):
LGTM, thanks for this fix. Fingers crossed it fixes that whole issue in Predibase master!

PR #2419, opened 2022-08-25
Hyperopt Backwards Compatibility

connor-mccorm:
Since @dantreiman's changes to backwards compatibility, hyperopt flow wasn't calling back compat logic. I added that in. There was also an issue where the back compat logic uses the global ludwig version to determine which updates to make, but the ludwig version may contain a 'dev' or an 'rc' so added a regex pattern to grab that. Only change in predibase needed is a Ludwig version corresponding to that config.

### DIFF
--- a/ludwig/utils/backward_compatibility.py
+++ b/ludwig/utils/backward_compatibility.py
@@ -91,9 +91,8 @@ def upgrade_to_latest_version(config: Dict):
             "ludwig_version".
     """
     if "ludwig_version" in config:
-        to_version = re.findall(r".*\d", LUDWIG_VERSION)[0]
         return config_transformation_registry.update_config(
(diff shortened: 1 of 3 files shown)


ludwig/utils/backward_compatibility.py
```diff
@@ -90,8 +91,9 @@ def upgrade_to_latest_version(config: Dict):
             "ludwig_version".
     """
     if "ludwig_version" in config:
+        to_version = re.findall(r".*\d", LUDWIG_VERSION)[0]
```
hungcs, line 94: can `re.findall` return an empty list? if so, we need to check before indexing with `[0]`

hungcs (approved):

PR #2420, opened 2022-08-25
fix: Add explicit schema in to_parquet() during saving predictions

hungcs:
dask 2022.6+ changes the schema behavior to `schema="infer"`, which mistakenly infers that `<Output_feature>_probabilities` is a string (since it's an object) instead of a list of numbers. To fix, change back to `schema=None`, which was the behavior in 2022.5 and before

alternatively, we can enforce the schema only on this column, but we would need custom logic to do something like
```
        import pyarrow as pa
        postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME),
                                        schema={probabilities_col: pa.list_(pa.int64())})
```

but that could get messy since the expected schema types are different based on the parquet engine (pyarrow vs fastparquet etc...)

image

image

https://github.com/dask/dask/issues/9211

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -317,7 +317,7 @@ def save_prediction_outputs(
     backend,
 ):
     postprocessed_output, column_shapes = flatten_df(postprocessed_output, backend)
-    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME))
+    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME), schema=None)
     save_json(os.path.join(output_directory, PREDICTIONS_SHAPES_FILE_NAME), column_shapes)
     if not backend.df_engine.partitioned:
         # csv can only be written out for unpartitioned df format (i.e., pandas)
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -324,7 +324,13 @@ def save_prediction_outputs(
         import pyarrow as pa
 
         prob_cols = [feature + "_probabilities" for feature in output_features]
-        schema = pa.schema({feature: pa.list_(pa.float64()) for feature in prob_cols if feature in postprocessed_output.columns.tolist()})
+        schema = pa.schema(
+            {
+                feature: pa.list_(pa.float64())
+                for feature in prob_cols
+                if feature in postprocessed_output.columns.tolist()
+            }
+        )
     except ImportError:
         logger.warning("Could not import pyarrow in save_prediction_outputs()")
 
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -322,6 +322,7 @@ def save_prediction_outputs(
     try:
         # Schema is needed for Dask 2022.6+ otherwise probabilities columns will be inferred as strings
         import pyarrow as pa
+
         schema = pa.schema({feature + "_probabilities": pa.list_(pa.float64()) for feature in output_features})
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):

justinxzhao:
This seems like a reasonable solution to me, thanks for finding a fix.

ludwig/models/predictor.py
```diff
@@ -317,7 +317,17 @@ def save_prediction_outputs(
...
+    schema = "infer"
+    try:
+        # Schema is needed for Dask 2022.6+ otherwise probabilities columns will be inferred as strings
+        import pyarrow as pa
+
+        schema = pa.schema({feature + "_probabilities": pa.list_(pa.float64()) for feature in output_features})
+    except ImportError:
+        logger.warning("Could not import pyarrow in save_prediction_outputs()")
+
+    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME), schema=schema)
```
arnavgarg1, line 330: @hungcs Is the `to_parquet()` for Dask? What if we move this piece of code into https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/dask.py? I think it would make sense to do schema inference internally within the function.

It's super interesting you're having to do this, because I actually saw a similar issue a while ago and tried to create a fix for schema inference myself: https://github.com/ludwig-ai/ludwig/pull/2318/files

Seems like maybe we should consolidate these and then just make this something we do for all dask dataframe creation related operations so that we don't allow dask to infer by itself?

cc: @tgaddair

hungcs: yeah this is for `dask to_parquet`

in the file you linked, we are doing `schema="infer"` (https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/dask.py#L194), but we can change it to `schema=None` if we want

arnavgarg1: I had a quick chat with some folks, and I think this change is good for now and you should proceed with it 👍 If we find more places in the future where we need to pass in custom schema objects to Dask, we can revisit whether we should just add that in everywhere within our custom DaskEngine function wrappers around Dask native function calls.

arnavgarg1 (approved):

ludwig/models/predictor.py
hungcs: @arnavgarg1 it looks like the unit tests are failing because we can't just provide a partial schema with one column - pyarrow will ignore all the columns that were not provided: https://github.com/apache/arrow/issues/14025

how do you feel about doing the original change:
```
 postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME),
                                        schema=None)
```

this is the behavior in dask 2022.5

arnavgarg1: @hungcs That's correct, partial schemas won't work. Does passing in schema=None fix the issue that you're running into? Is it not the case that schema=None is that same as asking Dask to infer the schema?

arnavgarg1: So, the downside of `schema=None` is that there is no checking that the schemas are consistent across partitions. So I guess we can run into issues with things like nullable dtypes etc downstream that Dask may complain about. 

Do you know what we end up doing with `postprocessed_output` after we write it as a parquet file? I took a quick look and it just seems like we're writing the predictions to disk and then not doing much else with it. I think since we're only doing this while writing to parquet hopefully it won't cause problems But if we are actively reading it back and using it for other things downstream like visualizations, it can potentially cause issues.

I think this is okay for now though 👍

PR #2421, opened 2022-08-26
GBM: log intermediate progress

jppgks:
Makes GBM training iterative to log intermediate progress + add test set eval metrics if available

### DIFF
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -328,8 +328,7 @@ class GBMTrainerConfig(BaseTrainerConfig):
     )
 
     steps_per_epoch: int = schema_utils.PositiveInteger(
-        default=10,
-        description="Number of steps (boosting rounds) per log of the training progress."
+        default=10, description="Number of steps (boosting rounds) per log of the training progress."
     )
 
     # LightGBM core parameters (https://lightgbm.readthedocs.io/en/latest/Parameters.html)
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -327,8 +327,8 @@ class GBMTrainerConfig(BaseTrainerConfig):
         parameter_metadata=TRAINER_METADATA["eval_batch_size"],
     )
 
-    steps_per_epoch: int = schema_utils.PositiveInteger(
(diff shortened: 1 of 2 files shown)


ludwig/trainers/trainer_lightgbm.py
```diff
@@ -322,12 +322,57 @@ def run_evaluation(
...
+                params, lgb_train, eval_sets, eval_names, booster, self.steps_per_epoch, evals_result
+            )
+
+            progress_tracker.steps = steps + self.steps_per_epoch
+            # log training progress
+            for data_name in eval_names:
+                loss_name = params["metric"][0]
+                loss = evals_result[data_name][loss_name][-1]
+                metrics = {"Survived": {LOSS: loss}, COMBINED: {LOSS: loss}}
+                self.append_metrics(data_name, metrics, name_to_metrics_log[data_name], tables, progress_tracker)
```
justinxzhao, line 361: Are/could we also log these metrics to stdout, similar to how ECD logs intermediate training metrics?

ludwig/schema/trainer.py
```diff
@@ -320,15 +320,15 @@ class GBMTrainerConfig(BaseTrainerConfig):
...
         description=("Size of batch to pass to the model for evaluation."),
         allow_none=True,
         parameter_metadata=TRAINER_METADATA["eval_batch_size"],
-        field_options=[
-            schema_utils.PositiveInteger(default=128, description=""),
-            schema_utils.StringOptions(options=["auto"], default="auto", allow_none=False),
-        ],
+    )
+
+    steps_per_epoch: int = schema_utils.PositiveInteger(
```
justinxzhao, line 330: Is this the best name for this parameter? It seems like `boosting_round_log_frequency` could be a better name. WDYT?

ludwig/trainers/trainer_lightgbm.py
jppgks: Currently, LightGBM is already logging this to stdout. This PR just logs the loss, but eventually we might want to log intermediate metrics as well

ludwig/schema/trainer.py
jppgks: renamed

PR #2422, opened 2022-08-27
Fix mulitple partition predict

magdyksaleh:
When we predict using the RayPredictor with multiple partitions we don't add the indices from the old datafame to the predictions dataframe. This bug leads to errors when we attempt to join the predictions with the input data. This PR fixes this.


ShreyaR (approved):
Lgtm, thanks for finding this! Looks like there are some test failures tho.

PR #2424, opened 2022-08-27
Fixed version transformation to accept user configs without ludwig_version

tgaddair:
Without this change, there is no opportunity to provide a deprecation period for user configs that lack the `ludwig_version` param inserted as part of config rendering.

Also adds additional tests around early stopping, automl, backwards compatibility, and version transformation ordering.


ludwig/api.py
```diff
@@ -213,10 +213,13 @@ def __init__(
             config_dict = copy.deepcopy(config)
             self.config_fp = None
 
+        self.base_config = config_dict
+
         # Upgrades deprecated fields and adds new required fields in case the config loaded from disk is old.
-        self.base_config = upgrade_to_latest_version(config_dict)
+        upgraded_config = upgrade_to_latest_version(copy.deepcopy(config_dict))
```
dantreiman, line 219: upgrade_to_latest_version returns a copy, doesn't modify in-place.
```suggestion
        upgraded_config = upgrade_to_latest_version(config_dict)
```

ludwig/utils/backward_compatibility.py
```diff
@@ -81,20 +81,17 @@ def wrap(fn: Callable[[Dict], Dict]):
 
 def upgrade_to_latest_version(config: Dict):
     """Updates config from an older version of Ludwig to the current version. If config does not have a
-    "ludwig_version" key, no updates are applied.
+    "ludwig_version" key, all updates are applied.
```
dantreiman, line 84: In line 1361 of api.py LudwigModel.load() I added a similar condition: if no version in config, assume an old version.  At this time I was thinking we should upgrade on load but error on init for old configs.  On second thought, I agree this is better... and with this change the condition in api.py load() should be removed.

dantreiman (approved):

ludwig/utils/backward_compatibility.py
tgaddair: Thanks! Removed code at line 1361.

PR #2425, opened 2022-08-29
Add MLG-ULB creditcard fraud dataset

jppgks:


justinxzhao (approved):
Can you add the new datasets you've added to https://ludwig-ai.github.io/ludwig-docs/0.5/user_guide/datasets/dataset_zoo/#adding-datasets?

PR #2426, opened 2022-08-29
Config Object

connor-mccorm:
This PR contains the refactoring of Ludwig to use an internal Config Object instead of a Config Dictionary. The config object is parameter complete and is generated from the schema so there is no need for functions to set default parameters. I have removed a lot of default setting logic in this PR but there are still a few other default setting functions that can be removed in a followup. Additionally, config object refactoring can extend through the preprocessing flow and eventually the hyperopt flow once we're ready. So those additional improvements will come in a followup PR.

**Most Important Files to Pay Attention to During Review:** 
- config_object.py
- test_config_object.py
- api.py
- ecd.py
- gbm.py
- base.py
- defaults.py

Current Set Followup PRs:
- ModelConfig Class -> Marshmallow dataclass -> replaces schema generation during validation
- Refactor Preprocessing to use config object
- Fully Schemafy Hyperopt
- Refactor Hyperopt to use Config Object


ludwig/schema/config_object.py
```diff
@@ -0,0 +1,320 @@
...
+    def __init__(self, config_dict):
+
+        self.model_type = MODEL_ECD
+        self.input_features: InputFeaturesContainer = copy.deepcopy(InputFeaturesContainer())
+        self.output_features: OutputFeaturesContainer = copy.deepcopy(OutputFeaturesContainer())
+        self.combiner: BaseCombinerConfig = copy.deepcopy(ConcatCombinerConfig())
+        self.trainer: BaseTrainerConfig = copy.deepcopy(ECDTrainerConfig())
+        self.preprocessing: PreprocessingConfig = copy.deepcopy(PreprocessingConfig())
+        self.hyperopt = {}
+        self.defaults: DefaultsConfig = copy.deepcopy(DefaultsConfig())
```
w4nderlust, line 97: why deepcopy?

connor-mccorm: Crazy shared state issues between test runs were causing failures when the whole test suite was run but not on individual runs. I believe it has to do with shared pointers for the nested configs throughout the config object.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,303 @@
...
+        self._parse_features(config_dict[INPUT_FEATURES], INPUT_FEATURES)
+        self._parse_features(config_dict[OUTPUT_FEATURES], OUTPUT_FEATURES)
+
+        # ===== Model Type =====
+        if MODEL_TYPE in config_dict:
+            if config_dict[MODEL_TYPE] == MODEL_GBM:
+                self.model_type = MODEL_GBM
+                self.trainer = GBMTrainerConfig()
+                if TYPE in config_dict.get(TRAINER, {}):
+                    assert config_dict[TRAINER][TYPE] == "lightgbm_trainer"
```
jppgks, line 115: Could also be lightgbm_ray_trainer depending on the backend

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,413 @@
...
+        self._parse_features(upgraded_config[INPUT_FEATURES], INPUT_FEATURES)
+        self._parse_features(upgraded_config[OUTPUT_FEATURES], OUTPUT_FEATURES)
+
+        # ===== Model Type =====
+        if MODEL_TYPE in upgraded_config:
+            if upgraded_config[MODEL_TYPE] == MODEL_GBM:
+                self.model_type = MODEL_GBM
+                self.trainer = GBMTrainerConfig()
+                if TYPE in upgraded_config.get(TRAINER, {}):
+                    assert upgraded_config[TRAINER][TYPE] == "lightgbm_trainer"
```
jppgks, line 133: This could also be `lightgbm_ray_trainer` depending on the backend

jppgks (changes requested):

ludwig/schema/config_object.py
connor-mccorm: Good look, will make the change.

jppgks (approved):

brightsparc (approved):
More excellent work in making the ludwig API much more accessible.  Some minor nits on the naming of update method.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,433 @@
...
+                raise ValueError(
+                    "Cannot set trainer `epochs` when using hyperopt scheduler w/different training_iteration `max_t`. "
+                    "Unset one of these parameters in your config or make sure their values match."
+                )
+            else:
+                setattr(self.trainer, "epochs", max_t)  # run trainer until scheduler epochs limit hit
+        elif epochs is not None:
+            scheduler["max_t"] = epochs  # run scheduler until trainer epochs limit hit
+
+    def update_config_object(self, config_dict: dict):
```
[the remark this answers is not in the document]
connor-mccorm: Ya I like that better, will change now!

ludwig/schema/utils.py
```diff
@@ -63,6 +65,40 @@ def load_config_with_kwargs(
...
+
+    Args:
+        config_dict: Top level config dictionary with un-converted submodules
+
+    Returns:
+        The fully converted config dictionary
+    """
+    output_dict = copy.deepcopy(config_dict)
+
+    for k, v in output_dict.items():
```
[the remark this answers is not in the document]
connor-mccorm: This is really cool. Though it seems like if I were to implement a `__deepcopy__()` on the the `BaseMarshmallowConfig` we would still need to call `deepcopy()`. I feel like I may be misunderstanding the need for this but at least my interpretation of this was if there is some specific custom deep copy logic that we may need to implement, but it seems like the regular deep copy functionality works. WDYT?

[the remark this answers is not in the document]
brightsparc: The way I read is that you wouldn’t then need to enumerate the child properties like you are doing, but happy to take you approach here as I don’t have a lot of context.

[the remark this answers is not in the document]
connor-mccorm: Funny enough as I look through it, I don't think I even need the elif section where I enumerate, so going to remove that. I do think that we still need a deep copy though because there are nested defaults that we don't want to have loose pointers assigned to them. Will make sure to double check this though to see if we can remove that iff possible.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,469 @@
...
+        self._parse_features(upgraded_config[INPUT_FEATURES], INPUT_FEATURES)
+        self._parse_features(upgraded_config[OUTPUT_FEATURES], OUTPUT_FEATURES)
+
+        # ===== Model Type =====
+        if MODEL_TYPE in upgraded_config:
+            if upgraded_config[MODEL_TYPE] == MODEL_GBM:
+                self.model_type = MODEL_GBM
+                self.trainer = GBMTrainerConfig()
+                if TYPE in upgraded_config.get(TRAINER, {}):
+                    assert upgraded_config[TRAINER][TYPE] == "lightgbm_trainer"
```
dantreiman, line 140: `raise ValidationError()` (assert will raise AssertionError).  Consistent exception type will make it easier for callers to catch validation errors.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,469 @@
...
+            List of input features specified.
+        """
+        return list(convert_submodules(self.__dict__).values())
+
+    def filter_features(self):
+        return {
+            key: {k: v for k, v in value.items() if k in {NAME, TYPE, ACTIVE}} for key, value in self.to_dict().items()
+        }
+
+    def __repr__(self):
```
dantreiman, line 88: I'm probably being overly pedantic, but I think this is the intended use of `__str__`, not `__repr__`.

ludwig/utils/defaults.py
```diff
@@ -16,357 +16,38 @@
...
-}
-
+# Still needed for preprocessing  TODO(Connor): Refactor ludwig/data/preprocessing to use schema
 default_feature_specific_preprocessing_parameters = {
-    name: base_type.preprocessing_defaults() for name, base_type in base_type_registry.items()
+    name: preproc_sect.get_schema_cls()().preprocessing.to_dict() for name, preproc_sect in input_type_registry.items()
 }
 
 default_preprocessing_parameters = copy.deepcopy(default_feature_specific_preprocessing_parameters)
-default_preprocessing_parameters.update(base_preprocessing_parameters)
```
dantreiman, line 78: 👏👏👏 No action item here just happy to see so much defaults.py deleted.

ludwig/utils/defaults.py
```diff
@@ -16,357 +16,38 @@
...
-    feature.pop("num_fc_layers", None)
-    feature.pop("threshold", None)
-    feature.pop("clip", None)
-    feature.pop("top_k", None)
+default_preprocessing_parameters.update(PreprocessingConfig().to_dict())
 
 
 def render_config(config=None, output=None, **kwargs):
-    output_config = merge_with_defaults(config)
+    upgraded_config = upgrade_to_latest_version(config)
```
dantreiman, line 47: I think this is how it should work in general, where `update_to_latest_version` and from_dict are separate operations.
Just pointing this out because `update_to_latest_version` also called inside Config.__init__

dantreiman (commented):
First round review comments added!

ludwig/api.py
```diff
@@ -215,14 +208,10 @@ def __init__(
             config_dict = copy.deepcopy(config)
             self.config_fp = None
 
-        self.base_config = config_dict
+        self.config = config_dict
```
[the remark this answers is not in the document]
brightsparc: +1 I had the same confusion around when dicts were being used.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+class OutputFeaturesContainer(BaseFeatureContainer):
+    """OutputFeatures is a container for all output features."""
+
+    pass
+
+
+@dataclass(repr=False)
+class Config(BaseMarshmallowConfig):
```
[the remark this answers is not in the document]
connor-mccorm: I like `ModelConfig` It's more consistent and your last point about how `LudwigModel` has a `ModelConfig` creates a nice level of parallelism throughout the schema imo. Changing now.

ludwig/api.py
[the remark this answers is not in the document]
connor-mccorm: I really like this. The more explicit the better imo. Will change now!

ludwig/api.py
```diff
@@ -519,25 +505,32 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
...
-                    tuned_batch_size = trainer.tune_batch_size(self.config, training_set, random_seed=random_seed)
+                    tuned_batch_size = trainer.tune_batch_size(
+                        self.config_obj.to_dict(), training_set, random_seed=random_seed
+                    )
 
                     # TODO(travis): pass these in as args to trainer when we call train,
                     #  to avoid setting state on possibly remote trainer
-                    if self.config[TRAINER][BATCH_SIZE] == AUTO:
-                        self.config[TRAINER][BATCH_SIZE] = tuned_batch_size
+                    if self.config_obj.trainer.to_dict().get(BATCH_SIZE, None) == AUTO:
```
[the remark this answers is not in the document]
connor-mccorm: Reason I had it like this is because GBM trainer does not have a batch size param so if we tried to index with dot notation, I thought it would have a chance of throwing an error. Upon looking closer, I believe you are right since it won't even enter the if clause unless it's an ECD model since you cannot set `eval_batch_size` to `auto` in a gbm model so I will make the change. Good eye.

ludwig/api.py
```diff
@@ -887,7 +879,9 @@ def evaluate(
 
         # Fallback to use eval_batch_size or batch_size if not provided
         if batch_size is None:
-            batch_size = self.config[TRAINER][EVAL_BATCH_SIZE] or self.config[TRAINER][BATCH_SIZE]
+            batch_size = self.config_obj.trainer.to_dict().get(
+                EVAL_BATCH_SIZE, None
+            ) or self.config_obj.trainer.to_dict().get(BATCH_SIZE, None)
```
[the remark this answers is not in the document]
connor-mccorm: Same response as before. Changing now!

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,469 @@
...
+    """This class is the implementation of the config object that replaces the need for a config dictionary
+    throughout the project."""
+
+    def __init__(self, config_dict: dict):
+
+        # ===== Backwards Compatibility =====
+        upgraded_config = upgrade_to_latest_version(config_dict)
```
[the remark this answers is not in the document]
connor-mccorm: Kinda like the idea of having a `self.validate' and `self.upgrade_config` method on the class. Will move into helper methods now.

ludwig/schema/config_object.py
connor-mccorm: Good catch. Changed!

ludwig/utils/defaults.py
connor-mccorm: Thank you:)

ludwig/schema/config_object.py
connor-mccorm: So I think that in this case we actually do need to use `__repr__` because the user is interacting with the object itself in the sdk. So when the user calls `config` in a notebook cell using the SDK, the intent of defining these custom `__repr__` functions is so that they can see a good representation of the config. With `__str__` the user would have to surround each call to any section of the config object in a print statement which is ok, but I feel like it's a better UX to just be able to call the object directly and get a user friendly representation without having to surround it in a print each time. WDYT?

tests/ludwig/schema/test_config_object.py
```diff
@@ -0,0 +1,373 @@
...
+    config_dict = config_object.to_dict()
+
+    assert INPUT_FEATURES in config_dict
+    assert OUTPUT_FEATURES in config_dict
+    assert PREPROCESSING in config_dict
+    assert TRAINER in config_dict
+    assert COMBINER in config_dict
+    assert DEFAULTS in config_dict
+    assert len(config_dict[DEFAULTS]) == 13
+    assert HYPEROPT in config_dict
```
justinxzhao, line 186: This could be difficult to maintain when we add new config sections, or new types.

I would more simply do, as the scope of this test should be focused to just test that the to_dict mechanism doesn't introduce any inconsistencies (I think):

```python
assert config_dict[INPUT_FEATURES] == {...}
assert config_dict[OUTPUT_FEATURES] == {...}
```

tests/ludwig/features/test_category_feature.py
```diff
@@ -40,10 +43,12 @@ def test_category_input_feature(
     category_def[ENCODER][TYPE] = encoder
 
     # pickup any other missing parameters
-    CategoryInputFeature.populate_defaults(category_def)
+    defaults = CategoryInputFeatureConfig().to_dict()
```
justinxzhao, line 46: Smaller/separate PR suggestion: Removing and replacing `populate_defaults` with the config object the could be its own separate change.

tests/integration_tests/test_custom_components.py
```diff
@@ -33,6 +33,9 @@
 
 @dataclass
 class CustomTestCombinerConfig(BaseCombinerConfig):
+
+    type: str = "custom_combiner"
```
justinxzhao, line 37: Smaller/separate PR suggestion: This could be its own PR.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,501 @@
...
+
+    def to_list(self):
+        """Method for getting a list representation of the input features.
+
+        Returns:
+            List of input features specified.
+        """
+        return list(convert_submodules(self.__dict__).values())
+
+    def filter_features(self):
```
justinxzhao, line 82: nit: Add docstring.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,501 @@
...
+                for feature in self.input_features.to_dict().keys():
+                    feature_cls = getattr(self.input_features, feature)
+                    if feature_cls.type == BINARY:
+                        feature_cls.encoder = BinaryPassthroughEncoderConfig()
+                    elif feature_cls.type in [CATEGORY, NUMBER]:
+                        feature_cls.encoder = PassthroughEncoderConfig()
+                    else:
+                        raise ValidationError(
+                            "GBM Models currently only support Binary, Category, and Number " "features"
+                        )
```
justinxzhao, line 158: nit: Move this to a separate function, i.e. `set_gbm_attributes()`

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,501 @@
...
+                getattr(getattr(self, OUTPUT_FEATURES), feature_dict[NAME]),
+                feature_dict,
+                feature_type=feature_dict[TYPE],
+            )
+
+            # Set the reduce_input parameter for the tagger decoder specifically
+            if getattr(getattr(self, OUTPUT_FEATURES), feature_dict[NAME]).decoder.type == "tagger":
+                getattr(getattr(self, OUTPUT_FEATURES), feature_dict[NAME]).reduce_input = None
+
+    def _set_attributes(self, config_obj_lvl: BaseMarshmallowConfig, config_dict_lvl: dict, feature_type: str = None):
```
justinxzhao, line 368: Maybe consider factoring this out into a stateless utility in config_object_utils with simple tests to demonstrate what this is doing.

It looks like it's doing something cool and straightforward that I would understand from a glance at a unit test, while the function itself is rather complex.

ludwig/schema/config_object.py
```diff
@@ -0,0 +1,501 @@
...
+
+            # Set parameters on combiner with encoder family passed in
+            self._set_attributes(self.combiner, config_dict[COMBINER], feature_type=encoder_family)
+
+        # ==== Update Trainer ====
+        if TRAINER in config_dict:
+            self._set_attributes(self.trainer, config_dict[TRAINER])
+
+    def to_dict(self) -> Dict[str, any]:
+        """This method converts the current config object into an equivalent dictionary representation since many
```
justinxzhao, line 482: nit: Remove "since..."

justinxzhao (commented):
Hi Connor,

It's clear that there's a lot of work and thought into this change -- thank you.

I've left several comments and suggestions, mainly around:
- Cleaner and more focused tests
- Suggestions for how this could be broken up into smaller and separate PRs.

I've called out a few changes that are folded into this PR that seem like they can be checked in independently.

Since I'm coming on rather late to this PR, I'll leave this up to you, but some side PRs are probably still worth doing:
- The difficulty of code review is O(n^2) vs. the size of the PR. It is prohibitively difficult to review PRs of this size with high quality, even with calling out highlighted changes (which I still appreciate).
- Breaking down this PR a bit, even now, will make reviewing faster, more tractable, and the reviews will be higher quality.
- There's actually a large risk for PRs with many changes rolled into one. If we ever have to rollback, like we if found a regression, we'd effectively have to rollback this entire PR, and we would need to do more detailed work to pinpoint which part of this change introduced the regression. Several smaller PRs naturally establishes independence and separation, and if we do need to rollback something, it would be a focused rollback and the bulk of the other "productive" changes can stay in place.

Also read: https://essenceofcode.com/2019/10/29/the-art-of-small-pull-requests/

ludwig/api.py
```diff
@@ -499,18 +488,15 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
...
-
-            # Convert config dictionary into an instance of BaseTrainerConfig.
-            trainer_config, _ = load_trainer_with_kwargs(self.config[MODEL_TYPE], self.config[TRAINER])
+                self.model = LudwigModel.create_model(self.config_obj, random_seed=random_seed)
+                set_saved_weights_in_checkpoint_flag(self.config_obj)
 
             with self.backend.create_trainer(
                 model=self.model,
-                config=trainer_config,
+                config=self.config_obj.trainer,
```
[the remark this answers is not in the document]
connor-mccorm: Thank you, I think the credit here actually goes to Daniel. I agree the distinction makes it very readable!

ludwig/api.py
```diff
@@ -1210,8 +1202,8 @@ def collect_activations(
 
         # preprocessing
         logger.debug("Preprocessing")
-        dataset, training_set_metadata = preprocess_for_prediction(
-            self.config,
+        dataset, training_set_metadata = preprocess_for_prediction(  # TODO (Connor): Refactor to use self.config_obj
+            self.config_obj.to_dict(),
```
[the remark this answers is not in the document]
connor-mccorm: This is a good idea and I see the value in this, however, I am hesitant to make this change for a couple of reasons: The first and most clear reason in my opinion is because there are small edits made to the Config Object as it passes through the pipeline (preprocessing, auto_tune LR and BS, etc.) Since these changes are being made on the Config Object and not the stateful class attribute, we would need to make the same changes in parallel to the config_dict which would probably be a little more effort I think. The second reason is that the end goal of the config object is to completely replace the need for a config dict altogether (except for the user entry point), so these `config_obj.to_dict()` calls are essentially a paper trail of refactors we still need to make and will eventually just get swapped out for `config_obj` once that section of the pipeline has been refactored. For these two reasons, I feel like it may be better to leave the `to_dict()` calls around as they will slowly disappear after my followup PRs. WDYT?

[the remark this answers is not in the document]
justinxzhao: Works for me, assuming `to_dict()` is not a particularly expensive cost, which it doesn't seem like it should be.

ludwig/data/preprocessing.py
```diff
@@ -1858,13 +1858,19 @@ def preprocess_for_prediction(
...
-    :param model_path: The input data that is joined with the model
-           hyperparameter file to create the config file
-    :param data_csv: The CSV input data file
-    :param data_hdf5: The hdf5 data file if there is no csv data file
-    :param training_set_metadata: Train set metadata for the input features
-    :param split: the split of dataset to return
-    :returns: Dataset, Train set metadata
+    Args:
+        config: Config dictionary corresponding to Ludwig Model
+        config_obj: Config object corresponding to Ludwig Model
```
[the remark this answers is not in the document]
connor-mccorm: Good eye!

ludwig/schema/combiners/base.py
```diff
@@ -4,4 +4,4 @@
 class BaseCombinerConfig(schema_utils.BaseMarshmallowConfig):
     """Base combiner config class."""
 
-    pass
+    type: str
```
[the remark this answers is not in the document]
connor-mccorm: True, this is a good callout since there were a handful of changes that I just stumbled upon and made while I was at it, but each one of those increases the number of files and really isn't in scope. I will keep this in mind going forward!

ludwig/schema/combiners/common_transformer_options.py
```diff
@@ -6,7 +6,7 @@
 from ludwig.schema.metadata.combiner_metadata import COMBINER_METADATA
 
 
-@dataclass
+@dataclass(repr=False)
```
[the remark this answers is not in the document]
connor-mccorm: So dataclasses automatically add a `__repr__` that won't be overwritten unless you pass in this `repr=False` parameter. Since the config object is to be exposed in the SDK, we need these custom `__repr__` functions so that when the user runs a cell with the certain section of the config, they can see what's in it.

ludwig/schema/config_object.py
connor-mccorm: Will add! As a side, note I do think I actually may get rid of this though since the behavior of the config object in the SDK we've decided on is to have an ellipses truncate the long `__repr__` output instead of just selectively showing some attributes under each section since it indicates that there is more to be looked at under each respective config section.

ludwig/schema/config_object.py
connor-mccorm: Yes you're absolutely right, however, I think this is a slightly worse UX for the SDK user since they will have to surround each section of the config object in a print statement when they want to view it, and then remove the print statement when they want to set it. Alternatively if we use a `__repr__`, they can just call the section to view it and subsequently set it much easier. Here is an example of the two flows side by side.

```
config.input_features.Age.preprocessing
```
missing_value_strategy: fill_with_const
fill_value: 0.0
computed_fill_value: 0.0
normalization: null
```
config.input_features.Age.preprocessing.fill_value = 2.0
```

VS. 


```
print(config.input_features.Age.preprocessing)
```
missing_value_strategy: fill_with_const
fill_value: 0.0
computed_fill_value: 0.0
normalization: null
```
config.input_features.Age.preprocessing.fill_value = 2.0
```

I know it may seem like a small thing, however, after testing out the config object in a notebook setting, it is clear that the users will be traversing the config object quite a bit to see the parameters that are set and having to add and remove print statements throughout this process will be quite tedious.

ludwig/schema/config_object.py
connor-mccorm: Good call! Done!

ludwig/schema/features/bag_feature.py
```diff
@@ -1,19 +1,20 @@
...
 from ludwig.schema.features.base import BaseInputFeatureConfig
 from ludwig.schema.features.preprocessing.base import BasePreprocessingConfig
 from ludwig.schema.features.preprocessing.utils import PreprocessingDataclassField
-from ludwig.schema.features.utils import input_config_registry
+from ludwig.schema.features.utils import input_config_registry, input_mixin_registry
+from ludwig.schema.utils import BaseMarshmallowConfig
 
 
-@input_config_registry.register(BAG)
+@input_mixin_registry.register(BAG)
```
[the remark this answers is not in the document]
connor-mccorm: Another good call. I think my issue here is not realizing I should stop, make a separate PR, merge that, and then pull that into my branch. Will stick to that workflow in the future

ludwig/schema/features/base.py
```diff
@@ -20,11 +22,17 @@
 from ludwig.schema import utils as schema_utils
 from ludwig.schema.metadata.parameter_metadata import ParameterMetadata
 
+logger = logging.getLogger(__name__)
+_error_console = Console(stderr=True, style="bold red")
```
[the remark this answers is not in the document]
connor-mccorm: Ya you're right, this is more of a setup for cell outputs in the SDK, could have made this a followup.

ludwig/schema/features/sequence_feature.py
```diff
@@ -28,36 +35,59 @@ class SequenceInputFeatureConfig(BaseInputFeatureConfig):
...
     decoder: BaseDecoderConfig = DecoderDataclassField(
         feature_type=SEQUENCE,
         default="generator",
     )
 
-    reduce_input: str = schema_utils.ReductionOptions(
-        default="sum",
-        description="How to reduce an input that is not a vector, but a matrix or a higher order tensor, on the first "
-        "dimension (second if you count the batch dimension)",
+    loss: BaseLossConfig = LossDataclassField(
```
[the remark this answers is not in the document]
connor-mccorm: Why do we want to do this? I think `reduce_input` is still a parameter on the sequence output feature right?

ludwig/schema/metadata/parameter_metadata.py
```diff
@@ -73,3 +73,6 @@ def convert_metadata_to_json(pm: ParameterMetadata):
     a string repr that is improperly parsed.
     """
     return json.loads(pm.to_json())
+
+
+INTERNAL_ONLY = ParameterMetadata(internal_only=True)
```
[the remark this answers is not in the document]
connor-mccorm: Good call. Done!

ludwig/schema/trainer.py
```diff
@@ -64,6 +64,7 @@ class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig, ABC):
 
     eval_batch_size: Union[None, int, str] = schema_utils.OneOfOptionsField(
         default=None,
+        allow_none=True,
```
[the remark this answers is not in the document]
connor-mccorm: Good catch!

ludwig/schema/features/sequence_feature.py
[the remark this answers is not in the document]
connor-mccorm: Going to refactor this into a separate PR!

requirements.txt
```diff
@@ -31,6 +31,7 @@ psutil
 protobuf==3.20.1 # https://github.com/databrickslabs/dbx/issues/257
 experiment_impact_tracker
 gpustat
+rich~=12.4.4
```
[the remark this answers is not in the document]
connor-mccorm: This dependency is specifically for console messages when using the config object in a notebook. It specifically allows for more robust messages after enabling and disabling features. It would be a little bit tricky to factor out into a separate PR like we discussed in standup because it would require the full config object lands before so it would be better as a followup PR. That being said, it is a very small change and if we are not on board with it, I'm ok just removing it altogether since it's a small UX improvement on the SDK side.

tests/ludwig/schema/test_config_object.py
```diff
@@ -0,0 +1,373 @@
...
+
+@pytest.mark.parametrize(
+    "session",
+    [
+        {"sess_id": 0, "encoder": "parallel_cnn", "loss": {"type": "mean_squared_error"}},
+        {"sess_id": 1, "encoder": "cnnrnn", "loss": {"type": "mean_absolute_error"}},
+        {"sess_id": 2, "encoder": "parallel_cnn", "loss": {"type": "mean_absolute_error"}},
+    ],
+)
+def test_shared_state(session):
```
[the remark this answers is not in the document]
connor-mccorm: So this test is a very tricky part of this PR but very necessary. It is better to explain over a call since it's quite complicated, but the short explanation is that when dataclasses (which we use throughout the schema to generate the full schema) have class attributes set to mutable defaults, those mutable defaults can be altered like class attributes. Thing is, we can't set these as instance attributes for the schema because marshmallow uses the dataclasses to generate the json schema. So what happens here is that these classes will get modified across test runs and cause tests to fail when run in a suite, but they will pass when run individually. It was a really tricky thing to nail down, but we solved it by adding deep copies in strategic places to just get the config classes with instance attributes initialized. This test makes sure that the shared state across test cases is handled appropriately. Here is a link that describes the python dataclass behavior that causes this issue: https://docs.python.org/3/library/dataclasses.html#mutable-default-values

tests/ludwig/schema/test_config_object.py
```diff
@@ -0,0 +1,373 @@
...
+        ],
+    }
+
+    config_object.update_with_dict(temp_config)
+
+    assert config_object.input_features.text_feature.encoder.max_sequence_length == 10
+
+
+@pytest.mark.parametrize("load_format", ["dict", "yaml"])
+def test_constructors(load_format):
```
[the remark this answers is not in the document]
connor-mccorm: Will do!

ludwig/schema/features/sequence_feature.py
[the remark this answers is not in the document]
justinxzhao: Sorry was unclear, but I think you groc'ed what I was commenting on -- basically the `BaseLossConfig` piece for all of the output features could be its own rather uncontroversial change.

ludwig/schema/features/bag_feature.py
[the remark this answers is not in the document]
justinxzhao: That works!

Another workflow I recommend that comes without developer interruption (discussed offline) is to build a prototype of the full change in all of it's size and glory (aka this PR), and then using that as a reference to generate small, focused, maximally digestible PRs for smooth easier reviewing for the team.

Development happens as usual, but then merging in "a big PR" becomes a more logistical matter of figuring out how to best present / breakdown the technical details for your team.

- This will only be in service of others being able to appreciate the technical depth of your work, and lowering the bar for other engineers to understand, accept, and help maintain(!) your fully materialized vision (implementation, implications, etc.).
- Uncontroversial changes with a tight scope can be reviewed, merged, and rebased on quickly.
- Changes can often be reviewed and merged in parallel.
- When changes are broken down and more visible, you'll discover that sometimes there are additional test cases, comments, or other context that were missed or expanded upon, and the separate PR may turn into a more enriched/mature version than your initial reference, which you can happily merge into your reference.
- You might get push back for one of these intermediate changes that might result in needing to fundamentally change your prototype, but presumably there would be a discussion and there would be good reason for it, and it's exactly the kind of review that we want.

tests/ludwig/schema/test_config_object.py
connor-mccorm: This is a good catch, changed up a test a little to capture the scope better!

tests/ludwig/features/test_category_feature.py
connor-mccorm: Agreed! In this case however, the removing of `populate_defaults` went along with the full removal of `merge_with_defaults` which would take quite a while to decouple into a separate PR, so for this I would opt to keep them together!

tests/integration_tests/test_custom_components.py
connor-mccorm: Agreed, will factor this one out!

ludwig/schema/features/bag_feature.py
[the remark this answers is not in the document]
connor-mccorm: Per this recommended workflow, I have factored this change out into its own PR: https://github.com/ludwig-ai/ludwig/pull/2628

ludwig/schema/features/sequence_feature.py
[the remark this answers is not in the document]
connor-mccorm: Per discussion in MLQ standup, I have factored this out into a separate PR: https://github.com/ludwig-ai/ludwig/pull/2628

ludwig/schema/utils.py
```diff
@@ -1,9 +1,11 @@
+import copy
```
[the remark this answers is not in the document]
connor-mccorm: I've factored the changes from this PR out into the defaults mixin PR since they are required for that change: https://github.com/ludwig-ai/ludwig/pull/2628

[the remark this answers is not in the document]
connor-mccorm: Need to keep the convert submodules and `__repr__` changes here need to stay in this PR since they are used in the config object.

ludwig/utils/misc_utils.py
```diff
@@ -84,6 +84,10 @@ def resolve_pointers(dict1, dict2, dict2_name):
             if key_in_dict2 in dict2.keys():
                 value = dict2[key_in_dict2]
                 resolved_dict[key] = value
+            if ENCODER in dict2.keys():  # Temporary fix, this whole function should be removed during preproc refactor
+                if key_in_dict2 in dict2[ENCODER].keys():
+                    value = dict2[ENCODER][key_in_dict2]
+                    resolved_dict[key] = value
```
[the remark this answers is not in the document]
connor-mccorm: It's a symptom of the `config_obj.to_dict()` returning a fully populated config. Since this PR passes in the return of `config_obj.to_dict()` to the preprocessing pipeline, there are now encoders in the feature configs, so this basically makes sure that the resolve pointers function works the same way with the additional params in the feature config. Will add a TODO to refactor this since it will go away once I refactor the preprocessing section with the Config Object.

tests/ludwig/schema/test_config_object.py
```diff
@@ -0,0 +1,373 @@
...
+    assert config_object.combiner.weights_initializer == "xavier_uniform"
+    assert config_object.combiner.fc_layers is None
+
+    assert config_object.trainer.epochs == 50
+    assert config_object.trainer.batch_size == "auto"
+
+    assert config_object.trainer.optimizer.type == "adam"
+    assert config_object.trainer.optimizer.beta1 == 0.8
+    assert config_object.trainer.optimizer.beta2 == 0.999
+    assert config_object.trainer.optimizer.epsilon == 5e-09
```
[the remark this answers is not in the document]
connor-mccorm: I think this actually may not be the best idea. Funny enough what you're suggesting is actually the original way I was running this test, however, because `to_dict()` returns a fully populated config dictionary, it ends up being way more lines (was like 300-400 lines or something originally). So i figured it would actually be more readable to just check that all the parameters that I am setting in this config dict are set in the config object as opposed to comparing a full config dict. WDYT?

tests/integration_tests/test_custom_components.py
connor-mccorm: Per our standup discussion, factored out into this PR: https://github.com/ludwig-ai/ludwig/pull/2627

ludwig/api.py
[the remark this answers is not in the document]
connor-mccorm: In the spirit of not complicating this PR any further, I would opt to discuss whether we want to do the Evaluator class in a followup meeting and then if we decide yes then implement in another PR. Does that work?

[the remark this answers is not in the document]
justinxzhao: Certainly! Refactoring out a separate evaluator class is certainly out of scope for this PR.

For now, let's just leave a comment explaining why `.to_dict().get(EVAL_BATCH_SIZE, None)` is required, as Daniel suggests.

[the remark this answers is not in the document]
connor-mccorm: Sounds good, will do!

ludwig/schema/config_object.py
connor-mccorm: In this instance, if we were to do this, I would probably want to refactor all the other util functions and port them over to a util file as well (i.e. initialize_input_features, set_input_features, output_features, get_config_cls, etc.) This would be a considerable amount of effort and would change the structure of the config object a bit. Since this function is kind of the core utility of the config object, it is tested pretty directly under the `test_config_object` test and indirectly  through the rest of the tests. Would you be open to taking 10-15 with me to do a walk through instead? Additionally, once this lands, I plan on either taking a tech talk slot or an all hands slot to walk the entire team through the config object flow so that everyone understands what's going on. How does this sound?

tests/ludwig/schema/test_config_object.py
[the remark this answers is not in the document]
justinxzhao: Makes sense. Let's go with what you have now.

justinxzhao (approved):

arnavgarg1 (approved):

ludwig/schema/combiners/common_transformer_options.py
ksbrar, line 9: Since this is everywhere it might be worth defining a lambda in some utils file.

ludwig/schema/combiners/comparator.py
```diff
@@ -7,15 +7,22 @@
...
 
     entity_1: List[str]
-    """TODO: Document parameters."""
+    """List of input features that compose the first entity to be compared."""
 
     entity_2: List[str]
-    """TODO: Document parameters."""
+    """List of input features that compose the second entity to be compared."""
+
+    type: str = schema_utils.StringOptions(
```
ksbrar, line 20: ah yes, wanted this for a long time - nice!

ludwig/schema/features/sequence_feature.py
```diff
@@ -66,6 +67,13 @@ class SequenceOutputFeatureConfig(BaseOutputFeatureConfig, SequenceOutputFeature
     """SequenceOutputFeatureConfig is a dataclass that configures the parameters used for a sequence output
     feature."""
 
+    default_validation_metric: str = schema_utils.StringOptions(
```
ksbrar, line 70: Is this metric parameter on every output feature?

ludwig/schema/model_config.py
```diff
@@ -0,0 +1,529 @@
...
+    def _validate_config(config_dict: dict) -> None:
+        """Helper function used to validate the config using the Ludwig Schema.
+
+        Args:
+            config_dict: Config Dictionary
+        """
+        validate_config(config_dict)
+
+    @staticmethod
+    def _get_config_cls(section: str, section_type: str, feature_type: str) -> BaseMarshmallowConfig:
```
ksbrar, line 241: Might rename this to `_get_config_nested_cls` to be clear at a glance

ludwig/schema/optimizers.py
```diff
@@ -32,7 +32,11 @@ def wrap(optimizer_config: BaseOptimizerConfig):
     return wrap
 
 
-@dataclass
+def get_optimizer_cls(name: str):
```
ksbrar, line 35: Nit: add comment just bc this can be read the wrong way at a glance

ludwig/schema/utils.py
```diff
@@ -56,6 +58,31 @@ def load_config_with_kwargs(
     }
 
 
+def convert_submodules(config_dict: dict) -> TDict[str, any]:
```
ksbrar, line 61: Nit: If its only used by the config object, feel as though this belongs in the config_object module itself

ludwig/utils/defaults.py
```diff
@@ -16,357 +16,38 @@
...
 
 default_preprocessing_parameters = copy.deepcopy(default_feature_specific_preprocessing_parameters)
-default_preprocessing_parameters.update(base_preprocessing_parameters)
-
-default_model_type = MODEL_ECD
-
-default_combiner_type = "concat"
-
-
-def _perform_sanity_checks(config):
```
ksbrar, line 85: nice.

ksbrar (approved):
brilliant!!

ludwig/schema/features/sequence_feature.py
connor-mccorm: Yes, it was previously stored as a class attribute on the output_features themselves but moved it to the schema in the spirit of SSoT.

ludwig/schema/combiners/comparator.py
connor-mccorm: Thanks!

ludwig/schema/features/sequence_feature.py
ksbrar: If it's required then I'd say stick it on some inherited base class.

connor-mccorm: Definitely will do! Though I would still have to leave it on each of the individual output features because each output feature has a different default validation metric so I would just set the one on the base class to none.

ludwig/schema/model_config.py
connor-mccorm: Good call!

ludwig/schema/optimizers.py
connor-mccorm: Good looks!

ludwig/schema/utils.py
connor-mccorm: So it actually gets used in the schema utils file as well in the `BaseMarshmallowConfig` since I wanted it to be inherited by all schema classes so that any conversion to dict will convert submodules.

ludwig/utils/defaults.py
connor-mccorm: So satisfying eh!

PR #2429, opened 2022-08-29
Added conditional check for UNK token insertion into category feature vocab

arnavgarg1:
This change conditionally adds the UNKNOWN symbol to the category feature vocab when the number of "most common" tokens is less than the total number of unique tokens in the vocab. This has a downstream effect on simplifying and producing more meaningful visualizations.

Co-authored-by: Justin Zhao @justinxzhao


ludwig/features/category_feature.py
```diff
@@ -59,7 +59,11 @@ class _CategoryPreprocessing(torch.nn.Module):
     def __init__(self, metadata: Dict[str, Any]):
         super().__init__()
         self.str2idx = metadata["str2idx"]
-        self.unk = self.str2idx[UNKNOWN_SYMBOL]
+        if UNKNOWN_SYMBOL in self.str2idx:
+            self.unk = self.str2idx[UNKNOWN_SYMBOL]
+        else:
+            # TODO: Replace with mode or something to that effect
+            self.unk = 0
```
arnavgarg1, line 66: I can replace this with the same thing I'm doing in lines 153-158, but want to make sure that this approach even makes sense to begin with and it is the cleanest way to do things @justinxzhao

justinxzhao: Another potential alternative to try would be to set `self.unk` as None, and add a conditional in `forward()`. This should be able to get around torchscript tracing (maybe?)

```python
if self.unk:
    indices = [self.str2idx.get(s.strip(), self.unk) for s in v]
else:
    indices = [self.str2idx[s.strip()] for s in v]
return torch.tensor(indices, dtype=torch.int32)
```

CC: @geoffreyangus

ludwig/features/category_feature.py
```diff
@@ -137,13 +141,26 @@ def get_feature_meta(column, preprocessing_parameters, backend):
...
-        return column.map(
-            lambda x: (
-                metadata["str2idx"][x.strip()]
-                if x.strip() in metadata["str2idx"]
-                else metadata["str2idx"][UNKNOWN_SYMBOL]
+        def __replace_token_with_idx(value: Any, metadata: Dict[str, Any], unknown_symbol_idx: int) -> int:
+            stripped_value = value.strip()
+            if stripped_value in metadata["str2idx"]:
+                return metadata["str2idx"][stripped_value]
+            logger.warning("Using unknown symbol during category feature processing.")
```
justinxzhao, line 148: Let's add more to this warning.

> Encountered unknown symbol '{stripped_value}' for '{column.name}' during category feature preprocessing. This should never happen during training. If this happens during inference, this may be an indication that not all possible symbols are present in your training set. Consider re-splitting your data to ensure full representation, or setting `preprocessing.most_common` to be smaller than this feature's total vocabulary size, {len(metadata["str2idx"])}, which will ensure that the model is architected and trained with an `UNKNOWN` symbol.

arnavgarg1: @justinxzhao Sounds good! Just to confirm, we're thinking of leaving this as a warning and not a Exception right? Should I still replace it with a default value like the mode, or maybe something even simpler like the first token?

ludwig/features/category_feature.py
```diff
@@ -137,13 +141,26 @@ def get_feature_meta(column, preprocessing_parameters, backend):
 
     @staticmethod
     def feature_data(column, metadata):
-        return column.map(
-            lambda x: (
-                metadata["str2idx"][x.strip()]
-                if x.strip() in metadata["str2idx"]
-                else metadata["str2idx"][UNKNOWN_SYMBOL]
+        def __replace_token_with_idx(value: Any, metadata: Dict[str, Any], unknown_symbol_idx: int) -> int:
```
justinxzhao, line 145: Rename `unknown_symbol_idx` to `fallback_symbol_idx` so that we can reserve the word "unknown" for the actual unknown symbol.

ludwig/features/category_feature.py
```diff
@@ -137,13 +141,26 @@ def get_feature_meta(column, preprocessing_parameters, backend):
 
     @staticmethod
     def feature_data(column, metadata):
-        return column.map(
-            lambda x: (
-                metadata["str2idx"][x.strip()]
-                if x.strip() in metadata["str2idx"]
-                else metadata["str2idx"][UNKNOWN_SYMBOL]
```
justinxzhao, line 144: Let's keep this code for the `else` case, since we only care about logging a warning when we are using the fallback symbol when the `UNKNOWN_SYMBOL` is not in the metadata, yet we have found that we need to use the fallback symbol.

ludwig/features/category_feature.py
justinxzhao: Warning SGTM since otherwise we'd need to change all of the tests using synthetic datasets with rows=25/vocab_size=10.

Updated warning:

> Encountered unseen symbol '{stripped_value}' for '{column.name}' during category feature preprocessing. This should never happen during training. If this happens during inference, this may be an indication that not all possible symbols are present in your training set. Consider re-splitting your data to ensure full representation, or setting preprocessing.most_common to be smaller than this feature's total vocabulary size, {len(metadata["str2idx"])}, which will ensure that the model is architected and trained with an explicit UNKNOWN symbol. For now, examples using '{stripped_value}' will be mapped to `{metadata["idx2str"][fallback_symbol_idx]}`.

ludwig/features/category_feature.py
```diff
@@ -137,13 +142,36 @@ def get_feature_meta(column, preprocessing_parameters, backend):
...
+                return metadata["str2idx"][stripped_value]
+            logger.warning(
+                f"""
+                Encountered unknown symbol '{stripped_value}' for '{column.name}' during category
+                feature preprocessing. This should never happen during training. If this happens during
+                inference, this may be an indication that not all possible symbols were present in your
+                training set. Consider re-splitting your data to ensure full representation, or setting
+                preprocessing.most_common parameter to be smaller than this feature's total vocabulary
+                size, {len(metadata["str2idx"])}, which will ensure that the model is architected and
+                trained with an UNKNOWN symbol. Returning the index for the most frequent symbol instead.
```
justinxzhao, line 157: We can actually include the most frequent symbol in this warning:

`metadata["idx2str"][fallback_symbol_idx]`

ludwig/features/category_feature.py
arnavgarg1: FYI I tried this and torchscript complained. My guess is that torchscript tracing can't handle condition jumps and needs a more clear data flow to infer things.

justinxzhao (approved):
Thanks!

PR #2439, opened 2022-08-30
Loss, Split, and Defaults Schema Additions

connor-mccorm:
This PR adds a loss schema, spllit schema, and a defaults schema. Additionally this changes some internals to use schema default properties. The next PR after this will land the config-object which will remove the need for default functions on the input and output features.


ludwig/schema/config_object.py
```diff
@@ -0,0 +1,82 @@
...
+    trainer: BaseTrainerConfig = ECDTrainerConfig()
+    preprocessing = PreprocessingConfig()
+    hyperopt = {}
+    defaults = {}
+
+    def __init__(self, config_dict):
+        self.parse_input_features(config_dict[INPUT_FEATURES])
+        self.parse_output_features(config_dict[OUTPUT_FEATURES])
+
+        if COMBINER in config_dict:
```
martindavis, line 48: Hey @connor-mccorm , is this IF here to deal with Trees where we don't have a combiner?

connor-mccorm: Not quite, basically how this Config class will work is it will take all attributes defined in the user defined `config_dict` and update the defaults with them. Since the combiner is an optional section, it may or may not be present in the user config so we have to check to see if it's there to tell if we need to run any updates. This does bring up a good point though since I am only accounting for ECD model trainer here and I need to have GBM support as well. Either way, this file is actually not being used in this PR, it is actually part of the config object PR so I can just remove it from here since it isn't relevant.

ludwig/features/feature_utils.py
```diff
@@ -100,7 +100,11 @@ def sanitize(name):
 
 
 def compute_feature_hash(feature: dict) -> str:
-    preproc_hash = hash_dict(feature.get(PREPROCESSING, {}))
+    preproc = feature.get(PREPROCESSING, {})
```
justinxzhao, line 103: Let's update the type hint and add a comment/docstring about when feature is a dict and when it's a schema object.

ludwig/schema/utils.py
```diff
@@ -853,7 +860,8 @@ def _deserialize(self, value, attr, obj, **kwargs):
             for option in field_options:
                 mfield_meta = option.metadata["marshmallow_field"]
                 try:
-                    mfield_meta.validate(value)
+                    if mfield_meta.validate:
```
justinxzhao, line 863: When is `mfield_meta.validate=None`?

justinxzhao (approved):

ludwig/features/feature_utils.py
connor-mccorm: Ah good eye, I remember noticing that during the call but pretty quickly forgot. Thanks for pointing that out. Will change now!

martindavis (approved):

PR #2441, opened 2022-08-31
Fix: Upgrade ludwig config before schema validation

connor-mccorm:
This fix adds a backwards compatibility step into the validate config step for the schema. The issue we we're dealing with was the frontend validating for 0.5 configs and Ludwig for 0.6. The solution in this PR is to update the configs in the validation step which will allow the frontend to validate old valid configs. These will get converted by backwards compatibility upon training anyways so we are just updating them before validation so that the frontend can validate without conflicting with Ludwig validation.

Note: the reason for importing `upgrade_to_latest_version` in the `validate_config` function is because circular imports start occurring if we import it at the top. This is something I will fix in the future, but may take a bit longer. If everyones ok with it, I will leave it like this for now.

### DIFF
--- a/ludwig/schema/__init__.py
+++ b/ludwig/schema/__init__.py
@@ -62,7 +62,8 @@ def validate_config(config):
     # Update config from previous versions to check that backwards compatibility will enable a valid config
     from ludwig.utils.backward_compatibility import upgrade_to_latest_version
 
-    config["ludwig_version"] = "0.4"
+    if "ludwig_version" not in config:
+        config["ludwig_version"] = "0.4"
     updated_config = upgrade_to_latest_version(config)
 
     type_checker = Draft7Validator.TYPE_CHECKER.redefine("array", custom_is_array)
--- a/ludwig/schema/__init__.py
+++ b/ludwig/schema/__init__.py
@@ -59,6 +59,11 @@ def validate_config(config):
     def custom_is_array(checker, instance):
         return isinstance(instance, list) or isinstance(instance, tuple)
 
+    # Update config from previous versions to check that backwards compatibility will enable a valid config
+    from ludwig.utils.backward_compatibility import upgrade_to_latest_version
+    config['ludwig_version'] = "0.4"
+    updated_config = upgrade_to_latest_version(config)
+
     type_checker = Draft7Validator.TYPE_CHECKER.redefine("array", custom_is_array)
     CustomValidator = extend(Draft7Validator, type_checker=type_checker)
-    validate(instance=config, schema=get_schema(), cls=CustomValidator)
+    validate(instance=updated_config, schema=get_schema(), cls=CustomValidator)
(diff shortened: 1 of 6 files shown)


ludwig/schema/__init__.py
```diff
@@ -59,6 +59,12 @@ def validate_config(config):
     def custom_is_array(checker, instance):
         return isinstance(instance, list) or isinstance(instance, tuple)
 
+    # Update config from previous versions to check that backwards compatibility will enable a valid config
+    from ludwig.utils.backward_compatibility import upgrade_to_latest_version
+
+    config["ludwig_version"] = "0.4"
```
jeffreyftang, line 65: Will this break if the config is in fact not from 0.4? As in either newer or older?

jeffreyftang (approved):

ludwig/schema/__init__.py
connor-mccorm: No it shouldn't, but I am actually adding a quick update to that section which basically checks if the ludwig version is already set so we don't have to overwrite it.

connor-mccorm: But worst case scenario, if the version is 0.4, all the back compat updates are ran and they have no affect if the config is already valid with 0.6

hungcs (approved):

ludwig/schema/__init__.py
dantreiman: This may be un-needed if we merge Travis's change here: https://github.com/ludwig-ai/ludwig/pull/2424

tests/ludwig/utils/test_validate_config_combiner.py
```diff
@@ -51,7 +51,7 @@ def test_config_bad_combiner():
             category_feature(encoder={"type": "dense", "vocab_size": 2}, reduce_input="sum"),
             number_feature(),
         ],
-        "output_features": [binary_feature(weight_regularization=None)],
+        "output_features": [binary_feature(deccoder={"weight_regularization": None})],
```
justinxzhao, line 54: This looks like a typo. `decoder`?

connor-mccorm: good catch Justin!

justinxzhao (approved):

ludwig/schema/decoders/utils.py
```diff
@@ -60,7 +60,6 @@ def _jsonschema_type_mapping():
                 },
                 "title": "decoder_options",
                 "allOf": get_decoder_conds(feature_type),
-                "required": ["type"],
```
martindavis, line 63: Is this something we'll need to bring back once we're using Ludwig 0.6 throughout our entire ecosystem? With the removal of `type` being required, this will affect both the JSON Schema we generate and the docs right?

connor-mccorm: So I thought about it a bit and we actually don't need to require type from the user on their config. Reason being, the default will just get filled in for them if they don't specify. So if they were to say just specify dropout, it would just switch out the dropout value for the default encoder.

PR #2445, opened 2022-09-02
Extended hyperopt to support nested configuration block parameters

tgaddair:


arnavgarg1:
Looks awesome, super excited for this change. Thanks for getting it in super quickly @tgaddair. I also think this greatly simplifies parameter substitution logic and gets default parameters to also work very nicely.

ludwig/hyperopt/run.py
```diff
@@ -215,7 +209,6 @@ def hyperopt(
     logging.info(
         "Features that may be updated in hyperopt trials if default parameters are specified in the search space"
     )
-    logging.info(pformat(dict(features_eligible_for_shared_params), indent=4))
     logging.info("\n")
```
arnavgarg1, line 212: nit: This is no longer needed - we can delete this.

ludwig/hyperopt/utils.py
```diff
@@ -51,3 +54,90 @@ def should_tune_preprocessing(config):
...
+
+    config = copy.copy(config)
+    config[INPUT_FEATURES] = input_features_dict
+    config[OUTPUT_FEATURES] = output_features_dict
+    return config
+
+
+def feature_dict_to_list(config: Dict[str, Any]) -> Dict[str, Any]:
+    # This works because Python dicts are order-preserving, so we do not need to
+    # do anything special to map from a key in the dict to an index in a list
```
arnavgarg1, line 94: I guess even if these were out of order, this would work because the ordering of the features in the config doesn't make a difference to the model itself right?

ludwig/hyperopt/utils.py
```diff
@@ -51,3 +54,90 @@ def should_tune_preprocessing(config):
...
+    # features by name, so convert temporarily to a dict to simplify the mergep process.
+    config = feature_list_to_dict(config)
+
+    # Merge parameters into the user configuration in order. As such, if there are conflicting
+    # params, the later params will take precedence.
+    for name, value in parameters.items():
+        # User params are provided as <feature_name>.<param>, but we group input / output features
+        # together during the merge to make it easier and unambiguous to convert back and forth
+        # TODO(travis): we should revisit the user format here, as it silently breaks situations
+        # where the user has a feature named "trainer", "combiner", etc.
```
arnavgarg1, line 130: This is a quick check to see if the top-level config section keywords of Ludwig are within the feature names from lines 117-118 right? Would the ideal outcome be to just raise an error and ask the user to rename their feature? 

I think that users shouldn't be allowed to have feature names that have any of the "special" keywords that we use within the Ludwig config.

ludwig/hyperopt/utils.py
```diff
@@ -51,3 +54,90 @@ def should_tune_preprocessing(config):
...
+    input_feature_names = {feature[NAME] for feature in config[INPUT_FEATURES]}
+    output_feature_names = {feature[NAME] for feature in config[OUTPUT_FEATURES]}
+
+    # Features in the user config are provided as a list, but in hyperopt we reference
+    # features by name, so convert temporarily to a dict to simplify the mergep process.
+    config = feature_list_to_dict(config)
+
+    # Merge parameters into the user configuration in order. As such, if there are conflicting
+    # params, the later params will take precedence.
+    for name, value in parameters.items():
```
arnavgarg1, line 126: One thought about updating the config object with sampled values incrementally while using the custom config parameter "." is that the order of parameters makes a difference, which may matter depending on how the user sets things up. 

For e.g., if there's a `trainer.learning_rate` within "." that also has its own top level parameter like "trainer.learning_rate" outside of ".",  then essentially where the "trainer.learning_rate" falls (before or after the ".") will affect the final value that's picked for that particular trial. 

So if the Ludwig user sets up their parameter space like you have in your tests, then there's no problem, but if they do it slightly differently like in the example above, then it's something that we should make sure they're aware of. Ideally, the guideline should be that they always add "." first and then add config-specific parameters afterwards. 

Just something that might be worth calling out in the docs for 0.6 once this is merged in!

arnavgarg1 (approved):
This is a great change!

ludwig/hyperopt/utils.py
arnavgarg1: Created a task for documenting hyperopting over nested configs: https://github.com/ludwig-ai/ludwig-docs/issues/207, I can take a stab at adding documentation for it and can add you as a reviewer. Should we also call this feature out in 0.6?

ludwig/hyperopt/utils.py
tgaddair: Yes, though I think preserving order if good to keep the config easy to read by the user.

ludwig/hyperopt/utils.py
tgaddair: I would probably want to change the way we reference features in hyperopt to include `input_features.` in front to resolve ambiguity. But it's a big breaking change, so probably not worth doing just now. I agree a check in schema validation would be useful to catch this as early as possible.

ludwig/hyperopt/utils.py
tgaddair: We may want to hold off on this feature for v0.6, as it's not a bug fix and the branch has already been cut. I think it's fine to wait for v0.7 here, as we don't have plans to use it immediately, and can give it some time to cook before officially releasing it.

I's definitely important that we call it in the docs to parameter ordering consideration you raised. There are some cool things you can do with this, but the user needs to know what they're doing.

arnavgarg1: @tgaddair Got it! In that case, I will need to re-open [this issue](https://github.com/ludwig-ai/ludwig/issues/2243) and get it merged into the 0.6 release branch before Thursday since we want this capability for default parameters to work correctly in 0.6 and there's currently insufficient support. 

Alternatively, we can merge this for now but only call out the defaults feature without mentioning being able to hyperopt over subsection configs.

Let me know what you think is the best way.

ludwig/hyperopt/utils.py
arnavgarg1: Got it! Might be a good change to include in 0.7 along with this change.

ludwig/hyperopt/utils.py
tgaddair: Ah, interesting. Yeah in that case I'm fine merging this into the 0.6 branch. We can also update the docs to include the nested configs stuff, since it's there.

tests/integration_tests/test_hyperopt.py
```diff
@@ -562,3 +491,94 @@ def test_hyperopt_old_config(csv_filename, tmpdir, ray_cluster):
...
+                "time_budget_s": 200,
+                "cpu_resources_per_trial": 1,
+                "num_samples": 4,
+                "scheduler": {TYPE: "fifo"},
+            },
+            "search_alg": {TYPE: "variant_generator"},
+            "parameters": {
+                ".": {
+                    "space": "choice",
+                    "categories": [
```
martindavis, line 519: So we're essentially hyperopting over the combo of a `combiner` with type `tabnet` and the `adam` optimizer OR a `combiner` with type `concat` and the default trainer with specific type of learning rate scaling?

justinxzhao: Correct.

justinxzhao (approved):
Implementation wise, this LGTM.

In documentation, it would be good to call out which circumstances we might encourage using this `.` over using separate full configs, which we are discussing offline.

Overall, seems worth checking in though, as more generally this gives us a way to hyperopt over grouped/related/dependent parameters.

arnavgarg1:
> Implementation wise, this LGTM.
> 
> In documentation, it would be good to call out which circumstances we might encourage using this `.` over using separate full configs, which we are discussing offline. Not sure we should encourage this usage pattern for users interested in searching over multiple model types at once, for example.
> 
> Overall, seems worth checking in though, as more generally this gives us a way to hyperopt over grouped/related/dependent parameters.

I'll take a stab at writing this out this afternoon and assign you guys as reviewers

PR #2446, opened 2022-09-02
fix: Switched Learning Rate to NonNegativeFloat Field

connor-mccorm:


tgaddair (approved):

ludwig/schema/optimizers.py
```diff
@@ -61,7 +61,7 @@ class SGDOptimizerConfig(BaseOptimizerConfig):
     """Must be 'sgd' - corresponds to name in `ludwig.modules.optimization_modules.optimizer_registry` (default:
        'sgd')"""
 
-    lr: float = FloatRange(default=1e-03, min=0.0, max=1.0, description="Learning rate.")
+    lr: float = NonNegativeFloat(default=1e-03, description="Learning rate.")
```
ksbrar, line 64: just occurred to me - but ideally `lr` should be part of `BaseOptimizerConfig`. 

3 questions:
1) I think `FtrlOptimizer` actually should have an `lr` (way back when I added it I think I overlooked this but...)
2) Is ftrl optimizer ever used? It's not added to the registry because pytorch doesn't have an in-house ftrl class but I am unfamiliar if the tensorflow backwards compat may use it.
3) Similarly I think nadam optimizer should be registered because it does have a corresponding pytorch class (https://pytorch.org/docs/stable/generated/torch.optim.NAdam.html) - I think it either did not or I overlooked this when I initially created the schemas

ksbrar (approved):

ludwig/schema/optimizers.py
connor-mccorm: Added `lr` to the `BaseOptimizerConfig`.  I don't actually know if FtrlOptimizer is actually used, I feel like @justinxzhao, @w4nderlust, @ShreyaR, or @dantreiman may know. Regarding the nadam optimizer, I agree it should be registered. It looks like you already had it registered at one point but commented that out, so I just uncommented that code again.

PR #2447, opened 2022-09-02
[tests] Added more explicit lifecycle management to Ray clusters during tests

tgaddair:


tests/integration_tests/test_preprocessing.py
```diff
@@ -25,9 +24,14 @@
...
 
 
-@pytest.mark.parametrize("backend", ["local", "ray"])
-@pytest.mark.distributed
-def test_sample_ratio(backend, tmpdir):
+@pytest.mark.parametrize(
+    "backend",
+    [
+        pytest.param("local", id="local"),
+        pytest.param("ray", id="ray", marks=pytest.mark.distributed),
```
justinxzhao, line 31: Cool that you can add marks within as part of the parameterization like this.

justinxzhao (approved):
Looks like there's some really nice changes here as the CI is down to 31 minutes!

PR #2459, opened 2022-09-06
Enh: quality-of-life improvements for `export_torchscript`

geoffreyangus:
This PR implements small improvements for the `export_torchscript` CLI command. Notable behavior changes are the following:

- `--output_path` is now an optional command line argument. If none, then the `output_path` defaults to being the same as `model_path`. This is in contrast to its initial implementation, which placed models in some `torchscript/` directory. The downside of using `torchscript/` as the default was that artifacts could be overridden if multiple torchscript arguments were passed in.
- `model_hyperparameters.json` and `training_set_metadata.json` are now copied into the resulting torchscript export directory. This makes it possible to use the `InferenceModule.predict` function– which requires `model_hyperparameters.json`– regardless of where the inference modules are saved.

### DIFF
--- a/ludwig/export.py
+++ b/ludwig/export.py
@@ -30,7 +30,7 @@ logger = logging.getLogger(__name__)
 
 
 def export_torchscript(
-    model_path: str, model_only: bool = False, output_path: str = "torchscript", device: Optional[str] = None, **kwargs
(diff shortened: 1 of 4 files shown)


brightsparc (approved):
LGTM

justinxzhao (approved):

abidwael (approved):

PR #1803, opened 2022-03-10
Move from epoch-based training and evaluation to step-based training and evaluation.

justinxzhao:
[Design doc](https://docs.google.com/document/d/1T_CJqvnzH_TAA57-4U-E6vK5IjN8zHKoRHChNUGjR1I/edit)

Notably, this PR adds a new flag, `steps_per_checkpoint`, which configures how frequently the model is check-pointed, as well as how frequently evaluation runs.

Other changes:
- The `ProgressTracker` was already tracking steps, so I switched over all of the epoch-based loops, actions, and callbacks, to be based on training steps.
- Training-in-progress metric dictionaries now hold `TrainerMetric` namedtuples instead of raw values. This allows us to associate metadata like step and epoch with specific metrics.
- The consumers of the dictionary-lists of evaluation results implicitly assume that there’s one entry per epoch. I’ve updated consumers of the dictionary-list like the best-validation-metric-finding code to use the augmented structure.
- Some processes like hyperopt on Ray tune don't need precise step information. For these flows, we use `reduce_dict_dict_trainer_metrics()` to reduce the nested dictionary of lists of `TrainerMetric` namedtuples into a dictionary with the same structure, but with raw lists of floats. Making hyperopt on Ray Tune use the augmented structure seems like a more involved change that I can look into in a follow up PR, if necessary.
- Instead of one training progress bar per epoch, we’ll have one long training progress bar over the whole run. Evaluation happens intermittently throughout, according to a new flag, `steps_per_checkpoint` , which also saves the model every N steps.
- Dataset evaluations are factored out and combined into a single method, `run_evaluation()`, which is called within `_train_loop()`, invoked whenever `steps % steps_per_checkpoint == 0`.

Next, we'll fully decouple the training and evaluation so that they can happen on separate processes/threads/workers, without interrupting each other. This should enable maximally fluid and continuous training and evaluation.

Example of new stdout:
```
╒════════════════════════╕
│ EXPERIMENT DESCRIPTION │
╘════════════════════════╛

╒══════════════════╤═════════════════════════════════════════════════════════════════╕
│ Experiment name  │ simple_experiment                                               │
├──────────────────┼─────────────────────────────────────────────────────────────────┤
│ Model name       │ simple_model                                                    │
├──────────────────┼─────────────────────────────────────────────────────────────────┤
│ Output directory │ /Users/justinzhao/ludwig/results/simple_experiment_simple_model │
├──────────────────┼─────────────────────────────────────────────────────────────────┤
[...]

PR #2468, opened 2022-09-09
Add backwards-compatibility logic for model progress tracker

jeffreyftang:
* Convert epoch-based stats into steps-based stats for compatibility with #1803. Epoch to steps conversion is approximate, as we don't have access to the full history of any batch size increases.
* Reformat train/test/validation metrics into `TrainerMetric` tuples, also with estimated epoch and step values.

### DIFF
--- a/tests/ludwig/utils/test_backward_compatibility.py
+++ b/tests/ludwig/utils/test_backward_compatibility.py
@@ -546,13 +546,7 @@ def test_upgrade_model_progress():
     # To do so, we modify the batch size value and re-run the upgrade on the otherwise-valid `new_model_progress` dict.
     new_model_progress["batch_size"] = 1
     unchanged_model_progress = upgrade_model_progress(new_model_progress)
-
-    for stat in ("improvement", "increase_batch_size", "learning_rate_reduction"):
-        assert unchanged_model_progress[f"last_{stat}_steps"] == new_model_progress[f"last_{stat}_steps"]
-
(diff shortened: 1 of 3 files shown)


tgaddair (changes requested):
Tests

tests/ludwig/utils/test_backward_compatibility.py
```diff
@@ -490,3 +492,109 @@ def test_update_missing_value_strategy(missing_value_strategy: str):
...
+    # To do so, we modify the batch size value and re-run the upgrade on the otherwise-valid `new_model_progress` dict.
+    new_model_progress["batch_size"] = 1
+    unchanged_model_progress = upgrade_model_progress(new_model_progress)
+
+    for stat in ("improvement", "increase_batch_size", "learning_rate_reduction"):
+        assert unchanged_model_progress[f"last_{stat}_steps"] == new_model_progress[f"last_{stat}_steps"]
+
+    unchanged_metric = unchanged_model_progress["validation_metrics"]["combined"]["loss"][0]
+    new_metric = new_model_progress["validation_metrics"]["combined"]["loss"][0]
+    assert unchanged_metric == new_metric
```
justinxzhao, line 555: I find that it's easier to read a single check for equality on the entire dictionary, wholesale. What do you think?

tgaddair: Agreed, just need to specify an expected config dict.

jeffreyftang: Done

tgaddair (approved):

PR #2472, opened 2022-09-10
Fixed backwards compatibility for training_set_metadata and bfill

tgaddair:
Also adds new tests for old model versions.

We should, very soon, find a way to decouple the config from the training set metadata to avoid these sorts of situations in the future.


ludwig/data/preprocessing.py
```diff
@@ -1500,7 +1501,11 @@ def shuffle(df):
 
 def load_metadata(metadata_file_path):
```
arnavgarg1, line 1502: nit: might be nice to add a type hint for this function

arnavgarg1 (approved):
LGTM, thanks for the fix! Agreed that we should look into decoupling the two

PR #2475, opened 2022-09-12
Fixed backwards compatibility for models with level metadata in saved configs

tgaddair:
Following #1859, some older models (pre-v0.5 official release) could contain these invalid params, which are no longer part of the valid preprocessing schema.


justinxzhao (approved):
Thanks for adding this!

PR #2490, opened 2022-09-14
GBM: support hyperopt

jppgks:
Runs evaluation after every `boosting_rounds_per_checkpoint` to populate training/validation metrics as required by hyperopt

Mimics as much as possible the ECD train loop logic

### DIFF
--- a/ludwig/trainers/trainer_lightgbm.py
+++ b/ludwig/trainers/trainer_lightgbm.py
@@ -1,5 +1,8 @@
 import logging
 import os
+import signal
+import sys
+import threading
 import time
 from collections import OrderedDict
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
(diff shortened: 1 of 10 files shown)


ludwig/trainers/trainer_lightgbm.py
```diff
@@ -319,51 +308,149 @@ def run_evaluation(
...
 
-        for epoch, steps in enumerate(range(0, self.num_boost_round, self.boosting_round_log_frequency), start=1):
-            progress_tracker.epoch = epoch
+        return should_break
 
-            evals_result = {}
-            booster = self.train_step(
-                params, lgb_train, eval_sets, eval_names, booster, self.boosting_round_log_frequency, evals_result
-            )
+    def check_progress_on_validation(
```
justinxzhao, line 392: Is this function the same as the ECD trainer? If so, we should factor this out, like you've done with append_metrics.

jppgks: Not entirely, ECD trainer has additional logic here for reducing learning rate and increasing batch size

justinxzhao (approved):

PR #2493, opened 2022-09-14
Fix flaky ray nightly image test

arnavgarg1:
When we create a tensor directly from a numpy array instead of from a copy, it creates unpredictable behavior and results in the creation of ragged tensors downstream which causes weird type mismatch issues that Ray doesn't like. This change ensures we create the tensor from a copy of the numpy array so that we don't run into this issue.


justinxzhao (approved):
Nice! Thanks for the fix.

tgaddair (approved):

Issue #2497, opened 2022-09-15
Examples to use large remote dataset(s3 or minio)


Jeffwan:
**Is your feature request related to a problem? Please describe.**

I want to use remote dataset hosted in S3 or minio. Do you have any examples? Seems most examples in ludwig website are inbuilt dataset or local files. Do you have guidance using S3 or minio?  

```
from ludwig.datasets import mushroom_edibility
dataset_df = mushroom_edibility.load()
```

```
import pandas as pd
dataset_df = pd.read_csv(dataset_path)
```

1. I personally tried it `import dask.dataframe as dd; dataset_df = dd.read_csv('s3://bucket/myfiles.*.csv')` but notice I have to handle `s3fs` (required by dask). Is this a right way or there's easier way?
2. I also notice dataset accepts string. I am using minio for testing and could I know if it supports minio here? I want to customize `endpoint` and `signature`  
![image](https://user-images.githubusercontent.com/4739316/190300062-0d7eb350-405c-4695-8ead-c8f0e57bd6f6.png)

```
aws configure set default.s3.signature_version s3v4
aws --endpoint-url http://minio-service:9090 s3 ls
```

**Describe the use case**
Use remote dataset

**Describe the solution you'd like**
Provide an easy to use wrapper.

tgaddair:
Hey @Jeffwan, yes we support s3 / minio and any remote object storage supported by [fsspec](https://filesystem-spec.readthedocs.io/en/latest/).

Reading the data from minio with Dask is one way to do it. This is the easiest way to go if your environment is not configured to automatically connect to the remote storage backend. We provide a wrapper `ludwig.utils.data_utils.use_credentials` that simplifies setting credentials:

```
creds = {
    'client_kwargs': {
            'aws_access_key_id': 'accessKey',
            'aws_secret_access_key': 'mySecretKey',
            'endpoint_url': 'http://minio-service:9000'
        }
}
with use_credentials(creds):
    df = dask.read_csv("s3://...")
```

The other option is to pass a string. This also works with Minio, but it assumes that your environment is already setup to connect to s3 / minio without specifying any additional credentials. However, the `endpoint_url` makes this somewhat tricky with s3fs (see: https://github.com/fsspec/s3fs/issues/432). So for now I recommend providing the credentials explicitly and reading from Dask.

One thing we could do, if it would make things easier, is allow you to provide credentials (either path to credentials file or directly) within the Ludwig config, similar to how we let the user specify the cache credentials:

[...]

PR #2508, opened 2022-09-16
Fixed support for distributed datasets in create_auto_config

### LINKED ISSUE #2499: TypeError: '>' not supported between instance of 'str' and 'int' when tune_for_memory is True
**Describe the bug**
I tried to set tune_for_memory to True and meet TypeError. 

image

**To Reproduce**

```
    automl_config = create_auto_config(
        dataset=dataset_df,
        target=target_column,
        time_limit_s=time_limit_s,
        tune_for_memory=tune_for_memory,
        user_config=None,
        use_reference_config=False,
    )


    auto_train_results = train_with_config(
        dataset=dataset_df,
        config=automl_config,
        output_directory=output_dir,
        random_seed= default_random_seed,
    )

```

**Expected behavior**
tune_for_memory should be working without any expected error

- OS: \[e.g. iOS\] ludwigai/ludwig:nightly
- Version \[e.g. 22\] 
- Python version 3.8
- Ludwig version  nightly

tgaddair:
Fixes #2499.
Related #2497.

Also changes exclusion logic to be less restrictive for index-looking fields.

### DIFF
--- a/ludwig/utils/automl/data_source.py
+++ b/ludwig/utils/automl/data_source.py
@@ -1,9 +1,13 @@
 from abc import ABC, abstractmethod
 from typing import List, Tuple
 
+import dask.dataframe as dd
+import pandas as pd
+
 from ludwig.utils.audio_utils import is_audio_score
 from ludwig.utils.automl.utils import avg_num_tokens
 from ludwig.utils.image_utils import is_image_score
+from ludwig.utils.misc_utils import memoized_method
 from ludwig.utils.types import DataFrame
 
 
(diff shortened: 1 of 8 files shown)


ludwig/utils/automl/data_source.py
```diff
@@ -1,9 +1,13 @@
 from abc import ABC, abstractmethod
 from typing import List, Tuple
 
+import dask.dataframe as dd
```
justinxzhao, line 4: This dask import should be ok since users using Ludwig automl are expected to `pip install ludwig[distributed]`. Is that right?

justinxzhao (approved):
Thank you for this change!

ludwig/utils/automl/data_source.py
tgaddair: Yeah, the automl library requires Ray and Dask to work.

PR #2509, opened 2022-09-16
Add in-memory dataset size calculation

arnavgarg1:
Based on a conversation with an OSS user, it is useful to surface the size of the in-memory dataset to Ludwig users after preprocessing so that they can better estimate their cluster memory. 

This will also be useful for profiling and debugging.

Cost of running `RayDataset.size_bytes()` is O(1), while cost of running `df.memory_usage(deep=True)` should be minimal (can't find a computational complexity approximation anywhere). 

In the future, we can pass a flag to decide whether to compute these values or not if we find this to be a bottleneck.

### DIFF
--- a/ludwig/data/dataset/base.py
+++ b/ludwig/data/dataset/base.py
@@ -31,6 +31,10 @@ class Dataset(ABC):
     def to_df(self):
         raise NotImplementedError()
 
+    @property
+    def in_memory_size_bytes(self):
+        raise NotImplementedError()
+
 
 class DatasetManager(ABC):
     @abstractmethod
--- a/ludwig/data/dataset/pandas.py
+++ b/ludwig/data/dataset/pandas.py
@@ -74,6 +74,12 @@ class PandasDataset(Dataset):
     def __len__(self):
         return self.size
 
+    @property
+    def in_memory_size_bytes(self):
+        df = self.to_df()
+        memory_usage = df.memory_usage(deep=True).sum() if not None else 0
+        return memory_usage
+
     @contextlib.contextmanager
     def initialize_batcher(self, batch_size=128, should_shuffle=True, seed=0, ignore_last=False, horovod=None):
         sampler = DistributedSampler(len(self), shuffle=should_shuffle, seed=seed, horovod=horovod)
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -130,6 +130,11 @@ class RayDataset(Dataset):
     def size(self):
         return len(self)
 
+    @property
+    def in_memory_size_bytes(self):
+        memory_usage = self.ds.size_bytes() if not None else 0
+        return memory_usage
+
     def to_df(self):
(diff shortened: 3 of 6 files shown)


jppgks:
nit: is it possible to add this as a column to the existing table? So columns would be:
- Dataset name 
- Datset size (rows)
- Dataset size (bytes)

ludwig/data/dataset/pandas.py
```diff
@@ -74,6 +74,12 @@ def get_dataset(self):
     def __len__(self):
         return self.size
 
+    @property
+    def in_memory_size_bytes(self):
+        df = self.to_df()
+        memory_usage = df.memory_usage(deep=True).sum() if not None else 0
```
jppgks, line 80: `if not None` will always evaluate to `True`

ludwig/data/dataset/base.py
```diff
@@ -31,6 +31,10 @@ def initialize_batcher(self, batch_size=128, should_shuffle=True, seed=0, ignore
     def to_df(self):
         raise NotImplementedError()
 
+    @property
+    def in_memory_size_bytes(self):
+        raise NotImplementedError()
```
jppgks, line 36: Can we add an implementation for dask too?

ludwig/data/dataset/pandas.py
arnavgarg1: Good catch, will change this

ludwig/data/dataset/base.py
arnavgarg1: @jppgks  believe Dask/Modin/Pandas all use the PandasDatasetManager class, and all 3 have the same API to check memory usage. I will double check just in case I missed this.

arnavgarg1:
> nit: is it possible to add this as a column to the existing table? So columns would be:
> - Dataset 
> - Size (rows)
> - Size (bytes)
> 
> Also, can use `format_memory` helper  from benchmarking utils to show human-readable units

Yup, I can make this change. Also good to know about the format_memory helper, I'll make sure to include that.

ludwig/data/dataset/ray.py
```diff
@@ -130,6 +130,11 @@ def __len__(self):
     def size(self):
         return len(self)
 
+    @property
+    def in_memory_size_bytes(self):
+        memory_usage = self.ds.size_bytes() if not None else 0
```
tgaddair, line 135: Should be `if self.ds is not None`. As written, this will always resolve to `True`.

tgaddair (commented):
Only one issue I see, then LGTM.

ludwig/data/dataset/ray.py
tgaddair: Hmm, maybe I'm wrong about this. Just tested it and seems to work...

arnavgarg1: So self.ds may be None, but I think even size_bytes() can be none for whatever reason (probably when ds is None)

tgaddair (approved):

PR #2523, opened 2022-09-18
Updated schedule logic for placement groups for ray backend

magdyksaleh:


ludwig/backend/ray.py
```diff
@@ -822,7 +834,9 @@ def provision_preprocessing_workers(self):
             yield
         else:
             num_cpu = self._preprocessor_kwargs["num_cpu_workers"]
-            self._preprocessor_pg = placement_group([{"CPU": num_cpu}])
+            bundles = self.generate_bundles(num_cpu)
```
tgaddair, line 837: This comment is for the code above, but we shouldn't be logging a warning if the user doesn't set this flag (warnings indicate there might be something wrong, but here it's just that some option wasn't enabled).

I would also remove the `use_preprocessing)placement_group` param, since it looks like this should be assumed when `num_cpu_workers` is set. So we can just check for the existence of `num_cpu_workers` instead.

Finally, I would consider renaming this param to just `num_cpus` since we're just asking for a pool of CPUs, not specifically a number of actors, etc.

ludwig/backend/ray.py
```diff
@@ -812,6 +812,18 @@ def initialize(self):
         # Disable placement groups on dask
         dask.config.set(annotations={"ray_remote_args": {"placement_group": None}})
 
+    def generate_bundles(self, num_cpu):
+        # Ray requires that each bundle be scheduleable on a single node.
+        # So a bundle of 320 cpus would never get scheduled. For now a simple heuristic
+        # to be used is to just request cpus 4 at a time.
+        if num_cpu <= 4:
```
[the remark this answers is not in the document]
tgaddair: We have an example here of setting 1 CPU per bundle:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/hyperopt/execution.py#L88

tgaddair:
@magdyksaleh I'm assuming you want this to be included in 0.6? In that case, you should add the `release-0.6` label to this PR.

tgaddair:
Also, you should add a test to make sure this works as expected.

ShreyaR (approved):

ludwig/backend/ray.py
magdyksaleh: done

magdyksaleh:
> Also, you should add a test to make sure this works as expected.

Added

ludwig/backend/ray.py
```diff
@@ -812,17 +812,24 @@ def initialize(self):
         # Disable placement groups on dask
         dask.config.set(annotations={"ray_remote_args": {"placement_group": None}})
 
+    def generate_bundles(self, num_cpu):
```
jeffreyftang, line 815: Nit: prefix with underscore since this is not really meant to be used outside this package.

jeffreyftang (approved):

PR #2524, opened 2022-09-19
Ensure resource availability for ray datasets workloads when running on cpu clusters

arnavgarg1:
This change ensures that there are always at least 2 CPUs available for Ray Dataset related when running hyperopt trials with a RayBackend to prevent trials from stalling. 

This is particularly helpful on RayBackends that are entirely composed of CPUs to prevent infinite stalls when running hyperopt when `num_samples` >= number of available CPUs on the machine or cluster.

This PR is in response to:
- https://github.com/ludwig-ai/ludwig/issues/2403
- https://github.com/ludwig-ai/ludwig/issues/2465


ludwig/backend/horovod.py
```diff
@@ -69,4 +69,16 @@ def is_coordinator(self):
...
+
+    @property
+    def num_cpus(self) -> int:
+        # TODO: Implement
+        return 1
+
+    @property
+    def num_gpus(self) -> int:
+        # TODO: Implement
+        return 0
```
arnavgarg1, line 84: @tgaddair I'm not sure what to do for HorovodBackend - any help would be great!

arnavgarg1: Would 
- CPUs = `self._horovod.size()`
- GPUs = `self._horovod.local_rank()`
?

ludwig/backend/horovod.py
```diff
@@ -69,4 +71,20 @@ def is_coordinator(self):
 
     @property
     def num_nodes(self) -> int:
+        if not self._horovod:
+            return 1
         return self._horovod.size()
+
+    @property
+    def num_cpus(self) -> int:
```
tgaddair, line 79: The Horovod backend is not used with Horovod on Ray (that's what the Ray bacend does). The correct way to implement this would be to compute number of CPUs / GPUs on each worker independently, then allreduce the results:

```
cpu_count = psutil.cpu_count()
return hvd.allreduce(cpu_count)
```

ludwig/backend/horovod.py
```diff
@@ -69,4 +71,20 @@ def is_coordinator(self):
...
+        # Since we use Horovod on Ray, return the number of CPUs on the Ray cluster.
+        if not ray.is_initialized():
+            return 1
+        return ray.cluster_resources().get("CPU", 1)
+
+    @property
+    def num_gpus(self) -> int:
```
tgaddair, line 86: Same as above.

ludwig/backend/ray.py
```diff
@@ -147,8 +147,9 @@ def _create_pandas_engine(**kwargs):
 
 
 def _get_df_engine(processor):
-    logger.info(f"Ray processor params: {processor}")
-    if processor is None:
+    if processor:
```
tgaddair, line 150: Can clean this up:

```
if processor is None:
    return _create_dask_engine()

logger.info(f"Ray processor params: {processor}")
processor_kwargs = processor.copy()
...
```

That way we don't need to keep extra if-else statements (too many if-elses are generally considered an anti-pattern).

ludwig/hyperopt/run.py
```diff
@@ -34,12 +37,16 @@
 from ludwig.utils.misc_utils import get_class_attributes, get_from_registry, set_default_value, set_default_values
 
 try:
+    from ludwig.backend.horovod import HorovodBackend
```
tgaddair, line 40: Seems this is unused. Remove import?

tgaddair (changes requested):
I'm not sure about this approach in general, using `max_concurrent_trials` to ensure resource availability for Ray Datasets. For one thing, this would not work well with an auto-scaling cluster, and I'm not sure how it would play out for a heterogenous cluster either.

Seems like this is something that would be better solved with placement groups. We should chat about it.

ludwig/backend/horovod.py
arnavgarg1: Got it, that makes a lot of sense!

arnavgarg1:
> I'm not sure about this approach in general, using `max_concurrent_trials` to ensure resource availability for Ray Datasets. For one thing, this would not work well with an auto-scaling cluster, and I'm not sure how it would play out for a heterogenous cluster either.
> 
> Seems like this is something that would be better solved with placement groups. We should chat about it.

That makes sense to me. I agree that this implementation would underutilize resources for an auto-scaling cluster. 
One option is to only set `max_concurrent_trials` like this is it's a CPU only cluster and skip this step altogether for heterogenous clusters, but this would still not be ideal for an auto-scaling cluster. 

With placement groups, are we able to separate out what resources will be used by what tasks/actors? For e.g. always reserve one CPU resource for RayDatasets and then use the rest of the resources to, say RayTune? In that case, do you know if RayTune would automatically limit `max_concurrent_trials` to the number of resources available for the placement group being used? If yes, then I agree that placement groups is the right way to go to account for both auto-scaling and al cluster configurations

ludwig/hyperopt/run.py
```diff
@@ -423,7 +429,80 @@ def log_warning_if_all_grid_type_parameters(hyperopt_parameter_config: Dict[str,
...
+    Set max_concurrent_trials in the hyperopt executor to ensure CPU resources are available for Ray Dataset related
+    tasks.
+    """
+    cpu_resources_per_trial = executor_config.get(CPU_RESOURCES_PER_TRIAL, 1)
+    num_samples = executor_config.get(NUM_SAMPLES)
+    max_concurrent_trials = executor_config.get(MAX_CONCURRENT_TRIALS)
+
+    # Default to num_samples if max_concurrent_trials isn't set, otherwise use the value from the config
+    is_auto = max_concurrent_trials == AUTO
+    max_concurrent_trials = num_samples if max_concurrent_trials == AUTO else max_concurrent_trials
```
ShreyaR, line 445: I don't think we should default `max_concurrent_trials` to `num_samples` if it is set. The default `max_concurrent_trials` value when set to `AUTO` should be determined by the number of available CPUs.

If `num_samples >= num_cpus_available`, then each trial takes up a CPU with no additional CPUs left for dataset stuff.

ludwig/hyperopt/run.py
```diff
@@ -423,7 +429,80 @@ def log_warning_if_all_grid_type_parameters(hyperopt_parameter_config: Dict[str,
...
+
+    # Max_concurrent_trials set to None, no need to infer
+    if not max_concurrent_trials:
```
ShreyaR, line 448: We should also add a return in cases when `max_concurrent_trials` is set to an integer value in the executor config by the user.

ludwig/hyperopt/run.py
```diff
@@ -423,7 +429,80 @@ def log_warning_if_all_grid_type_parameters(hyperopt_parameter_config: Dict[str,
     )
 
 
-def update_hyperopt_params_with_defaults(hyperopt_params):
+def update_or_set_max_concurrent_trials(executor_config: dict, backend: Backend) -> None:
+    """Datasets read tasks request 0.5 CPUs and all transformation tasks request 1 CPU, so if there are no cores
+    available, trials won't be able to run.
```
arnavgarg1, line 434: Make this more readable. 3 conditions:
- Auto
- None
- Set by the user

arnavgarg1:
Created this PR to address the test failure: https://github.com/ludwig-ai/ludwig/pull/2610, so this PR is now ready to review @ShreyaR @tgaddair

tests/ludwig/hyperopt/test_hyperopt.py
```diff
@@ -75,3 +76,96 @@
 def test_substitute_parameters(parameters, expected):
     actual_config = substitute_parameters(BASE_CONFIG, parameters)
     assert actual_config == expected
+
```
ShreyaR, line 79: Thanks for the comprehensive testing!

tests/integration_tests/test_preprocessing.py
```diff
@@ -374,7 +374,6 @@ def test_empty_training_set_error(backend, tmpdir, ray_cluster_2cpu):
         ludwig_model.preprocess(dataset=df)
 
 
-@pytest.mark.distributed
```
ShreyaR, line 377: nit: Can you move general test housekeeping to a separate PR?

tests/integration_tests/test_hyperopt.py
```diff
@@ -622,3 +633,51 @@ def test_hyperopt_grid_search_more_than_one_sample(csv_filename, tmpdir, ray_clu
 
     with pytest.warns(RuntimeWarning):
         hyperopt(config, dataset=rel_path, output_directory=tmpdir, experiment_name="test_hyperopt")
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize(
+    "max_concurrent_trials", ["auto", None], ids=["max_concurrent_trials_auto", "no_max_concurrent_trials_specified"]
+)
+def test_hyperopt_with_infer_max_concurrent_trials(csv_filename, tmpdir, max_concurrent_trials, ray_cluster):
```
ShreyaR, line 642: How expensive is it to run this test? My concern is that we already test out that `update_max_concurrent_trials` works as expected in the hyperopt unit tests below. Since this test requires running a full hyperopt experiment, it ends up being an expensive test to run to test out functionality that is already mostly covered elsewhere. This is also in the recent context that we've seen of a lot of timeouts in hyperopt testing.

What do you think?

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -318,3 +334,42 @@ def run_hyperopt(
 
     # check for existence of the hyperopt statistics file
     assert os.path.isfile(os.path.join(tmpdir, experiment_name, HYPEROPT_STATISTICS_FILE_NAME))
+
+
+@pytest.mark.distributed
+def test_hyperopt_with_infer_max_concurrent_trials(csv_filename, tmpdir, ray_cluster_4cpu):
```
ShreyaR, line 340: Same comment as above for this test.

ShreyaR (commented):
Looks great! Left some small comments. Also looks like some tests are failing.

arnavgarg1:
@ShreyaR Thanks for the review - will update!

As for the failing test - created this PR to address the test failure: https://github.com/ludwig-ai/ludwig/pull/2610
It just requires reducing the level of precision for `np.isclose()`

tests/integration_tests/test_hyperopt.py
arnavgarg1: @ShreyaR Great points! 

I agree that we may not need the entire test. My incentive to keep this test is that it ensures that `hyperopt_statistics.json` does actually reflect that this value was set. One very easy way of stripping this test down to the bare bones is to basically remove the check for number of trials run, set time_budget_s to something like 15 seconds, and then just do a check for `max_concurrent_trials` in `hyperopt_statistics.json`. What do you think?

ShreyaR (approved):

ludwig/hyperopt/utils.py
```diff
@@ -141,3 +173,105 @@ def substitute_parameters(
...
+        raise RuntimeError("Atleast 2 CPUs are required for hyperopt")
+
+    if backend.BACKEND_TYPE == "local":
+        max_possible_trials = (num_cpus_available - 1) // cpus_per_trial
+    else:
+        # Ray requires at least 2 free CPUs to ensure trials don't stall
+        max_possible_trials = (num_cpus_available - 2) // cpus_per_trial
+
+    if max_possible_trials < 1:
+        raise ValueError("Not enough CPUs available for hyperopt, reduce the number of CPUs requested per trial")
```
tgaddair, line 264: We cannot make this assumption outright, as the user may be using an auto-scaling cluster. This is why we should allow them to set it explicitly, even if it doesn't "look" like it will work.

ludwig/hyperopt/utils.py
```diff
@@ -141,3 +173,105 @@ def substitute_parameters(
...
+        raise ValueError(f"Set '{CPU_RESOURCES_PER_TRIAL}' to > 0")
+
+    if num_cpus_available < 2:
+        raise RuntimeError("Atleast 2 CPUs are required for hyperopt")
+
+    if backend.BACKEND_TYPE == "local":
+        max_possible_trials = (num_cpus_available - 1) // cpus_per_trial
+    else:
+        # Ray requires at least 2 free CPUs to ensure trials don't stall
+        max_possible_trials = (num_cpus_available - 2) // cpus_per_trial
```
tgaddair, line 261: Should probably be:

```
max_concurrent_trials = num_cpus_available // cpus_per_trial - 1
```

So basically, always leave one trial out to make room for Ray Datasets, which only strictly requires 1 CPU to succeed.

ludwig/hyperopt/utils.py
```diff
@@ -141,3 +173,105 @@ def substitute_parameters(
...
+
+    executor = get_from_registry(hyperopt_params[EXECUTOR][TYPE], executor_registry)
+    executor_defaults = {k: v for k, v in executor.__dict__.items() if k in get_class_attributes(executor)}
+    set_default_values(
+        hyperopt_params[EXECUTOR],
+        executor_defaults,
+    )
+
+
+def update_or_set_max_concurrent_trials(hyperopt_config: dict, backend: Backend) -> None:
```
tgaddair, line 229: Let's add unit (not integration) tests for this function. Can put them in `tests/ludwig/backend.ray.py` if we make the change I suggested to move this into the Backend.

ludwig/hyperopt/utils.py
```diff
@@ -141,3 +173,105 @@ def substitute_parameters(
...
+
+    if max_concurrent_trials == AUTO:
+        if max_possible_trials > num_total_trials:
+            # If all trials easily fit on the available CPUs, then remove max_concurrent_trials
+            del hyperopt_config[EXECUTOR][MAX_CONCURRENT_TRIALS]
+            return
+        max_concurrent_trials = max_possible_trials
+    else:
+        # Use min to respect value set in user config incase it is smaller
+        max_concurrent_trials = min(max_concurrent_trials, max_possible_trials)
```
tgaddair, line 274: As said above, we should not override the user's explicit setting.

ludwig/hyperopt/utils.py
```diff
@@ -141,3 +173,105 @@ def substitute_parameters(
...
+        max_possible_trials = (num_cpus_available - 1) // cpus_per_trial
+    else:
+        # Ray requires at least 2 free CPUs to ensure trials don't stall
+        max_possible_trials = (num_cpus_available - 2) // cpus_per_trial
+
+    if max_possible_trials < 1:
+        raise ValueError("Not enough CPUs available for hyperopt, reduce the number of CPUs requested per trial")
+
+    if max_concurrent_trials == AUTO:
+        if max_possible_trials > num_total_trials:
```
tgaddair, line 267: I don't see any harm in this, but I don't think it's necessary either. If `max_concurrent_trials > num_total_trials`, then it's effectively a no-op, which is fine. If we remove this logic, it simplifies the code and allows us to avoid having to compute `num_total_trials`.

tgaddair (changes requested):
Some issues I see with the implementation here. I think there's an opportunity to dramatically simplify, which we should do. The blocker is that we should not be overriding the user's explicit setting of this param, as they may have auto-scaling enabled.

ludwig/hyperopt/utils.py
tgaddair: Looks like we can just take the unit tests from `test_hyperopt.py` and adapt them to the new API.

ludwig/hyperopt/utils.py
arnavgarg1: Makes sense, I will just return None so it doesn't get set

ludwig/backend/horovod.py
```diff
@@ -83,3 +84,6 @@ def get_available_resources(self) -> Resources:
         gpus = self._horovod.allreduce(gpus, op=self._horovod.Sum).item()
 
         return Resources(cpus=cpus, gpus=gpus)
+
+    def get_max_concurrent_trials(self, hyperopt_config: Dict[str, Any]) -> Union[int, None]:
+        return None
```
arnavgarg1, line 89: Need to add the same logic from the RayBackend here

ludwig/backend/horovod.py
```diff
@@ -83,3 +84,23 @@ def get_available_resources(self) -> Resources:
...
+            raise RuntimeError("At least 2 CPUs are required for hyperopt when using a HorovodBackend.")
+
+        # Ray requires at least 2 free CPUs to ensure trials don't stall
+        max_possible_trials = (num_cpus_available - 2) // cpus_per_trial
+
+        # Users may be using an autoscaling cluster, so return None
+        if max_possible_trials < 1:
+            return None
+
+        return max_possible_trials
```
arnavgarg1, line 106: Refactor to pull this function out for RayBackend and HorovodBackend to prevent code duplication

ludwig/hyperopt/utils.py
arnavgarg1: Unfortunately, this seems to still cause trials to compete for resources from all of the testing I've done. However, if we leave two free CPUs, things seem to be fine so that is why I've implemented the logic this way.

ludwig/backend/ray.py
```diff
@@ -1008,6 +1017,26 @@ def get_available_resources(self) -> Resources:
...
+    def max_concurrent_trials(self, hyperopt_config: Dict[str, Any]) -> Union[int, None]:
+        cpus_per_trial = hyperopt_config[EXECUTOR].get(CPU_RESOURCES_PER_TRIAL, 1)
+        num_cpus_available = self.get_available_resources().cpus
+
+        # No actors will compete for ray datasets tasks dataset tasks are cpu bound
+        if cpus_per_trial == 0:
+            return None
+
+        if num_cpus_available < 2:
+            raise RuntimeError("At least 2 CPUs are required for hyperopt when using a RayBackend.")
```
tgaddair, line 1029: Could be an auto-scaling cluster in this event as well. So I would suggest logging a warning here with this message and returning `None`. Maybe make the message:

```
At least 2 CPUs are required for hyperopt when using a RayBackend, but only found {num_cpus_available}. If you are not using an auto-scaling Ray cluster, your hyperopt trials may hang.
```

ludwig/backend/ray.py
```diff
@@ -1008,6 +1017,26 @@ def get_available_resources(self) -> Resources:
...
+
+        # Ray requires at least 2 free CPUs to ensure trials don't stall
+        max_possible_trials = int((num_cpus_available - 2) // cpus_per_trial)
+
+        # Users may be using an autoscaling cluster, so return None
+        if max_possible_trials < 1:
+            return None
```
tgaddair, line 1036: We should log a warning, like: 

```
Hyperopt trials will request {cpus_per_trial} CPUs in addition to CPUs needed for Ray Datasets, but only {num_cpus_available} CPUs are currently available. If you are not using an auto-scaling Ray cluster, your hyperopt trials may hang.
```

ludwig/backend/ray.py
```diff
@@ -1008,6 +1017,26 @@ def get_available_resources(self) -> Resources:
...
+
+        # No actors will compete for ray datasets tasks dataset tasks are cpu bound
+        if cpus_per_trial == 0:
+            return None
+
+        if num_cpus_available < 2:
+            raise RuntimeError("At least 2 CPUs are required for hyperopt when using a RayBackend.")
+
+        # Ray requires at least 2 free CPUs to ensure trials don't stall
+        max_possible_trials = int((num_cpus_available - 2) // cpus_per_trial)
```
tgaddair, line 1032: Are you sure we need 2 free CPUs? I believe it should just be 1. But please correct me if I'm missing something!

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -273,8 +276,8 @@ def test_hyperopt_run_hyperopt(csv_filename, ray_mock_dir, ray_cluster_7cpu):
                 "lower": 0.001,
                 "upper": 0.1,
             },
-            output_feature_name + ".output_size": {"space": "randint", "lower": 2, "upper": 32},
-            output_feature_name + ".num_fc_layers": {"space": "randint", "lower": 2, "upper": 6},
+            output_feature_name + ".output_size": {"space": "randint", "lower": 2, "upper": 8},
```
tgaddair, line 279: `test_hyperopt_run_hyperopt` is a good one to test this feature out with. You can set `max_concurrent_trials=auto`, the bump up the `cpus_per_trial=2`. If your code works, then the test should pass, but if it doesn't, it will hang. So we can test this logic for free without needing to add a new test.

tgaddair (commented):
Looks great! Just a couple remaining points before landing.

tests/integration_tests/test_hyperopt_ray_horovod.py
arnavgarg1: @tgaddair So that test is currently skipped because of some sort of issues with dynamic resource allocation.

Is it worth adding to `test_hyperopt_executor`, specifically the test with variant generator?

tgaddair (approved):

PR #2531, opened 2022-09-19
Cleanup: move to per-module loggers instead of the global logging object.

justinxzhao:


arnavgarg1:
@justinxzhao Out of curiosity, what's the reason for this change? Does it help with performance / better readability for when errors/info are logged?

justinxzhao:
@arnavgarg1 I think so. This is motivated based on some recent work training GBM models locally using Jupyter notebooks.

It appears that the global logging statements aren't properly governed by the logging_level set in the `LudwigModel` constructor.

arnavgarg1:
Oh that's pretty interesting, and definitely makes sense. I haven't thought about that before. Good change 👍

arnavgarg1 (approved):

PR #2535, opened 2022-09-20
Hyperopt: Log warning with num_extra_trials if all grid search parameters and num_samples > 1

arnavgarg1:
As the title of this PR says, this logs a warning if `num_samples` > 1 and there are only grid_search type parameters. This is to make sure the user is aware that extra, duplicate trials are being created.


justinxzhao (approved):
It looks like an old test (that I think was fixed) is broken. Do you need to rebase?

arnavgarg1:
@justinxzhao That was a failure I was seeing for Ray nightly with Python 3.9. It's weird that its now showing up for 3.8. I can take a look and create a follow-up PR! I don't think it's related to this change.

PR #2540, opened 2022-09-20
Adds model configs to ludwig.datasets

dantreiman:
Model configs are stored in `ludwig.datasets.model_configs`

Each model config must be prefixed by the name of its dataset, ex. `higgs_default.yaml`, `higgs_best.yaml`.
`<dataset>_default.yaml` is a special name which marks this config as the default config for the dataset.

Adds two APIs to get model configs for a dataset:
1. Functional api `ludwig.datasets.model_configs_for_dataset`
```python
higgs_configs = ludwig.datasets.model_configs_for_dataset("higgs")
```

2.  Accessor method of DatasetLoader:
```python

titanic_model_config = titanic.default_model_config

model = ludwig.api.LudwigModel(titanic_model_config)
```


ludwig/datasets/__init__.py
```diff
@@ -25,21 +25,47 @@ def _get_dataset_configs() -> Dict[str, DatasetConfig]:
...
 
+def _load_model_config(model_config_filename: str):
+    """Loads a model config."""
+    model_config_path = os.path.join(os.path.dirname(model_configs.__file__), model_config_filename)
+    with open(model_config_path) as f:
+        return yaml.safe_load(f)
+
+
+@lru_cache(maxsize=3)
+def _get_model_configs(dataset_name: str) -> Dict[str, Dict]:
```
justinxzhao, line 48: Can the scope of this be expanded to fetch the "best" model config, like the configs that @abidwael and @jppgks have been checking in?

e.g. https://github.com/ludwig-ai/ludwig/pull/2536

justinxzhao (commented):
Nice.

ludwig/datasets/__init__.py
dantreiman: Good idea!  Added another special model config name "best", similar to "default" - what do you think of this?

justinxzhao (approved):

ludwig/datasets/__init__.py
justinxzhao: Looks good to me. The key-lookup API seems nicely general enough to expand to other categories as well, if/when we introduce them.

ludwig/datasets/README.md
```diff
@@ -14,24 +15,82 @@ dataset_df = titanic.load()
...
+
+```python
 print(ludwig.datasets.describe_dataset("titanic"))
+```
+
+______________________________________________________________________
+
+### get_dataset
+
+Get a dataset by name
```
abidwael, line 48:
```suggestion
Get a dataset module by name
```

ludwig/datasets/README.md
```diff
@@ -14,24 +15,82 @@ dataset_df = titanic.load()
...
+______________________________________________________________________
+
+### model_configs_for_dataset
+
+Gets a dictionary of model configs for the specified dataset.  Keys are the config names, and may
+contain the special keys:
+
+- `default` - The default config for the dataset.  Should train to decent performance under 10 minutes on a tyipcal
+  laptop without GPU.
+- `best` - The best known config for the dataset.  Should be replaced when a better config is found.  This is a good
```
abidwael, line 65: Could we split this into `best_ecd` and `best_gbm`? A user could be interested in a more lightweight model than a neural network.

ludwig/datasets/model_configs/adult_census_income_default.yaml
```diff
@@ -0,0 +1,42 @@
...
+combiner:
+  type: concat
+  num_fc_layers: 3
+  fc_size: 128
+  dropout: 0.1
+training:
+  batch_size: 256
+  learning_rate: .001
+  epochs: 1
+  steps_per_checkpoint: 1
```
abidwael, line 42: For default configs, I was thinking that they would only contain `input_features` and `output_features`. The rest would be taken from the defaults we define, and if we find better general defaults than the ones we currently have, we can update those. This way the default config could be a blank canvas for experimentation.

tests/ludwig/datasets/model_configs/train_all_model_configs.py
```diff
@@ -0,0 +1,133 @@
...
+class TrainingResults:
+    """Results of a training run for a dataset."""
+
+    dataset_version: str
+    dataset_name: str
+    has_config: bool
+    output_directory: Optional[str] = None
+    splits: Optional[str] = None
+    metric: Optional[str] = None
+    performance: Optional[float] = None
```
abidwael, line 29: These two should probably be two lists or a dict mapping from metric to its value. As an example, a regression dataset like `sarcos` has the following metrics:
- `mean_absolute_error`
- `mean_squared_error`
- `r2`
- `root_mean_squared_error`
- `root_mean_squared_percentage_error`

tests/ludwig/datasets/model_configs/train_all_model_configs.py
```diff
@@ -0,0 +1,133 @@
...
+from typing import Optional
+
+import pandas as pd
+
+from ludwig import datasets, visualize
+from ludwig.api import LudwigModel
+
+
+@dataclass
+class TrainingResults:
```
abidwael, line 20: We can also add the current commit hash of the repo, as performance could change based on that.

tests/ludwig/datasets/model_configs/train_all_model_configs.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _train_dataset_process(dataset_name, results_queue):
+    """Runs each train job in a new process."""
+    load_start_time = time.time()
+    dataset = datasets.get_dataset(dataset_name)
+    config = dataset.default_model_config
+    df = dataset.load()
+    load_end_time = time.time()
+    if "split" not in df:
+        df["split"] = 0
```
abidwael, line 43: As splits might change based on the dataset config, should we consider using `from ludwig.utils.dataset_utils import get_repeatable_train_val_test_split`

abidwael (commented):
Nice addition to the datasets API!

ludwig/datasets/README.md
dantreiman: Lets keep `best` as "the best config we've seen so far" to keep it simple.  Adding GBM configs would be a good follow-up though.

ludwig/datasets/model_configs/adult_census_income_default.yaml
dantreiman: I think this is a good goal to iterate towards.  As it stands, training would take too long on some datasets and the default encoders for some datatypes will use a lot of memory.

tests/ludwig/datasets/model_configs/train_all_model_configs.py
dantreiman: Some of the datasets here have splits which match the splits used in the literature, so I don't want to re-split those.

Also, this test script is more of a sanity check than a benchmark, so I think its OK that some datasets don't have proper splits.

tests/ludwig/datasets/model_configs/train_all_model_configs.py
dantreiman: Done.

tests/ludwig/datasets/model_configs/train_all_model_configs.py
dantreiman: I considered this, but I wanted a simple table of outputs to look at, and putting many metrics in a table makes things harder to read.  This isn't a real benchmark, just a quick-and-simple test script which can be run on a laptop.

PR #2543, opened 2022-09-22
ZScore Normalization Failure When Using Constant Value Number Feature

arnavgarg1:
This PR raises a warning when a number feature is being used that has a constant value across all rows of the dataset, along with preprocessing being set to `zscore`. This results in a std deviation of 0 and a divide by zero error. This results in the feature being processed as NaNs, the downstream metrics to be NaNs, etc. which isn't a good outcome. 

Questions:
- Is this a good enough reason to raise a RuntimeError and cause the model execution to stop?
- Would it instead make sense to:
    - Log a warning stating the reason why zscore normalization failed
    - Just return the same value back, i.e., `self.mu` which is the average of the column == the value itself?

### DIFF
--- a/ludwig/features/number_feature.py
+++ b/ludwig/features/number_feature.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import copy
 import logging
 import random
 from typing import Any, Dict, Union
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):
Raising an error seems like the right behavior to me. If it's a constant feature, it should absolutely be removed.

PR #2545, opened 2022-09-22
Adds class names to calibration plot title, reformats Brier scores as grouped bar chart

dantreiman:
This is the result of running `train_mushroom_edibility_calibrated.py`, which generates two models "Uncalibrated" and "Calibrated" for the dataset with two classes "e" and "p" (edible and poisonous)

Before:
calibration_1_vs_all_0
calibration_1_vs_all_1
calibration_1_vs_all_brier

---

After:

calibration_1_vs_all_0
calibration_1_vs_all_1
calibration_1_vs_all_brier


justinxzhao (approved):
LGTM!

PR #2547, opened 2022-09-23
Remove global torchtext version-specific tokenizer availability warnings.

justinxzhao:
Currently, a tokenizer availability warning is always logged unless `torchtext>=0.13.0` is installed, regardless of what kind of model you have.

Instead, we remove the pan-warning in favor of raising a `KeyError` during tokenizer registry access, with additional help text if the requested tokenizer would be available for a different version of torchtext.

Valid tokenizers are also enforced by the schema.


dantreiman (approved):
Accessor functions instead of directly importing registry objects.  Yes!

I agree we shouldn't warn on import for features the user hasn't asked for yet.

LGTM.

PR #2549, opened 2022-09-23
Update version to 0.6.

justinxzhao:


arnavgarg1:
Excited for this change! We should probably wait until:

1. https://github.com/ludwig-ai/ludwig/pull/2548 finishes running so that we know there's no more Ray nightly failures
2. Root fix for the `tests/integration_tests/test_ray.py::test_ray_image[pandas+numpy_images]` before merging this in?

arnavgarg1 (approved):

justinxzhao:
I don't think we should block on the `tests/integration_tests/test_ray.py::test_ray_image[pandas+numpy_images]`. We can always push 0.6.1 when that gets fixed.

arnavgarg1:
@justinxzhao Got it! I think I've figured out the cause for the test failure, just working on a fix :) Will get it merged in today! 
If that doesn't happen by today evening, let's skip the test and release 0.6.

arnavgarg1:
@justinxzhao I think we might need to cherry pick https://github.com/ludwig-ai/ludwig/commit/519be85d9883117bd4377ce16f3fec43f38481cd for the Ray 3.9 failures

PR #2550, opened 2022-09-24
int: bump Dask version

geoffreyangus:
This PR updates the version of Dask and implements PyArrow schema inference so that both cached inputs and predictions can be saved to parquet files. This code path was relatively under-tested, particularly for Dask with Ray, and so this PR also implements a unit test.

Some things worth calling out:
1. The unit test shows that we only have partial coverage for caching inputs and outputs. Notably, the H3 input feature and the sequence and text output features are incompatible with this code path.
2. NaN handling for date input features make them incompatible with the ray test, which does an equality check between the local and ray backends to verify correctness. Because date input features fill using the present time, they cannot be equal.
3. There is no NaN handling for vector input features.

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -22,7 +22,7 @@ from ludwig.globals import (
 )
 from ludwig.models.base import BaseModel
 from ludwig.progress_bar import LudwigProgressBar
-from ludwig.utils.data_utils import save_csv, save_json
+from ludwig.utils.data_utils import save_csv, save_json, get_pa_schema
 from ludwig.utils.dataframe_utils import flatten_df, from_numpy_dataset
 from ludwig.utils.horovod_utils import return_first
 from ludwig.utils.print_utils import repr_ordered_dict
@@ -317,7 +317,8 @@ def save_prediction_outputs(
     backend,
 ):
     postprocessed_output, column_shapes = flatten_df(postprocessed_output, backend)
-    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME), schema=None)
+    schema = get_pa_schema(postprocessed_output)
+    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME), schema=schema)
     save_json(os.path.join(output_directory, PREDICTIONS_SHAPES_FILE_NAME), column_shapes)
     if not backend.df_engine.partitioned:
         # csv can only be written out for unpartitioned df format (i.e., pandas)
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -33,6 +33,7 @@ from typing import Any, Dict, List, Tuple, Union
 
 import numpy as np
 import pandas as pd
+import pyarrow as pa
 import yaml
 from fsspec.config import conf, set_conf_files
 from pandas.errors import ParserError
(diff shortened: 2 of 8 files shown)


justinxzhao (approved):

ludwig/utils/data_utils.py
```diff
@@ -827,6 +829,24 @@ def is_ludwig_ndarray_string(parm_string):
...
+    if np.isscalar(obj):
+        return pa.from_numpy_dtype(np.array(obj).dtype)
+    elif isinstance(obj, np.ndarray) or isinstance(obj, list) or isinstance(obj, tuple):
+        return pa.list_(get_pa_dtype(obj[0]))
+    else:
+        raise ValueError(f"Unsupported type for pyarrow dtype: {type(obj)}")
+
+
+def get_pa_schema(df: DataFrame):
+    head = df.head(1)
```
ShreyaR, line 842: Would there be scenarios in which inferring schema from the first row may not be sufficient? E.g. in the case of `nan`s in the first row.

ludwig/models/predictor.py
```diff
@@ -317,7 +317,8 @@ def save_prediction_outputs(
     backend,
 ):
     postprocessed_output, column_shapes = flatten_df(postprocessed_output, backend)
-    postprocessed_output.to_parquet(os.path.join(output_directory, PREDICTIONS_PARQUET_FILE_NAME), schema=None)
+    schema = get_pa_schema(postprocessed_output)
```
arnavgarg1, line 320: Do we need to manually pass in the schema to `to_parquet()` here given that we already make this call within `to_parquet()` here? https://github.com/ludwig-ai/ludwig/pull/2550/files#diff-fec65d5ff8d68e96ebf11e5114a25e0ffa187739296f1e89d9949ae04cef0cc1R188-R195

ludwig/utils/data_utils.py
arnavgarg1: Agreed, I think it might be worth taking a larger sample of rows or finding the first non-null row and use that.

geoffreyangus: Good callout– I believe that in the vanilla Ludwig codepath, any calls to this function would only occur after NaNs are handled. That said, it is possible that to_parquet could be called in something like Predibase, so I've added a larger buffer to minimize the likelihood of failure here.

ludwig/models/predictor.py
geoffreyangus: Good callout– didn't realize that the `backend` variable was being passed in. Used that instead. Thanks!

tgaddair (approved):

arnavgarg1 (approved):

tgaddair:
I have some concerns that the new tests are among the slowest in the suite:

```
96.39s call     tests/integration_tests/test_ray.py::test_ray_save_inputs_with_nans[parquet]
96.29s call     tests/integration_tests/test_ray.py::test_ray_save_inputs_with_nans[csv]
```

That said, I am fine to land this as it is potentially a major issue for users of certain Dask versions.

geoffreyangus:
Yup– as a quick fix, I've commented out the `csv` test for now, but left the `parquet` test in place– we'll see if that causes the CI to time out. 

With the introduction of this test, there is duplicated behavior now among the ray tests– maybe I can incorporate the validation being conducted in those tests with the existing tests to eliminate the new slow test and amortize its cost across the existing tests. Will let you know how that goes.

PR #2553, opened 2022-09-26
Ensure bfill/ffill leave no residual NaNs in the dataset during preprocessing

arnavgarg1:
If the dataset has the first few rows or last few rows as NaNs (especially the first row in the case of ffill, or last row in case of bfill), it can result in a dataset where there are NaNs even after we try to handle the missing value strategy. This causes ragged tensors downstream (particularly in the case of image or audio features) where the missing value strategy is `bfill` by default.

This fix adds a conditional check to see if there are still rows with NaNs, and then applies a secondary missing value strategy to fix it. For e.g., if `bfill` doesn't get rid of all of the NaNs, we're going also apply `ffill` since this will guarantee all rows are non-None. The same is true if the primary missing_value_strategy is `ffill` and there are NaNs left over, which will then create a secondary call to `bfill`. 

This is a good measure to ensure that we don't have NaNs anymore, and also prevents issues like https://github.com/ludwig-ai/ludwig/issues/2452 from showing up in the future because of this.

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -100,7 +100,6 @@ from ludwig.utils.data_utils import (
     TSV_FORMATS,
     use_credentials,
 )
-from ludwig.utils.dataframe_utils import is_dask_series
 from ludwig.utils.defaults import default_preprocessing_parameters, default_random_seed
 from ludwig.utils.fs_utils import file_lock, path_exists
 from ludwig.utils.misc_utils import get_from_registry, merge_dict, resolve_pointers
@@ -1154,7 +1153,7 @@ def build_dataset(
 
     for feature_config in feature_configs:
         preprocessing_parameters = feature_name_to_preprocessing_parameters[feature_config[NAME]]
-        handle_missing_values(dataset_cols, feature_config, preprocessing_parameters)
+        handle_missing_values(dataset_cols, feature_config, preprocessing_parameters, backend)
 
     # Happens after missing values are handled to avoid NaN casting issues.
     logger.debug("cast columns")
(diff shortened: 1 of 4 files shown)


justinxzhao (approved):

ludwig/data/preprocessing.py
```diff
@@ -1468,9 +1469,27 @@ def handle_missing_values(dataset_cols, feature, preprocessing_parameters):
...
+        # applied. This causes downstream errors with Dask (https://github.com/ludwig-ai/ludwig/issues/2452)
+        # To get around this issue, apply the primary missing value strategy (say bfill) first, and then follow it
+        # up with the other missing value strategy (ffill) to ensure all NaNs are filled
         dataset_cols[feature[COLUMN]] = dataset_cols[feature[COLUMN]].fillna(
             method=missing_value_strategy,
         )
+
+        # Add conditional check since executing another missing value strategy is expensive
+        requires_secondary_fill = False
+        if is_dask_series(dataset_cols[feature[COLUMN]]):
```
tgaddair, line 1482: We should be using `backend` for this:

```
backend.compute(dataset_cols[feature[COLUMN]].isna().sum())
```

ludwig/utils/dataframe_utils.py
```diff
@@ -17,6 +17,16 @@ def is_dask_backend(backend: Optional["Backend"]) -> bool:  # noqa: F821
     return backend is not None and is_dask_lib(backend.df_engine.df_lib)
 
 
+def is_dask_series(series) -> bool:
```
tgaddair, line 20: Please remove this and use the Backend abstraction.

tgaddair (changes requested):

tests/integration_tests/utils.py
```diff
@@ -760,6 +760,18 @@ def to_fwf(df, fname):
     return dataset_to_use
 
 
+def augment_dataset_with_none(
+    df: pd.DataFrame, first_row_none: bool = False, last_row_none: bool = False, nan_cols: List = []
+) -> pd.DataFrame:
```
justinxzhao, line 765: Let's add a docstring to briefly describe how the columns are used.

justinxzhao (approved):

tgaddair (approved):

PR #2556, opened 2022-09-27
Define custom Ludwig types and replace Dict[str, Any] type hints with them.

justinxzhao:


dantreiman (changes requested):
Great idea, left a few comments

ludwig/types.py
```diff
@@ -0,0 +1,7 @@
+"""Public Ludwig types."""
+from typing import Any, Dict
+
+LudwigFeature = Dict[str, Any]
```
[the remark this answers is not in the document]
justinxzhao: Done.

ludwig/types.py
```diff
@@ -0,0 +1,7 @@
...
+LudwigConfig = Dict[str, Any]
```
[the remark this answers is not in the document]
justinxzhao: Done.

ludwig/typing.py
```diff
@@ -0,0 +1,28 @@
+"""Common typing for Ludwig dictionary parameters."""
+from typing import Any, Dict
+
+# Dictionary of parameters used to configure an input or output feature.
+# https://ludwig.ai/latest/configuration/features/supported_data_types/
+FeatureConfigDict = Dict[str, Any]
```
dantreiman, line 6: Though there is not a standard way to provide a doc-string for a variable, Sphinx, VSCode, and PyCharm support this: (they will show the doc string when you hover over the variable name).  Maybe we should use this convention:
```
FeatureConfigDict = Dict[str, Any]
"""
Dictionary of parameters used to configure an input or output feature.
https://ludwig.ai/latest/configuration/features/supported_data_types/
"""
```

ludwig/typing.py
```diff
@@ -0,0 +1,28 @@
...
+# Dictionary representation of the ModelConfig object.
+# https://ludwig.ai/latest/configuration/
+ModelConfigDict = Dict[str, Any]
```
dantreiman, line 9: 2 newlines between definitions, to visually separate them.

ludwig/typing.py
```diff
@@ -0,0 +1,28 @@
+"""Common typing for Ludwig dictionary parameters."""
```
dantreiman, line 1: I'd like to imagine how this looks as a module in the future.  Once we define these types they will be imported throughout our code and user code as well.  Then, later on, we may replace these with dataclasses or something (which might be defined in their own files).

For that reason, I'd recommend to:
1.  call this module `types` instead of `typing`.      `from ludwig.types import ModelConfigDict`
AND
2. maybe import these into the `__init__.py`.       `from ludwig import ModelConfigDict`
Making these framework-level types like`torch.Tensor`

dantreiman (approved):

ludwig/typing.py
dantreiman, line 6: On our call I recommended using @PublicAPI annotations, however these type aliases are basically global variables (not classes or functions) so annotations won't work here.  Recommend just adding `# Public API` to the module-level doc string instead.

justinxzhao: Great idea! Done.

ludwig/typing.py
justinxzhao: Done.

ludwig/typing.py
justinxzhao: Done.

ludwig/typing.py
justinxzhao: I've gone with the first suggestion and will hold off on the second. There are other `types.py` files in the repo that could be worth considering more before the global unification.

PR #2560, opened 2022-09-27
Only create hyperopt pair plots when there is more than 1 parameter

arnavgarg1:
This ensures that hyperopt pair plots are only created if there is > 1 parameter in the hyperopt parameter search space. Otherwise, in the case of only a single parameter, this creates an empty pair plot.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,7 +57,11 @@ def yaml_filename():
 
 @pytest.fixture(scope="module")
 def hyperopt_results(request):
-    """This function generates hyperopt results."""
+    """This function generates hyperopt results.
+
+    The request argument is used to allow indirect passing of parameters. In this test, it is a boolean that indicates
+    whether to use a single parameter or multiple parameters.
+    """
     input_features = [
         text_feature(name="utterance", encoder={"cell_type": "lstm", "reduce_output": "sum"}),
         category_feature(encoder={"vocab_size": 2}, reduce_input="sum"),
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -77,7 +77,8 @@ def hyperopt_results_single_parameter():
             "type": "variant_generator",
         },
     }
-    hyperopt(config, dataset=rel_path, output_directory="results", experiment_name="hyperopt_test")
+    # Prevent resume from failure since this results in failures in other tests
+    hyperopt(config, dataset=rel_path, output_directory="results", experiment_name="hyperopt_test", resume=False)
     return os.path.join(os.path.abspath("results"), "hyperopt_test")
 
 
@@ -106,7 +107,8 @@ def hyperopt_results_multiple_parameters():
             "type": "variant_generator",
         },
     }
-    hyperopt(config, dataset=rel_path, output_directory="results", experiment_name="hyperopt_test")
+    # Prevent resume from failure since this results in failures in other tests
+    hyperopt(config, dataset=rel_path, output_directory="results", experiment_name="hyperopt_test", resume=False)
     return os.path.join(os.path.abspath("results"), "hyperopt_test")
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,7 +56,7 @@ def yaml_filename():
 
 
 @pytest.fixture(scope="module")
-def hyperopt_results():
+def hyperopt_results(request):
     """This function generates hyperopt results."""
     input_features = [
         text_feature(name="utterance", encoder={"cell_type": "lstm", "reduce_output": "sum"}),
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -56,8 +56,12 @@ def yaml_filename():
...
 
 @pytest.fixture(scope="module")
-def hyperopt_results():
-    """This function generates hyperopt results."""
+def hyperopt_results(request):
+    """This function generates hyperopt results.
+
+    The request argument is used to allow indirect passing of parameters. In this test, it is a boolean that indicates
+    whether to use a single parameter or multiple parameters.
+    """
```
justinxzhao, line 64: While overloading is a relatively smaller code change footprint, it makes this fixture and users of this fixture more convoluted.

I would recommend defining an entirely separate fixture, refactoring the shared components into helper functions.

```python
def get_sample_config():
    """Returns a sample config."""
    input_features = [
        text_feature(name="utterance", encoder={"cell_type": "lstm", "reduce_output": "sum"}),
        category_feature(encoder={"vocab_size": 2}, reduce_input="sum"),
    ]
    output_features = [category_feature(decoder={"vocab_size": 2}, reduce_input="sum")]
    csv_filename = uuid.uuid4().hex[:10].upper() + ".csv"
    rel_path = generate_data(input_features, output_features, csv_filename)
    config = {
        INPUT_FEATURES: input_features,
        OUTPUT_FEATURES: output_features,
        COMBINER: {TYPE: "concat", "num_fc_layers": 2},
        TRAINER: {EPOCHS: 2, "learning_rate": 0.001},
    }
    return config


def get_hyperopt_results(config_with_hyperopt) -> str:
    """Returns path to hyperopt results directory."""
    hyperopt(config, dataset=rel_path, output_directory="results", experiment_name="hyperopt_test")
    return os.path.join(os.path.abspath("results"), "hyperopt_test")


def hyperopt_results():
    config = get_sample_config()
    output_feature_name = config[OUTPUT_FEATURES][0][NAME]
    config[HYPEROPT] = {
        "parameters": {
            "trainer.learning_rate": {
                "space": "loguniform",
                "lower": 0.0001,
                "upper": 0.01,
             }
        }, 
        "goal": "minimize",
        "validation_metrics": "loss",
        "executor": {
            "type": "ray",
            "num_samples": 2,
        },
        "search_alg": {
            "type": "variant_generator",
        },
    }
    return get_hyperopt_results(config)


def hyperopt_results_multiple_parameters():
    config = get_sample_config()
    output_feature_name = config[OUTPUT_FEATURES][0][NAME]
    config[HYPEROPT] = {
        "parameters": {
            "trainer.learning_rate": {
                "space": "loguniform",
                "lower": 0.0001,
                "upper": 0.01,
             f"{output_feature_name}.decoder.output_size": {"space": "choice", "categories": [32, 64, 128, 256]},
             f"{output_feature_name}.decoder.num_fc_layers": {"space": "randint", "lower": 1, "upper": 6},
        }, 
        "goal": "minimize",
        "validation_metrics": "loss",
        "executor": {
            "type": "ray",
            "num_samples": 2,
        },
        "search_alg": {
            "type": "variant_generator",
        },
    }
    return get_hyperopt_results(config)

```

arnavgarg1: @justinxzhao Gotcha, that makes a lot of sense. Will refactor.

Out of curiosity, when (if ever) does it make sense to use parameterized fixtures?

justinxzhao: If output feature name wasn't dynamically determined, this maybe could've been an appropriate usage for a parameterized fixture, for example if you could do something like:

```python
def hyperopt_results(hyperopt_config):
    config = get_sample_config()
    config[HYPEROPT] = hyperopt_config
    return get_hyperopt_results(config)


@pytest.mark.parametrize("hyperopt_results", ["parameters": {
            "trainer.learning_rate": {
                "space": "loguniform",
                "lower": 0.0001,
                "upper": 0.01,
             f"{output_feature_name}.decoder.output_size": {"space": "choice", "categories": [32, 64, 128, 256]},
             f"{output_feature_name}.decoder.num_fc_layers": {"space": "randint", "lower": 1, "upper": 6},
        }, 
        "goal": "minimize",
        "validation_metrics": "loss",
        "executor": {
            "type": "ray",
            "num_samples": 2,
        },
        "search_alg": {
            "type": "variant_generator",
        },], indirect=True)
def test_hyperopt_report_vis_api_no_pairplot(hyperopt_results, tmpdir):
    vis_dir = os.path.join(tmpdir, "visualizations")

    visualize.hyperopt_report(os.path.join(hyperopt_results, HYPEROPT_STATISTICS_FILE_NAME), output_directory=vis_dir)

    figure_cnt = glob.glob(os.path.join(vis_dir, "*"))

    # Only create plot for single parameter and skip pairplot creation
    assert len(figure_cnt) == 1
```

It's a big parameter, so this isn't really that elegant, but at least there's no complicating control flow at play that would require reading through the fixture function to understand what's happening, and the fixture itself is a trivial skeleton.

arnavgarg1: @justinxzhao Nice, that totally makes sense! Thanks for sharing. 

It seems like you have to use the `request` keyword if you want to use parameterized fixtures, which feels kinda strange but it didn't even work for me locally unless I used that keyword as the argument. Otherwise I think this could have been more clearly set. 

In either case, your example makes a tremendous amount of sense and it clearly disambiguates what the role of the argument is, so I will keep that in mind.

Thanks for your suggestion on refactoring the pytest fixtures.

justinxzhao (approved):

PR #2563, opened 2022-09-28
Fix ray_utils: return Resources object instead of dict

jppgks:


justinxzhao (approved):

tgaddair (commented):
Was this causing a bug? If so, can we add a test to catch it?

dantreiman (approved):

tgaddair:
Closing in favor of #2570.

PR #2564, opened 2022-09-28
Explanation API and feature importance for GBM

jppgks:
Introduces abstract `Explainer` class with the following method for explaining model predictions:
```python
def explain(self, **kwargs) -> Tuple[List[Explanation], List[float]]:
    """Explain the model's predictions.

    # Return

    :return: (Tuple[List[Explanation], List[float]]) `(explanations, expected_values)`
        `explanations`: (List[Explanation]) A list of explanations, one for each row in the input data. Each
        explanation contains the feature attributions for each label in the target feature's vocab.

        `expected_values`: (List[float]) of length [output feature cardinality] Expected value for each label in
        the target feature's vocab.
    """
```

`Explanation` is a dataclass for storing explanations for a single row of input data.

Additionally:
- refactors Integrated Gradients as an implementation of the `Explainer` base class 
- adds `GBMExplainer` for explaining GBM models using LightGBM feature importance


ludwig/explain/gbm.py
```diff
@@ -0,0 +1,40 @@
...
+        feat_imp = feat_imp / feat_imp.sum() if feat_imp.sum() > 0 else feat_imp
+
+        expected_values = []
+        for target_idx in range(self.vocab_size):
+            for explanation in self.explanations:
+                # Add the feature attributions to the explanation object for this row.
+                explanation.add(feat_imp)
+
+            # TODO:
+            expected_values.append(0.0)
```
jppgks, line 38: @tgaddair what should the expected values be for GBM?

jppgks: From Travis: value of `0` OK for now, can look into this again depending on visualizations

ludwig/explain/util.py
```diff
@@ -1,9 +1,49 @@
...
+@dataclass
+class LabelExplanation:
+    """Stores the feature attributions for a single label in the target feature's vocab."""
+
+    # The attribution for each input feature.
+    feature_attributions: npt.NDArray[np.float64]
+
+
+@dataclass
+class Explanation:
```
dantreiman, line 21: Explanation probably deserves its own file (explanation.py)

ludwig/explain/base.py
```diff
@@ -0,0 +1,74 @@
...
+
+from ludwig.api import LudwigModel
+from ludwig.constants import BINARY, CATEGORY, TYPE
+from ludwig.explain.util import Explanation, prepare_data
+from ludwig.utils.torch_utils import get_torch_device
+
+DEVICE = get_torch_device()
+
+
+class Explainer(metaclass=ABCMeta):
```
dantreiman, line 14: We could call this file `explainer.py` since its main export is the `Explainer` class

ludwig/explain/base.py
```diff
@@ -0,0 +1,74 @@
...
+        """Constructor for the explainer.
+
+        # Inputs
+
+        :param model: (LudwigModel) The LudwigModel to explain.
+        :param inputs_df: (pd.DataFrame) The input data to explain.
+        :param sample_df: (pd.DataFrame) A sample of the ground truth data.
+        :param target: (str) The name of the target to explain.
+        """
+        model.model.to(DEVICE)
```
dantreiman, line 25: Perhaps this should be a check that `model.model.device == DEVICE`?   The reason I say that is because I would be surprised to see that constructing Explainer(model, ...) would in-place modify model.
If we need to keep `model.to(DEVICE)` here, mention it in the doc string.

ludwig/explain/base.py
```diff
@@ -0,0 +1,74 @@
...
+        For regression (number) this is 1, for binary it is 2, and for category it is the vocab size.
+        """
+        if self.is_category_target:
+            return self.model.training_set_metadata[self.target_feature_name]["vocab_size"]
+        elif self.is_binary_target:
+            return 2
+        return 1
+
+    @abstractmethod
+    def explain(self, **kwargs) -> Tuple[List[Explanation], List[float]]:
```
dantreiman, line 63: AFAICT the kwargs catch-all is not used.  Variable-length args can stealthily consume parameters and hide type errors... lets remove it and use named and typed arguments unless this is needed.

dantreiman (changes requested):
Code looks great, very positive improvement to`ludwig.explain`.  Made a few optional refactoring suggestions

ludwig/explain/base.py
jppgks: Moved device placement from base class to `IntegratedGradientsExplainer.explain`, as it is not used by `GBMExplainer`.

dantreiman (approved):
LGTM!

PR #2565, opened 2022-09-28
fix: Limit frequency array to top_n_classes in F1 viz

hungcs:
f1 currently fails if total # classes exceeds `top_n_classes` (which =100 in this example)

Screen Shot 2022-09-28 at 2 01 52 PM

### DIFF
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -3605,8 +3605,13 @@ def frequency_vs_f1(
             per_class_stats = test_stats[of_name]["per_class_stats"]
             class_names = metadata[of_name]["idx2str"]
 
+            # get np arrays of frequencies, f1s and labels
+            idx2freq = {metadata[of_name]["str2idx"][key]: val for key, val in metadata[of_name]["str2freq"].items()}
+            freq_np = np.array([idx2freq[class_id] for class_id in sorted(idx2freq)], dtype=np.int32)
+
             if k > 0:
                 class_names = class_names[:k]
+                freq_np = freq_np[:k]
 
             f1_scores = []
             labels = []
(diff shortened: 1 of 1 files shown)


arnavgarg1:
I think this change makes sense, great catch @hungcs

arnavgarg1 (approved):

ludwig/visualize.py
```diff
@@ -3618,7 +3618,7 @@ def frequency_vs_f1(
 
             # get np arrays of frequencies, f1s and labels
             idx2freq = {metadata[of_name]["str2idx"][key]: val for key, val in metadata[of_name]["str2freq"].items()}
-            freq_np = np.array([idx2freq[class_id] for class_id in sorted(idx2freq)], dtype=np.int32)
+            freq_np = np.array([idx2freq[class_id] for class_id in sorted(idx2freq)][:k], dtype=np.int32)
```
dantreiman, line 3621: I find it weird that in line 3608 there is:
```
if k > 0:
     class_names = class_names[:k]
```

This implies to me that if k == 0 we should be showing all classes.... like k==0 means don't truncate.  If so, then maybe we should change that to:
```
k = top_n_classes[i]
if k <= 0:
    k = len(class_names)
class_names = class_names[:k]
```

dantreiman (approved):
LGTM, added 2 minor comments.

ludwig/visualize.py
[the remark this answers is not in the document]
arnavgarg1: Ah good catch, I missed this as well @dantreiman

arnavgarg1: @dantreiman Afaik, `top_n_classes` is a list with a single item, which is just the number of top classes. I don't know why it was written this way but I noticed this in many of the other visualization tests, so I think this may not work, @hungcs you may want to double check to be sure

arnavgarg1: So maybe something like:

``` 
if k <= len(class_names):
    class_names = class_names[:k]
```
?

arnavgarg1: But then yeah, I agree that if k == 0, then this doesn't quite work because it won't produce anything.

dantreiman: I'm just assuming that `if k > 0:` was put in there for a good reason :-).

We could also do
```
if k > 0:
    freq_np = freq_np[:k]
```

up to you @hungcs, and feel free to ignore the part where I wrote top_n_classes[i].

PR #2566, opened 2022-09-29
specify seed in RayDataset shuffling

abidwael:
- Training the same model twice on the Ray backend yields different metric value results. This PR fixes that.
- It also modifies the `ludwig-ray` image to get the torch wheel from `https://download.pytorch.org/whl/cpu`

### DIFF
--- a/tests/ludwig/models/test_training_determinism.py
+++ b/tests/ludwig/models/test_training_determinism.py
@@ -3,6 +3,8 @@ import logging
 import os
 
 import pytest
+import numpy as np
+from ludwig.utils.data_utils import flatten_dict
 
 from ludwig.api import LudwigModel
 from ludwig.constants import TRAINER
@@ -32,12 +34,8 @@ def test_training_determinism_ray_backend(csv_filename, tmpdir):
     eval_stats_1, train_stats_1, _, _ = experiment_output_1
     eval_stats_2, train_stats_2, _, _ = experiment_output_2
 
-    assert json.dumps(eval_stats_1, cls=NumpyEncoder, sort_keys=True, indent=4) == json.dumps(
-        eval_stats_2, cls=NumpyEncoder, sort_keys=True, indent=4
-    )
-    assert json.dumps(train_stats_1, cls=NumpyEncoder, sort_keys=True, indent=4) == json.dumps(
-        train_stats_2, cls=NumpyEncoder, sort_keys=True, indent=4
-    )
+    np.testing.assert_equal(eval_stats_1, eval_stats_1)
+    np.testing.assert_equal(train_stats_1, train_stats_1)
 
 
 def test_training_determinism_local_backend(csv_filename, tmpdir):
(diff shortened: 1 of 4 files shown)


tests/ludwig/models/test_training_determinism.py
```diff
@@ -0,0 +1,109 @@
...
+
+@pytest.mark.distributed
+def test_training_determinism_ray_backend(csv_filename, tmpdir):
+    experiment_output_1, experiment_output_2 = train_twice("ray", csv_filename, tmpdir)
+
+    eval_stats_1, train_stats_1, _, _ = experiment_output_1
+    eval_stats_2, train_stats_2, _, _ = experiment_output_2
+
+    assert json.dumps(eval_stats_1, cls=NumpyEncoder, sort_keys=True, indent=4) == json.dumps(
+        eval_stats_2, cls=NumpyEncoder, sort_keys=True, indent=4
```
justinxzhao, line 36: Do we need the extra json.dumps sorting step? I would expect pytest to be able to compare dicts (with differently sorted keys), as is.

abidwael: I resorted to `json.dumps` beacuse direct dict comparison fails in case there are nans.

justinxzhao: Hmm, patching your PR, I'm seeing that 

```
assert eval_stats_1 == eval_stats_2
assert train_stats_1 == train_stats_2
```

Passes.

justinxzhao: Nevermind, looks like it's not a consistent pass.

tests/ludwig/models/test_training_determinism.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+    assert json.dumps(train_stats_1, cls=NumpyEncoder, sort_keys=True, indent=4) == json.dumps(
```
ShreyaR, line 38: Are the training stats won't always be the exact same, even with full determinism? I'd assume that there's some decimal errors, in which case doing an `np.allclose` might be more appropriate instead of an `equals` assertion.

ShreyaR (commented):
Lgtm! Can approve once the comment is addressed.

justinxzhao (commented):
This looks like it should be broken up into separate PRs:
1. Update ludwig-ray docker image with torch CPU wheel.
2. Specify seed in RayDataset shuffling.
3. Customize logged message when early stopping is disabled.

While each change is small, we should get into the habit of making PRs cleanly self-encapsulated. Could you create separate PRs for the other changes?

tests/ludwig/models/test_training_determinism.py
abidwael: Used `numpy.testing.assert_equal` which consider `np.nan` as if they were normal numbers.

tests/ludwig/models/test_training_determinism.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@pytest.mark.distributed
+def test_training_determinism_ray_backend(csv_filename, tmpdir):
+    experiment_output_1, experiment_output_2 = train_twice("ray", csv_filename, tmpdir)
+
+    eval_stats_1, train_stats_1, _, _ = experiment_output_1
+    eval_stats_2, train_stats_2, _, _ = experiment_output_2
+
+    np.testing.assert_equal(eval_stats_1, eval_stats_1)
```
justinxzhao, line 34: Nice!

justinxzhao (approved):

PR #2568, opened 2022-09-29
Model performace in GitHub actions

abidwael:
Adds pytest to run ECD and GBM model benchmarks in GitHub actions with the `ray` backend on the following datasets: `ames_housing`, `mercedes_benz_greener`, `adult_census_income`, `sarcos`. Asserts model performance values. Benchmarking pytests run after the branch has been merged into master.


justinxzhao (commented):
Thanks for getting this work started @abidwael! Left a couple of comments.

.github/workflows/pytest.yml
```diff
@@ -163,6 +164,125 @@ jobs:
...
+        if: runner.os == 'linux'
+        run: |
+          sudo apt-get install -y cmake libsndfile1
+
+      - name: Setup macOS
+        if: runner.os == 'macOS'
+        run: |
+          brew install libuv
+
+      - name: pip cache
```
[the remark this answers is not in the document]
abidwael: Split the benchmark into its separate workflow `.github/workflows/pytest_benchmarks.yml`. 

I did a quick dive on workflows and I'm not sure they could be helpful in this specific case as the dependencies for the full pytest suite and the benchmarking are different (I also cut down more installation steps in the new workflow file). But I think they could help us better organize the `Install dependencies` step of the `pytest` job in `.github/workflows/pytest.yml` instead of having if conditions. I can do that in a follow up PR.

tests/regression_tests/benchmark/process_config.py
```diff
@@ -0,0 +1,10 @@
+def process_config(ludwig_config: dict, experiment_dict: dict) -> dict:
```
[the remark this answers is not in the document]
abidwael: It's a function that is used to modify multiple configs programmatically (https://github.com/ludwig-ai/ludwig/blob/master/ludwig/benchmarking/README.md#programmatically-editing-ludwig-configs)

I removed it, though, as it's not being used here.

tests/regression_tests/benchmark/test_model_performance.py
```diff
@@ -0,0 +1,116 @@
...
+profiler:
+  enable: true
+  use_torch_profiler: false
+  logging_interval: 0.1
+experiments:
+  - dataset_name: {dataset}
+    config_path: {config_path}
+"""
+
+dataset_name_to_metric = {
```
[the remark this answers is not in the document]
abidwael: Dataclasses are always a good idea! Added them as well as the YAMLs.

The tests up until your comment were on the local backend. I switched to running these tests on the Ray backend. However, since time and resource tracking aren't supported in LudwigProfiler on the Ray backend yet (bug [here](https://github.com/ludwig-ai/ludwig/pull/2462#issuecomment-1260337734)), I only added these YAMLs and dataclasses for the model performance metric. The plan is to add time and resource usage later.

tests/regression_tests/benchmark/test_model_performance.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+
+    benchmarking_config_fp = tmpdir + "/benchmarking_config.yaml"
+    with open(benchmarking_config_fp, "w") as f:
+        f.write(benchmarking_config)
+
+    benchmark(benchmarking_config_fp)
+
+    test_statistics_fp = os.path.join(tmpdir, dataset, experiment_name, "experiment_run", "test_statistics.json")
+    test_statistics = load_json(test_statistics_fp)
```
justinxzhao, line 87: Hmm, I feel like I have seen this code somewhere already!

In the spirit of anticipating public APIs, it could be a good idea to define a library for common results-unpacking like this in the benchmark module, i.e. `benchmark.get_test_statistics(benchmarking_config)`, and using that here.

tests/regression_tests/benchmark/test_model_performance.py
```diff
@@ -0,0 +1,111 @@
...
+            self.regression_direction = 1
+        else:
+            raise ValueError(
+                "Regression direction in the expected performance YAML file should be one" "of 'LOWER', 'HIGHER'."
+            )
+
+
+@pytest.mark.benchmark
+@pytest.mark.parametrize("model_type", [MODEL_GBM, MODEL_ECD])
+@pytest.mark.parametrize("dataset", ["ames_housing", "mercedes_benz_greener", "adult_census_income", "sarcos"])
```
justinxzhao, line 67: If we can standardize on a config name like `<dataset>.<descriptors>.yaml`, then we can get the dataset name from the config with `path.split('.')[0]`.

Then, by parameterizing over listdir, users could add new benchmarking configs without touching this file whatsoever.

Here's one idea:

```python
def get_dataset_from_benchmark_config_path(config_path: str) -> str:
    """path/to/config/<dataset>.<descriptors>.yaml -> dataset"""
    return os.path.basename(config_path).split(".")[0]


def assert_no_regression(expected_metric: ExpectedMetric, test_statistic: Union[int, float], dataset: str) -> None:
    """Assert that the test statistic has not regressed from the expected metric."""
    tolerance = (
            expected_metric.percent_change_sensitivity
            * expected_metric.expected_value
        )
    if expected_metric.regression_direction == MetricRegressionDirection.LOWER.value:
        assert test_statistic >= expected_metric.expected_value - tolerance, f"The observed '{expected_metric.metric_name}': {test_statistic} for output feature '{expected_metric.output_feature_name}' was regressively lower than the expected value '{expected_metric.expected_value}' with tolerance '{expected_metric.percent_change_sensitivity}'."
    else:
        assert test_statistic <= expected_metric.expected_value + tolerance, f"The observed '{expected_metric.metric_name}': {test_statistic} for output feature '{expected_metric.output_feature_name}' was regressively higher than the expected value '{expected_metric.expected_value}' with tolerance '{expected_metric.percent_change_sensitivity}'."


@pytest.mark.parametrize("config_path", os.listdir("tests/regression_tests/benchmark/configs"))
def test_performance(config_path, tmpdir):
    if not config_path.endswith(".yaml"):
        # Skip non-YAML files.
        return True

    dataset = get_dataset_from_benchmark_config_path(config_path)
    benchmarking_config = {
  "hyperopt": False, 
  "experiment_name": "regression_test",
  "experiments": [
    {
      "dataset_name": dataset,
      "config_path": config_path,
    }
  ], 
  "export": {
    "export_base_path": tmpdir,
    "export_artifacts": True,
  }, 
  "profiler": {
    "enable": false, 
    "use_torch_profiler": false, 
    "logging_interval": 0.1,
  }
}

    benchmark.benchmark(benchmarking_config_dict=benchmarking_config)

    test_statistics = benchmark.get_test_statistics(benchmarking_config)

    expected_metrics_path = os.path.join("tests/regression_tests/benchmark/expected_metrics", os.path.basename(config_path))
    if not os.path.exists(expected_metrics_path):
        raise FileNotFoundError(f"No corresponding expected metrics found for benchmarking config '{config_path}'. Please add a new metrics yaml file '{expected_metrics_path}'. Suggested content:\n{ ...<would be helpful to print out a suggested expected metrics dict that can be easily copy-pasted into a new file>... }")
    
    expected_metrics: Dict = load_yaml(expected_metrics_path)
    expected_metrics: List[ExpectedMetric] = [
        ExpectedMetric.from_dict(expected_metric) for expected_metric in expected_metrics["metrics"]
    ]

    for expected_metric in expected_metrics:
        assert_no_regression(expected_metric, test_statistics[expected_metric.output_feature_name][expected_metric.metric_name])
```

In this block, I've also suggested more detailed logs when the file does not exist, as well as when we detect a regression, which seems like a nice convenience for users who want to add more configs, and since we're exporting benchmark results to a tmpdir (which is the right thing to do), and which will be hard to inspect after the test cleans up.

tests/regression_tests/benchmark/test_model_performance.py
```diff
@@ -0,0 +1,111 @@
...
+        export_base_path=tmpdir,
+        dataset=dataset,
+        config_path=config_path,
+    )
+
+    benchmarking_config_fp = tmpdir + "/benchmarking_config.yaml"
+    with open(benchmarking_config_fp, "w") as f:
+        f.write(benchmarking_config)
+
+    benchmark(benchmarking_config_fp)
```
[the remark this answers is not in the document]
abidwael: PR to address this comment: https://github.com/ludwig-ai/ludwig/pull/2626

tests/regression_tests/benchmark/test_model_performance.py
abidwael: PR to address this comment: https://github.com/ludwig-ai/ludwig/pull/2626

tests/regression_tests/benchmark/test_model_performance.py
```diff
@@ -0,0 +1,111 @@
...
+  enable: false
+  use_torch_profiler: false
+  logging_interval: 0.1
+experiments:
+  - dataset_name: {dataset}
+    config_path: {config_path}
+"""
+
+
+class MetricRegressionDirection(Enum):
```
jppgks, line 29: Can implement this using existing logic in Ludwig metrics: https://github.com/ludwig-ai/ludwig/blob/f000e73ea1d3a28d32575c066437a26db10607c7/ludwig/modules/metric_modules.py#L120

tests/regression_tests/benchmark/expected_metric.py
```diff
@@ -0,0 +1,41 @@
...
+    metric_name: str
+
+    # Expected metric value.
+    expected_value: Union[int, float]
+
+    # Which direction is considered a regression.
+    regression_direction: int
+
+    # The percentage change that would trigger a notification/failure.
+    percent_change_sensitivity: float
```
jppgks, line 31: Let's use the term `tolerance` here, to be consistent with other usages of this concept

tests/regression_tests/benchmark/expected_metric.py
```diff
@@ -0,0 +1,41 @@
...
+    output_feature_name: str
+
+    # Metric name.
+    metric_name: str
+
+    # Expected metric value.
+    expected_value: Union[int, float]
+
+    # Which direction is considered a regression.
+    regression_direction: int
```
jppgks, line 28: Let's infer this using the Ludwig Metric

jppgks (changes requested):

tests/regression_tests/benchmark/test_model_performance.py
abidwael: Using `abs(expected - actual) < tolerance` instead of minimize/maximize.

tests/regression_tests/benchmark/expected_metric.py
abidwael: changed!

tests/regression_tests/benchmark/expected_metric.py
abidwael: removed this

tests/regression_tests/benchmark/test_model_performance.py
abidwael: Made suggested changes

jppgks (approved):

tests/regression_tests/benchmark/test_model_performance.py
```diff
@@ -0,0 +1,62 @@
...
+        raise FileNotFoundError(
+            """No corresponding expected metrics found for benchmarking config '{config_path}'.
+            Please add a new metrics YAML file '{expected_test_statistics_fp}'. Suggested content:
+
+            metrics:
+              - output_feature_name: <YOUR_OUTPUT_FEATURE e.g. SalePrice>
+                metric_name: <YOUR METRIC NAME e.g. accuracy>
+                expected_value: <A FLOAT VALUE>
+                tolerance_percent: 0.15"""
+        )
```
justinxzhao, line 40: Is there a way to automatically derive a reasonable suggested `ExpectedMetric` from a `BenchmarkingArtifact`?

Printing this out in a well-formatted way, alongside the "expected metric file is missing" error, would make it super clear and convenient to add a new expected metric file. WDYT?

abidwael: I wanted to check for the existence of this file before the training happens since the test will fail anyway if the expected metric YAML is not found.

The `BenchmarkingResult` object has a lot of attributes and we can definitely generate a message automatically, but this will require us to do the training first, then extract metrics from `test_statistics`. I think the current example should illustrates the YAML without giving specific values.

justinxzhao: Sounds good.

I anticipate that it will be useful to provide a separate utility (or extend an existing one) to run benchmarking config for a new benchmarking config and which prints out an `ExpectedMetric` yaml file that can be feasibly copy-pasted wholesale into a new file in the expected metrics dir.

justinxzhao (approved):

PR #2574, opened 2022-09-30
GBM bugfix: matching predictions LightGBM, hummingbird

jppgks:
This PR:

1. Fixes a bug where category variable with 2 classes gave wrong predictions
2. Asserts predictions from LightGBM and converted Hummingbird model match
3. Extends test coverage (incl. for the above cases)
4. Uses cleaner sklearn API for LightGBM

There are quite some changes, the PR is best reviewed commit-by-commit

### DIFF
--- a/ludwig/models/gbm.py
+++ b/ludwig/models/gbm.py
@@ -7,8 +7,9 @@ import numpy as np
 import torch
 import torchmetrics
 from hummingbird.ml import convert
+from hummingbird.ml.operator_converters import constants as hb_constants
 
-from ludwig.constants import BINARY, CATEGORY, LOGITS, MODEL_GBM, NAME, NUMBER
+from ludwig.constants import BINARY, CATEGORY, LOGITS, MODEL_GBM, NAME, NUMBER, TYPE
 from ludwig.features.base_feature import OutputFeature
 from ludwig.globals import MODEL_WEIGHTS_FILE_NAME
 from ludwig.models.base import BaseModel
@@ -28,6 +29,12 @@ class GBM(BaseModel):
         random_seed: int = None,
         **_kwargs,
     ):
+        if len(output_features) > 1:
+            raise ValueError("Only single task currently supported")
+        feat_types = {f[TYPE] for f in output_features + input_features}
+        if len(feat_types - {NUMBER, CATEGORY, BINARY}) != 0:
+            raise ValueError("Model type GBM only supports numerical, categorical, or binary features")
+
         super().__init__(random_seed=random_seed)
 
         self._input_features_def = copy.deepcopy(input_features)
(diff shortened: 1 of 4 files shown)


ludwig/models/gbm.py
```diff
@@ -28,6 +30,12 @@ def __init__(
         random_seed: int = None,
         **_kwargs,
     ):
+        if len(output_features) > 1:
```
justinxzhao, line 33: Nice. Thanks for adding these asserts.

tests/integration_tests/test_gbm.py
```diff
@@ -38,21 +35,39 @@ def ray_backend():
...
-    output_features = [text_feature()]
-
+def _train_and_predict_gbm(input_features, output_features, tmpdir, backend_config):
     csv_filename = os.path.join(tmpdir, "training.csv")
     dataset_filename = generate_data(input_features, output_features, csv_filename, num_examples=100)
 
-    config = {MODEL_TYPE: "gbm", "input_features": input_features, "output_features": output_features}
+    config = {
+        MODEL_TYPE: "gbm",
+        "input_features": input_features,
```
justinxzhao, line 44: nit: These are also available constants.

justinxzhao (approved):

dantreiman (approved):

ludwig/models/gbm.py
```diff
@@ -28,6 +30,12 @@ def __init__(
...
+            raise ValueError("Only single task currently supported")
```
tgaddair, line 34: For these kinds of errors, I always find it's helpful to include the failing state so the user knows what is wrong. Like:

```
Only single task currently supported, found: {feature_name_1: NUMBER, feature_name_2: CATEGORY}
```

And:

```
Model type GBM only supports number, category, or binary features, found: {feature_name: TEXT}
```
```

File: ludwig/trainers/trainer_lightgbm.py [hunk 4]
```diff
@@ -719,17 +722,17 @@ def set_steps_to_1_or_quit(self, signum, frame):
     def _construct_lgb_params(self) -> Tuple[dict, dict]:
         output_params = {}
         feature = next(iter(self.model.output_features.values()))
-        if feature.type() == CATEGORY:
+        if feature.type() == BINARY or (hasattr(feature, "num_classes") and feature.num_classes == 2):
```
tgaddair, line 725: Why do we need to check the num_classes stuff here when the feature is binary?

tgaddair (approved):

ludwig/trainers/trainer_lightgbm.py
jppgks: a user can specify a category variable with only two classes, this is to catch that case and explicitly use the binary objective

PR #2578, opened 2022-10-03
Changes learning_curves to use "step" or "epoch" as x-axis label.

dantreiman:
- Adds frequency and units to training stats
- Adds dataclasses: TrainingResults, TrainingStats, PreprocessedDataset
- training_statistics.json has evaluation frequency added, so visualize learning curves from saved model uses the right units in x axis label

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -136,6 +136,7 @@ class TrainingStats:
     test: Dict[str, Any]
     evaluation_frequency: EvaluationFrequency = dataclasses.field(default_factory=EvaluationFrequency)
 
+    # TODO(daniel): deprecate multiple return value unpacking and dictionary-style element access
     def __iter__(self):
         return iter((self.training, self.test, self.validation))
 
@@ -158,6 +159,7 @@ class PreprocessedDataset:
     test_set: Dataset
     training_set_metadata: Dict[str, Any]
 
+    # TODO(daniel): deprecate multiple return value unpacking and indexed access
     def __iter__(self):
         return iter((self.training_set, self.validation_set, self.test_set, self.training_set_metadata))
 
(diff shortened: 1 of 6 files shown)


ludwig/api.py
```diff
@@ -104,6 +106,86 @@
...
+    def __contains__(self, key):
+        return (
+            (key == TRAINING and self.training)
+            or (key == VALIDATION and self.validation)
+            or (key == TEST and self.test)
+        )
+
+    def __getitem__(self, key):
+        # Supports dict-style [] element access for compatibility.
+        return {TRAINING: self.training, VALIDATION: self.validation, TEST: self.test}[key]
```
justinxzhao, line 151: Should we add a Todo here to get rid of these auxiliary/non-standard accessor methods?

justinxzhao (approved):

justinxzhao (approved):

ludwig/api.py
```diff
@@ -104,6 +106,88 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
```
connor-mccorm, line 109: Should we maybe move these dataclasses into an auxilary file and import them to maintain the api file flow? I feel like these data classes could be pretty easily compartmentalized into a dedicated file which may help with the mental model associated with the api file. WDYT?

connor-mccorm (commented):
This is good stuff, had one comment on whether we should be putting the new data classes in the api file.

ludwig/api.py
dantreiman: I put them here because they are used as data types in method of api.py, and I couldn't think of a better idea.

On that note, Arnav and I have discussed further refactoring of api.py this sprint, perhaps defining LudwigModel in `ludwig_model.py` and pulling it into api.py.  So maybe we can separate that out into a follow-up PR.

PR #2579, opened 2022-10-03
feat: adds `max_batch_size` to auto batch size functionality

geoffreyangus:
This PR adds the `max_batch_size` parameter, which limits the `Trainer.tune_batch_size` to ensure that the batch size chosen is less than some user-defined value. This is useful for manually preventing known OOM errors.


ludwig/schema/metadata/trainer_metadata.py
```diff
@@ -24,6 +24,22 @@
                                      expected_impact=ExpectedImpact.HIGH,
                                      literature_references=None,
                                      internal_only=False),
+      'auto_batch_size_ceiling': ParameterMetadata(ui_display_name='Auto Batch Size Ceiling',
```
tgaddair, line 27: Nice! Maybe call it `max_batch_size` and have it also apply to `increase_batch_size_on_plateau`? Then we can have the auto batch size process also set this and then remove `increase_batch_size_on_plateau_max`. What do you think?

geoffreyangus: Sounds reasonable! Will do this.

geoffreyangus: Following up on this: if we unify `auto_batch_size_ceiling` with `increase_batch_size_on_plateau_max`, what should the default value be? Right now the default for `increase_batch_size_on_plateau_max` is 512, which I don't believe we want as the default for `auto_batch_size_ceiling`. Tentatively, we set the default for the new parameter `max_batch_size` to `None`, and we simply expect the user to pass this value in if they are interested in using the `increase_batch_size_on_plateau` functionality. WDYT? cc @dantreiman

ludwig/trainers/trainer.py
```diff
@@ -480,7 +480,15 @@ def tune_batch_size(
 
             def _is_valid_batch_size(batch_size):
                 # make sure that batch size is valid (e.g. less than size of ds)
-                return batch_size < len(training_set)
+                is_smaller_than_training_set = batch_size < len(training_set)
+                is_under_max_batch_size = self.max_batch_size is None or batch_size < self.max_batch_size
```
ShreyaR, line 484: nit: Looks like `batch_size == max_batch_size` is an invalid value, which seems counter-intuitive. I'd assume that the valid values of `batch_size` would be `<= max_batch_size`

ShreyaR (approved):

ludwig/trainers/trainer.py
geoffreyangus: Good catch. Part of the rationale of this in the first place was that, on the Predibase side, `max_batch_size` was set to `prev_rendered_config['trainer']['batch_size']`. This meant that you needed a strict `<` to limit `tune_batch_size`'s search to values strictly less than the previous attempt's batch size.

I made the change here, and ensured that `max_batch_size` was manually set to 1/2 on the Predibase side.

ludwig/schema/metadata/trainer_metadata.py
```diff
@@ -361,18 +374,6 @@
                                                          expected_impact=ExpectedImpact.MEDIUM,
                                                          literature_references=None,
                                                          internal_only=False),
-     'increase_batch_size_on_plateau_max': ParameterMetadata(ui_display_name='Batch Size Increase On Plateau: Cap',
```
tgaddair, line 364: @geoffreyangus we should also add a backwards compatibility hook as we do for other removed params so we rename `increase_batch_size_on_plateau_max` to `max_batch_size` when we detect it in an old config.

geoffreyangus: ah okay, I was wondering about this. I'll take a look there. Thanks!

geoffreyangus: Done.

ludwig/schema/trainer.py
```diff
@@ -158,6 +159,16 @@ class ECDTrainerConfig(BaseTrainerConfig):
         ],
     )
 
+    max_batch_size: Optional[int] = schema_utils.PositiveInteger(
+        default=sys.maxsize,
```
dantreiman, line 163: `sys.maxsize` may vary depending on the architecture its running on, IMO it would be better to make this a large constant.   Or it could default to None, since allow_none=True.

ludwig/trainers/trainer.py
```diff
@@ -1317,6 +1325,9 @@ def increase_batch_size(
         increase_batch_size_eval_split: str = TRAINING,
     ):
         """Uses the progress tracker to determine if the batch size should be increased."""
+        if increase_batch_size_on_plateau_max is None:
+            increase_batch_size_on_plateau_max = float("inf")  # If None, no hard limit on batch size.
```
dantreiman, line 1329: I feel weird about assigning this to a float when its typed as an int, though it works fine maybe we could use some really big integer constant instead?  Not super important

dantreiman (approved):
LGTM, some minor comments, looks good!

ludwig/trainers/trainer.py
geoffreyangus: This makes sense. To address both of the above comments, I think it could make sense to have some constant `MAX_POSSIBLE_BATCH_SIZE` that we use in both locations. I'll add that in for now– let me know what you think in the meantime. Thanks!

ludwig/schema/trainer.py
geoffreyangus: See below comment.

ludwig/constants.py
```diff
@@ -1,5 +1,5 @@
 #! /usr/bin/env python
-# Copyright (c) 2019 Uber Technologies, Inc.
+# Copyright (c) 2019 Uber Technologies, BATCHS.
```
justinxzhao, line 2: nit: bad replacement?

justinxzhao (approved):

ludwig/constants.py
geoffreyangus: yup fixed in latest push, thanks!

ludwig/utils/backward_compatibility.py
```diff
@@ -598,15 +598,27 @@ def upgrade_missing_value_strategy(config: Dict[str, Any]) -> Dict[str, Any]:
     return config
 
 
+@register_config_transformation("0.6", ["trainer"])
+def _upgrade_max_batch_size(trainer: Dict[str, Any]) -> Dict[str, Any]:
+    if "increase_batch_size_on_plateau_max" in trainer:
+        trainer["max_batch_size"] = trainer.pop("increase_batch_size_on_plateau_max")
```
tgaddair, line 604: We should check that `trainer["max_batch_size"]` is not already set. If so, we should discard `increase_batch_size_on_plateau_max` and log a warning.

tgaddair (commented):
LGTM, only one comment to address before landing.

tgaddair (approved):

PR #2589, opened 2022-10-04
Log warning for failed ray imports

arnavgarg1:
I started running into some bespoke test failures and figured out it was because the version of Ray I was using in my container was different from the version now officially supported by Ludwig (Ray >= 1.13). 

I realized that we don't actually log warnings when an import to ray fails and we also set this to None, which often causes downstream issues for tests in particular when we do a `ray.init()` call using test fixtures.

This PR logs warnings to make sure users running their tests are made of ray import failures.

### DIFF
--- a/ludwig/automl/automl.py
+++ b/ludwig/automl/automl.py
@@ -55,7 +55,7 @@ try:
 
     _ray_113 = parse_version(ray.__version__) >= parse_version("1.13.0")
 except ImportError:
-    raise ImportError(" ray is not installed. In order to use auto_train please run pip install ludwig[ray]")
+    raise ImportError("ray is not installed. In order to use auto_train please run pip install ludwig[ray]")
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


tgaddair (changes requested):
Not sure we want these warnings. The issue here is that these files get imported regardless of whether the user has ray installed, selected the ray backend, or anything else. So imagine a user who runs `pip install ludwig` and then just wants to train something. With these changes, they will see a warning even though they're not doing anything wrong. So we shouldn't confuse them by giving them warnings when there's no issue.

arnavgarg1:
@tgaddair I see, I didn't realize that.

I think a better fix then might be to catch `ray.init()` calls within Ludwig to ensure that ray is not None when making that call and provide a more informative logging message there if it is. How does that sound? Otherwise it throws an error with something like `NoneType object has no method init`

tgaddair:
@arnavgarg1 definitely we should be raising appropriate errors if the user is trying to do something with ray but ray isn't installed, etc. Can you provide a repro for the behavior you're describing? I'm not sure I know the specific issue being addressed.

arnavgarg1:
@tgaddair Can't seem to repro the error I was running into anymore. It was something along the lines of me trying to do something with Ray but the Ray import failing (Ray didn't exist), so it set `ray = None` silently. Then, when I tried to run one of the tests, it just said `NoneType object has no attribute init()`. I was just trying to make sure the error would make sense/have more context in case this ever came up again for anyone else. 

Will just close this issue for now.

PR #2596, opened 2022-10-04
Add API Annotations to Ludwig

arnavgarg1:
This PR introduces `api_annotations.py` that can be used to delineate the boundary of Ludwig APIs that are public, for developers, and those that are deprecated.

Co-authored-by: Daniel Treiman <daniel@predibase.com> (@dantreiman)

### DIFF
--- a/ludwig/api_annotations.py
+++ b/ludwig/api_annotations.py
@@ -1,0 +1,185 @@
+#! /usr/bin/env python
+#
+# Code in api_annotations.py adapted from https://github.com/ray-project/ray (Apache-2.0 License)
+#
+# ==============================================================================
+
+
+from typing import Optional
+
+
+def PublicAPI(*args, **kwargs):
+    """Annotation for documenting public APIs. Public APIs are classes and methods exposed to end users of Ludwig.
+
+    If 'stability="stable"', the APIs will remain backwards compatible across minor Ludwig releases
+    (e.g., Ludwig 0.6 -> Ludwig 0.5).
+
+    If 'stability="experimental"', the APIs can be used by advanced users who are tolerant to and expect
+    breaking changes. This will likely be seen in the case of incremental new feature development.
+
+    Args:
+        stability: One of {"stable", "experimental"}.
+        message: A message to help users understand the reason for the deprecation, and provide a migration path.
+    Examples:
+        >>> from api_annotations import PublicAPI
+        >>> @PublicAPI
+        ... def func1(x):
+        ...     return x
+        >>> @PublicAPI(stability="beta")
+        ... def func2(y):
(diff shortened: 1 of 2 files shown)


ludwig/api_annotations.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def inner(obj):
+        _append_doc(obj, message=message, directive="warning")
+        _mark_annotated(obj)
+        return obj
+
+    return inner
+
+
+def _append_doc(obj, *, message: str, directive: Optional[str] = None) -> str:
```
dantreiman, line 119: Also I'm not sure why there is a * here in the arg list, I don't see anywhere we pass in more arguments.  how about
```python
def _append_doc(obj, message: str, directive: Optional[str] = None) -> str:
```

dantreiman (changes requested):

ludwig/api_annotations.py
```diff
@@ -0,0 +1,183 @@
...
+
+    return inner
+
+
+def _append_doc(obj, message: str, directive: Optional[str] = None) -> str:
+    """
+    Args:
+        message: An additional message to append to the end of docstring for a class
+                 or method that uses one of the API annotations
+        directive:
```
arnavgarg1, line 115: @dantreiman Do you know what the `directive` argument does exactly? I can see that you can pass in things like `warning` or `exception`, but just want to make sure I understand this right.

dantreiman: AFAICT it just prepends it on a line of the docstring and indents the message.


**directive is not None**
```
directive:  
    message
```

**directive is None**
```
message
```

dantreiman (approved):

ludwig/api_annotations.py
arnavgarg1: Yup that's what I could tell as well, sounds good!

ludwig/api_annotations.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Args:
+        stability: One of {"stable", "experimental"}
+
+    Examples:
+        >>> from api_annotations import PublicAPI
+        >>> @PublicAPI
+        ... def func1(x):
+        ...     return x
+        >>> @PublicAPI(stability="beta")
```
jppgks, line 21: we define the choices above as `{"stable", "experimental"}`, here we mention `"beta"` too

jppgks (approved):

ludwig/api_annotations.py
arnavgarg1: Great catch, just fixed!

justinxzhao (approved):

PR #2597, opened 2022-10-04
Shutdown ray if initialized in _ray_start fixture

ShreyaR:
This PR adds a check to shutdown ray if it was previously initialized in a separate `_ray_start` call. Occasionally, the teardown code is not executed which causes future `ray.init` cause an error. This PR shuts down ray before calling `ray.init` to avoid that error.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -143,6 +143,9 @@ def _ray_start(request, **kwargs):
         yield None
         return
 
+    if ray.is_initialized():
+        ray.shutdown()
+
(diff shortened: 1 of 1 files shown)


arnavgarg1 (approved):
Should we leave an explanation/comment on why we're doing this (ray test fixture doesn't run the teardown step incase of a test failure)?

ShreyaR:
Closing -- looks like this doesn't fix the core issue and would cause downstream issues when the fixture scope is being exited.

PR #2598, opened 2022-10-05
Fixed issue when uploading output directory artifacts to remote filesystems

tgaddair:
Also adds tests to ensure that s3 works as a remote storage layer.


justinxzhao (approved):
LGTM

PR #2599, opened 2022-10-05
Speed up horovod hyperopt tests and solve OOMs

arnavgarg1:
This PR takes a stab at solving a variety of problems that show up with hyperopt Horovod tests:

1. We're trying to train some relatively large trial runs (some of which end up having 6 fully connected layers in the combiner) which push the overall memory per trial, so we can strip this down pretty easily by reducing the size/scope of some of these parameters (added to this PR).

2. When running grid search with `num_samples=2`, we're essentially running 6 trials on a 7 CPU cluster. This by itself is not a problem, except that the trials compete for the last remaining CPU for dataset-related tasks. This causes the scheduler to constantly complain about resource scheduling requests until the single extra CPU frees up. This causes three problems:
    - Each trial takes much longer to finish running
    - 6/6 trials are running at the same time, using > 95% of the total memory on the cluster/node when all 6 are running in parallel, which causes warnings with potentials for OOMs.
    - The trials end up blocking each other very much in the style of thread locks where the trials are waiting for the "lock" to open up so that they can lock it.

In this PR, I've stripped down the number of parameters, reduced the grid search space, and reduced the number of trials to a reasonable amount. This helps speed up these tests without losing the essence of the tests themselves.

I've tested this locally and I see a 3x speed up (450 seconds to about 150 seconds) with both of these changes in parallel, peak memory usage of about 70%, and no resource request scheduling issues/failed requests.


tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -71,8 +66,8 @@ def mock_storage_client(path):
             "lower": 0.001,
             "upper": 0.1,
         },
-        "combiner.num_fc_layers": {"space": "randint", "lower": 2, "upper": 6},
-        "combiner.num_steps": {"space": "grid_search", "values": [3, 4, 5]},
```
arnavgarg1, line 75: num_steps is only for `tabnet` combiners, but we use concat in all of our test so this wasn't actually doing anything

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -32,10 +32,8 @@
 try:
     import ray
 
-    # Ray nightly version is always set to 3.0.0.dev0
-    _ray_nightly = version.parse(ray.__version__) >= version.parse("3.0.0.dev0")
-    _ray_114 = version.parse(ray.__version__) >= version.parse("1.14")
-    if _ray_114:
+    _ray_200 = version.parse(ray.__version__) >= version.parse("1.13")
```
dantreiman, line 35: Why 1.13 as opposed to >= 2.0?

dantreiman (approved):
Changes to search spaces are sensible.  Had a question about ray version flags, otherwise LGTM.

tests/integration_tests/test_hyperopt_ray_horovod.py
arnavgarg1: @dantreiman Ah yeah good catch, that should've been > 1.13. Will update

justinxzhao (approved):

PR #2601, opened 2022-10-05
Extend test coverage for non-conventional booleans

jppgks:
Tests for non-conventional binary feature that 

1. true/false labels are inferred correctly when `fallback_true_label` is provided
2. a warning is logged in case no `fallback_true_label` is provided
3. AutoML treats them as `CATEGORY`

### DIFF
--- a/tests/integration_tests/test_preprocessing.py
+++ b/tests/integration_tests/test_preprocessing.py
@@ -1,5 +1,7 @@
+import logging
 import os
 import random
+import re
 import string
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tests/integration_tests/test_preprocessing.py
```diff
@@ -398,3 +400,94 @@ def test_in_memory_dataset_size(backend, tmpdir, ray_cluster_2cpu):
...
+
+    # Check that a warning is logged.
+    warning_text = re.compile(
+        f"Binary feature {bin_feature[NAME]} has at least 1 unconventional boolean value:"
+        r" Cannot automatically map value '(>50K|<=50K)'"
+        " to a boolean and no `preprocessing.fallback_true_label` specified."
+        " We will now interpret <=50K as 1 and the other values as 0. If this is incorrect, please use the category"
+        " feature type or manually specify the true value with `preprocessing.fallback_true_label`."
+    )
+    assert re.search(warning_text, caplog.text) is not None
```
justinxzhao, line 493: I would recommend against checking on such a specific regex. If at any point someone changes the text of the warning, then they have to also update this test.

Maybe check against "*unconventional boolean value"

justinxzhao (approved):

PR #2606, opened 2022-10-06
Fix assertions in training_determinism tests

arnavgarg1:
These tests should be comparing the values from both of the runs, rather than to themselves.


justinxzhao (approved):

abidwael (approved):
Sorry about the typo. Thanks for the fix!

PR #2607, opened 2022-10-07
Ensure no ghost ray instances are running in tests

arnavgarg1:
This PR ensures that tests that use a Ray backend and initialize Ray always call `ray.shutdown()` so that there are no ghost ray instances running in subsequent tests. This causes a problem for downstream tests that want to either initialize ray manually or through a test fixture, because they try to re-initialize ray. 

The fix is just as simple as adding this text fixture to tests, so that Ray isn't initialized by the test (only by the fixture) and it is calls the teardown function automatically at the end of each module

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,7 +56,8 @@ def yaml_filename():
 
 
 @pytest.fixture(scope="module")
-def hyperopt_results_single_parameter():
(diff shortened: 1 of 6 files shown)


tgaddair (approved):
Great work!

PR #2614, opened 2022-10-07
Unpin Ray nightly in CI

tgaddair:


tgaddair:
Landing this now so we can fix nightly tests with newest ray version.

PR #2615, opened 2022-10-10
Add splitter that deterministically splits on an ID column

tgaddair:
This is useful for deterministically splitting on a unique ID. Even when additional rows are added to the dataset in the future, each ID will retain its original split assignment.

This approach does not guarantee that the split proportions will be assigned exactly, but the larger the dataset, the more closely the assignment should match the given proportions.

This approach can be used on a column with duplicates, but it will further skew the assignments of rows to splits. However, this may be useful if you need to ensure that the particular ID only appears in a single split (for example, a user ID with multiple rows, where we want to ensure the user isn't in both train and test sets).

The CRC32 approach was from: https://towardsdatascience.com/improve-the-train-test-split-with-the-hashing-function-f38f32b721fb

### DIFF
--- a/tests/ludwig/data/test_split.py
+++ b/tests/ludwig/data/test_split.py
@@ -292,7 +292,7 @@ def test_hash_split(df_engine, ray_cluster_2cpu):
     assert len(splits) == 3
     if isinstance(df_engine, DaskEngine):
         splits = [split.compute() for split in splits]
-        
+
     # IDs should not overlap between splits
     assert all([set(split1["id"]).isdisjoint(set(split2["id"])) for split1, split2 in combinations(splits, 2)])
 
--- a/tests/ludwig/data/test_split.py
+++ b/tests/ludwig/data/test_split.py
@@ -288,18 +288,13 @@ def test_hash_split(df_engine, ray_cluster_2cpu):
     backend = Mock()
     backend.df_engine = df_engine
     splits = splitter.split(df, backend)
-
     assert len(splits) == 3
-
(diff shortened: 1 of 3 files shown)


tests/ludwig/data/test_split.py
```diff
@@ -262,3 +262,59 @@ def random_date(*args, **kwargs):
...
+
+    backend = Mock()
+    backend.df_engine = df_engine
+    splits = splitter.split(df, backend)
+
+    assert len(splits) == 3
+
+    for split, p in zip(splits, probabilities):
+        if isinstance(df_engine, DaskEngine):
+            split = split.compute()
```
jppgks, line 296:
```suggestion
    if isinstance(df_engine, DaskEngine):
        splits = [split.compute() for split in splits]
    for split, p in zip(splits, probabilities):
```

tests/ludwig/data/test_split.py
```diff
@@ -262,3 +262,59 @@ def random_date(*args, **kwargs):
...
+
+        # Should be approximately the same size as the desired proportion
+        assert nrows * p - 5 <= len(split["id"]) <= nrows * p + 5
+
```
jppgks, line 300: Check IDs are disjoint across splits

Needs `from itertools import combinations`
```suggestion

    # IDs should not overlap between splits
    assert all([set(split1["id"]).isdisjoint(set(split2["id"])) for split1, split2 in combinations(splits, 2)])

```

jppgks (approved):

PR #2616, opened 2022-10-10
fix(explain): missing columns for fixed split

jppgks:
Preprocessing config with fixed splitting without explicit column was breaking explanations. Fixed by keeping all columns required by splitter


connor-mccorm (approved):
Good fix. LGTM

PR #2617, opened 2022-10-10
Fixed hyperopt trial syncing to remote filesystems for Ray 2.0

tgaddair:
This also revealed that there are issues with remote syncing for Ray 1.13, so we will only be supporting this feature with Ray 2.0 and above.

This PR uses the `RemoteSyncer` from #2386. The changes to support injecting credentials will come in a follow-up PR.

### DIFF
--- a/tests/integration_tests/test_hyperopt.py
+++ b/tests/integration_tests/test_hyperopt.py
@@ -15,6 +15,7 @@
 import contextlib
 import json
 import os.path
+import uuid
 from typing import Any, Dict, Optional, Tuple, Union
 
 import pytest
@@ -39,6 +40,7 @@ from ludwig.constants import (
 from ludwig.globals import HYPEROPT_STATISTICS_FILE_NAME
 from ludwig.hyperopt.results import HyperoptResults
 from ludwig.hyperopt.run import hyperopt, update_hyperopt_params_with_defaults
+from ludwig.utils import fs_utils
 from ludwig.utils.data_utils import load_json
 from ludwig.utils.defaults import merge_with_defaults
 from tests.integration_tests.utils import category_feature, generate_data, private_param, remote_tmpdir, text_feature
@@ -55,7 +57,7 @@ except ImportError:
     _ray113 = None
 
 
-RANDOM_SEARCH_SIZE = 4
+RANDOM_SEARCH_SIZE = 2
 
 HYPEROPT_CONFIG = {
     "parameters": {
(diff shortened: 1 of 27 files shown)


tests/integration_tests/test_hyperopt.py
```diff
@@ -370,14 +353,19 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
         "goal": "minimize",
         "output_feature": output_feature_name,
         "validation_metrics": "loss",
-        "executor": {TYPE: "ray", "num_samples": 1 if search_space == "grid" else RANDOM_SEARCH_SIZE},
+        "executor": {
+            TYPE: "ray",
+            "num_samples": 1 if search_space == "grid" else RANDOM_SEARCH_SIZE,
+            "max_concurrent_trials": 1,
```
arnavgarg1, line 359: @tgaddair Is there a reason to set max_concurrent_trials to 1?

tgaddair: Yeah, when you run multiple trials on these runners / locally it often ends up making them compete for resources, which slows everything down. Limiting to 1 trial at a time helps to avoid this resource contention.

arnavgarg1: Got it, that makes sense! Similar to the problem I was seeing as well - personally I even noticed this happening on my local at times even when not using a RayBackend.

arnavgarg1:
One of the other things we'll need to do is wrap writing some of the `hyperopt_statistics` to remote storage in a `use_credentials` block as well over here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/hyperopt/run.py#L385 (which happens after `tune.run()` returns).  This will ensure that `hyperopt_statistics.json` is also saved to the same remote storage location for retrieval since we write this manually.

It might be worth either modifying an existing test (or adding a new one, although we should probably avoid that) to make sure we can also retrieve hyperopt_statistics.json from the same remote location that we sync to when running hyperopt E2E.

tests/integration_tests/test_hyperopt.py
tgaddair: It's a good indicator that 1 CPU per trial is probably too low. In practice, we may want to bump this up.

tgaddair:
> Maybe this is not in the scope of this PR, but one of the other things we'll need to do is wrap writing some of the `hyperopt_statistics` to remote storage in a `use_credentials` block as well over here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/hyperopt/run.py#L385 (which happens after `tune.run()` returns). This will ensure that `hyperopt_statistics.json` is also saved to the same remote storage location for retrieval since we write this manually.
> 
> It might be worth either modifying an existing test (or adding a new one, although we should probably avoid that) to make sure we can also retrieve hyperopt_statistics.json from the same remote location that we sync to when running hyperopt E2E.

@arnavgarg1 we do have a test for the existence of this file in the remote [here](https://github.com/ludwig-ai/ludwig/pull/2617/files#diff-53bd1d32793cc8ee9b0a79ef7829dccfdf59d54a7d9f6f7138a6d4dd6502c3b1R382).

It's true that we'll need to do the `use_credentials` trick if the user doesn't have the credentials in their environment. But I purposely left that out of scope for this PR to focus on just getting syncing working when credentials are being set correctly. In a follow-up, I will add in the plumbing to make use of `use_credentials` to allow overriding the creds in the environment.

arnavgarg1:
Sounds good! This looks great, thanks @tgaddair!

arnavgarg1 (approved):
Approving this for now, but we should probably land this once the plumbing PR is in as well!

tgaddair:
Will avoid cherry-picking into the release branch until we have the credential PR in.

PR #2619, opened 2022-10-10
Set commonly used parameters

connor-mccorm:


ludwig/schema/combiners/utils.py
```diff
@@ -16,15 +19,17 @@ def get_combiner_jsonschema():
...
+    parameter_metadata = convert_metadata_to_json(COMBINER_METADATA[TYPE])
     return {
         "type": "object",
         "properties": {
             "type": {
                 "type": "string",
                 "enum": combiner_types,
                 "default": "concat",
-                "title": "type",
+                "title": "Combiner Type",
```
ksbrar, line 30: Actually I think we should revert this change. Otherwise the combiner page will read Combiner as a super heading and then "Combiner Type" as a sub-heading. I think we'll do the prefixing on the frontend side (CC @martindavis ).

Also as a general note we should keep things on the Ludwig side in snake case IMO.

ksbrar (approved):
Nice! I'm fine with this change but considering how many things are toggled false, just want to throw out there that we could just add another attribute to `ParameterMetadata` that's called `frontend_commonly_used` or something. Thoughts?

ludwig/schema/combiners/utils.py
connor-mccorm: Ah ok, yes this is indeed your wheelhouse so will change back!

connor-mccorm:
Ya I thought about that, but when looking at the actual parameters that had `commonly_used=True` about half weren't actually commonly used. I think there is going to be a documentation workstream in the future anyways where we start filling out the parameter metadata more and actually choosing the params that we want commonly used. So for now I think it's ok since the original `commonly_used` was really already corrupted.

justinxzhao (approved):

martindavis (approved):

PR #2623, opened 2022-10-11
Adds minimum split size, ensures random split is never smaller than minimum for local backend

dantreiman:
Should fix NaN metrics caused by too-small validation set.

Problem: if the validation set is too small (1 row, for example) some metrics evaluate to NaN, which causes tests to fail.

This situation is mainly encountered in test cases where we use small synthetic datasets with random splits.

The solution here is to add a constant `MIN_DATASET_SPLIT_ROWS`, and ensure that each nonempty split contains at least `MIN_DATASET_SPLIT_ROWS`

### DIFF
--- a/ludwig/data/split.py
+++ b/ludwig/data/split.py
@@ -22,7 +22,7 @@ import numpy as np
 from sklearn.model_selection import train_test_split
 
 from ludwig.backend.base import Backend
-from ludwig.constants import BINARY, CATEGORY, COLUMN, DATE, SPLIT, TYPE
+from ludwig.constants import BINARY, CATEGORY, COLUMN, DATE, MIN_DATASET_SPLIT_ROWS, SPLIT, TYPE
 from ludwig.schema.split import (
     DateTimeSplitConfig,
     FixedSplitConfig,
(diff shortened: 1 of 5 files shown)


ludwig/data/split.py
```diff
@@ -69,21 +103,23 @@ def __init__(self, probabilities: List[float] = DEFAULT_PROBABILITIES, **kwargs)
     def split(
         self, df: DataFrame, backend: Backend, random_seed: float = default_random_seed
     ) -> Tuple[DataFrame, DataFrame, DataFrame]:
+        probabilities = _make_fractions_ensure_minimum_rows(self.probabilities, len(df))
```
tgaddair, line 106: Use value of `n` below instead of recomputing.

tgaddair, line 106: `len(df)` is pretty expensive in Dask (requires a full pass over the data). Maybe we can only perform this optimization for Pandas for the time being?

dantreiman: This should be ok, as long as the distributed tests have enough rows.

dantreiman: Actually, I'm going to take a stab at implementing this for the dask ddf case and see how it goes.  There are some distributed tests that could be fixed by guaranteeing sufficient split size.

dantreiman: Implementing this in the distributed case is a lot more complicated...  so I decided to revert this PR back to local-only.  Will try to find a better solution for distributed and do that in a separate PR.

ludwig/data/split.py
```diff
@@ -61,6 +61,38 @@ def required_columns(self) -> List[str]:
...
 
 
+def _make_divisions_ensure_minimum_rows(divisions, n_examples, min_val_rows, min_test_rows):
+    """Revises divisions to ensure no dataset split has too few examples."""
+    result = list(divisions)
+    n = [dn - dm for dm, dn in zip((0,) + divisions, divisions + (n_examples,))]  # Number of examples in each split.
+    if 0 < n[2] < min_test_rows and n[0] > 0:
+        # Test set is nonempty but too small, take examples from training set.
+        shift = min(min_test_rows - n[2], n[0])
+        result = [d - shift for d in result]
```
arnavgarg1, line 71: @dantreiman Just to make sure I understand this correctly:

Let's say we have 4 rows in our dataset with split probabilities [0.7, 0, 0.3]. In this case, d1 = 3 and d2 = 5 (because of the `math.ceil` calls). Then, in this function, we're going to see that n = [3, 2, -1]. With the first condition, we're going to see that it's not satisfied since n[2] = -1. With the second condition, n[1] = 2 < min_val_rows so we adjust the first division to be 2. So the new division will then be (2, 5). This doesn't match with what we want right?

Let me know if I misunderstood the code.

dantreiman: With 4 rows, split fractions (0.7, 0, 0.3):   In this situation, d1 and d2 are both 3.  Reason:
- `d1 = int(np.ceil(probabilities[0] * n_rows))` evaluates to 3.  `0.7 * 4` is 2.8, rounded up to 3.
- `n2 = int(probabilities[1] * n_rows)` evaluates to 0, since probabilities[1] is 0
- `d2 = d1 + n2`

So then we call `_make_divisions_ensure_minimum_rows((3, 3), 4)`, which will return `(1, 1)`, because it enforces the minimum size 3 of the test set.

This is working as intended, though now we have a training set of size 1 which is not desirable.  I thought about guaranteeing the size of the training set, but it might make more sense to enforce a minimum dataset size for splitting instead.... that could be a good follow-up.

arnavgarg1: Ah got it! I missed `probabilities[1]`, thought it was `probabilities[-1] * n_rows` 
Yeah, that makes sense - my biggest concern was that the training set size gets depleted which is a problem only for very small datasets (sub 10 rows), or tasks that build up datasets row by row (definitely seen a few cases within the Ludwig open source community). 

Could be okay to go ahead with this for now and then add a PR to ensure a minimum dataset size for splitting :)

tgaddair (approved):

arnavgarg1 (approved):

PR #2626, opened 2022-10-11
accepting dictionary as input to `benchmarking.benchmark`

abidwael:
Addressing the following comments:
- https://github.com/ludwig-ai/ludwig/pull/2568#discussion_r992500174
- https://github.com/ludwig-ai/ludwig/pull/2568#discussion_r992525203


ludwig/benchmarking/artifacts.py
```diff
@@ -0,0 +1,25 @@
+import os
+
+from ludwig.utils.data_utils import load_json, load_yaml
+
+
+class BenchmarkingArtifact:
```
justinxzhao, line 6: I like this idea of defining a compositional return type that users can subsequently interface with after calling `benchmark()`.

`ludwig experiment` and `ludwig train` should move to this pattern as well, if/when we move all of the modeling artifacts move to dataclasses, CC: @dantreiman 

It looks like, besides the `load_yaml` unpacking, this could be as simple `dataclass`.

Suggestion: define a dataclass, i.e. `BenchmarkingResult`, and a method that creates that dataclass, i.e. `build_benchmarking_results(benchmarking_config: dict, experiment_idx: int)`.

Then the return of `benchmark` could be a dict of experiment name (or other id) (str) -> `BenchmarkingResult`.

ludwig/benchmarking/utils.py
```diff
@@ -155,6 +155,46 @@ async def download_one(
...
+    ):
+        raise ValueError("You must either specify a global experiment name or an experiment name for each experiment.")
+    if "export" not in benchmarking_config:
+        raise ValueError(
+            """You must specify export parameters. Example:
+            export:
+              export_artifacts: true
+              export_base_path: s3://benchmarking.us-west-2.ludwig.com/bench/    # include the slash at the end.
+        """
+        )
```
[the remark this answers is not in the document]
abidwael: Not sure I understand, but put the parenthesis on the same line with the quote.

ludwig/benchmarking/artifacts.py
```diff
@@ -0,0 +1,60 @@
+import os
+from dataclasses import dataclass
+from typing import Any, Dict
+
+from ludwig.utils.data_utils import load_json, load_yaml
+
+
+@dataclass
+class BenchmarkingResult:
```
justinxzhao, line 9: Much cleaner! :)

ludwig/benchmarking/benchmark.py
```diff
@@ -94,44 +97,43 @@ def benchmark_one(experiment: Dict[str, Union[str, Dict[str, str]]]) -> None:
...
 
 
-def benchmark(bench_config_path: str) -> None:
+def benchmark(benchmarking_config: Union[dict, str]) -> Dict[str, BenchmarkingResult]:
     """Launch benchmarking suite from a benchmarking config.
 
-    bench_config_path: config for the benchmarking tool. Specifies datasets and their
-        corresponding Ludwig configs, as well as export options.
+    Args:
+        bench_config_path: config for the benchmarking tool. Specifies datasets and their
```
justinxzhao, line 115: Update docstring.

ludwig/benchmarking/benchmark.py
```diff
@@ -94,44 +97,43 @@ def benchmark_one(experiment: Dict[str, Union[str, Dict[str, str]]]) -> None:
...
             skip_save_unprocessed_output=True,
             skip_save_predictions=True,
             skip_collect_predictions=True,
         )
-        delete_model_checkpoints(output_directory)
+        delete_model_checkpoints(experiment["experiment_name"])
 
 
-def benchmark(bench_config_path: str) -> None:
+def benchmark(benchmarking_config: Union[dict, str]) -> Dict[str, BenchmarkingResult]:
```
justinxzhao, line 111: `Dict` instead of `dict`?

ludwig/benchmarking/benchmark.py
```diff
@@ -94,44 +97,43 @@ def benchmark_one(experiment: Dict[str, Union[str, Dict[str, str]]]) -> None:
...
-                experiment["process_config_file_path"] = benchmarking_config["process_config_file_path"]
-            if "profiler" in benchmarking_config:
-                experiment["profiler"] = benchmarking_config["profiler"]
             benchmark_one(experiment)
+            dataset_name = experiment["dataset_name"]
+            experiment_artifacts[dataset_name] = build_benchmarking_result(benchmarking_config, experiment_idx)
         except Exception:
             logger.exception(
-                f"Experiment *{experiment['experiment_name']}* on dataset *{experiment['dataset_name']}* failed"
+                f"Experiment *{experiment['experiment_name']}* on " f"dataset *{experiment['dataset_name']}* failed"
```
justinxzhao, line 130: It looks like this can be one string "".

ludwig/benchmarking/utils.py
```diff
@@ -155,12 +159,57 @@ async def download_one(
...
+    if "experiments" not in benchmarking_config:
+        raise ValueError("You must specify a list of experiments.")
```
justinxzhao, line 181: nit: Move these three and the raise on 191 to a separate `validate_benchmarking_config()` function.

This will require two traversals through the benchmarking_config (one in validation and one in propagation), but the separation of validation and propagation is clearer.

justinxzhao (approved):
Thanks for the changes!

ludwig/benchmarking/benchmark.py
```diff
@@ -94,44 +97,43 @@ def benchmark_one(experiment: Dict[str, Union[str, Dict[str, str]]]) -> None:
...
-        corresponding Ludwig configs, as well as export options.
+    Args:
+        benchmarking_config: config or config path for the benchmarking tool. Specifies datasets and their
+            corresponding Ludwig configs, as well as export options.
     """
-    benchmarking_config = load_yaml(bench_config_path)
-    for experiment in benchmarking_config["experiments"]:
+    if isinstance(benchmarking_config, str):
+        benchmarking_config = load_yaml(benchmarking_config)
+    benchmarking_config = propagate_global_parameters(benchmarking_config)
```
justinxzhao, line 120: nit: I would move the validation step to be called at this level:

```python
validate_benchmarking_config(benchmarking_config)
benchmarking_config = populate_benchmarking_config_with_defaults(benchmarking_config)
benchmarking_config = propagate_global_parameters(benchmarking_config)
```

ludwig/benchmarking/utils.py
```diff
@@ -155,12 +159,80 @@ async def download_one(
     return dataset_name, local_dir
 
 
+def validate_benchmarking_config(benchmarking_config: Dict[str, Any]) -> None:
```
justinxzhao, line 162: nit: Consider adding a simple unit test for this function. And/or a test that validates the benchmarking config that's currently checked in.

abidwael: Done

tests/ludwig/benchmarking/test_benchmarking.py
```diff
@@ -0,0 +1,30 @@
...
+    return [(config_fp, does_not_raise()) for config_fp in valid_bench_configs] + [
+        (config_fp, pytest.raises(ValueError)) for config_fp in invalid_bench_configs
+    ]
+
+
+@pytest.mark.parametrize("benchmarking_config_fp,error_expectation", get_benchamrking_config_validity_pair())
+def test_benchmarking(benchmarking_config_fp, error_expectation):
+    benchmarking_config = load_yaml(benchmarking_config_fp)
+
+    with error_expectation:
```
justinxzhao, line 29: If you want to keep this test, I would  split it two separate tests, one for valid configs and one for invalid configs, i.e. `test_benchmarking_configs_valid` and `test_benchmarking_configs_invalid`.

The extra method to pair up configs error expectations is more complicated than these tests need to be.

In general, avoid logic in tests:
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests

justinxzhao (approved):

PR #2628, opened 2022-10-11
Factor out defaults mixin change

connor-mccorm:
Factoring out this code change from Config Object PR : https://github.com/ludwig-ai/ludwig/pull/2426#discussion_r991723691


justinxzhao (approved):
Thanks for factoring out this change from the main PR!

This change is more clear (and less controversial) and I'm glad we can check it in independently.

PR #2630, opened 2022-10-12
Added storage backend API to allow injecting dynamic credentials

tgaddair:
This PR adds a basic structure for specifying different cloud credentials for different entities in the system. Currently supported is the cache and artifacts (for hyperopt), but this will be extended to support all artifacts (outputs) and datasets (inputs) in the future.

This also deprecates the `cache_credentials` backend config param and moves it to `credentials.cache`. The structure is:

```
backend:
  credentials:
    artifacts: ...
    cache: ...
```

Also removes s3 as a test dep and uses an internal minio container instead, so forks can still run tests successfully.

### DIFF
--- a/ludwig/backend/base.py
+++ b/ludwig/backend/base.py
@@ -17,13 +17,14 @@
 from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
-from typing import Callable, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 import numpy as np
 import pandas as pd
 import psutil
 import torch
 
+from ludwig.backend.utils.storage import StorageManager
 from ludwig.data.cache.manager import CacheManager
 from ludwig.data.dataframe.pandas import PANDAS
 from ludwig.data.dataset.base import DatasetManager
@@ -42,17 +43,23 @@ class Backend(ABC):
         self,
         dataset_manager: DatasetManager,
         cache_dir: Optional[str] = None,
-        cache_credentials: Optional[Union[str, dict]] = None,
+        credentials: Optional[Dict[str, Any]] = None,
     ):
+        credentials = credentials or {}
         self._dataset_manager = dataset_manager
-        self._cache_manager = CacheManager(self._dataset_manager, cache_dir, cache_credentials)
+        self._storage_manager = StorageManager(**credentials)
+        self._cache_manager = CacheManager(self._dataset_manager, cache_dir)
 
     @property
-    def cache(self):
+    def storage(self) -> StorageManager:
+        return self._storage_manager
+
+    @property
+    def cache(self) -> CacheManager:
         return self._cache_manager
 
     @property
-    def dataset_manager(self):
+    def dataset_manager(self) -> DatasetManager:
         return self._dataset_manager
 
     @abstractmethod
--- a/ludwig/backend/base.py
+++ b/ludwig/backend/base.py
@@ -43,7 +43,7 @@ class Backend(ABC):
         self,
         dataset_manager: DatasetManager,
         cache_dir: Optional[str] = None,
-        credentials: Optional[Dict[str, Any]] = None,
+        credentials: Optional[Dict[str, Dict[str, Any]]] = None,
     ):
         credentials = credentials or {}
         self._dataset_manager = dataset_manager
--- a/ludwig/backend/utils/storage.py
+++ b/ludwig/backend/utils/storage.py
@@ -2,7 +2,6 @@ import contextlib
 from typing import Any, Dict, Optional, Union
 
 from ludwig.utils import data_utils
-from ludwig.utils.fs_utils import RemoteFilesystem
 
 CredInputs = Optional[Union[str, Dict[str, Any]]]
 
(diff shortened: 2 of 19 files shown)


ludwig/backend/base.py
```diff
@@ -42,17 +43,23 @@ def __init__(
         self,
         dataset_manager: DatasetManager,
         cache_dir: Optional[str] = None,
-        cache_credentials: Optional[Union[str, dict]] = None,
+        credentials: Optional[Dict[str, Dict[str, Any]]] = None,
     ):
+        credentials = credentials or {}
         self._dataset_manager = dataset_manager
-        self._cache_manager = CacheManager(self._dataset_manager, cache_dir, cache_credentials)
+        self._storage_manager = StorageManager(**credentials)
```
dantreiman, line 50: Nice refactoring here!  Separation of concerns caching vs. storage

ludwig/utils/backward_compatibility.py
```diff
@@ -161,6 +161,22 @@ def _traverse_dicts(config: Any, f: Callable[[Dict], None]):
             _traverse_dicts(v, f)
 
 
+@register_config_transformation("0.6")
```
dantreiman, line 164: Slightly simpler option:

```
@register_config_transformation("0.6", "backend")
def _update_backend_cache_credentials(backend: Dict[str, Any]) -> Dict[str, Any]:
    if "cache_credentials" in backend:
        ...
    return backend
```

ludwig/backend/utils/storage.py
```diff
@@ -0,0 +1,69 @@
...
+DATASETS = "datasets"
+CACHE = "cache"
+
+
+class Storage:
+    def __init__(self, creds: Optional[Dict[str, Any]]):
+        self._creds = creds
+
+    @contextlib.contextmanager
+    def use(self):
```
dantreiman, line 20: Maybe this should be called `use_credentials()`, i.e.

`with backend.storage.cache.use_credentials():`

Alternatively, `def use(self):` could have a doc string explaining what it does.

dantreiman (approved):
LGTM

ludwig/backend/utils/storage.py
```diff
@@ -0,0 +1,69 @@
...
+
+from ludwig.utils import data_utils
+
+CredInputs = Optional[Union[str, Dict[str, Any]]]
+
+
+DEFAULTS = "defaults"
+ARTIFACTS = "artifacts"
+DATASETS = "datasets"
+CACHE = "cache"
```
arnavgarg1, line 12: nit: maybe pull these in from `constants.py` instead of defining them inline? I think some of these might already be there.

tgaddair: These I actually want to keep separate, as they're only used internally within this file.

ludwig/utils/backward_compatibility.py
tgaddair: Good idea! Fixed.

ludwig/backend/utils/storage.py
tgaddair: I agree, changed to `use_credentials`

ludwig/backend/utils/storage.py
arnavgarg1: Got it!

ludwig/utils/fs_utils.py
```diff
@@ -192,9 +192,6 @@ def copy(src, tgt, recursive=False):
 def makedirs(url, exist_ok=False):
     fs, path = get_fs_and_path(url)
     fs.makedirs(path, exist_ok=exist_ok)
-    if not path_exists(url):
-        with fsspec.open(url, mode="wb"):
-            pass
```
arnavgarg1, line 197: I thought this was here to ensure that in case we can't make a "dir", we just create a file in write mode since S3 storage works that way? But I guess this may break for other types of storage. I just checked and it seems like fsspec does seem to automatically do this exact thing for filesystems that need it, so I guess it should be ok to remove this

arnavgarg1 (approved):
Looks great, I like this structure for managing credentials for different parts of Ludwig!

tests/integration_tests/test_hyperopt.py
```diff
@@ -379,17 +386,33 @@ def test_hyperopt_run_hyperopt(csv_filename, search_space, tmpdir, ray_cluster):
     assert isinstance(hyperopt_results, HyperoptResults)
 
     # check for existence of the hyperopt statistics file
-    assert fs_utils.path_exists(os.path.join(tmpdir, experiment_name, HYPEROPT_STATISTICS_FILE_NAME))
+    with use_credentials(minio_test_creds()):
+        assert fs_utils.path_exists(os.path.join(tmpdir, experiment_name, HYPEROPT_STATISTICS_FILE_NAME))
+        for trial in hyperopt_results.experiment_analysis.trials:
+            assert fs_utils.path_exists(os.path.join(tmpdir, experiment_name, f"trial_{trial.trial_id}"))
```
arnavgarg1, line 392: Is it also possible to add additional checks for ensuring that `model_hyperparameters.json` and `training_set_metadata.json` are also in this folder / exist / are non-empty? Just to make sure this works for hyperopt end-to-end since those files are super important.

tgaddair: Yes, we can check those as well. Let's do that in a follow-up.

Issue #2344, opened 2022-08-03
Return trial parameters from hyperopt before training starts


arnavgarg1:
**Describe the use case**
It would be nice to have a way to get access to the sampled hyperopt space for each trial before the trials begin, rather than only seeing them during RayTune execution or after the trials are done.

arnavgarg1:
@jeffkinnison

Issue #2343, opened 2022-08-02
Get trial status from outside the program


Jeffwan:
**Is your feature request related to a problem? Please describe.**

Yes. I would like to know how many trials are pending/running/finished. How can I get the status from external?

**Describe the use case**

As a user, once I kickoff the job, I want to understand the progress of the program. 

**Describe the solution you'd like**
ludwig is more like a SDK, seems it's very hard to get program status from external? If there's no elegant way to make it, could we update a file so external program can read the file to get latest updates?

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

ShreyaR:
Hi @Jeffwan, great question!

The most straightforward way to get hyperopt trial status would be the `experiment_state-<datetime>.json` file that is created inside the results logdir. This file contains a list of checkpoints, where each checkpoint is a dictionary that contains information about a trial. You can take a look at the `status` key for each dictionary, and the checkpoints list overall, to query the status of the hyperopt experiment. Would that be sufficient for unblocking your use case?

Another way to do this programatically would be using [Ray Tune Callbacks](https://docs.ray.io/en/latest/tune/api_docs/internals.html#ray.tune.callback.Callback), where we update the trial status each time the callback is used.

Jeffwan:
I feel it's good to write status in file but it's hard to check status real time. Currently, I copy all files to remote once it's done. If I provide a service to check trial status, that means my backend service needs to find the file and parse it to get status which doesn't seem very promising. But I think it's still helpful to understand the statistic. For example, how many trials failed etc

PR #2640, opened 2022-10-12
Allow Ray Tune callbacks to be passed into hyperopt and log model config

jeffkinnison:
This update makes three changes to hyperopt in Ludwig:

1.  Adds the `tune_callbacks` parameter to hyperopt to allow passing [Ray Tune Callbacks](https://docs.ray.io/en/latest/tune/api_docs/internals.html#callbacks) into Ray Tune
2. Updates the logic of the RayTuneExecutor to pass the callbacks into Ray Tune
3. Logs user-supplied and rendered model configs to trial logdirs

These changes support error surfacing, as well as #2343 and #2344

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -449,16 +449,13 @@ class RayTuneExecutor:
         trial_id = tune.get_trial_id()
         trial_dir = Path(tune.get_trial_dir())
 
-        # Write out the unmerged config to the trial's local directory.
-        with open(os.path.join(trial_dir, 'user_hyperparameters.json'), 'w') as f:
-            json.dump(hyperopt_dict['config'], f)
-
         modified_config = substitute_parameters(copy.deepcopy(hyperopt_dict["config"]), config)
 
-        modified_config = ModelConfig.from_dict(modified_config).to_dict()
+        # Write out the unmerged config to the trial's local directory.
+        with open(os.path.join(trial_dir, "trial_hyperparameters.json"), "w") as f:
+            json.dump(hyperopt_dict["config"], f)
 
-        with open(os.path.join(trial_dir, 'model_hyperparameters.json'), 'w') as f:
-            json.dump(hyperopt_dict['config'], f)
+        modified_config = ModelConfig.from_dict(modified_config).to_dict()
 
         hyperopt_dict["config"] = modified_config
         hyperopt_dict["experiment_name "] = f'{hyperopt_dict["experiment_name"]}_{trial_id}'
@@ -770,7 +767,6 @@ class RayTuneExecutor:
             )
 
         tune_config = {}
-        # tune_callbacks = []
         for callback in callbacks or []:
             run_experiment_trial, tune_config = callback.prepare_ray_tune(
                 run_experiment_trial,
@@ -816,7 +812,7 @@ class RayTuneExecutor:
         # https://docs.ray.io/en/latest/tune/tutorials/tune-stopping.html
         should_resume = "AUTO" if resume is None else resume
 
-        raise_on_failed_trial = kwargs.get('raise_on_failed_trial', True)
+        raise_on_failed_trial = kwargs.get("raise_on_failed_trial", True)
 
         try:
             analysis = tune.run(
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -451,7 +451,7 @@ class RayTuneExecutor:
 
         modified_config = substitute_parameters(copy.deepcopy(hyperopt_dict["config"]), config)
 
-        # Write out the unmerged config to the trial's local directory.
+        # Write out the unmerged config with sampled hyperparameters to the trial's local directory.
         with open(os.path.join(trial_dir, "trial_hyperparameters.json"), "w") as f:
             json.dump(hyperopt_dict["config"], f)
 
@@ -812,8 +812,6 @@ class RayTuneExecutor:
         # https://docs.ray.io/en/latest/tune/tutorials/tune-stopping.html
         should_resume = "AUTO" if resume is None else resume
 
-        raise_on_failed_trial = kwargs.get("raise_on_failed_trial", True)
-
         try:
             analysis = tune.run(
                 f"trainable_func_f{hash_dict(config).decode('ascii')}",
@@ -840,7 +838,6 @@ class RayTuneExecutor:
                 verbose=hyperopt_log_verbosity,
                 resume=should_resume,
                 log_to_file=True,
-                raise_on_failed_trial=raise_on_failed_trial,
             )
         except Exception as e:
             # Explicitly raise a RuntimeError if an error is encountered during a Ray trial.
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -449,9 +449,16 @@ class RayTuneExecutor:
         trial_id = tune.get_trial_id()
         trial_dir = Path(tune.get_trial_dir())
 
+        # Write out the unmerged config to the trial's local directory.
+        with open(os.path.join(trial_dir, 'user_hyperparameters.json'), 'w') as f:
+            json.dump(hyperopt_dict['config'], f)
+
         modified_config = substitute_parameters(copy.deepcopy(hyperopt_dict["config"]), config)
 
         modified_config = ModelConfig.from_dict(modified_config).to_dict()
+
+        with open(os.path.join(trial_dir, 'model_hyperparameters.json'), 'w') as f:
+            json.dump(hyperopt_dict['config'], f)
 
         hyperopt_dict["config"] = modified_config
         hyperopt_dict["experiment_name "] = f'{hyperopt_dict["experiment_name"]}_{trial_id}'
@@ -659,6 +666,7 @@ class RayTuneExecutor:
         gpu_memory_limit=None,
         allow_parallel_threads=True,
         callbacks=None,
+        tune_callbacks=None,
         backend=None,
         random_seed=default_random_seed,
         debug=False,
@@ -762,7 +770,7 @@ class RayTuneExecutor:
             )
 
         tune_config = {}
-        tune_callbacks = []
+        # tune_callbacks = []
         for callback in callbacks or []:
             run_experiment_trial, tune_config = callback.prepare_ray_tune(
                 run_experiment_trial,
@@ -808,6 +816,8 @@ class RayTuneExecutor:
         # https://docs.ray.io/en/latest/tune/tutorials/tune-stopping.html
         should_resume = "AUTO" if resume is None else resume
 
+        raise_on_failed_trial = kwargs.get('raise_on_failed_trial', True)
+
         try:
             analysis = tune.run(
                 f"trainable_func_f{hash_dict(config).decode('ascii')}",
@@ -834,6 +844,7 @@ class RayTuneExecutor:
                 verbose=hyperopt_log_verbosity,
                 resume=should_resume,
                 log_to_file=True,
+                raise_on_failed_trial=raise_on_failed_trial,
             )
         except Exception as e:
             # Explicitly raise a RuntimeError if an error is encountered during a Ray trial.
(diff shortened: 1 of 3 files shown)


ludwig/hyperopt/execution.py
```diff
@@ -806,6 +814,8 @@ def _register(name, trainable):
         # https://docs.ray.io/en/latest/tune/tutorials/tune-stopping.html
         should_resume = "AUTO" if resume is None else resume
 
+        raise_on_failed_trial = kwargs.get("raise_on_failed_trial", True)
```
arnavgarg1, line 817: Should this be false by default?

ludwig/hyperopt/execution.py
```diff
@@ -447,10 +447,17 @@ def _run_experiment(
...
+        # Write out the unmerged config to the trial's local directory.
+        with open(os.path.join(trial_dir, "user_hyperparameters.json"), "w") as f:
+            json.dump(hyperopt_dict["config"], f)
+
         modified_config = substitute_parameters(copy.deepcopy(hyperopt_dict["config"]), config)
 
         modified_config = merge_with_defaults(modified_config)
 
+        with open(os.path.join(trial_dir, "model_hyperparameters.json"), "w") as f:
+            json.dump(hyperopt_dict["config"], f)
```
arnavgarg1, line 459: @jeffkinnison We actually write this during `model.train()` here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/api.py#L547

Is the reason to do this to get `model_hyperparameters.json` before we even get to `model.train()`? This would be useful is there's failures that may happen during preprocessing time since we only write `model_hyperparameters.json` after preprocessing.

ludwig/hyperopt/execution.py
```diff
@@ -447,10 +447,17 @@ def _run_experiment(
         trial_id = tune.get_trial_id()
         trial_dir = Path(tune.get_trial_dir())
 
+        # Write out the unmerged config to the trial's local directory.
+        with open(os.path.join(trial_dir, "user_hyperparameters.json"), "w") as f:
+            json.dump(hyperopt_dict["config"], f)
+
```
arnavgarg1, line 453: @jeffkinnison Would it make sense to potentially rename this to `trial_hyperparameters.json`?

arnavgarg1: Also, would it make more sense to do this with `modified_config` (pre-merge with defaults, post merge with sampled values) as supposed to just `hyperopt_dict["config"]`? I'm assuming we also want to capture the values that were actually sampled for this trial, which I believe is stored in config and then merged into `hyperopt_dict["config"]` in the substitute_parameters call below. Let me know if I'm missing something!

ludwig/hyperopt/run.py
```diff
@@ -7,6 +7,7 @@
 
 import pandas as pd
 import yaml
+from ray.tune import Callback as TuneCallback
```
arnavgarg1, line 10: @jeffkinnison We'll want to move this into the try-except block on line 46, and then set `TuneCallback` to None or something like that. The reason is because Ray/Dask falls under "distributed" tests, so this will cause failures in the non-distributed tests when collecting all the tests to run

ludwig/hyperopt/execution.py
arnavgarg1: Would also encourage using `MODEL_HYPERPARAMETERS_FILE_NAME` from `ludwig.globals` just incase this file name changes :)

arnavgarg1:
Are we able to modify an existing test and pass in a very simple TuneCallback to make sure that tune_callbacks works as intended? Any of the ones in `test_hyperopt.py` or `test_hyperopt_ray.py` might be good candidates!

ludwig/hyperopt/execution.py
jeffkinnison: Yeah, I see that we make some modifications to the config at the start of training. It probably makes more sense to read the rendered config on training start.

ludwig/hyperopt/execution.py
jeffkinnison: Sure, I can do that.

jeffkinnison:
> Are we able to modify an existing test and pass in a very simple TuneCallback to make sure that tune_callbacks works as intended? Any of the ones in `test_hyperopt.py` or `test_hyperopt_ray.py` might be good candidates!

I can do that for both.

ludwig/hyperopt/execution.py
jeffkinnison: That was a good call, and an oversight on my part, thanks!

ludwig/hyperopt/execution.py
```diff
@@ -806,6 +810,8 @@ def _register(name, trainable):
         # https://docs.ray.io/en/latest/tune/tutorials/tune-stopping.html
         should_resume = "AUTO" if resume is None else resume
 
+        raise_on_failed_trial = kwargs.get("raise_on_failed_trial", True)
```
ShreyaR, line 813: nit: Why not just add this as an optional argument to the function? That was, you can also add documentation for it in the docstring that explains why it's needed.

ShreyaR (approved):
lgtm! Feel free to merge once comment is addressed and tests are passing.

ludwig/hyperopt/execution.py
jeffkinnison: This is an artifact of an earlier build I'm going to remove it now.

ludwig/hyperopt/execution.py
jeffkinnison: This should be addressed by [7b40418](https://github.com/ludwig-ai/ludwig/pull/2640/commits/7b40418caeba34b8ab3c914984282f07542a75cf)

ludwig/hyperopt/execution.py
```diff
@@ -449,6 +449,10 @@ def _run_experiment(
 
         modified_config = substitute_parameters(copy.deepcopy(hyperopt_dict["config"]), config)
 
+        # Write out the unmerged config to the trial's local directory.
```
arnavgarg1, line 452: nit: Can we update this comment to 

```Write out the unmerged config with sampled parameters to the trial's local directory```

File: ludwig/hyperopt/execution.py (hunk 5 above)
Comment (arnavgarg1) [reply ludwig/hyperopt/execution.py:813]: @jeffkinnison Was this supposed to be updated based on your response to Shreya's comment? Seeing the same thing as before - let me know!

File: ludwig/hyperopt/run.py (hunk 4 above)
Comment (arnavgarg1) [reply ludwig/hyperopt/run.py:10]: @jeffkinnison I think you'll still need to address this to prevent the test failures in the non-distributed tests

File: tests/integration_tests/test_hyperopt_ray.py [hunk 7]
```diff
@@ -123,6 +129,37 @@ def ray_cluster_4cpu():
...
+        self.trial_ids = set()
+        self.user_config = {}
+        self.rendered_config = {}
+
+    def on_trial_start(self, iteration: int, trials: List["Trial"], trial: "Trial", **info):
+        super().on_trial_start(iteration, trials, trial, **info)
+        self.trial_ids.add(trial.trial_id)
+
+    def on_trial_complete(self, iteration: int, trials: List["Trial"], trial: "Trial", **info):
+        super().on_trial_start(iteration, trials, trial, **info)
```
arnavgarg1, line 143: I think this should be 

```
super().on_trial_complete(iteration, trials, trial, **info)
```

arnavgarg1 (changes requested):
This is almost there. I think there's a couple of small changes

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -305,16 +342,22 @@ def run_hyperopt(
...
+        tune_callbacks=[tune_test_callback],
     )
 
     # check for return results
     assert isinstance(hyperopt_results, HyperoptResults)
 
     # check for existence of the hyperopt statistics file
     assert os.path.isfile(os.path.join(tmpdir, experiment_name, HYPEROPT_STATISTICS_FILE_NAME))
+
+    # check for evidence that the callback
```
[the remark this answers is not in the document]
jeffkinnison: Ah, somehow that was left out of the last commit.

ludwig/hyperopt/run.py
jeffkinnison: I removed it entirely since it was only used for type hints. Replaced with a string in [a073cc0](https://github.com/ludwig-ai/ludwig/pull/2640/commits/a073cc00aec3e3915c4cdd143aabbee7a247c816)

ludwig/hyperopt/execution.py
jeffkinnison: Removed in [a073cc0](https://github.com/ludwig-ai/ludwig/pull/2640/commits/a073cc00aec3e3915c4cdd143aabbee7a247c816)

ludwig/hyperopt/execution.py
jeffkinnison: Updated in [a073cc0](https://github.com/ludwig-ai/ludwig/pull/2640/commits/a073cc00aec3e3915c4cdd143aabbee7a247c816)

tests/integration_tests/test_hyperopt_ray.py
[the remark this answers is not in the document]
jeffkinnison: The code now exists in [a073cc0](https://github.com/ludwig-ai/ludwig/pull/2640/commits/a073cc00aec3e3915c4cdd143aabbee7a247c816)

tests/integration_tests/test_hyperopt_ray.py
jeffkinnison: Fixed in [a073cc0](https://github.com/ludwig-ai/ludwig/pull/2640/commits/a073cc00aec3e3915c4cdd143aabbee7a247c816)

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -13,29 +13,37 @@
...
 import logging
 import os.path
+import sys
+from typing import List
 
 import mlflow
 import pandas as pd
 import pytest
 from mlflow.tracking import MlflowClient
+from pyexpat import model
```
arnavgarg1, line 26: Accidental import?

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -13,29 +13,37 @@
...
+from ray import tune
```
arnavgarg1, line 27: Wondering if this import is needed and if it'll cause the same error that the callback was throwing - would be good to double check!

jeffkinnison: Pre-commit CI bot added that too.

arnavgarg1 (approved):
Looks great, thanks @jeffkinnison ! Would just resolve merge conflicts and make sure the tests pass before merging!

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -125,6 +130,44 @@ def ray_cluster_4cpu():
...
+        super().on_trial_complete(iteration, trials, trial, **info)
+        self.trial_status[trial.trial_id] = trial.status
+
+        model_hyperparameters = os.path.join(
+            trial.logdir, f"{self.exp_name}_{self.model_type}", "model", MODEL_HYPERPARAMETERS_FILE_NAME
+        )
+        if os.path.isfile(model_hyperparameters):
+            try:
+                with open(model_hyperparameters) as f:
+                    json.load(f)
```
arnavgarg1, line 156: Is it worth adding a check here to make sure the loaded json isn't empty?

arnavgarg1: Same for the json load below

jeffkinnison: Sure, I can add that.

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -125,6 +132,46 @@ def ray_cluster_4cpu():
...
+class HyperoptTestCallback(TuneCallback):
+    def __init__(self, exp_name: str, model_type: str):
+        self.exp_name = exp_name
+        self.model_type = model_type
+        self.trial_ids = set()
+        self.trial_status = {}
+        self.user_config = {}
+        self.rendered_config = {}
+
+    def on_trial_start(self, iteration: int, trials: List["Trial"], trial: "Trial", **info):
```
arnavgarg1, line 144: This might just need a # noqa

PR #2642, opened 2022-10-13
Skip Horovod 0.26 installation, add packaging to requirements.txt

arnavgarg1:


requirements_distributed.txt
```diff
@@ -3,7 +3,7 @@ dask[dataframe]
 pyarrow==6.0.1 # https://github.com/ray-project/ray/issues/22310
 
 # requirements for horovod
-horovod[pytorch]>=0.24.0
+horovod[pytorch]>=0.24.0,<0.26.0
```
tgaddair, line 6: You shouldn't need to pin horovod if you have the packaging requirement.

arnavgarg1: I tried adding packaging just before the horovod install but it still caused the same error [add packaging to distributed](https://github.com/ludwig-ai/ludwig/pull/2642/commits/6c910c1e19aa2d7e7da5ad2e88ccfe61c5cede1a)

arnavgarg1: Got it! Let me add that in

arnavgarg1: Didn't actually know we could do `!=` in requirements.txt to make sure a certain version was never installed

tgaddair:
Hopefully we should be able to get v0.26.1 out today:

https://github.com/horovod/horovod/pull/3741

tgaddair (approved):

PR #2647, opened 2022-10-13
Remove Trainer `type` Param

connor-mccorm:
The frontend is no longer adding a trainer `type` before sending a validation request. This causes validation issues because without a `type` on the trainer, the schema automatically validates agains the first schema in the trainer enum which happens to be gbm which is a problem especially since ecd is the default. This PR actually removes the `type` param from the trainer schemas and subsequently from Ludwig since the trainer type is actually determined by the model type [ecd, gbm]. It also changes the trainer registry structure to link a trainer to a certain model type.

### DIFF
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -26,9 +26,9 @@ from ludwig.utils.registry import Registry
 trainer_schema_registry = Registry()
 
 
-def register_trainer_schema(name: str):
+def register_trainer_schema(model_type: str):
     def wrap(trainer_config: BaseTrainerConfig):
-        trainer_schema_registry[name] = trainer_config
+        trainer_schema_registry[model_type] = trainer_config
         return trainer_config
 
     return wrap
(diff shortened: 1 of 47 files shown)


ludwig/schema/trainer.py
```diff
@@ -102,21 +91,12 @@ class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig, ABC):
     )
 
 
-@register_trainer_schema("trainer")
+@register_trainer_schema("ecd_ray_legacy")
```
justinxzhao, line 94: There's two registrations here. Just checking -- is that intentional?

justinxzhao (approved):

ludwig/schema/trainer.py
connor-mccorm: Yes it is, thanks for checking! Basically this allows the `ray_legacy_trainer` to be accessed by the user. Since it uses the same schema as the regular trainer, I just registered it twice here.

PR #2649, opened 2022-10-14
Categorical passthrough encoder training failure fix

abidwael:
Fix for ECD training failure when we set
```
defaults:
  category:
    encoder:
      type: passthrough
```

### DIFF
--- a/ludwig/encoders/generic_encoders.py
+++ b/ludwig/encoders/generic_encoders.py
@@ -17,7 +17,7 @@ import logging
 
 import torch
 
-from ludwig.constants import BINARY, CATEGORY, NUMBER, VECTOR
+from ludwig.constants import BINARY, NUMBER, VECTOR
 from ludwig.encoders.base import Encoder
 from ludwig.encoders.registry import register_encoder
 from ludwig.modules.fully_connected_modules import FCStack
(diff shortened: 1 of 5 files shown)


ludwig/encoders/generic_encoders.py
```diff
@@ -26,7 +26,7 @@
 logger = logging.getLogger(__name__)
 
 
-@register_encoder("passthrough", [CATEGORY, NUMBER, VECTOR])
+@register_encoder("passthrough", [NUMBER, VECTOR])
```
dantreiman, line 29: I think I'm missing some context; what is the use case for this?

Categorical variables aren't continuous so I don't know why we would want a passthrough encoder for them.

dantreiman (changes requested):

ludwig/encoders/generic_encoders.py
abidwael: I don't think there's a new use case for this. We already support passthrough encoder for categorical features, but specifying it as an encoder in the config leads to an error.

justinxzhao: The passthrough encoder for category features is used by the gbm model type. CC: @jppgks

justinxzhao (approved):

ludwig/encoders/generic_encoders.py
dantreiman: Aha, that explains it.  Makes sense for decision trees.

dantreiman (approved):
LGTM

PR #2659, opened 2022-10-17
Set internal only flags

connor-mccorm:


ludwig/schema/features/preprocessing/utils.py
```diff
@@ -52,6 +52,7 @@ def _jsonschema_type_mapping():
             return {
                 "type": "object",
                 "properties": props,
+                "title": "preprocessing_options",
```
ksbrar, line 55: Honestly, we might as well add these missing titles elsewhere even if they aren't being used. What do you think about this?
* change title to `model_type` in `get_model_type_jsonschema`
* add title to `get_input/output_feature_jsonschema` that is just `input/output_features`
* add title to `get_combiner_jsonschema` that is just `combiner_options`
* add title to (global) `get_preprocessing_jsonschema` that is `global_preprocessing_options`
* after the JSON schema is unloaded in `get_defaults_jsonschema` manually set the title to `global_defaults_options`

ksbrar (approved):

martindavis (approved):

ludwig/schema/features/preprocessing/utils.py
martindavis: @ksbrar , what does appending `jsonschema` to help us achieve?

connor-mccorm: Ya good call. Better for consistency and if we ever need them.

PR #2662, opened 2022-10-17
fix: `Encountered unknown symbol 'foo'` warning in Category feature preprocessing

geoffreyangus:
This PR fixes the `Encountered unknown symbol 'foo'` warning that appears during Category feature preprocessing.

The origin of this warning seems to be related to https://github.com/dask/dask/issues/4172. My guess is that Dask passes in dummy input ("foo") into the `Series.map` function before applying changes to the Series in order to collect some metadata about the function itself. However, this step can be avoided by simply passing in the `meta` parameter with the appropriate output type.


tgaddair (approved):

arnavgarg1 (approved):

abidwael (approved):

arnavgarg1:
I think it might be ok to merge this even with the test failures in nightly

PR #2665, opened 2022-10-18
Stopgap solution for test_training_determinism

connor-mccorm:


arnavgarg1 (approved):
I think this should work since the validation/test set should have > 1 row even after the random split, which should prevent NaNs from showing up. 

I think this is a good stopgap solution, but we're going to have to also think of a long-term solution to prevent this from happening. One simple way is to add checks to enforce some minimum dataset sizes (like @dantreiman added for the local backend) post-split, but I'm wondering if there are better ways to deal with this outside of our own Ludwig tests.

justinxzhao (approved):

PR #2672, opened 2022-10-19
Documents running local tests with act

dantreiman:
Also skips cache and upload actions when running in act environement.


arnavgarg1 (approved):
Looks good to me! Thanks for putting this together.

justinxzhao (approved):

PR #2674, opened 2022-10-19
Add protos to the Ludwig project: DatasetProfile messages and Whylogs messages.

justinxzhao:
We copy the whylogs proto wholesale, as this seems preferable to adding the entire whylogs project as a submodule, and AFAIK there's not an elegant way to reference another project's proto.

I've also included adjustments to the pre-commit and other linting tooling so that protos are linted, and autogenerated `protoc` proto python code is ignored.


arnavgarg1:
@justinxzhao Is the plan to also push the pb2 files to Ludwig?

dantreiman (approved):

justinxzhao:
> @justinxzhao Is the plan to also push the pb2 files to Ludwig?

That's correct. The raw source files generated via `betterproto` look a lot cleaner, but we can't use them because `betterproto` doesn't support proto `any` types yet, which is used by the whylogs proto.

PR #2675, opened 2022-10-19
Add Kaggle test splits

abidwael:
Adds missing test splits from a number of Kaggle datasets.

### DIFF
--- a/ludwig/datasets/configs/allstate_claims_severity.yaml
+++ b/ludwig/datasets/configs/allstate_claims_severity.yaml
@@ -5,6 +5,7 @@ archive_filenames: allstate-claims-severity.zip
 sha256:
   allstate-claims-severity.zip: 165f7b4bc5ed40f43656dc958da6572143a7e126e2d37bcd41f1299bfbaa68e2
 train_filenames: train.csv
+test_filenames: test.csv
 loader: allstate_claims_severity.AllstateClaimsSeverityLoader
 description: |
   Allstate Claims Severity.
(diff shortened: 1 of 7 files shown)


dantreiman (changes requested):
I think we might want to define a different category for these "test" sets, because these are the test sets for submission to Kaggle and they don't have truth labels included.  Thus we can't use them as a "test" split in Ludwig.

Note that `train.csv` has a `loss` column (representing the monetary loss from the insurance claim), while `test.csv` does not

ludwig/datasets/configs/mercedes_benz_greener.yaml
```diff
@@ -5,6 +5,7 @@ archive_filenames: mercedes-benz-greener-manufacturing.zip
 sha256:
   mercedes-benz-greener-manufacturing.zip: 91143716085345a84dc4991b8eb1d5ff80d8aa134930de946b3b24be0f2e5d1a
 train_filenames: train.csv
+test_filenames: test.csv
```
dantreiman, line 8: Likewise, `train.csv` has a "y" column, while test.csv does not.  Perhaps, instead of including these as the "test set", we should add a new category .... `kaggle_test_filenames` perhaps?

Or maybe 'unlabeled_filenames` to be more general and less kaggle-specific?

ludwig/datasets/configs/allstate_claims_severity.yaml
```diff
@@ -5,6 +5,7 @@ archive_filenames: allstate-claims-severity.zip
 sha256:
   allstate-claims-severity.zip: 165f7b4bc5ed40f43656dc958da6572143a7e126e2d37bcd41f1299bfbaa68e2
 train_filenames: train.csv
+test_filenames: test.csv
```
dantreiman, line 8: This test.csv does not have a ground truth label.  note that `train.csv` has a "loss" column (representing the monetary loss of the insurance claim), but `test.csv` does not.  I'm concerned that people might think this `test.csv` is usable as a test split in Ludwig, but it is not.

dantreiman (changes requested):
I'm concerned that it might mislead users to call splits without a label column "test" -- since in Ludwig's process the test set is a held-out labeled set.  We should probably to introduce a new category to support unlabeled data for inference or contest submissions.

abidwael:
Good catch @dantreiman! I removed the test files from some of the other datasets that had them.

dantreiman (approved):

PR #2676, opened 2022-10-19
Fix GBM ray nightly test

jppgks:
batch_size=None to ensure map_batches doesn't change num_blocks


arnavgarg1:
@jppgks Not sure, but I'm thinking this leads to a bit of a performance hit potentially. Would be worth running the test suite again to be sure? If yes, it might be worth making this a conditional check to see if the overall dataset size is < 4096 and then set the batch_size to None, because otherwise should behave correctly.

arnavgarg1 (approved):

PR #2679, opened 2022-10-19
Remove Duplicate Schema Fields

connor-mccorm:
The JSON schema actually had duplicates fields for `type`, `name` and `column` since they were defined both at the json schema validation level within `get_input_feature_jsonschema()` and `get_output_feature_jsonschema()` and on the schema classes themselves. This creates potential issues downstream which we are avoiding by removing these duplicate fields upon conditional generations. The reason we need to keep the fields on the schema classes themselves is because they are used by the config object and consequently in Ludwig pipeline logic.


ksbrar:
@connor-mccorm Lets see if it works 😅 I think you might also need to do this for the combiner type

ksbrar (approved):
Looks like it did work!

Just wanna double-confirm, though: is this change only affecting input features? Do we need to do this at every nesting level (e.g. encoders, decoders, preprocessing, combiner) - or are those other `type` fields not duplicated?

connor-mccorm:
Ya good callout. Forgot that I added duplicate types with the config object PR. Will add now!

PR #2680, opened 2022-10-20
Speed up tune_batch_size by using synthetic batches

tgaddair:


arnavgarg1:
Looking at the tests, it seems that there's a big slowdown with our 3.8 distributed tests where it only got through 20% of the tests in the hour. I can't see why this change is causing this slowdown

tgaddair:
@arnavgarg1 I doubt the slowdown was caused by this test, rather I suspect there is a more general flakiness occurring leading to stalling.

PR #2685, opened 2022-10-20
Use fully rendered config dictionary when accessing model.config

tgaddair:


ludwig/api.py
```diff
@@ -293,16 +293,16 @@ def __init__(
             config_dict = copy.deepcopy(config)
             self.config_fp = None
 
-        self.config_dict = config_dict
+        self._user_config = config_dict
```
justinxzhao, line 296: I like this idea of more formally distinguishing the user config from fully filled out Ludwig configs with all defaults in them.

A private variable seems appropriate, though in the future it might make sense to expose a way to fetch this from the API?

justinxzhao (approved):

ludwig/api.py
tgaddair: Yeah, I agree we may wish to do that in a follow-up.

ludwig/api.py
```diff
@@ -1649,6 +1649,15 @@ def set_logging_level(logging_level: int) -> None:
         else:
             set_disable_progressbar(False)
 
+    @property
+    def config(self) -> Dict[str, Any]:
+        return self.config_obj.to_dict()
```
dantreiman, line 1654: This should have a doc string to make it clear that this returns the config with defaults, like 

"""Returns the fully-rendered config of this model including default values."""

ludwig/api.py
```diff
@@ -1649,6 +1649,15 @@ def set_logging_level(logging_level: int) -> None:
...
+
+    @config.setter
+    def config(self, user_config: Dict[str, Any]):
```
dantreiman, line 1657: Perhaps the setter should be private, or at least have a warning in the doc string explaining its use.

Setting the config of an already-constructed LudwigModel could cause unexpected behavior

dantreiman (approved):

PR #2687, opened 2022-10-20
Check for nans before testing equality in test_training_determinism

dantreiman:
If nans appear in metrics, the equality check fails `(nan == nan) == False`, but this gives an error which may be misleading.

This adds checks for nans and infinite values first, and provides an informative error message showing where the nan was encountered

### DIFF
--- a/ludwig/utils/numerical_test_utils.py
+++ b/ludwig/utils/numerical_test_utils.py
@@ -23,7 +23,7 @@ def _dict_like(x):
     """Returns true if an object is a dict or convertible to one, false if not."""
     try:
         _ = dict(x)
-    except TypeError:
+    except (TypeError, ValueError):
         return False
     return True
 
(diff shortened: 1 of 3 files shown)


abidwael:
@dantreiman I believe [`numpy.testing.assert_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html) covers the cases of scalars, lists, tuples, dictionaries or numpy arrays and it "handles NaN comparisons as if NaN was a “normal” number. That is, AssertionError is not raised if both objects have NaNs in the same positions".

dantreiman:
> The documentation of `numpy.testing.assert_equal` says that, but in practice it does fail when there are `nan`s.  It may be because they are nested inside dicts, and assert_equal is using `==` to compare the dicts.

> I want to explicitly check for `nan` or `inf` and log an error message that says "featABC.metrics.r2[0] is nan"

PR #2688, opened 2022-10-20
Dataset size checks in preprocess_for_training

dantreiman:
Check dataset sizes after preprocessing, warn if validation or test set is too small to compute metrics.

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1670,9 +1670,8 @@ def preprocess_for_training(
                 elif validation_set_size < MIN_DATASET_SPLIT_ROWS:
                     logger.warning(
                         f"Validation set too small to compute metrics. Need at least {MIN_DATASET_SPLIT_ROWS} rows, got"
-                        f" {validation_set_size} after preprocessing. Evaluation of validation set will be skipped."
+                        f" {validation_set_size} after preprocessing."
                     )
-                    validation_dataset = None
 
             test_dataset = None
             if test_set is not None:
@@ -1687,9 +1686,8 @@ def preprocess_for_training(
                 elif test_set_size < MIN_DATASET_SPLIT_ROWS:
                     logger.warning(
                         f"Test set too small to compute metrics. Need at least {MIN_DATASET_SPLIT_ROWS} rows, got"
-                        f" {test_set_size} after preprocessing. Evaluation of test set will be skipped."
+                        f" {test_set_size} after preprocessing."
                     )
-                    test_dataset = None
 
         return (training_dataset, validation_dataset, test_dataset, training_set_metadata)
 
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -38,6 +38,7 @@ from ludwig.constants import (
     FILL_WITH_MEAN,
     FILL_WITH_MODE,
     FULL,
+    MIN_DATASET_SPLIT_ROWS,
     NAME,
     NUMBER,
     PREPROCESSING,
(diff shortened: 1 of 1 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1647,32 +1648,45 @@ def preprocess_for_training(
         with backend.storage.cache.use_credentials() if cached else contextlib.nullcontext():
             logger.debug("create training dataset")
             training_dataset = backend.dataset_manager.create(training_set, config, training_set_metadata)
-            if not len(training_set):
+            training_set_size = len(training_dataset)
+            if training_set_size == 0:
                 raise ValueError("Training data is empty following preprocessing.")
+            elif training_set_size < MIN_DATASET_SPLIT_ROWS:
+                raise ValueError(f"Training dataset has only {training_set_size} rows following preprocessing.")
```
arnavgarg1, line 1655: Should we add that it needs a minimum of X rows as a part of the error message?

ludwig/data/preprocessing.py
```diff
@@ -1647,32 +1648,45 @@ def preprocess_for_training(
...
                     logger.warning(
-                        "Encountered empty validation dataset. If this is unintentional, please check the "
-                        "preprocessing configuration."
+                        "Validation set empty. If this is unintentional, please check the preprocessing configuration."
+                    )
+                    validation_dataset = None
+                elif validation_set_size < MIN_DATASET_SPLIT_ROWS:
+                    logger.warning(
+                        f"Validation set ({validation_set_size}) too small to compute metrics. Evaluation of validation"
+                        " set will be skipped"
```
arnavgarg1, line 1670: Would it help to tell the user the number of rows it has and the number of rows it should have to not skip validation?

ludwig/data/preprocessing.py
```diff
@@ -1647,32 +1648,45 @@ def preprocess_for_training(
...
-                        "Encountered empty test dataset. If this is unintentional, please check the "
-                        "preprocessing configuration."
+                        "Test set empty. If this is unintentional, please check the preprocessing configuration."
                     )
                     test_dataset = None
+                elif test_set_size < MIN_DATASET_SPLIT_ROWS:
+                    logger.warning(
+                        f"Test set ({test_set_size}) too small to compute metrics. Evaluation of test set will be"
+                        " skipped"
+                    )
```
arnavgarg1, line 1688: Same as above

ludwig/data/preprocessing.py
dantreiman: Good call, thanks!

ludwig/data/preprocessing.py
```diff
@@ -1647,32 +1648,48 @@ def preprocess_for_training(
...
-                        "preprocessing configuration."
+                        "Test set empty. If this is unintentional, please check the preprocessing configuration."
                     )
                     test_dataset = None
+                elif test_set_size < MIN_DATASET_SPLIT_ROWS:
+                    logger.warning(
+                        f"Test set too small to compute metrics. Need at least {MIN_DATASET_SPLIT_ROWS} rows, got"
+                        f" {test_set_size} after preprocessing. Evaluation of test set will be skipped."
+                    )
+                    validation_dataset = None
```
arnavgarg1, line 1692: Typo: This should probably be `test_dataset = None`?

dantreiman: Copy/paste error, thanks for reading carefully!

arnavgarg1 (approved):
Looks good to me - would just wait for the tests to run since I have a feeling this may result in a few test failures where the number of examples being used is 25 instead of maybe like >= 50.

dantreiman:
> Looks good to me - would just wait for the tests to run since I have a feeling this may result in a few test failures where the number of examples being used is 25 instead of maybe like >= 50.

You are right; test_preprocessing.py fails for exactly that reason.  Maybe it should only warn for small validation/test set but not set them to None as if they were empty?

PR #2692, opened 2022-10-21
Fix validator for reduce_learning_rate_on_plateau

carlogrisetti:
validator was set up incorrectly and effectively blocked usage of such parameter.
This reverts the functionality as it was before.

### DIFF
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -177,13 +177,11 @@ class ECDTrainerConfig(BaseTrainerConfig):
         parameter_metadata=TRAINER_METADATA["should_shuffle"],
     )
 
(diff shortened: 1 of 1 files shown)


ludwig/schema/trainer.py
```diff
@@ -177,13 +177,11 @@ class ECDTrainerConfig(BaseTrainerConfig):
         parameter_metadata=TRAINER_METADATA["should_shuffle"],
     )
 
-    reduce_learning_rate_on_plateau: float = schema_utils.FloatRange(
-        default=0.0,
-        min=0,
-        max=1,
+    reduce_learning_rate_on_plateau: int = schema_utils.NonNegativeInteger(
```
justinxzhao, line 180: Great catch! Thanks for the fix.

justinxzhao (approved):

PR #2695, opened 2022-10-22
Fix InputDataError to be serializeable

tgaddair:
This results in Ray being unable to surface the error to the user when sending it back to the driver from the worker processes.

See https://stackoverflow.com/questions/49715881/how-to-pickle-inherited-exceptions


dantreiman (commented):
Retroactively LGTM

PR #2701, opened 2022-10-24
Adds DeveloperAPI and PublicAPI annotations to AutoML

dantreiman:
Also adds `dataset_utils.get_repeatable_train_val_test_split` as public API, since it is referenced in the AutoML user guide.  https://ludwig.ai/latest/user_guide/automl/

### DIFF
--- a/ludwig/automl/automl.py
+++ b/ludwig/automl/automl.py
@@ -19,7 +19,7 @@ import pandas as pd
 import yaml
 
 from ludwig.api import LudwigModel
-from ludwig.api_annotations import DeveloperAPI
+from ludwig.api_annotations import PublicAPI
 from ludwig.automl.auto_tune_config import memory_tune_config
 from ludwig.automl.base_config import _create_default_config, _get_reference_configs, DatasetInfo, get_dataset_info
 from ludwig.backend import Backend, initialize_backend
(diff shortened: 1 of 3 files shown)


ludwig/automl/automl.py
```diff
@@ -142,6 +143,7 @@ def auto_train(
     return train_with_config(dataset, config, output_directory=output_directory, random_seed=random_seed, **kwargs)
 
 
+@DeveloperAPI
```
arnavgarg1, line 146: Do we feel like these should potentially be PublicAPI?

dantreiman: You are right, also `auto_train` should be PublicAPI as well.  They are referenced in the AutoML user guide:  https://ludwig.ai/latest/user_guide/automl/

ludwig/utils/dataset_utils.py
```diff
@@ -3,11 +3,13 @@
 import pandas as pd
 from sklearn.model_selection import train_test_split
 
+from ludwig.api_annotations import PublicAPI
 from ludwig.constants import TEST_SPLIT, TRAIN_SPLIT, VALIDATION_SPLIT
 from ludwig.data.dataset.base import Dataset
 from ludwig.utils.defaults import default_random_seed
 
 
+@PublicAPI
```
arnavgarg1, line 12: @dantreiman Should this be DeveloperAPI since we don't expose this directly?

dantreiman: I made it @PublicAPI because it is currently referenced in the Ludwig AutoML user guide here:

Screen Shot 2022-10-25 at 1 55 13 PM

arnavgarg1: Got it!

arnavgarg1 (approved):

PR #2708, opened 2022-10-25
Add H&M fashion recommendation dataset

jppgks:


ludwig/datasets/loaders/dataset_loader.py
```diff
@@ -279,7 +279,7 @@ def load(self, split=False, kaggle_username=None, kaggle_key=None) -> pd.DataFra
         :param split: (bool) splits dataset along 'split' column if present. The split column should always have values
         0: train, 1: validation, 2: test.
         """
-        self._download_and_process()
+        self._download_and_process(kaggle_username=kaggle_username, kaggle_key=kaggle_key)
```
dantreiman, line 282: Very important fix!  Thanks for noticing this

ludwig/datasets/kaggle.py
```diff
@@ -40,8 +41,12 @@ def download_kaggle_dataset(
         api = create_kaggle_client()
         api.authenticate()
     with upload_output_directory(download_directory) as (tmpdir, _):
-        if kaggle_competition:
-            api.competition_download_files(kaggle_competition, path=tmpdir)
+        dataset_or_competition = kaggle_competition or kaggle_dataset_id
+        if filenames:
+            download_fn = api.competition_download_file if kaggle_competition else api.dataset_download_file
```
dantreiman, line 46: This causes test_titanic_workflow to fail because it is mocking `competition_download_files`, but now we're calling `competition_download_file` instead.

See `test_titanic_workflow.py:88`:
`mock_kaggle_api.competition_download_files = download_files`

ludwig/datasets/loaders/hm_fashion_recommendations.py
```diff
@@ -0,0 +1,128 @@
...
+    The split is done in a chronological manner based on the year_month column. The split is done by customer_id,
+    so that interactions for a given customer are present in all splits.
+
+    Params:
+        df: The dataframe to split.
+
+    Returns:
+        A tuple of (train_df, validation_df, test_df).
+    """
+    backend = Mock()
```
dantreiman, line 63: Mocks shouldn't be used outside of test code.  Could this be `LOCAL_BACKEND` instead?

ludwig/datasets/loaders/utils.py
```diff
@@ -0,0 +1,106 @@
...
+    npos = interaction_row.shape[1] - len(neg_items)
+    samples_required = npos * neg_pos_ratio + extra_samples
+    should_sample = samples_required <= available_samples
+
+    neg_items = np.random.choice(neg_items, samples_required, replace=False) if should_sample else neg_items
+
+    return neg_items.tolist(), max(0, samples_required - available_samples)
+
+
+def negative_sample(
```
dantreiman, line 39: I think we're exceeding the scope of the datasets API by performing negative sampling here.  The datasets API intent is to return standard benckmark datasets 'as-is' (or as close to the original as we can get).

IMO the sampling implementation belongs in `ludwig.data`, and should be performed as a separate phase after the dataset is loaded (also sampling has hyperparameters `neg_pos_ratio` and `log_pct`).

Is this possible, or would it be too inefficient to merge article and customer features prior to negative sampling?

ludwig/datasets/loaders/hm_fashion_recommendations.py
```diff
@@ -0,0 +1,128 @@
...
+
+        train_dfs.append(train_df)
+        val_dfs.append(val_df)
+        test_dfs.append(test_df)
+
+    return pd.concat(train_dfs), pd.concat(val_dfs), pd.concat(test_dfs)
+
+
+class HMLoader(DatasetLoader):
+    def load_unprocessed_dataframe(self, file_paths: List[str], sample=True) -> pd.DataFrame:
```
dantreiman, line 84: Need doc string for sample, not obvious what it does or why its default is True.  It looks like it sub-samples the dataset and chooses 100 unique customers?

dantreiman (changes requested):

jppgks:
I extracted the negative sampling into a [separate PR](https://github.com/ludwig-ai/ludwig/pull/2711)

ludwig/datasets/loaders/utils.py
jppgks: I agree - removed the negative sampling from the loader and moved it to `ludwig.data` in [this PR](https://github.com/ludwig-ai/ludwig/pull/2711)

dantreiman (approved):

PR #2709, opened 2022-10-26
Support Distributed Training And Ray Tune with Ray 2.1

arnavgarg1:
This PR migrates Ludwig Trainer from Ray 1.x/Ray 2.0 to Ray 2.1 APIs. Here's a list of notable changes:

- Trainer Migration: We’ve migrated from the [Trainer](https://github.com/ray-project/ray/blob/releases/1.13.0/python/ray/train/trainer.py#L100) class in Ray 1.x to the [HorovodTrainer](https://docs.ray.io/en/releases-2.1.0/train/api.html?highlight=HorovodTrainer#ray.train.horovod.HorovodTrainer) class in Ray 2.0/2.1 which sets up workers with the necessary config/libraries required for distributed training with Horovod.
    - We deprecated the use of environment variables to decide whether to use spread vs pack scheduling for training. This is now handled by [ScalingConfig](https://docs.ray.io/en/latest/ray-air/package-ref.html#ray.air.config.ScalingConfig).
    -  This PR also introduces a new function, `get_window_size_bytes`, that determines whether to create an infinite pipeline or a windowed pipeline.
    - We’ve migrated to the [session](https://docs.ray.io/en/latest/tune/api_docs/trainable.html#session-function-api) API within Ray for reporting results during regular model training since results don’t get returned from the Trainer.fit() call
- Hyperopt:
    - New Resource Type: “hyperopt_resources” has been introduced as a new custom resource for Hyperopt. 
    - Ray Cluster Config: Users who use Ludwig+Ray+Kubernetes need to now specify “hyperopt_resources” as a custom resource upon `ray start`. 
- We no longer support training using the `RayLegacyTrainer` class
- All callbacks passed into functions in Ray 2.1 are subclasses of `ray.tune.callbacks`, so we’ve migrated the TQDM callback to subclass `ray.tune.callbacks.Callbacks`

Co-authored-by: Geoffrey Angus <geoffrey@predibase.com>


ludwig/backend/ray.py
```diff
@@ -187,25 +167,30 @@ def train_fn(
...
         )
 
         try:
-            val_shard = rt.get_dataset_shard("val")
+            val_shard = session.get_dataset_shard("val")
         except KeyError:
             val_shard = None
 
+        eval_data_loader_kwargs = copy.deepcopy(data_loader_kwargs)
+        eval_data_loader_kwargs["shuffle"] = False
```
arnavgarg1, line 182: Don't make this assumption, but pass this in first and overwrite with data_loader_kwargs so that incase it isn't specified, only in those scenarios we fall back to `shuffle = False`

ludwig/backend/ray.py
```diff
@@ -303,15 +282,65 @@ def tune_learning_rate_fn(
...
         super().__init__()
         self.progess_bars = {}
 
+    def on_trial_result(iteration: int, trials: List["Trial"], trial: "Trial", result: Dict, **info):
+        print("On_Trial_Result")
+        print(f">>> iteration: {iteration}")
+        print(f">>> trials: {trials}")
+        print(f">>> trial: {trial}")
+        print(f">>> result: {result}")
+        print(f">>> info: {info}")
```
arnavgarg1, line 341: Seems like this is called everytime `session.report()` is called in subprocesses.

ludwig/backend/ray.py
```diff
@@ -366,13 +395,45 @@ def spread_env(use_gpu: bool = False, num_workers: int = 1, **kwargs):
...
+                "val": DatasetConfig(split=True),
+                "test": DatasetConfig(split=True),
+            },
+            # run_config=RunConfig(
+            #     callbacks=callbacks,
+            #     verbose=2,
+            # ),
+            # resume_from_checkpoint=None,
+        )
+        # tuner = ray.tune.Tuner(trainer)
```
arnavgarg1, line 435: Only way to pass in callbacks is through the RunConfig, but that internally starts saying that the trainer needs to be wrapped in the `ray.tune.Tuner()` call, likely because callbacks are shared between the trainer and ray tune in Ray 2.0. Once I do this though, I keep seeing this error: 

```
ray.tune.error.TuneError: Tune run failed. Please use tuner = Tuner.restore("/home/vscode/ray_results/HorovodTrainer_2022-10-26_16-33-44") to resume.
```

ludwig/data/dataset/ray.py
```diff
@@ -174,14 +137,46 @@ def data_format(self):
...
-        self.epoch_iter = dataset_shard.iter_epochs()
+        self.epoch_iter = self.pipeline(**data_loader_kwargs)
+
+    def pipeline(
+        self,
+        shuffle: bool = True,
+        fully_executed: bool = True,
+        window_size_bytes: Optional[int] = None,
+        shuffle_seed: int = default_random_seed,
+    ) -> DatasetPipeline:
```
arnavgarg1, line 156: Had to move this into RayDatasetShard so that we can still keep granular control over how we want too set the DatasetPipeline objects for train/test/val. The alternative was to do this from the DatasetConfig, but that would require always enabling windowing, which likely will cause performance degradation in terms of model quality.

My suspicion for the increase in 2 CPUs is from moving this code block from RayDataset to RayDatasetShard, since we're now working on the split dataset, but need to validate and understand why.

ludwig/backend/ray.py
arnavgarg1: Have a fix that's working locally, will push in a moment

ludwig/backend/ray.py
```diff
@@ -366,13 +325,44 @@ def spread_env(use_gpu: bool = False, num_workers: int = 1, **kwargs):
...
+        trainer_kwargs.pop("backend", None)
+
+        # When training on GPU, you want to pack workers together to limit network latency during
+        # allreduce. Conversely, for CPU training you want to spread out the workers to limit
+        # CPU and memory contention.
+        strategy = "PACK" if trainer_kwargs.get("use_gpu") else "SPREAD"
+        # Set _max_cpu_fraction_per_node to 80% to prevent contention with Ray dataset tasks
+        # when passing in Datasets into Trainers.
+        self.scaling_config = ScalingConfig(
+            placement_strategy=strategy, _max_cpu_fraction_per_node=0.8, **trainer_kwargs
```
tgaddair, line 343: I don't think we should set `_max_cpu_fraction_per_node`.

ludwig/backend/ray.py
```diff
@@ -366,13 +325,44 @@ def spread_env(use_gpu: bool = False, num_workers: int = 1, **kwargs):
 def create_runner(**kwargs):
     trainer_kwargs = get_trainer_kwargs(**kwargs)
     with spread_env(**trainer_kwargs):
-        trainer = Trainer(**trainer_kwargs)
+        yield RayAirRunner(trainer_kwargs)
```
tgaddair, line 328: No longer need `spread_env` with the ScalingConfig.

tests/integration_tests/test_postprocessing.py
```diff
@@ -59,7 +59,7 @@ def random_set_logits(*args, num_predict_samples, vocab_size, pct_positive, **kw
     ],
 )
 @pytest.mark.parametrize("distinct_values", [(False, True), ("No", "Yes")])
-def test_binary_predictions(tmpdir, backend, distinct_values, ray_cluster_2cpu):
+def test_binary_predictions(tmpdir, backend, distinct_values, ray_cluster_4cpu):
```
tgaddair, line 62: I would prefer to keep it to 2 cpu, as it should work. If it's not, we should figure out what's causing the spike in resource demands.

arnavgarg1: Agreed, this was just a temporary change to make sure all the tests work with the new trainer

ludwig/data/dataset/ray.py
```diff
@@ -34,12 +34,10 @@
 from ludwig.data.batcher.base import Batcher
 from ludwig.data.dataset.base import Dataset, DatasetManager
 from ludwig.utils.data_utils import DATA_TRAIN_HDF5_FP, DATA_TRAIN_PARQUET_FP
-from ludwig.utils.defaults import default_random_seed
 from ludwig.utils.fs_utils import get_fs_and_path
 from ludwig.utils.misc_utils import get_proc_features
 from ludwig.utils.types import DataFrame, Series
 
-_ray113 = version.parse(ray.__version__) == version.parse("1.13.0")
 _ray_nightly = version.parse(ray.__version__) > version.parse("1.13")
```
arnavgarg1, line 41: Fix ray flag, this is incorrect

ludwig/hyperopt/execution.py
```diff
@@ -776,23 +767,19 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
...
-            # but we do want to enable it when GPUs are specified
             resources_per_trial = PlacementGroupFactory(
                 [{}] + ([{"CPU": 0, "GPU": 1}] * self._gpu_resources_per_trial_non_none)
                 if self._gpu_resources_per_trial_non_none
-                else [{}] + [{"CPU": self._cpu_resources_per_trial_non_none}]
+                else [{}] + [{"CPU": self._cpu_resources_per_trial_non_none}],
+                # When training on GPU, you want to pack workers together to limit network latency during
+                # allreduce. Conversely, for CPU training you want to spread out the workers to limit
+                # CPU and memory contention and avoid too many workers on a single machine.
+                strategy="SPREAD" if self._gpu_resources_per_trial_non_none else "PACK",
```
arnavgarg1, line 777: @arnavgarg1 to change this to be PACK for GPU and SPREAD for CPU

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -225,7 +225,7 @@ def test_hyperopt_executor_with_metric(use_split, csv_filename, tmpdir, ray_clus
 
 
 @pytest.mark.distributed
-@pytest.mark.parametrize("backend", ["local", "ray"])
+@pytest.mark.parametrize("backend", ["local"])  # "ray"
```
dantreiman, line 228: Is this meant to be committed?  If so, need comment describing reason for commenting out ray backend version of this test.

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -1,306 +1,306 @@
-# Copyright (c) 2019 Uber Technologies, Inc.
```
dantreiman, line 1: I can't approve merging this file commented out without explanation or documentation.  What would we need to do to un-comment this?  Is it fixed in another PR?

dantreiman (changes requested):

tests/integration_tests/test_hyperopt_ray.py
arnavgarg1: Yes, it is meant to be committed. It will be reverted when some more fixes are in place.

tests/integration_tests/test_hyperopt_ray_horovod.py
arnavgarg1: I'll be uncommenting this out once I have a fix in place for this file. This was just commented out to make sure the tests pass.

tests/integration_tests/test_hyperopt.py
```diff
@@ -123,30 +118,9 @@ def _setup_ludwig_config(dataset_fp: str) -> Tuple[Dict, str]:
...
-        ray.shutdown()
-
-
-@pytest.fixture(scope="module")
-def ray_cluster():
-    gpus = [i for i in range(torch.cuda.device_count())]
-    with ray_start(num_gpus=len(gpus)):
-        yield
-
-
```
geoffreyangus, line 146: Switched to using the fixtures in conftest.py to ensure that the ray cluster was started with `hyperopt_resources: 1000`.

tests/integration_tests/test_hyperopt_ray_horovod.py
```diff
@@ -106,7 +106,7 @@ def mock_storage_client(path):
 
 
 # TODO ray: replace legacy mode when Ray Train supports placement groups
-RAY_BACKEND_KWARGS = {"processor": {"parallelism": 4}}
+RAY_BACKEND_KWARGS = {"processor": {"parallelism": 1}}
```
geoffreyangus, line 109: We reduce `parallelism` in this workaround to ensure that Ray Datasets doesn't reserve all available CPUs ahead of the other trials being scheduled.

Before this change, all CPUs for the `train_fn` of each trial were scheduled up front by the Tuner. In this change, only the dummy `hyperopt_resources` are scheduled by the Tuner. The inner Tuners then schedule CPUs ad-hoc as they are called and executed by each trial.

The danger with this is in its interaction with Ray Datasets, which grabs resources opportunistically. If an inner Tuner is scheduled and its Ray Datasets tasks grab the remaining CPUs, other trials may be prevented from starting, causing the test to double in duration (since some trials are executed in sequence instead of all at once).

Setting `parallelism` to 1 here ensures that the number of CPUs requested by Ray Datasets is limited to 1 per trial.

tests/integration_tests/test_automl.py
```diff
@@ -232,8 +228,7 @@ def test_train_with_config_remote(fs_protocol, bucket, test_data_tabular_large,
     }
 
     with remote_tmpdir(fs_protocol, bucket) as tmpdir:
-        with pytest.raises(ValueError) if not _ray200 else contextlib.nullcontext():
```
geoffreyangus, line 235: @ShreyaR I removed this context manager because we are moving to support only Ray>=2.0. Wanted to call this out in case you think there are issues with this.

ludwig/data/dataset/ray.py
```diff
@@ -250,7 +205,15 @@ def __init__(
...
-        self.epoch_iter = dataset_shard.iter_epochs()
+        self.create_epoch_iter()
+
+    def create_epoch_iter(self) -> None:
+        if isinstance(self.dataset_shard, DatasetPipeline):
+            self.epoch_iter = self.dataset_shard.iter_epochs()
+        else:
+            # Dataset shard is a Ray Dataset object
+            # Convert Ray Dataset to a DatasetPipeline object before enabling epoch iteration
+            self.epoch_iter = self.dataset_shard.repeat().iter_epochs()
```
tgaddair, line 216: How does this work with shuffling / windowing from earlier in the pipeline? If we're repeating over the shard, seems we will never perform any shuffling. Am I missing something?

ludwig/hyperopt/execution.py
```diff
@@ -506,7 +504,7 @@ def _checkpoint_progress(self, trainer, progress_tracker, save_path) -> None:
                 """Checkpoints the progress tracker."""
                 if is_using_ray_backend:
                     trainer_ckpt = Checkpoint.from_directory(save_path)
-                    ckpt_ref = trainer_ckpt.to_object_ref()
+                    ckpt_ref = ray.put(trainer_ckpt.to_dict(), _owner=ray_queue.actor)
```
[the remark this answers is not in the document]
tgaddair: Or is this needed due to some reference counting issue on the Ray side (it wasn't an issue before, was it a regression?)?

Would be useful to add a comment explaining and linking to relevant Ray issue if its exists.

ludwig/hyperopt/execution.py
```diff
@@ -806,44 +787,73 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
...
-                stop=CallbackStopper(callbacks),
-                verbose=hyperopt_log_verbosity,
-                resume=should_resume,
-                log_to_file=True,
-            )
+            tuner_cls = TunerRay210 if not _ray220 else Tuner
+
+            if resume:
+                # Restore from experiment directory
+                tuner = tuner_cls.restore(
```
tgaddair, line 813: Does this work with our custom syncer?

ludwig/hyperopt/execution.py
```diff
@@ -806,44 +787,73 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
...
+                        **tune_config,
+                    },
+                    tune_config=TuneConfig(
+                        metric=metric,
+                        mode=mode,
+                        search_alg=search_alg,
+                        scheduler=self.scheduler,
+                        num_samples=self.num_samples,
+                        time_budget_s=self.time_budget_s,
+                        reuse_actors=True,
```
tgaddair, line 830: What are the implications of `reuse_actors` here? For GPU training, this can result in GPU memory leaks if we're not careful (https://github.com/ray-project/ray/issues/29624).

One thing we may want to do it set this equal to `not use_gpu`.

ludwig/hyperopt/utils.py
```diff
@@ -207,7 +207,13 @@ def update_hyperopt_params_with_defaults(hyperopt_params: HyperoptConfigDict) ->
     set_default_value(hyperopt_params, METRIC, LOSS)
     set_default_value(hyperopt_params, GOAL, MINIMIZE)
 
-    set_default_values(hyperopt_params[EXECUTOR], {TYPE: RAY, NUM_SAMPLES: 1, MAX_CONCURRENT_TRIALS: AUTO})
+    set_default_values(
+        hyperopt_params[EXECUTOR],
+        {TYPE: RAY, NUM_SAMPLES: 1, MAX_CONCURRENT_TRIALS: AUTO},
+    )
+
+    if hyperopt_params[EXECUTOR].get("trial_function_resources") is None:
```
tgaddair, line 215: Is this the same as `set_default_values(hyperopt_params[EXECUTOR]["trial_function_resources"], {"CPU": 1, "GPU": 0})`?

ludwig/backend/ray.py
```diff
@@ -244,11 +199,23 @@ def train_fn(
 
         torch.cuda.empty_cache()
 
-        train_results = results, trainer.validation_field, trainer.validation_metric
-
+        # Passing objects containing Torch tensors as metrics is not supported as it will throw an
```
tgaddair, line 202: You should be able to get around this by putting torch tensors into the object store with `ray.put`, and if necessary converting to a numpy array. I am not sure how the checkpoint API here works. Is it writing to object store, or to disk somehow? If the latter, then how does it handle copies across nodes? I think this merits some additional comments / links to documentation.

tgaddair (commented):
Overall fantastic work! Just a few questions.

ludwig/data/dataset/ray.py
arnavgarg1: You are right! 

What I should've called out in the comments here is that `self.dataset_shard` will only be a `ray.data.Dataset` object when we're doing auto batch size tuning or learning rate tuning, in which case we don't actually need to shuffle or window. In all other cases (regular training, prediction etc), it will be a DatasetPipeline object which will allow windowing/shuffles, etc through the DatasetConfig.

Will call this out in a better comment.

tests/integration_tests/test_hyperopt_ray_horovod.py
geoffreyangus: Right– so what we're seeing is not hanging behavior– instead what we are seeing is that the first `n` trials scheduled have Ray Datasets tasks that then take up the resources in the cluster. Eventually they let the resources go, at which point the `n+1`th trial begins executing. This is typically fine; however, for the sake of unit tests, this causes the test to take much longer, due to the fact that all `n+1` trials cannot execute exactly in parallel.

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
geoffreyangus: This is needed because there are actually three different processes operating for a given trial– (1) the main process, which is what calls `tuner.fit()`, (2) the trial function process, which spawns the `ray_queue` object, and (3) the remote task `HorovodTrainer` creates to ultimately call our function `ludwig.backend.ray.train_fn`. The original checkpoint object `trainer_ckpt` is produced/owned by process #3, but we are trying to grab it from process #2 with `ray_queue.get_nowait_batch` here: https://github.com/ludwig-ai/ludwig/blob/402511f2db6eb2737ed1ca538533abfa85b3485e/ludwig/hyperopt/execution.py#L617

I believe this is failing because process #3 is terminated before we can get the object reference, causing an Owner Died error on the now orphaned object reference. The solution to this was using the `ray.put` call above with the `_owner` kwarg, which specifies `ray_queue.actor` as the owner of the checkpoint object reference. By doing this, we ensure the checkpoint's lifecycle is tied to `ray_queue`, which itself is tied to process #2.

I'm not entirely clear with the behavior prior to this change– perhaps process #2 and #3 used to be the same process, therefore avoiding this sort of error– @arnavgarg1 may be able to provide some context. In any case, I'll add a clarification comment.

ludwig/hyperopt/execution.py
```diff
@@ -806,44 +787,73 @@ def run_experiment_trial(config, local_hyperopt_dict, checkpoint_dir=None):
...
         run_experiment_trial_params = tune.with_parameters(run_experiment_trial, local_hyperopt_dict=hyperopt_dict)
 
+        if _is_ray_backend(backend):
+            # If Ray backend, only request custom resource at trial level (inner Tuner will request resources)
+            resources = [self.trial_function_resources]
+        else:
+            # If not Ray backend, request all of the resources required at the trial level
+            use_gpu = bool(self._gpu_resources_per_trial_non_none)
+            num_cpus, num_gpus = _get_num_cpus_gpus(use_gpu)
+            resources = [self.trial_function_resources, {"CPU": num_cpus, "GPU": num_gpus}]
```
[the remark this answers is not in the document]
geoffreyangus: If we have a single bundle, then we can only schedule the trials on nodes with all three resources. Here, we ensure that the trial driver is scheduled on the head node, and that the worker can be scheduled wherever it needs to be.

ludwig/schema/hyperopt/executor.py
```diff
@@ -29,6 +29,15 @@ class ExecutorConfig(schema_utils.BaseMarshmallowConfig):
         default=3600, description="The number of seconds for the entire hyperopt run."
     )
 
+    trial_function_resources: Dict[str, float] = schema_utils.Dict(
+        default=None,
+        description=(
+            "The resources reserved by each trial function. This differs from cpu_resources_per_trial and "
```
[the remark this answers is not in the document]
geoffreyangus: Hm, I think that trial function is close to what is correct, because technically, the trial driver is what calls `tuner.fit`: https://docs.ray.io/en/latest/tune/tutorials/tune-lifecycle.html#what-happens-in-tuner-fit

The trial function resources are for the process between the trial driver and the Ray trainer it spawns. IIUC, there are three processes, each spawning the next:
1. the trial driver, which calls tuner.fit(). Each trial is then spawned in the form of
2. the trial function process, which is `run_experiment` and manages the `ray_queue` for the experiment. Within `run_experiment`, the trainer then spawns
3. `HorovodTrainer` process, which actually executes `ludwig.backend.ray.train_fn`. This process reserves its own placement groups.

This param gives resources to process #2.

ludwig/hyperopt/execution.py
geoffreyangus: Hm, good question– @arnavgarg1 do we have existing tests for this? I didn't test it explicitly myself.

ludwig/hyperopt/execution.py
geoffreyangus: @arnavgarg1 do you have context here?

ludwig/hyperopt/utils.py
geoffreyangus: I believe this is different because we want to wipe out all other keys with our custom dictionary. I think when I tried the above, it just added the `CPU` and `GPU` keys alongside the `hyperopt_resources` key.

ludwig/backend/ray.py
geoffreyangus: Good question. What we have here is roughly how Antoni recommended we do this– I can ask for clarification on how `session.report` actually handles the artifact. My guess is that it does copies, but not sure how that would work in a multi-node setting.

tests/integration_tests/test_hyperopt_ray_horovod.py
geoffreyangus: Prior to this change, the trial driver would reserve the resources for all `n+1` trials at the onset. However, since there is this trial-within-trial scenario, only the top-level resources for the trial function (i.e. `hyperopt_resources`) are reserved at the onset. The `HorovodTrainer` resources (i.e. CPU/GPU) are reserved as the trial functions each hit that point in the code path.

ludwig/backend/ray.py
geoffreyangus: According to Antoni:
> The checkpoint is put in the object store and then retrieved by the Trainable actor to be reported to Tune. It is also persisted on disk by the Trainable (and synced to cloud, if configured to do so)

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
justinxzhao: @geoffreyangus Seems worth adding this as a comment in the code.

tests/integration_tests/test_hyperopt_ray_horovod.py
justinxzhao: This also seems like a valuable comment to add to the code!

justinxzhao (approved):
LGTM. Just a few comments, questions, and suggestions.

How confident do we feel about landing this as-is?

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
tgaddair: @geoffreyangus for local execution mode, the trial driver is the worker, so I think you would just want the one bundle.

ludwig/backend/ray.py
```diff
@@ -904,24 +811,18 @@ def initialize_pytorch(self, **kwargs):
...
-            return trainer_cls(**all_kwargs)
-        else:
-            if model.name == MODEL_GBM:
-                raise RuntimeError("Legacy trainer not supported for GBM models.")
 
-            # TODO: deprecated 0.5
-            return RayLegacyTrainer(self._horovod_kwargs, executable_kwargs)
+        trainer_cls = get_from_registry(model.type(), ray_trainers_registry)
+
+        # Deep copy to workaround https://github.com/ray-project/ray/issues/24139
```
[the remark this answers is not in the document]
geoffreyangus: Good point, will remove from the PR. The fix for the issue added a [Ray unit test](https://github.com/ray-project/ray/pull/24141/files#diff-8a074c7df8d7f43f696fbb5e68010130476206262cab66671273a21f3323754aR1168) for this specific bug, so we probably don't need to write our own.

ludwig/hyperopt/execution.py
arnavgarg1: @tgaddair Good callout, yeah I think we might want to set this to `not use_gpu` to prevent memory leaks. I didn't realize that GPU memory isn't cleaned up on successive runs. Looks like this is fixed in Ray 2.2 so we can probably enable this again once we bump Ludwig to 2.2. Will add a TODO comment

ludwig/backend/ray.py
```diff
@@ -675,7 +579,7 @@ def __init__(
     ):
         self.batch_size = predictor_kwargs["batch_size"]
         self.trainer_kwargs = trainer_kwargs
-        self.data_loader_kwargs = data_loader_kwargs
+        self.data_loader_kwargs = data_loader_kwargs or {}
```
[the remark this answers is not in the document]
geoffreyangus: @arnavgarg1 could you take a look at this?

ludwig/data/dataset/ray.py
```diff
@@ -250,7 +205,15 @@ def __init__(
         self.dataset_shard = dataset_shard
         self.features = features
         self.training_set_metadata = training_set_metadata
-        self.epoch_iter = dataset_shard.iter_epochs()
+        self.create_epoch_iter()
+
+    def create_epoch_iter(self) -> None:
```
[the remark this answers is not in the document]
geoffreyangus: I think @arnavgarg1 has more context here, so will leave it to him to follow-up on this 🙂

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
geoffreyangus: Right that makes sense– I've added a couple of comments to this code block that explains the contents of the `resources` list. Copied here for convenience:

```
if _is_ray_backend(backend):
    # NOTE(geoffrey): as of PR #2709, when we are using a Ray backend, there are two main processes
    # spawned during a Ray Tune trial.
    #
    # The first process is the `trial_function`, which is the function that is passed to `tune.run`.
    # The `tuner` object allocates resources for this function using `self.trial_function_resources` below.
    # Note that `trial_function` is a very lightweight process whose main purpose is to spawn the `trainer`
    # process and communicate with the `tuner` object.
    #
    # The second process is the actual `trainer`, which is spawned by `trial_function`. The `tuner` object
    # does not know about this process, so we cannot allocate resources for it ahead of time. Instead, the
    # `trainer` will reserve its own resources downstream based on its config.
    #
    # In summary, we should only request resources for `trial_function` at this level.
    # `trainer` will request its own resources.
    resources = [self.trial_function_resources]
else:
    # If we are NOT using the Ray backend, the `trial_function` and the `trainer` are executed in the same
    # process. Since the `trial_function` doesn't really require its own resources, we can just request
    # resources for the `trainer` here.
    use_gpu = bool(self._gpu_resources_per_trial_non_none)
    num_cpus, num_gpus = _get_num_cpus_gpus(use_gpu)
    resources = [{"CPU": num_cpus, "GPU": num_gpus}]
```

ludwig/backend/ray.py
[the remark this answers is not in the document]
arnavgarg1: I think this is only because of the way some of our tests are written, but I don't think this is actually needed otherwise. Will remove

ludwig/hyperopt/execution.py
arnavgarg1: Will test this today, but I believe it should work

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
geoffreyangus: For local execution mode, I initially requested resources for `self.trial_function_resources` in addition to the resources required for the trainer to ensure consistent behavior between the two execution modes. 

In the latest change, I've removed `self.trial_function_resources` in the above code because you are right, it isn't explicitly necessary to request those additional resources in local execution mode. That said, we can't put `self.trial_function_resources` in the same bundle as the other trainer resources. If we did that, we would restrict task placement to pods with all three resource types (trial_function_resources, CPUs, GPUs), which is not necessarily desirable.

tests/conftest.py
```diff
@@ -168,7 +161,8 @@ def _ray_start(request, **kwargs):
 
     init_kwargs = _get_default_ray_kwargs()
     init_kwargs.update(kwargs)
-    res = ray.init(**init_kwargs)
+    # HACK(geoffrey): `hyperopt_resources` is a required resource for hyperopt to prevent deadlocks in Ludwig tests.
+    res = ray.init(**init_kwargs, resources={"hyperopt_resources": 1000})
```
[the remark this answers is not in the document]
geoffreyangus: Added additional context to the HACK comment– copied below for convenience.

```python
# HACK(geoffrey): `hyperopt_resources` is a required resource for hyperopt to prevent deadlocks in Ludwig tests.
#   For context, if there are 4 hyperopt trials scheduled and 7 CPUs available, then the trial function will require
#   some resource to run *in addition* to the resources required by the trainer downstream. If we use 1 CPU
#   (default trial function request), then the trial will be scheduled on 1 CPU and the trainer will later request
#   an additional 1 CPU. Across all 4 trials, this will possibly consume >7 CPUs, causing a deadlock since
#   Ray Datasets will not be able to grab resources for data preprocessing.
#
#   By adding a `hyperopt_resources` resource, we can ensure that the trial function will be scheduled without
#   consuming any CPU resources. This allows each trial's trainer to request 1 CPU without starving Ray Datasets.
# TODO(geoffrey): remove for Ray 2.2
res = ray.init(**init_kwargs, resources={"hyperopt_resources": 1000})
```

ludwig/schema/hyperopt/executor.py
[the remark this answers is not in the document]
geoffreyangus: Just synced with @jeffkinnison and @arnavgarg1 to clarify our shared terminology. Here are the three processes again (each spawning the next):
1. the **tune** driver process, which calls `tuner.fit()`. This name is in line with the [Ray docs](https://docs.ray.io/en/latest/tune/tutorials/tune-lifecycle.html#what-happens-in-tuner-fit).
2. the **trial** driver (formerly called the trial function) process. This process executes `run_experiment`, which creates the Ray Queue.
3. the `trainer` process, which executes `train_fn` in its own placement group.

I'll update the references to `trial_function` to `trial_driver` to make sure this is clear.

ludwig/hyperopt/execution.py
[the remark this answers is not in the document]
geoffreyangus: Updated the above comment to use the terms defined here: https://github.com/ludwig-ai/ludwig/pull/2709#discussion_r1062880369

ludwig/data/dataset/ray.py
```diff
@@ -82,20 +81,11 @@ def __init__(
...
-        shuffle_seed: int = default_random_seed,
-    ) -> DatasetPipeline:
-        """
-        Args:
-            shuffle: If true, the entire dataset is shuffled in memory before batching.
-            fully_executed: If true, force full evaluation of the Ray Dataset by loading all blocks into memory.
-            window_size_bytes: If not None, windowing is enabled and this parameter specifies the window size in bytes
-                    for the dataset. If None and the dataset is large, set to the window size determined at init.
-        """
+    def get_window_size_bytes(self, window_size_bytes: Optional[int] = None):
```
jeffkinnison, line 84: nit: We could add a type annotation for the return value, `def get_window_size_bytes(self, window_size_bytes: Optional[int] = None) -> int:`

ludwig/distributed/ddp.py
```diff
@@ -77,7 +77,17 @@ def gather_all_tensors_fn(cls) -> Optional[Callable]:
 
     @classmethod
     def get_ray_trainer_backend(cls, **kwargs) -> Optional[Any]:
-        return "torch"
+        from ray.train.torch import TorchConfig
+
+        backend = kwargs.get("backend", None)
```
tgaddair, line 82: It looks like this `backend` conflicts with the current definition of backend defined in `get_trainer_kwargs` here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L132

Am I missing something?

ludwig/backend/ray.py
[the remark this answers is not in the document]
justinxzhao: nit: Remove the comment.

justinxzhao (approved):
LGTM!

jeffkinnison (approved):

ludwig/distributed/ddp.py
```diff
@@ -77,7 +77,9 @@ def gather_all_tensors_fn(cls) -> Optional[Callable]:
 
     @classmethod
     def get_ray_trainer_backend(cls, **kwargs) -> Optional[Any]:
-        return "torch"
+        from ray.train.torch import TorchConfig
+
+        return TorchConfig()
```
arnavgarg1, line 82: @tgaddair Just updated - I think this should be fine for now. `TorchConfig` does have some custom attributes but I think we can skip overriding them from now. Eventually, we can add this to the config.

tgaddair (approved):
LGTM!

PR #2710, opened 2022-10-26
Use whylogs to generate dataset profiles for pandas and dask dataframes.

justinxzhao:


ludwig/profiling/dataset_profile.py
```diff
@@ -0,0 +1,74 @@
+import time
+from typing import Dict, Tuple, Union
+
+import whylogs as why
+from whylogs.core.proto import ColumnMessage
```
dantreiman, line 5: This line failed for me (whylogs 0.7.9 installed):
```
tests/ludwig/profiling/test_dataset_profile.py:3: in <module>
    from ludwig.profiling.dataset_profile import (
ludwig/profiling/dataset_profile.py:5: in <module>
    from whylogs.core.proto import ColumnMessage
E   ModuleNotFoundError: No module named 'whylogs.core.proto'
```

After upgrading whylogs (1.1.9), the error goes away.  Probably should add a minimum version to `requirements.txt`

ludwig/profiling/dataset_profile.py
```diff
@@ -0,0 +1,74 @@
...
+def get_column_profile_summaries_from_proto(
+    dataset_profile_proto: dataset_profile_pb2.DatasetProfile,
+) -> Dict[str, ColumnProfileSummary]:
+    """Returns a mapping of feature name to ColumnProfileView."""
+    column_profile_views: Dict[str, ColumnProfileView] = {}
+    for feature_name, feature_profile in dataset_profile_proto.feature_profiles.items():
+        whylogs_metrics_proto = ColumnMessage()
+        # Extra copy+deserialization to avoid TypeError.
+        whylogs_metrics_proto.ParseFromString(feature_profile.whylogs_metrics.SerializeToString())
+        column_profile_view: ColumnProfileView = ColumnProfileView.from_protobuf(whylogs_metrics_proto)
```
dantreiman, line 63: Nit: type annotation redundant here
```suggestion
        column_profile_view = ColumnProfileView.from_protobuf(whylogs_metrics_proto)
```

dantreiman (changes requested):

ludwig/profiling/dataset_profile.py
```diff
@@ -0,0 +1,74 @@
...
+PROFILING_CAP = 100000
+
+
+def get_dataset_profile_view(dataset: Union[str, DataFrame]) -> Tuple[DatasetProfileView, int]:
+    """Returns whylogs dataset profile view."""
+    dataframe = load_dataset(dataset)
+    size_bytes = sum(dataframe.memory_usage(deep=True))
+
+    # Manual cap, also takes care of converting dask to pandas.
+    dataframe = dataframe.head(PROFILING_CAP)
```
brightsparc, line 24: Do we need to have a cap, and if so should this be an optional argument?

If we were to cap, instead of taking `.head` here should we use `.sample(PROFILING_CAP)` if the `shape[0]> PROFILING_CAP`  so that we a better distribution when the data is sorted like in the case of timeseries data?

brightsparc (commented):
I'm not super clear on the utility of converting the DatasetProfileView into a dataset profile pb object to return column profile file

The `dataset_profile_view` class provides methods to like [get_column](https://whylogs.readthedocs.io/en/latest/api/whylogs/core/view/dataset_profile_view/index.html#whylogs.core.view.dataset_profile_view.DatasetProfileView.get_column) which in turn haas methods like [get_metric](https://whylogs.readthedocs.io/en/latest/api/whylogs/core/view/column_profile_view/index.html#whylogs.core.view.column_profile_view.ColumnProfileView.get_metric) which can be used to introspect this object model.  To get an example of how this is used you can see some of the utils methods for example [frequent_items_calculations](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/viz/utils/frequent_items_calculations.py).

If we require read/writing this object model, there are also [read](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/view/dataset_profile_view.py#L198) and [write](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/view/dataset_profile_view.py#L106) methods on the dataset profile class which are used by the various writers such as s3 for pushing to the cloud.

ludwig/profiling/dataset_profile.py
justinxzhao: Good idea, done.

ludwig/profiling/dataset_profile.py
justinxzhao: `sample()` is a great idea, though using `.head()` here to be consistent with what we're doing in pbase, and avoids a few issues detailed below:

1. If you do `.sample(n)`, `n` needs to be strictly less than the total number of rows in the dataframe, else we run into this error.

> `ValueError: Cannot take a larger sample than population when 'replace=False'`

`replace=False` seems not ideal.

2. `sample()` with a dask dataframe seems to require a ratio instead of a pure number, which could be tricky to get right without checking the `len()` of the overall dataframe, which I believe is an expensive call that we want to try to avoid (cc: @dantreiman @tgaddair). The error:

> `results = ValueError("sample does not support the number of sampled items parameter, 'n'. Please use the 'frac' parameter instead.")`

Making `cap` an optional parameter SGTM though.

justinxzhao:
> I'm not super clear on the utility of converting the DatasetProfileView into a dataset profile pb object to return column profile file

> If we require read/writing this object model, there are also [read](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/view/dataset_profile_view.py#L198) and [write](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/view/dataset_profile_view.py#L106) methods on the dataset profile class which are used by the various writers such as s3 for pushing to the cloud.

Great call out! The main reason why I've included these additional to/from proto methods is that the plan is to swap this out with how we're using/storing `DatasetInfo` for data connections in pbase. AFAIU, the whylogs-provided read/write methods won't be sufficient for reading/writing to databases.

It also seems valuable to retain the flexibility of adding additional metrics on top of what whylogs profiles provides, which is also the motivation for defining our own [dataset profile proto](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/profiling/proto/dataset_profile.proto), which would be the primary payload in pbase.

> The `dataset_profile_view` class provides methods to like [get_column](https://whylogs.readthedocs.io/en/latest/api/whylogs/core/view/dataset_profile_view/index.html#whylogs.core.view.dataset_profile_view.DatasetProfileView.get_column) which in turn haas methods like [get_metric](https://whylogs.readthedocs.io/en/latest/api/whylogs/core/view/column_profile_view/index.html#whylogs.core.view.column_profile_view.ColumnProfileView.get_metric) which can be used to introspect this object model. To get an example of how this is used you can see some of the utils methods for example [frequent_items_calculations](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/viz/utils/frequent_items_calculations.py).

This is certainly a reasonable suggestion. So far, I've found converting `ColumnProfileView`s wholesale to python dictionaries with their `.to_summary_dict()` method upfront sufficiently straightforward. Are there another benefits to using the `ColumnProfileView` getter methods? FYI, [this branch](https://github.com/ludwig-ai/ludwig/compare/master...whylogs_prototype#diff-d9dd1f7ac62674e33cbeccf288043a9a36c6c509ea6b7c58e16b7fceb992d66fR1) has a sketch of how I'm interfacing with the extracted dictionaries.

justinxzhao (commented):
Thanks for the comments!

justinxzhao:
Discussed with @brightsparc last week:

- We want to keep the current design, which uses our own `DatasetProfile` proto with the whylogs proto as a subfield.
- The proto format is important for efficient storage, and backwards-compatible saving and loading (vs. a traditional dataclass `DatasetInfo`).
- Having the whylogs proto as a subfield enables us to deserialize from storage into whylogs `ProfileView` and `ColumnView` objects, which is important, for example, for loading a coldly stored dataset profiles into whylogs objects that work with whylogs' drift detection libraries.
- We want to wrap the whylogs proto as there may be / are additional dataset profiling metrics not covered by whylogs that we'll want to add ourselves, and have a way to store them alongside the whylogs profile.

Further, I'll propose that we keep `.head(N)` call for now, to unblock other PRs waiting on this one, as this is not a regression from what we do today, and there's perhaps a more involved discussion about the most efficient way to sample from a large dask dataset, with or without needing to call `len()`.

CC: @tgaddair

brightsparc (approved):
Thanks for adding that extra context. Agree to land for now and possible implement a data source specific sample method for large datasets in future.

.github/workflows/pytest.yml
```diff
@@ -22,21 +22,21 @@ jobs:
...
             ray-version: 2.0.0
-          - python-version: 3.8
-            pytorch-version: 1.12.1
-            torchscript-version: 1.10.2
-            ray-version: 2.0.0
-          - python-version: 3.9
-            pytorch-version: nightly
-            torchscript-version: 1.10.2
-            ray-version: nightly
+          # - python-version: 3.8
```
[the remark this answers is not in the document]
justinxzhao: Reverted.

tests/ludwig/profiling/test_dataset_profile.py
```diff
@@ -1,11 +1,50 @@
...
+            "weight": [14.3, 11.8, 4.3, 30.1, 2.0, 120.0, 2.7, 2.2, 1.2],
+        }
+    )
 
-    assert from_serialized.num_examples == 10
+    dataset_profile_view = get_dataset_profile_view(df)
+    dataset_profile_view_proto = get_dataset_profile_proto(dataset_profile_view)
+    column_profile_summaries = get_column_profile_summaries_from_proto(dataset_profile_view_proto)
+
+    assert set(column_profile_summaries.keys()) == {
```
martindavis, line 25: Hey @justinxzhao , a question: where do we want to handle `NaN`s and making them strings? Should that only be a concern when converting this data to JSON?

justinxzhao: Hey @martindavis! 

I imagine that something like generating valid JSON from the whylogs proto for profile visualizations would be handled outside of Ludwig like in the engine, closer to the frontend.

PR #2711, opened 2022-10-26
Add negative sampling to ludwig.data

jppgks:
Adds Pandas implementation for negative sampling to ludwig.data. This can be used for creating recommendation system datasets from implicit feedback datasets (e.g. customer transactions).

### DIFF
--- a/ludwig/data/negative_sampling.py
+++ b/ludwig/data/negative_sampling.py
@@ -43,7 +43,7 @@ def negative_sample(
     label_col: str = "label",
     neg_pos_ratio: int = 1,
     neg_val: Any = 0,
-    log_pct: int = 50,
+    log_pct: int = 0,
 ):
     """Negative sampling for implicit feedback datasets.
 
@@ -54,7 +54,7 @@ def negative_sample(
         label_col: column name for interaction labels (e.g. 1 for positive interaction)
         n_neg: number of negative samples per positive sample
         neg_val: label value for the negative samples
-        percent_print: print progress every percent_print percent
+        percent_print: print progress every percent_print percent. 0 to disable
     Returns:
         Input DataFrame with negative samples appended
 
@@ -83,7 +83,7 @@ def negative_sample(
     user_indices, item_indices = [], []
     extra_samples = 0
     for user_idx, interaction_row in enumerate(interactions_dense):
-        if user_idx % niter_log == 0:
+        if log_pct > 0 and user_idx % niter_log == 0:
             logging.info(
                 f"Negative sampling progress: {float(user_idx) * 100 / nrows:0.0f}% in {time.time() - start_time:0.2f}s"
             )
--- a/ludwig/data/negative_sampling.py
+++ b/ludwig/data/negative_sampling.py
@@ -1,0 +1,106 @@
+import logging
+import time
+from typing import Any, List, Tuple
+
+import numpy as np
+import pandas as pd
+import scipy
+
+from ludwig.utils.types import DataFrame
+
+
+def _negative_sample_user(interaction_row: np.array, neg_pos_ratio: int, extra_samples: int) -> Tuple[List[int], int]:
+    """Returns a list of negative item indices for given user-item interactions.
+
+    If there are not enough negative items, takes all of them and adds the difference to the extra_samples
+    otherwise, samples with replacement.
+
+    Params:
+        interaction_row: user-item interaction row
+        neg_pos_ratio: number of negative samples per positive sample
+        extra_samples: number of additional samples to add to the negative sample list
+    Returns:
+        Tuple of list of negative item indices and number of extra samples
+    """
+    # Find all items that are not interacted with by the user
+    neg_items = np.where(interaction_row == 0)[1]
+    available_samples = len(neg_items)
+
+    # Randomly sample negative items
+    npos = interaction_row.shape[1] - len(neg_items)
+    samples_required = npos * neg_pos_ratio + extra_samples
+    should_sample = samples_required <= available_samples
+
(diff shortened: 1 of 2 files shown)


ludwig/data/negative_sampling.py
```diff
@@ -0,0 +1,106 @@
...
+    available_samples = len(neg_items)
+
+    # Randomly sample negative items
+    npos = interaction_row.shape[1] - len(neg_items)
+    samples_required = npos * neg_pos_ratio + extra_samples
+    should_sample = samples_required <= available_samples
+
+    neg_items = np.random.choice(neg_items, samples_required, replace=False) if should_sample else neg_items
+
+    return neg_items.tolist(), max(0, samples_required - available_samples)
```
arnavgarg1, line 36: Wondering if we can be more opinionated here and actually make our negative samples even more specific. Instead of sampling from the entire world of negative labels, what if we only create negative labels from other positive labels? 

For e.g.,

user A: [1, 2, 3]
user B: [1, 2, 6]
user C: [1, 2, 5, 8]
All items: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]

In this case, as I understand it, for user A, neg_items will be [4, 5, 6, 7, 8, 9, 10] and we will randomly sample from this list. However, if items 7, 9, and 10 have never seen any positive interaction, I probably wouldn't even consider them as negative candidates for any user. I'd instead, look at all of the items that were positive across all users (in this case, [1, 2, 3, 5, 6, 8] and then sample from the diff between what user A interacted with and everything that all users interacted with = [5, 6, 8].

The advantage of this is that it really helps when the item space is large but actual positive interactions happen with just a smaller subset of the item space. The disadvantage is that doesn't really support discovery outside of what's already been interacted with by at least 1 user. 

I've seen this opinionated approach work pretty well on average, but might be consider when to use what strategy for negative sampling. 

What do you think?

arnavgarg1: I guess the case I'm making is that maybe in the future, we can create different negative sampling strategies because I assume that some will be more performant / better for certain recommendation-system use cases.

jppgks: I really like this idea. It could be worth adding an abstract `NegativeSamplingStrategy` class. Implementations can then differ in how they select negative samples. E.g. `FromAll(NegativeSamplingStrategy)` or `FromAllPositive(NegativeSamplingStrategy)`

arnavgarg1: @jppgks I would definitely be in support of a NegativeSamplingStrategy abstract class implementation and abstraction for strategies :) I have a sense that we may discover even more ideas on the way for negative sampling strategies. 

Could we do your proposed implementation as a part of this PR, or would that be too much additional work?

arnavgarg1: @jppgks I can already think of a third strategy that combines the two strategies with some sort of balance parameter 'ratio', where for e.g., 80% of the negative samples come from FromAllPositive and 20% comes from FromAll (and we ensure no overlap between the two sets) to ensure a way to balance what's been interacted with some sort of discovery element.

arnavgarg1: In the future, we can come up with a heuristic to pick these strategies based on the dataset profile.

ludwig/data/negative_sampling.py
```diff
@@ -0,0 +1,106 @@
...
+    user_id_cat = df[user_id_col].astype("category").cat
+    user_id_codes = user_id_cat.codes.values
+
+    item_id_cat = df[item_id_col].astype("category").cat
+    item_id_codes = item_id_cat.codes.values
+
+    interactions_sparse = scipy.sparse.coo_matrix((df[label_col], (user_id_codes, item_id_codes)))
+
+    # Convert to dense user-item matrix so we can iterate
+    interactions_dense = interactions_sparse.todense()
```
dantreiman, line 77: This will use O(users X items) memory I think, which could be large.  Could we skip conversion to dense and loop over:
```
for user_idx in range(nrows):
    interaction_row = interactions_sparse.getrow(i).todense()
```

ludwig/data/negative_sampling.py
```diff
@@ -0,0 +1,106 @@
...
+    should_sample = samples_required <= available_samples
+
+    neg_items = np.random.choice(neg_items, samples_required, replace=False) if should_sample else neg_items
+
+    return neg_items.tolist(), max(0, samples_required - available_samples)
+
+
+def negative_sample(
+    df: DataFrame,
+    user_id_col: str = "customer_id",
```
dantreiman, line 41: "user" and "item" could be other things, maybe they should have more generic names?

TBH I'm internally debating this, because I think what you have done is more clear than 'entity1' or 'entity2', which seems too generic.  So, we don't need to change naming now, but consider incrementally making it more general as we add more use cases.

ludwig/data/negative_sampling.py
```diff
@@ -0,0 +1,106 @@
...
+    label_col: str = "label",
+    neg_pos_ratio: int = 1,
+    neg_val: Any = 0,
+    log_pct: int = 0,
+):
+    """Negative sampling for implicit feedback datasets.
+
+    Params:
+        df: DataFrame containing user-item interactions
+        user_id_col: column name for user ids
```
dantreiman, line 52: Is there a difference between user and item s.t. this is most efficient when `|users| < |items|`?  If so, add to doc string.

dantreiman (approved):
Looks good!  Added a few optional suggestions, apply if they make sense.

jppgks:
There are many helpful suggestions from PR reviewers. Merging as is for now due to other priorities.

PR #2721, opened 2022-10-28
Enable windowed shuffle

jeffkinnison:
This update enables windowed shuffle when the available memory is less than 5x dataset size. When the dataset is too large, the shuffle window size is set to `ray.cluster_resources()["memory"] / 5` bytes. If a user-specified window size is provided, that size is used in place of the heuristically-determined size.

### DIFF
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -64,7 +64,13 @@ def read_remote_parquet(path: str):
 
 
 class RayDataset(Dataset):
-    """Wrapper around ray.data.Dataset."""
+    """Wrapper around ray.data.Dataset.
+
+    Attributes:
+        df_engine:
+
+        window_size_bytes: Automatically determined shuffle window size for large datasets.
+    """
 
     def __init__(
         self,
@@ -82,6 +88,8 @@ class RayDataset(Dataset):
         self._processed_data_fp = df if isinstance(df, str) else None
         self.window_size_bytes = None
 
+        # Enable windowed shuffle if the available memory is less than
+        # 5x larger than the dataset
         ds_memory_size = self.in_memory_size_bytes
         cluster_memory_size = ray.cluster_resources()["memory"]
         if cluster_memory_size < 5.0 * ds_memory_size:
@@ -106,8 +114,10 @@ class RayDataset(Dataset):
             shuffle: If true, the entire dataset is shuffled in memory before batching.
             fully_executed: If true, force full evaluation of the Ray Dataset by loading all blocks into memory.
             window_size_bytes: If not None, windowing is enabled and this parameter specifies the window size in bytes
-                    for the dataset.
+                    for the dataset. If None and the dataset is large, set to the window size determined at init.
         """
+        # If the user does not supply a window size, use the
+        # size determined at initialization.
         if window_size_bytes is None:
             window_size_bytes = self.window_size_bytes
 
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -80,6 +80,12 @@ class RayDataset(Dataset):
         self.data_hdf5_fp = training_set_metadata.get(DATA_TRAIN_HDF5_FP)
         self.data_parquet_fp = training_set_metadata.get(DATA_TRAIN_PARQUET_FP)
         self._processed_data_fp = df if isinstance(df, str) else None
+        self.window_size_bytes = None
+
+        ds_memory_size = self.in_memory_size_bytes
+        cluster_memory_size = ray.cluster_resources()["memory"]
+        if cluster_memory_size < 5.0 * ds_memory_size:
+            self.window_size_bytes = cluster_memory_size // 5
 
         # TODO ray 1.8: convert to Tensors before shuffle
         # def to_tensors(df: pd.DataFrame) -> pd.DataFrame:
@@ -102,6 +108,9 @@ class RayDataset(Dataset):
             window_size_bytes: If not None, windowing is enabled and this parameter specifies the window size in bytes
                     for the dataset.
         """
+        if window_size_bytes is None:
+            window_size_bytes = self.window_size_bytes
+
         if fully_executed:
             if _ray113:
                 # Workaround for: https://github.com/ray-project/ray/issues/25643


ludwig/data/dataset/ray.py
```diff
@@ -80,6 +86,14 @@ def __init__(
         self.data_hdf5_fp = training_set_metadata.get(DATA_TRAIN_HDF5_FP)
         self.data_parquet_fp = training_set_metadata.get(DATA_TRAIN_PARQUET_FP)
         self._processed_data_fp = df if isinstance(df, str) else None
+        self.window_size_bytes = None
```
ShreyaR, line 89: Instead of setting the window size bytes here, can you instead use a parameter for `auto_window`? If the parameter is set to `True`, then we can set `window_size_bytes` inside the `pipeline` function to use the heuristic below.

ShreyaR (commented):
Looks good! Can you also add a test for the change in this function?

arnavgarg1:
Things might change a bit based on the Ray trianer upgrade PR: https://github.com/ludwig-ai/ludwig/pull/2709

ludwig/data/dataset/ray.py
```diff
@@ -100,8 +107,16 @@ def pipeline(
...
-                    for the dataset.
+                    for the dataset. If None and the dataset is large, set to the window size determined at init.
         """
+        # If the user does not supply a window size and the dataset is large,
+        # set the window size to `<available memory> // 5`.
+        if self.auto_window and not window_size_bytes:
+            ds_memory_size = self.in_memory_size_bytes
+            cluster_memory_size = ray.cluster_resources()["memory"]
+            if cluster_memory_size < 5.0 * ds_memory_size:
+                window_size_bytes = cluster_memory_size // 5
```
arnavgarg1, line 118: Would be good to add a logging statement letting the user know that we're setting a window size since this will improve throughput/reduce disk spillage at the cost of model performance.

ludwig/data/dataset/ray.py
```diff
@@ -64,14 +64,20 @@ def read_remote_parquet(path: str):
...
+            automatically set window size to `<available memory> // 5`.
+    """
 
     def __init__(
         self,
         df: Union[str, DataFrame],
         features: Dict[str, Dict],
         training_set_metadata: Dict[str, Any],
         backend: Backend,
+        auto_window: bool = True,
```
ShreyaR, line 80: @jeffkinnison for now, we can set the default value of `auto_window` to False.

In order to set `auto_window` to True, we'd need to do some benchmarking in order to figure out the impact of windowing. The `5 * dataset_size` heuristic is useful for determining the object store memory in order ensure that no disk spillage occurs. However, it's possible that even if the object store memory < `5 * dataset_size`, windowing is still expensive and doesn't lead to overall latency gains. In practice, the Anyscale team had mentioned that this is somewhere between `2 * dataset_size` and `3 * dataset_size`. So determining what that threshold looks like would be necessary before setting `auto_window` to True.

Separately, because of the new engines workstream, I expect that we would set `auto_window` to True if we're trying to cut down on resource requirements for an engine, in which case we would also need to have benchmarking. However, priority-wise, that would be a P1 for the new engines workstream that would be good to work on once the initial blockers for engines are resolved.

ludwig/data/dataset/ray.py
```diff
@@ -64,14 +67,20 @@ def read_remote_parquet(path: str):
...
+            automatically set window size to `<available memory> // 5`.
+    """
 
     def __init__(
         self,
         df: Union[str, DataFrame],
         features: Dict[str, Dict],
         training_set_metadata: Dict[str, Any],
         backend: Backend,
+        auto_window: bool = False,
```
arnavgarg1, line 83: If we want to configure this, should we propagate this parameter here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataset/ray.py#L157? 

This is where we initialize the RayDataset object after preprocessing.

ludwig/data/dataset/ray.py
```diff
@@ -100,8 +110,17 @@ def pipeline(
...
-                    for the dataset.
+                    for the dataset. If None and the dataset is large, set to the window size determined at init.
         """
+        # If the user does not supply a window size and the dataset is large,
+        # set the window size to `<available memory> // 5`.
+        if self.auto_window and window_size_bytes is None:
+            ds_memory_size = self.in_memory_size_bytes
+            cluster_memory_size = ray.cluster_resources()["object_store_memory"]
+            if ds_memory_size > cluster_memory_size // 5:
+                logger.info("Dataset size is larger than object store, enabling windowed shuffle.")
```
arnavgarg1, line 121: @jeffkinnison I don't think this is actually an accurate log message, since it's technically not larger, but it's greater than 20% of the object store memory. I'm wondering if we can change this to something like:

```
In-memory dataset size is greater than 20% of object store memory. Enabling windowed shuffling of data to prevent chances of OOMs. 
Read more here: <link to what windowing is, how to prevent this from happening, and the modeling performance impact>
```

And then we can add a section on windowed shuffling and its impact to the Ludwig docs. What do you think?

tests/integration_tests/test_ray.py
```diff
@@ -782,3 +784,104 @@ def test_ray_preprocessing_placement_group(tmpdir, ray_cluster_2cpu):
             skip_save_log=True,
         )
         preds, _ = model.predict(dataset=dataset)
+
+
+@pytest.mark.distributed
```
ShreyaR, line 789: Super thorough tests -- thanks for adding them in!

ShreyaR (approved):
lgtm pending @arnavgarg1's comments!

ludwig/data/dataset/ray.py
jeffkinnison: That sounds good, do you have any suggestions for where to add the auto windowing info to the docs?

ludwig/data/dataset/ray.py
jeffkinnison: This should be addressed in [dc3f08e](https://github.com/ludwig-ai/ludwig/pull/2721/commits/dc3f08ea0ab0f3aa39d8fcdcbde20b27c5b8c835). I agree that we should probably do an auto-sizing investigation at some point and see if we can come up with better window size heuristics.

ludwig/data/dataset/ray.py
jeffkinnison: The updated logging statement should be added in here: [24f36f5](https://github.com/ludwig-ai/ludwig/pull/2721/commits/24f36f5829d2197a39ff847dc016fd2108341052)

ludwig/data/dataset/ray.py
jeffkinnison: Added the updated logging message in [24f36f5](https://github.com/ludwig-ai/ludwig/pull/2721/commits/24f36f5829d2197a39ff847dc016fd2108341052)

ludwig/data/dataset/ray.py
jeffkinnison: Added the passthrough in [24f36f5](https://github.com/ludwig-ai/ludwig/pull/2721/commits/24f36f5829d2197a39ff847dc016fd2108341052) and updated the tests to use it in [1c71308](https://github.com/ludwig-ai/ludwig/pull/2721/commits/1c71308fc403044402d53ab9e9d39da94ef71794).

tests/integration_tests/test_ray.py
jeffkinnison: Thanks!

PR #2728, opened 2022-11-07
Add custom to_dask() to infer Dask metadata from Datasets schema.

arnavgarg1:
This PR adds a custom implementation of `to_dask()` to ensure forward compatibility with Ray 2.0/Ray 2.1 schema inference. This ensures that Ray 2.1 will work correctly even with image/audio datasets.

### DIFF
--- a/ludwig/data/dataframe/dask.py
+++ b/ludwig/data/dataframe/dask.py
@@ -14,20 +14,25 @@
 # limitations under the License.
 # ==============================================================================
 
+import collections
 import logging
 import os
-from typing import Dict
+from typing import Any, Dict, Iterable, Tuple, Union
 
 import dask
 import dask.array as da
 import dask.dataframe as dd
+import numpy as np
 import pandas as pd
+import pyarrow as pa
 import ray
 from dask.diagnostics import ProgressBar
 from packaging import version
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):

ludwig/data/dataframe/dask.py
```diff
@@ -248,11 +255,68 @@ def to_ray_dataset(self, df):
...
+        # Infer Dask metadata from Datasets schema.
+        schema = dataset.schema()
+        if isinstance(schema, PandasBlockSchema):
+            meta = pd.DataFrame(
+                {
+                    col: pd.Series(dtype=(dtype if not isinstance(dtype, TensorDtype) else np.object_))
+                    for col, dtype in zip(schema.names, schema.types)
+                }
+            )
+        elif pa is not None and isinstance(schema, pa.Schema):
```
dantreiman, line 301: When would pa be None?  Should we log a warning or error if it is?

dantreiman (approved):
LGTM

ludwig/data/dataframe/dask.py
justinxzhao: I'm not sure.

I've gone ahead and removed the check since we're importing `pyarrow as pa` directly, and the import will either succeed for fail.

PR #2737, opened 2022-11-08
Moves image decoding out of Ray Datasets to Dask Dataframe

geoffreyangus:
This PR fixes cases where bad path entries cause dtype errors in image columns, which Ray Datasets tries to cast into a Tensor type column internally. Example error: 

```
TypeError: Tried to convert an ndarray of ndarray pointers (object dtype) to a well-typed ndarray but this failed; convert the ndarray to a well-typed ndarray before casting it as a TensorArray, and note that ragged tensors are NOT supported by TensorArray. First 5 subndarray types: [dtype('uint8'), dtype('uint8'), dtype('uint8'), dtype('O'), dtype('O')]
```

We circumvent this issue by moving byte decoding from Ray Datasets to Dask via its `map` function.


ludwig/backend/ray.py
```diff
@@ -985,29 +983,16 @@ def read_binary_files(
...
-            ds = self.df_engine.to_ray_dataset(column_df)
-
-        def map_batches_fn(df: pd.DataFrame, fn: Callable) -> pd.DataFrame:
-            # HACK: Workaround for https://github.com/modin-project/modin/issues/4686
-            if "value" in df:
-                key = "value"
-            else:
-                key = column.name
-
-            # We need to explicitly pass the credentials stored in fsspec.conf since the operation occurs on Ray.
```
geoffreyangus, line 1002: @jeffreyftang do I need to somehow keep this decorator now that the operation is occurring on Dask (instead of Ray)?

jeffreyftang: I *think* it should be fine.

jeffreyftang: My understanding was that running the function as remote via Ray was dropping the credentials in the original context. If the remote function is no longer in the picture, then things should be fine.

jeffreyftang: Although one question: seems like we could still wind up running things remotely through Dask-on-Ray, no?

geoffreyangus: Okay interesting– what exactly are the credentials for? 

At this point, we'll have already loaded the files from S3, GCP, etc.

This step is just for decoding the returned bytes objects into image tensors.

In any case, is there an easy way to validate this change either way? I tested this on the raw Twitter Bots dataset (not the self-contained version in S3) and it seemed to be fine.

geoffreyangus: Just ran the self-contained S3 version of the dataset. I'm seeing this message in the logs a bunch:
```
(explain_task pid=1183) Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))': /twitter_bots/profile_images/8e209998a920de227a213f0100871b0e5e287d94.jpeg
(explain_task pid=1183) Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', OSError(0, 'Error'))': /twitter_bots/profile_images/622e71720f5f6943613f791ffd3c9b820330e84d.jpg
(explain_task pid=1183) Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))': /twitter_bots/profile_images/46852d48e259e1b6ec842f3d0bc1b48280ea9d7c.jpg
```
Not sure if this is related. Let me know either way!

jeffreyftang: I assume that missing or invalid credentials wouldn't show up as "connection reset".

As for what the credentials are for, they handle the case when the URL for an image requires authentication (e.g., a private S3 bucket). So in addition to loading the tabular portion of the dataset, we could need a second set of credentials for accessing those image paths.

justinxzhao (approved):

PR #2738, opened 2022-11-08
Added exponential retry for mlflow, remote dataset loading

ShreyaR:


ludwig/utils/fs_utils.py
```diff
@@ -34,6 +34,8 @@
 from filelock import FileLock
 from fsspec.core import split_protocol
 
+from .error_handling_utils import default_retry, default_retry_call
```
arnavgarg1, line 37: Can we change this to `ludwig.utils.error_handling_utils`?

ShreyaR: Done

ludwig/utils/fs_utils.py
```diff
@@ -257,6 +272,7 @@ def put_fn():
         yield url, None
 
 
+# TODO(shreya): Add retry for open context.
```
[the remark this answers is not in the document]
ShreyaR: @arnavgarg1 it's not possible to execute the code in a context manager multiple times. More details [here](https://stackoverflow.com/a/36741101).

My plan to add retries for this are to change the context so that `open_file` calls `fs.open` on entering the context (I'll have to double check if an `fs.close` is needed on exit), and then wrap the `open` (and `close`) calls in `default_retry_call` wrappers.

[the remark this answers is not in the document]
arnavgarg1: That makes sense!

arnavgarg1:
LGTM! 🚢

tgaddair (commented):
For mlflow changes, did you look into configuring mlflow-native solutions like playing around with these environment variables?

https://www.mlflow.org/docs/latest/python_api/mlflow.environment_variables.html#mlflow.environment_variables.MLFLOW_HTTP_REQUEST_BACKOFF_FACTOR

If so, did it work, or were there still issues?

ludwig/utils/fs_utils.py
```diff
@@ -126,11 +130,13 @@ def abspath(url):
     return os.path.abspath(url)
 
 
+@default_retry()
```
tgaddair, line 133: I'm a little bit concerned about adding blanket retries for all these functions. If we set the retry too low, it could have little affect if there's a network partition lasting a minute or more. If it's too high, it could take a while for a real failure to bubble up to the users. Do we have some indication of what types of errors are surfaced when a connection issue arises?

There is one open issue in fsspec [here](https://github.com/fsspec/filesystem_spec/issues/776) that indicates there might be something for S3 and GCS we can use to implement retry logic at the filesystem level. But not sure what those params are or how are they are configured.

ShreyaR: I looked through the `s3fs` and `gcfs` libraries.

For `s3fs`:
It seems like there's a default retry for many (all?) `s3fs` methods. The retry policy is exponential backoff, with 5 requests with up to ~0.8 seconds of backoff between requests. You can check it out [here](https://s3fs.readthedocs.io/en/latest/_modules/s3fs/core.html#) and cmd+F `async def _error_wrapper`. Within the same file, there's also `S3_RETRYABLE_ERRORS ` which we can use if needed.

For `gcfs`:
Similar behavior, except that the max backoff is 32 seconds. Also seems like it applies to most if not all methods of the class. You can take a look at how retry behavior is configured [here](https://github.com/fsspec/gcsfs/blob/b4eabc01fe5d48ff5511d4fbf543fcbd55435acd/gcsfs/retry.py#L110). 

Given this info:
1. I'm leaning towards just removing retry decorators from any `fsspec` method. However,
2. It's unclear if all filesystems implement their own retry methods. And,
3. The retry policies seem to vary across filesystem, which may be ok I guess?

Wdyt?

tgaddair: Can we configure the retry policy for s3 and gcs via environment variables? If we could do the same for mlfow, that would be ideal, I think.

ShreyaR: @tgaddair Done!

arnavgarg1 (approved):

PR #2740, opened 2022-11-09
Use clearer error messages in ludwig serving, and enable serving to work with configs that have stratified splitting on target columns.

justinxzhao:


connor-mccorm (approved):
LGTM

PR #2746, opened 2022-11-10
Fix https://github.com/ludwig-ai/ludwig/issues/1735

herrmann:
Multiply the `output_size` by two if the RNN is bidirectional


justinxzhao:
Confirming that this fixes the issue.

Thanks for the contribution @herrmann!

PR #2748, opened 2022-11-10
Fix benchmark tests, update latest metrics, and use the local backend for GBM benchmark tests

abidwael:
- Changed GBM benchmark tests to use the local backend because on the ray backend, the number of workers must be explicitly set to match the GitHub runner’s resources (`num_workers = 2`)
- We decrease the benchmark r2 score for GBM on the `mercedes_benz_greener` dataset. The source of the regression is unclear, but it seems worth making benchmark tests green again for now so that we can notice new failures. We defer deeper investigation in favor of prioritizing other tasks.


ludwig/benchmarking/benchmark.py
```diff
@@ -88,7 +88,7 @@ def benchmark_one(experiment: Dict[str, Union[str, Dict[str, str]]]) -> None:
         )
         delete_hyperopt_outputs(experiment["experiment_name"])
     else:
-        backend = "ray"
```
justinxzhao, line 91: Is it worth adding a TODO here to change this back to `ray` in the future?

justinxzhao:
The PR LGTM, but let's update the PR title and description with more information about why we're making this change and what the change is exactly, (including if we're not sure) -- can maybe copy in over from slack communication.

This will make it easy for us to refer back to in the future to see what was changed and why, i.e.

justinxzhao (approved):

PR #2749, opened 2022-11-10
Annotate MLFlow callback, and utility functions

dantreiman:
Upgrades `_get_or_create_experiment_id`, `log_saved_model` to developer API.

Makes MLFlowCallback Public API, for consistency with other callbacks.

### DIFF
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -4,7 +4,7 @@ import queue
 import threading
 from typing import Any, Dict
 
-from ludwig.api_annotations import DeveloperAPI
+from ludwig.api_annotations import DeveloperAPI, PublicAPI
 from ludwig.callbacks import Callback
 from ludwig.constants import TRAINER
 from ludwig.data.dataset.base import Dataset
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):

ludwig/contribs/mlflow/__init__.py
```diff
@@ -21,14 +21,21 @@ def _get_runs(experiment_id: str):
...
     experiment = mlflow.get_experiment_by_name(experiment_name)
     if experiment is not None:
         return experiment.experiment_id
     return mlflow.create_experiment(name=experiment_name, artifact_location=artifact_uri)
 
 
-@DeveloperAPI
+# Included for backwards compatibility, Deprecated.
+# TODO(daniel): delete this.
+_get_or_create_experiment_id = get_or_create_experiment_id
```
arnavgarg1, line 35: Should we add DeprecatedAPI to this?

dantreiman: Unfortunately decorators don't work for global variables, only for function-like objects (functions, classes, enums).  This will be a problem too for constants.py or globals.py.

PR #2752, opened 2022-11-11
Public datasets

connor-mccorm:
**Part 1 of the Public Datasets**
- Add task entries to all ludwig dataset configs
- Add ludwig datasets method to get names and tasks


ludwig/datasets/dataset_config.py
```diff
@@ -29,6 +29,9 @@ class DatasetConfig:
     # The readable description of the dataset
     description: str = ""
 
+    # The learning task(s) associated with this dataset.
+    task: str = ""
```
dantreiman, line 33: Recommend calling it tasks, even when there is only one item.
```suggestion
    tasks: Union[str, List[str]] = field(default_factory=list)
```


Also IMO it would be cleaner to make this a `Union[str, List[str]]`, similar to `download_urls`. Then we could do:

```
task: Regression
```
or
```
task:
  - Regression
  - Classification
```

ludwig/datasets/configs/ae_price_prediction.yaml
```diff
@@ -12,3 +12,4 @@ description: |
   Innerwear Data from Victoria's Secret and Others
   600,000+ innerwear product data extracted from popular retail sites
   https://www.kaggle.com/PromptCloudHQ/innerwear-data-from-victorias-secret-and-others
+task: Regression, Classification
```
dantreiman, line 15: IMO lets use yaml list instead of comma separated list for consistency
```
task(s):
  - Regression
  - Classification
```

ludwig/datasets/configs/agnews.yaml
```diff
@@ -15,3 +15,4 @@ columns:
 loader: agnews.AGNewsLoader
 description: |
   News articles categorized as "World", "Sports", "Business", and "Science".
+task: Muli-Class Classification
```
dantreiman, line 18:
```suggestion
task: Multi-Class Classification
```

ludwig/datasets/dataset_config.py
```diff
@@ -29,6 +29,9 @@ class DatasetConfig:
     # The readable description of the dataset
     description: str = ""
 
+    # The learning task(s) associated with this dataset.
```
dantreiman, line 32: Provide reason for this field - what is it for?
```suggestion
    # Optional.  The types of learning tasks this dataset is intended to test. Helps users discover new datasets and
    # filter for relevance to a specific machine learning setting.
```

ludwig/datasets/__init__.py
```diff
@@ -80,6 +80,11 @@ def list_datasets() -> List[str]:
     return sorted(_get_dataset_configs().keys())
 
 
+def list_dataset_configs() -> Dict[str, DatasetConfig]:
```
dantreiman, line 83: Could you use an accessor function or function(s) like: 

```
@PublicAPI
def tasks_for_dataset(dataset_name) -> List[str]
```

and/or 

```
@PublicAPI
def datasets_for_task(dataset_name) -> List[str]
```

or whatever we need on the backend.  Otherwise we are implicitly making `DatasetConfig` part of the public API, which is something I've tried to avoid.  Once client code is using a data structure, we must worry about what we are breaking every time we change it.

See for example `describe_dataset` in `datasets/__init__.py` - its purpose is to provide the requested information **without** exposing the underlying data structure.

ludwig/datasets/configs/goodbooks_books.yaml
```diff
@@ -13,3 +13,4 @@ description: |
     book metadata (author, year, etc.)
     tags/shelves/genres
   https://github.com/zygmuntz/goodbooks-10k
+task: Multi-Class Classification
```
dantreiman, line 16: Is "Multi-Class Classification" different from "Classification"?

dantreiman (changes requested):
Once we decide on the API additions, and other review comments are done, please update `ludwig/datsets/README.md`

ludwig/datasets/configs/ae_price_prediction.yaml
connor-mccorm: Good look. Done!

ludwig/datasets/configs/goodbooks_books.yaml
connor-mccorm: It's just a more specific way of saying it's a classification problem. Since there are a bunch of binary classification problems in the zoo, I figured I would distinguish between those with this label.

ludwig/datasets/__init__.py
connor-mccorm: Made the change to get as close to this as I can. We need the task and the name and in order. I realize it's slightly bespoke but fundamentally we need this data so I feel like this is a lean way of doing this. Let me know what you think!

ludwig/datasets/configs/goodbooks_books.yaml
dantreiman: This is my assumption I make (I wanted to check that this is correct):

"Multi-Class Classification" - 3 or more classes
"Binary Classification" - 2 classes
"Classification" - 2 or more classes

Lets document what these mean in the Readme.md, docstring of the `types` field, or both.

justinxzhao (approved):

PR #2755, opened 2022-11-12
fix: refactor `IntegerOptions` field

ksbrar:
Simplifies impl. and adds optional `allow_none` and default-value validation back to this field.


dantreiman (approved):
LGTM

PR #2756, opened 2022-11-12
Added ray datasets repartitioning in cases of multiple train workers

ShreyaR:
For distributed GBMs, data sharding is done by number of blocks of the ray dataset. Ray dataset blocks are distributed across the parallel GBM workers.

This PR explicitly partition a RayDataset so that it has at least as many blocks as the total number of parallel workers.


tgaddair (approved):

jppgks (approved):
Thanks Shreya, I'm assuming this fixes an existing test, in which case this looks good to me! Otherwise can we add one?

ShreyaR:
> Thanks Shreya, I'm assuming this fixes an existing test, in which case this looks good to me! Otherwise can we add one?

@jppgks yes, this fixes the benchmark test failure in Higgs split.

PR #2762, opened 2022-11-15
Allow single value category features

abidwael:


arnavgarg1 (changes requested):
Let's log a warning or info block letting the user know that this feature isn't going to add much to the model, and that they should consider dropping it (although it will not actually cause an error per say)

ludwig/features/category_feature.py
```diff
@@ -126,8 +125,8 @@ def get_feature_meta(column, preprocessing_parameters, backend):
         )
         vocab_size = len(str2idx)
         if vocab_size <= 1:
-            raise InputDataError(
-                column.name, CATEGORY, f"At least 2 distinct values are required, column only contains {str(idx2str)}"
+            logger.warning(
+                f"Feature {column.name} of type {CATEGORY} only contains {vocab_size} distinct values: {str(idx2str)}."
```
arnavgarg1, line 129: This warning by itself doesn't say much. I'd probably expand and add in that this feature will not be useful for the machine learning model and that they should consider dropping it

tgaddair (changes requested):
There was actually a bug that occurred when this was supported. We should (1) make sure we have a workaround if we're going to support this, and (2) make sure we have a test that verified correctness.

ludwig/features/category_feature.py
```diff
@@ -126,8 +125,9 @@ def get_feature_meta(column, preprocessing_parameters, backend):
         )
         vocab_size = len(str2idx)
         if vocab_size <= 1:
-            raise InputDataError(
-                column.name, CATEGORY, f"At least 2 distinct values are required, column only contains {str(idx2str)}"
+            logger.warning(
+                f"Feature {column.name} of type {CATEGORY} only contains {vocab_size} distinct "
+                "values: {str(idx2str)}. Consider dropping the feature."
```
arnavgarg1, line 130: I think you may need to add an `f` before `"values: {str(idx2str)}. Consider dropping the feature."`, but if not, this looks good! Thanks for making the change!

abidwael: Good catch!

abidwael:
Closing since it's more useful to throw an error if a categorical feature with a single distinct value is specified.

PR #2764, opened 2022-11-15
Upload Datasets to Remote Location

connor-mccorm:
Ability to get byte buffer for ludwig datasets allowing for upload to remote locations such as S3 or Minio.


ludwig/datasets/configs/adult_census_income.yaml
```diff
@@ -28,3 +28,4 @@ loader: adult_census_income.AdultCensusIncomeLoader
 description: |
   Predict whether income exceeds $50K/yr based on census data
   https://archive.ics.uci.edu/ml/datasets/adult
+task: Binary Classification
```
tgaddair, line 31: Seem odd to specify the task in this way. In Ludwig, it would seem more appropriate to just list the output feature types, which is more aligned to how Ludwig thinks about tasks. Something like:

```
output_features:
  - name: <whatever>
     type: binary
```

connor-mccorm: Good call, I'll switch over to this method since it conveys the info more clearly!

ludwig/datasets/loaders/dataset_loader.py
```diff
@@ -281,8 +281,10 @@ def load(self, split=False, kaggle_username=None, kaggle_key=None) -> pd.DataFra
 
         :param split: (bool) splits dataset along 'split' column if present. The split column should always have values
         0: train, 1: validation, 2: test.
+        :param kaggle_username: (str) Kaggle username for downloading datasets from Kaggle.
+        :param kaggle_key: (str) Kaggle key for downloading datasets from Kaggle.
         """
-        self._download_and_process()
+        self._download_and_process(kaggle_username=kaggle_username, kaggle_key=kaggle_key)
```
justinxzhao, line 287: I'm surprised that this wasn't here before. Does this mean we weren't able to load the titanic dataset?

justinxzhao (approved):

abidwael (approved):

PR #2765, opened 2022-11-16
Correctly infer bool and object types in autoML

arnavgarg1:
Fixes an issue where `is_field_boolean` always return True if there are greater than 3 distinct values. This should return False by default if there are greater than 3 values. 

Additionally, some of the type checking fails because `source.get_distinct_values()` actually gets unique values after dropping NaNs. This prevents us from catching the case we're looking for, i.e., 3 distinct values of which one is a None/NaN. This logic is modified to check for None and Nans, and return True if either of those are found in the case that there are 3 distinct values.

Co-authored-by: @jppgks

### DIFF
--- a/ludwig/automl/base_config.py
+++ b/ludwig/automl/base_config.py
@@ -222,20 +222,24 @@ def get_dataset_info(df: Union[pd.DataFrame, dd.core.DataFrame]) -> DatasetInfo:
 
 
 def is_field_boolean(source: DataSource, field: str) -> bool:
-    num_unique_values, unique_values, _ = source.get_distinct_values(field, max_values_to_return=4)
+    """
+    Returns a boolean indicating whether the object field should have a bool dtype.
+
+    Columns with object dtype that have 3 distinct values of which one is Nan/None is a bool type column.
+    """
+    unique_values = source.df[field].unique()
+    num_unique_values = len(unique_values)
     if num_unique_values <= 3:
-        for entry in unique_values:
+        if None in unique_values:
+            return True
+        for unique_value in unique_values:
             try:
-                if np.isnan(entry):
-                    continue
+                if np.isnan(unique_value):
+                    return True
             except TypeError:
-                # For some field types such as object arrays np.isnan throws a TypeError
-                # we catch it since we know in this case it is not a bool.
-                return False
-            if isinstance(entry, bool):
-                continue
-            return False
-    return True
+                pass
+        return False
+    return False
 
 
 @DeveloperAPI
--- a/ludwig/automl/base_config.py
+++ b/ludwig/automl/base_config.py
@@ -234,7 +234,7 @@ def is_field_boolean(source: DataSource, field: str) -> bool:
                     continue
             except TypeError:
                 # For some field types such as object arrays, np.isnan throws a TypeError
-                # If this is the case, we can just proceed to checking if the entry is a bool object
+                # In this case, do nothing and proceed to checking if the entry is a bool object
                 pass
             if isinstance(entry, bool):
                 continue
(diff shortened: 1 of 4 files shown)


ludwig/automl/base_config.py
```diff
@@ -222,20 +222,23 @@ def get_dataset_info(df: Union[pd.DataFrame, dd.core.DataFrame]) -> DatasetInfo:
...
-    num_unique_values, unique_values, _ = source.get_distinct_values(field, max_values_to_return=4)
+    """Returns a boolean indicating whether the object field should have a bool dtype.
+
+    Columns with object dtype that have 3 distinct values of which one is Nan/None is a bool type column.
+    """
+    unique_values = source.df[field].unique()
+    num_unique_values = len(unique_values)
     if num_unique_values <= 3:
-        for entry in unique_values:
+        if None in unique_values:
```
magdyksaleh, line 232: Why does this mean that it is automatically True?

ludwig/automl/base_config.py
```diff
@@ -222,20 +222,23 @@ def get_dataset_info(df: Union[pd.DataFrame, dd.core.DataFrame]) -> DatasetInfo:
...
+            return True
+        for unique_value in unique_values:
             try:
-                if np.isnan(entry):
-                    continue
+                if np.isnan(unique_value):
+                    return True
```
magdyksaleh, line 237: Also why are we returning True here?

magdyksaleh (commented):
I would have assumed that the fix would be
```
def is_field_boolean(source: DataSource, field: str) -> bool:
    num_unique_values, unique_values, _ = source.get_distinct_values(field, max_values_to_return=4)
    if num_unique_values <= 3:
        for entry in unique_values:
            try:
                if np.isnan(entry):
                    continue
            except TypeError:
                # For some field types such as object arrays np.isnan throws a TypeError
                # we catch it since we know in this case it is not a bool.
                return False
            if isinstance(entry, bool):
                continue
            return False
        return True   
    return False
```

ludwig/automl/base_config.py
arnavgarg1: So we know the field has an object dtype, and we're trying to see if we should manually set the type to bool. Bool types have 2 distinct values, but if there is a None or a NaN, then it had 3 distinct values which is why it's being inferred as an object. I added the None conditional check incase the field is just empty and not an np.nan, but I can also just remove this check.

ludwig/automl/base_config.py
arnavgarg1: Because one of the three distinct values is a NaN, which means that the others are valid, which therefore makes this a valid bool type

ludwig/automl/base_config.py
magdyksaleh: what if unique values are [NaN, "Bananas", None]? Wouldn't we then see the nan, return True from the function even though it isn't true?

arnavgarg1: Hmm, that is an interesting edge case. Would the dtype in this case just be object?

arnavgarg1: >The bug is valid but I thought the only case where we want to change from type object to bool is if we have two bool values and a NaN

That makes sense, will update

arnavgarg1: @magdyksaleh just updated

magdyksaleh (approved):

ludwig/automl/base_config.py
```diff
@@ -222,20 +222,25 @@ def get_dataset_info(df: Union[pd.DataFrame, dd.core.DataFrame]) -> DatasetInfo:
...
+    """
+    unique_values = source.df[field].unique()
+    if len(unique_values) <= 3:
         for entry in unique_values:
             try:
                 if np.isnan(entry):
                     continue
             except TypeError:
                 # For some field types such as object arrays np.isnan throws a TypeError
                 # we catch it since we know in this case it is not a bool.
```
jeffreyftang, line 237: Nit: fix the comment here as well :)

arnavgarg1: Good catch! Addressed!

justinxzhao (approved):

PR #2774, opened 2022-11-21
Ensure correct output reduction for text encoders like MT5 and add warning messages when not supported

arnavgarg1:
This PR fixes an issue where `reduce_output==cls_pooled` was causing index out-of-range errors. This happens because the MT5Encoder doesn't have `pooler_outputs` returned from the Transformer. 

As a part of this PR, I reviewed all of the outputs for each of the text encoders to ensure that we only support `cls_pooled` for compatible text encoders, and log errors in cases where it doesn't work for that particular encoder.

This also adds a test for XLNetEncoder, which was missing.

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -50,6 +50,8 @@ logger = logging.getLogger(__name__)
 
 
 def _cls_pooled_error_message(encoder: str):
+    # TODO(Arnav): Remove this once we have reduce_output options set for
+    # each encoder type in the schema
     logger.error(f"reduce_output cannot be cls_pooled for {encoder}")
     sys.exit(1)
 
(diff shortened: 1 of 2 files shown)


ludwig/encoders/text_encoders.py
```diff
@@ -49,6 +49,11 @@
 logger = logging.getLogger(__name__)
 
 
+def _cls_pooled_error_message(encoder: str):
```
justinxzhao, line 52: Let's add a TODO (and can you file an issue to MLF?) to enforce specific options at the schema level.

justinxzhao (approved):
Thanks for the blazing fast fix!

ludwig/encoders/text_encoders.py
arnavgarg1: Yup!

PR #2778, opened 2022-11-22
Drop incomplete batches for Ray and Pandas to prevent Batchnorm computation errors

arnavgarg1:
This PR implements logic to drop the last batch for Pandas and Ray dataset batchers to prevent issues when there is only 1 row in a batch. 

A single row in a batch causes issues when computing Batchnorm (such as in FC layers, Tabnet, etc.) because it can't compute the mean/stddev for a single sample. Additionally, Ludwig has some logic for Tabnet that has a conditional check for sample_size == 1 that we can now safely remove.

Skipping adding tests since our current test suite already has tests that use the batcher classes for model training, so if the tests pass, then this logic should work as well. I modified individual tests locally to be sure.

Note: This only drops incomplete training batches if the batch size < total number of rows in the dataset. Additionally, it will always keep incomplete training batches for validation/test sets.

To follow:

- [ ]  Remove check for batch_size == 1 in Tabnet

### DIFF
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -144,13 +144,21 @@ class RayDataset(Dataset):
         return pipe
 
     @contextlib.contextmanager
-    def initialize_batcher(self, batch_size=128, should_shuffle=True, seed=0, ignore_last=False, horovod=None):
+    def initialize_batcher(
+        self,
+        batch_size=128,
+        should_shuffle=True,
+        seed=0,
+        ignore_last=False,
+        horovod=None,
+    ):
         yield RayDatasetBatcher(
             self.ds.repeat().iter_datasets(),
             self.features,
             self.training_set_metadata,
             batch_size,
             self.size,
+            ignore_last,
         )
 
     def __len__(self):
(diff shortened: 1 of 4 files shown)


ludwig/data/dataset/ray.py
```diff
@@ -30,6 +30,7 @@
 from ray.data.dataset_pipeline import DatasetPipeline
```
ShreyaR, line 30: An alternative way to implement the changes in this file would be to update `RayDatasetBatcher._fetch_next_batch` method. Specifically, in (this line)[https://github.com/ludwig-ai/ludwig/blob/2b81fcfc8c0329ed81ec723235cfd71384f70135/ludwig/data/dataset/ray.py#L339], you can add a check to raise `StopIteration` if the batch only has a single item.

This will allow you to remove the `_set_ignore_last` function, and have no need to pass in the `ignore_last` argument in `iter_batches`. Wdyt?

justinxzhao (approved):

ludwig/data/dataset/ray.py
arnavgarg1: Updated!

ShreyaR (approved):
lgtm!

PR #2780, opened 2022-11-22
feat: Hyperopt schema v0, part 3: Enable optional min/max support for `FloatTupleMarshmallowField` fields

ksbrar:
Lets `min` and `max` be optional kwargs on `FloatTupleMarshmallowField` (currently, you are forced to provide them or otherwise except [0,1] as your range).


justinxzhao (approved):
LGTM, but would you be able update the PR title to be more specific about what's being changed, and update the description with what+why?

connor-mccorm (approved):
LGTM

PR #2781, opened 2022-11-22
feat: Hyperopt schema v0, part 4: Add and use new hyperopt registry, search algorithm instantiation

ksbrar:
instantiate_search_algorithm + register_search_algorithm + hyperopt registry + RayTuneExecutor using registry

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -31,8 +31,8 @@ from ludwig.backend import initialize_backend, RAY
 from ludwig.backend.ray import initialize_ray
 from ludwig.callbacks import Callback
 from ludwig.constants import MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
+from ludwig.hyperopt.registry import instantiate_search_algorithm
 from ludwig.hyperopt.results import HyperoptResults, TrialResults
-from ludwig.hyperopt.search_algos import get_search_algorithm
 from ludwig.hyperopt.utils import load_json_values, substitute_parameters
 from ludwig.modules.metric_modules import get_best_function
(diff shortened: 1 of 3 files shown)


ludwig/hyperopt/registry.py
```diff
@@ -0,0 +1,22 @@
...
+        return cls
+
+    return wrap
+
+
+def get_search_algorithm_cls(name: str):
+    return search_algorithm_registry[name]
+
+
+def instantiate_search_algorithm(search_alg: dict):
```
justinxzhao, line 19: nit: `Dict`

ludwig/hyperopt/execution.py
```diff
@@ -32,7 +32,7 @@
 from ludwig.callbacks import Callback
 from ludwig.constants import MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
 from ludwig.hyperopt.results import HyperoptResults, TrialResults
-from ludwig.hyperopt.search_algos import get_search_algorithm
+from ludwig.hyperopt.search_algos import instantiate_search_algorithm  # implicit import
```
justinxzhao, line 35: Should this be

`from ludwig.hyperopt.registry import instantiate_search_algorithm`?

ludwig/hyperopt/registry.py
ksbrar: fixed

ludwig/hyperopt/execution.py
ksbrar: yup, I thought that this implicit import would work but this function is actually not imported in `search_algos` so it fails.

justinxzhao (approved):

arnavgarg1 (approved):

PR #2783, opened 2022-11-22
Proceed with model training even if saving preprocessed data fails.

justinxzhao:
Running experiments in parallel, sometimes I run into an error where saving the data as a 
parquet file fails (perhaps there's a duplicate or pre-existing file).

Perhaps it's reasonable to catch the exception and log a warning, simply skipping saving preprocessed data, and continuing to proceed to training, rather than crashing.

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1791,9 +1791,6 @@ def _preprocess_file_for_training(
     else:
         raise ValueError("either data or data_train have to be not None")
 
-    # print("backend", backend)
-    # print("data", backend.df_engine.compute(data))
-
     logger.debug("split train-val-test")
     training_data, validation_data, test_data = split_dataset(data, preprocessing_params, backend, random_seed)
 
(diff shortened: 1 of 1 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1791,17 +1791,20 @@ def _preprocess_file_for_training(
...
         logger.debug("writing split file")
         splits_df = concatenate_splits(training_data, validation_data, test_data, backend)
         split_fp = get_split_path(dataset or training_set)
-        backend.df_engine.to_parquet(splits_df, split_fp, index=True)
+        try:
+            backend.df_engine.to_parquet(splits_df, split_fp, index=True)
+        except Exception as e:
+            logger.warning(
+                f"Encountered error: '{e}' while writing data to parquet during saving processed input. "
+                "Skipping saving processed input."
```
arnavgarg1, line 1806: nit: `preprocessed` instead of `processed`?

ludwig/data/preprocessing.py
```diff
@@ -1791,17 +1791,20 @@ def _preprocess_file_for_training(
     else:
         raise ValueError("either data or data_train have to be not None")
 
-    # print("backend", backend)
-    # print("data", backend.df_engine.compute(data))
-
```
arnavgarg1, line 1796: wow can't believe this made it into Ludwig, good catch

arnavgarg1 (approved):

ludwig/data/preprocessing.py
justinxzhao: Done.

ludwig/data/preprocessing.py
arnavgarg1: Definitely does :)

PR #2784, opened 2022-11-23
feat: Hyperopt schema v0, part 5: Add basic search algorithm, scheduler, executor, and hyperopt schemas.

ksbrar:
This PR adds the schemas necessary for hyperopt v0 but does not enable validation against them. It incorporates some of the feedback from https://github.com/ludwig-ai/ludwig/pull/2346 and is based off the branch for https://github.com/ludwig-ai/ludwig/pull/2781 (so it includes those changes for now but must be updated with a merge commit once that PR is merged into master).

A longer set of notes:

There are four parts to this module. 
*`ludwig.schema.hyperopt.__init__` assembles the full hyperopt schema, in accordance with 0.6/0.7 conventions. The top level parameters are (with varying levels of validation): `output_feature` (basic string validation), `goal` (minimize or maximize), `metric` (enum validation but not fully restricted based on output feature type), `split` (train, validation, test), `search_alg` (full, separate schema validation), `executor` (full, separate schema validation), and `parameters` (simple type validation to check if it is a dict or not).
* `ludwig.schema.hyperopt.executor` - basic Ray (ONLY) executor settings
* `ludwig.schema.hyperopt.scheduler` - full schema validation with separate configs for each Ray scheduler supported in Ludwig (organized in a new `scheduler_config_registry`). Certain constants were copied over from Ray to prevent a direct Ray import here, see comments for further details. Certain parameters are marked with `TODO`(s) because they cannot be truly represented by JSON schemas (e.g. `Callable` parameters), so they may require further testing.
* `ludwig.schema.hyperopt.search_alg`. - a very basic schema that forces `search_alg` to be a dict with a `type` key (that must be set to a value registered in the `search_algorithm_registry` from above).

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -145,9 +145,6 @@ class RayTuneExecutor:
         scheduler: Optional[Dict] = None,
         **kwargs,
     ) -> None:
-        # Force-populate the search algorithm registry
-        import ludwig.hyperopt.search_algos  # noqa
-
         if ray is None:
             raise ImportError("ray module is not installed. To install it, try running pip install ray")
         self.output_feature = output_feature
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -31,8 +31,8 @@ from ludwig.backend import initialize_backend, RAY
 from ludwig.backend.ray import initialize_ray
 from ludwig.callbacks import Callback
 from ludwig.constants import MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
-from ludwig.hyperopt.registry import instantiate_search_algorithm
 from ludwig.hyperopt.results import HyperoptResults, TrialResults
+from ludwig.hyperopt.search_algos import instantiate_search_algorithm  # implicit import
 from ludwig.hyperopt.utils import load_json_values, substitute_parameters
 from ludwig.modules.metric_modules import get_best_function
 from ludwig.schema.model_config import ModelConfig
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -135,7 +135,7 @@ class RayTuneExecutor:
         metric: str,
         goal: str,
         split: str,
-        search_alg: Optional[Dict] = None,
+        search_alg: Dict,
         cpu_resources_per_trial: int = None,
         gpu_resources_per_trial: int = None,
         kubernetes_namespace: str = None,
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -31,8 +31,8 @@ from ludwig.backend import initialize_backend, RAY
 from ludwig.backend.ray import initialize_ray
 from ludwig.callbacks import Callback
 from ludwig.constants import MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
+from ludwig.hyperopt.registry import instantiate_search_algorithm
 from ludwig.hyperopt.results import HyperoptResults, TrialResults
-from ludwig.hyperopt.search_algos import instantiate_search_algorithm  # implicit import
 from ludwig.hyperopt.utils import load_json_values, substitute_parameters
 from ludwig.modules.metric_modules import get_best_function
 from ludwig.schema.model_config import ModelConfig
@@ -145,6 +145,9 @@ class RayTuneExecutor:
         scheduler: Optional[Dict] = None,
         **kwargs,
     ) -> None:
+        # Force-populate the search algorithm registry
+        import ludwig.hyperopt.search_algos  # noqa
+
         if ray is None:
             raise ImportError("ray module is not installed. To install it, try running pip install ray")
         self.output_feature = output_feature
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -145,6 +145,9 @@ class RayTuneExecutor:
         scheduler: Optional[Dict] = None,
         **kwargs,
     ) -> None:
+        # Force-populate the search algorithm registry
+        import ludwig.hyperopt.search_algos  # noqa
+
         if ray is None:
             raise ImportError("ray module is not installed. To install it, try running pip install ray")
         self.output_feature = output_feature
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -31,8 +31,8 @@ from ludwig.backend import initialize_backend, RAY
 from ludwig.backend.ray import initialize_ray
 from ludwig.callbacks import Callback
 from ludwig.constants import MAXIMIZE, TEST, TRAINER, TRAINING, TYPE, VALIDATION
+from ludwig.hyperopt.registry import instantiate_search_algorithm
 from ludwig.hyperopt.results import HyperoptResults, TrialResults
-from ludwig.hyperopt.search_algos import get_search_algorithm
 from ludwig.hyperopt.utils import load_json_values, substitute_parameters
 from ludwig.modules.metric_modules import get_best_function
 from ludwig.schema.model_config import ModelConfig
@@ -154,7 +154,7 @@ class RayTuneExecutor:
         self.search_space, self.decode_ctx = self._get_search_space(parameters)
         self.num_samples = num_samples
         self.goal = goal
-        self.search_algorithm = get_search_algorithm(search_alg)
+        self.search_algorithm = instantiate_search_algorithm(search_alg)
         self.scheduler = None if scheduler is None else tune.create_scheduler(scheduler[TYPE], **scheduler)
         self.output_feature = output_feature
         self.metric = metric
--- a/ludwig/schema/hyperopt/__init__.py
+++ b/ludwig/schema/hyperopt/__init__.py
@@ -1,0 +1,83 @@
+from abc import ABC
+from typing import Dict
+
+from marshmallow_dataclass import dataclass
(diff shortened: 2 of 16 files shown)


ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,509 @@
...
+            "passed."
+        ),
+    )
+
+
+@dataclass
+class CommonSchedulerOptions:
+    time_attr: str = time_attr_alias()
+
+    metric: Optional[str] = metric_alias()
```
ksbrar, line 93: @connor-mccorm addressing your comment again (https://github.com/ludwig-ai/ludwig/pull/2346#discussion_r1023370744), I am fine to change `metric` and `mode` to not be `Optional` - I just want to double confirm that removing `None` as an allowed value is not a breaking change.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,509 @@
...
+
+    mode: Optional[str] = schema_utils.StringOptions(
```
ksbrar, line 95: @connor-mccorm See above.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,509 @@
...
+        default=default,
+        description=(
+            "max time units per trial. Trials will be stopped after max_t time units (determined by time_attr) have "
+            "passed."
+        ),
+    )
+
+
+@dataclass
+class CommonSchedulerOptions:
```
ksbrar, line 90: @arnavgarg1 This is a class to hold some common scheduler options, re: your comment (https://github.com/ludwig-ai/ludwig/pull/2346#discussion_r1028045715). I also defined a couple of aliases to reduce verbosity.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,509 @@
...
+        schema_utils.remove_duplicate_fields(other_props)
+        preproc_cond = schema_utils.create_cond(
+            {"type": scheduler_config},
+            other_props,
+        )
+        conds.append(preproc_cond)
+    return conds
+
+
+def SchedulerDataclassField(default={"type": "fifo"}, description="Hyperopt scheduler settings."):
```
ksbrar, line 438: @arnavgarg1 (and CC @ShreyaR ) I have set the default scheduler to `fifo` as requested in a prior comment (https://github.com/ludwig-ai/ludwig/pull/2346#discussion_r1028051666).

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -0,0 +1,509 @@
...
+            if isinstance(value, dict):
+                if "type" in value and value["type"] in scheduler_config_registry:
+                    scheduler_config_cls = scheduler_config_registry[value["type"].lower()]
+                    try:
+                        return scheduler_config_cls.Schema().load(value)
+                    except (TypeError, ValidationError) as e:
+                        raise ValidationError(
+                            f"Invalid params for scheduler: {value}, see `{opt}` definition. Error: {e}"
+                        )
+                raise ValidationError(
```
ksbrar, line 464: @arnavgarg1 As you noticed in a prior comment (https://github.com/ludwig-ai/ludwig/pull/2346#discussion_r1028048497), this schema as configured would prevent empty scheduler configs. Like I said, I feel that we should restrict this as such but can I get double confirmation? Otherwise I can change it so that empty configs are accepted.

ludwig/hyperopt/execution.py
```diff
@@ -145,6 +145,9 @@ def __init__(
         scheduler: Optional[Dict] = None,
         **kwargs,
     ) -> None:
+        # Force-populate the search algorithm registry
+        import ludwig.hyperopt.search_algos  # noqa
```
ksbrar, line 149: @justinxzhao Not sure who to ping but I think you'll have a good answer. I have to import this module somewhere before search algorithm initialization so that the registry is not empty. 

In my original giant PR I had put this inside the declaration for `BaseSearchAlgorithmConfig` (see: https://github.com/ludwig-ai/ludwig/blob/26cda9d8d2a8ec747a012fe4efa92a447a39f632/ludwig/schema/hyperopt/search_algorithm.py#L13). But when breaking up that PR into parts, in #2781 I had to import it inside of the `init` for `RayTuneExecutor` (as seen here). Is there one place or another that is better for this?

justinxzhao: Does `ludwig/hyperopt/__init__.py` work?

ksbrar:
@connor-mccorm What changes (if any) do I need to make for config object integration? Do I need to add `repr=False`/`order=True` to these new schemas?

ludwig/schema/hyperopt/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    """Basic hyperopt settings."""
+
+    def get_hyperopt_metric_options():
+        all_metrics = []
+        for oftype in output_config_registry:
+            ofcls = output_config_registry[oftype]
+            all_metrics += ofcls.get_output_metric_functions().keys()
+        return all_metrics
+
+    output_feature: str = "combined"  # TODO: make more restrictive
```
arnavgarg1, line 24: Does this mean that it can be any string at the moment? That's totally ok for now, just want to make sure!

ludwig/schema/hyperopt/scheduler.py
arnavgarg1: Awesome, this is neat

ludwig/schema/hyperopt/scheduler.py
arnavgarg1: I think it might be ok to prevent empty scheduler configs, but I think there are two main things to keep in mind:
1. We should update the Ludwig docs to ensure that users know it is required, and also list out all of the options on Ludwig docs itself instead of users having to go to Ray docs.
2. (will message you)

arnavgarg1: Is my understanding correct that users must specify the scheduler config, but we also have it default to Fifo? If it must be specified, does it make sense to have a default at all?

ksbrar: @arnavgarg1 It's a bit more subtle. Specifically a user cannot set `scheduler: None` - ergo if they set a `scheduler` config they have to go all the way with the configuration. If they choose not to specify it at all then the default `scheduler` is chosen. This is the same thing we do for the optimizer, for example.

ludwig/schema/hyperopt/scheduler.py
connor-mccorm: ^^

ludwig/schema/hyperopt/scheduler.py
ksbrar: @connor-mccorm Yes, the way `metric` and `mode` are currently configured, the default is `None` regardless of whether `Optional` is there or not because `StringOptions` will set the JSON schema appropriately. But for consistency if the field supports `None` we should have the type annotation be `Optional` or `Union[str, None]`.

connor-mccorm (approved):
Looks good to me! Regarding the `Optional` I think taking it out should be fine, since it will default but ok keeping it in if you have reasoning since I think it doesn't make a huge difference either way!

ludwig/schema/hyperopt/scheduler.py
ksbrar: @arnavgarg1 yup

ludwig/schema/hyperopt/scheduler.py
arnavgarg1: @ksbrar @connor-mccorm my understanding here is that if we pass `metric` and `mode` at the top level (to RayTune), then RayTune will automatically propagate `metric` and `mode` to the scheduler and search algorithm. 
In our current implementation, we set this when we call `ray.tune.run()`, so I think it's None is a valid option for mode/metric at the Scheduler level

arnavgarg1: In our Ludwig implementation, we use `goal` in the top level hyperopt config to map to a `mode`, and `metric` from the top level hyperopt config. It looks like we default to `loss` for metric, so I think this is definitely ok and probably the right thing to do :)

ksbrar: @arnavgarg1 Okay, so keeping `None` for both is current behavior and works if I understand you correctly. Great!

That being said, I think we should remove `None` where we can as these schemas harden over time. It sounds like for now `mode` should definitely support `None` given the interaction with `goal`. We could set the default `metric` to `loss` and remove `None` as a valid option for it. WDYT?

ludwig/schema/hyperopt/scheduler.py
ksbrar: will keep this `Optional`.

ludwig/schema/hyperopt/scheduler.py
connor-mccorm: Sounds good to me!

justinxzhao (approved):

ludwig/hyperopt/execution.py
ksbrar: yup! the simplest of places in hindsight...

ludwig/schema/hyperopt/scheduler.py
ksbrar: actually I take that back about `metric` - Ray explicitly supports `None` as an important code path. So I will keep things as they are with no further changes (`None` supported by both, `Optional` remains on both annotations - CC @connor-mccorm @arnavgarg1 ).

PR #2785, opened 2022-11-23
feat: Hyperopt schema v0, part 6: Enable new hyperopt schema

ksbrar:
This branch is based off of #2784 so it includes the changes from that branch for now - once the former is merged, a merge commit should bring the affected number of files down to two: just `ludwig/schema/__init__` and the backward compatibility module.

The key changes to look at are the backward compatibility ones. In particular, I had to make two changes in the upgrade script that slightly change expectations:
- `search_alg` was previously not consistently being upgraded to a dict (e.g. it was possible to set `search_alg: 'bohb'` rather than the now schema-enforced `search_alg: {'type': 'bohb'}`). Just to be clear, is this what we want?
- It had been possible to manually set `scheduler` to be an empty dict. I don't think this should be allowed, so the upgrade script will now delete this key from the hyperopt dict if it is empty. Again, is this acceptable? @arnavgarg1

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,6 +71,6 @@ repos:
           - mdformat_frontmatter
         exclude: CHANGELOG.md
   - repo: https://github.com/yoheimuta/protolint
-    rev: v0.42.0
+    rev: v0.42.2
     hooks:
       - id: protolint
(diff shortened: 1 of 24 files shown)


connor-mccorm (approved):

arnavgarg1:
> 1. search_alg was previously not consistently being upgraded to a dict (e.g. it was possible to set search_alg: 'bohb' rather than the now schema-enforced search_alg: {'type': 'bohb'}). Just to be clear, is this what we want?

Ah good catch. Yes, we want this to always be a dict that has `type` = `something` because that's how we do the lookup in our SA registry.

> It had been possible to manually set scheduler to be an empty dict. I don't think this should be allowed, so the upgrade script will now delete this key from the hyperopt dict if it is empty. Again, is this acceptable? @arnavgarg1

Scheduler should never be an empty dict, I agree. It can be not set at all, but if it is, it should be a dict with at least a `type` key.

justinxzhao (approved):

PR #2787, opened 2022-11-23
Start fresh training run if files for resuming training are missing

arnavgarg1:
If training fails during the first epoch, then `training_progress.json` will be missing and `training_checkpoints` will not have the `latest.ckpt` file. In this case, Ludwig throws errors. However, it might be better to indicate that model resumption failed, and that a fresh model training run is being created.

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -51,6 +51,7 @@ from ludwig.trainers.registry import register_trainer
 from ludwig.utils import time_utils
 from ludwig.utils.checkpoint_utils import Checkpoint, CheckpointManager
 from ludwig.utils.defaults import default_random_seed
+from ludwig.utils.fs_utils import path_exists
 from ludwig.utils.horovod_utils import return_first
 from ludwig.utils.math_utils import exponential_decay, learning_rate_warmup, learning_rate_warmup_distributed
 from ludwig.utils.metric_utils import get_metric_names, TrainerMetric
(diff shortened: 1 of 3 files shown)


ludwig/trainers/trainer.py
```diff
@@ -1236,6 +1240,25 @@ def set_steps_to_1_or_quit(self, signum, frame):
                 signal.signal(signal.SIGINT, self.original_sigint_handler)
             sys.exit(1)
 
+    def resume_files_exist(
+        self,
+        training_progress_tracker_path: str,
+        training_checkpoint_path: str,
+    ) -> bool:
```
justinxzhao, line 1247: nit: Let's add a docstring here.

ludwig/trainers/trainer.py
```diff
@@ -1236,6 +1240,25 @@ def set_steps_to_1_or_quit(self, signum, frame):
...
+        missing_files = []
+        # training_progress.json
+        if not path_exists(training_progress_tracker_path):
```
justinxzhao, line 1250: Is there a reason why this doesn't need to be is_coordinator() guarded?

justinxzhao (approved):

tests/integration_tests/test_experiment.py
```diff
@@ -658,6 +658,44 @@ def test_experiment_model_resume(tmpdir):
...
+        if missing_file == "training_progress.json":
+            os.remove(missing_file_path)
+        else:
+            shutil.rmtree(missing_file_path)
+    finally:
+        # Training should start a fresh model training run without any errors
+        experiment_cli(config, dataset=rel_path, model_resume_path=output_dir)
+
+    predict_cli(os.path.join(output_dir, "model"), dataset=rel_path)
+    shutil.rmtree(output_dir, ignore_errors=True)
```
justinxzhao, line 696: nit: Remove? It's a tmpdir, is this automatically cleaned up?

ludwig/trainers/trainer.py
arnavgarg1: Fixed! It should be guarded

PR #2789, opened 2022-11-23
Add synthetic test data integration test utils, and use them for loss value decrease tests.

justinxzhao:


tests/integration_tests/synthetic_test_data.py
```diff
@@ -0,0 +1,71 @@
+"""Utilities for producing synthetic test data that is convergence-friendly."""
+
```
arnavgarg1, line 2: Should we move this into `integration_tests/utils.py?`. Alternatively, we can rename this to `util_synthetic_test_data.py`. I'm personally in favor of the former

tests/integration_tests/synthetic_test_data.py
```diff
@@ -0,0 +1,71 @@
...
+    input_features = [
+        {"name": "x", "type": "number"},
+    ]
+    output_features = [
+        {
+            "name": "y",
+            "type": "number",
+            "loss": {"type": "mean_squared_error"},
+            "decoder": {
+                "num_fc_layers": 5,
```
arnavgarg1, line 25: nit: Let's make this smaller so the tests that use synthetic data run quicker, I'd probably use 1 or 2 at most

tests/integration_tests/synthetic_test_data.py
```diff
@@ -0,0 +1,71 @@
+"""Utilities for producing synthetic test data that is convergence-friendly."""
+
+from collections import namedtuple
+
+import numpy as np
+import pandas as pd
+from sklearn.model_selection import train_test_split
+
+RANDOM_SEED = 42
+NUMBER_OBSERVATIONS = 500
```
arnavgarg1, line 10: Should we make this 100 to stay in line with what happens in most of the tests? They typically use 25 or 100 afaik

tests/integration_tests/synthetic_test_data.py
justinxzhao: Done.

tests/integration_tests/synthetic_test_data.py
justinxzhao: Done.

tests/integration_tests/synthetic_test_data.py
justinxzhao: Since this is a self-encapsulated set of library functions, I'd actually prefer to keep it in a separate file, also since `integration_tests/utils.py` is already quite large.

PR #2790, opened 2022-11-23
move preprocess callbacks inside model.preprocess (#2772)

jeffreyftang:
Cherry-pick #2772 into release branch.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -1280,7 +1280,7 @@ class LudwigModel:
         print_boxed("PREPROCESSING")
 
         for callback in self.callbacks:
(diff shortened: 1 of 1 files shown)


arnavgarg1:
@jeffreyftang Was this a cherry-pick after creating a branch from `release-0.6`? I'm seeing most of the tests fail because of the config_dict object, which I _believe_ didn't make it to `release-0.6` yet. cc: @connor-mccorm

connor-mccorm:
Correct, we don't have the config object code in the 0.6 release yet. I believe we were planning on moving it into a 0.7 release so may need to wait for that unless we want to pull it into a minor release to unblock this.

tgaddair (approved):

PR #2791, opened 2022-11-24
Enables Set output feature on Ray

geoffreyangus:
This PR enables the Set output feature on Ray, including validation via new tests.

The test for Set output features was previously disabled because Ray Datasets did not support postprocessing Set predictions. The Ludwig API has always post-processed Set predictions into a column of lists, where the contents of each list are the predicted positive classes in the Set for each sample. 

In general, Ray Datasets attempts to cast every column into a `TensorArray`. Also, Ray Datasets has very limited support for Ragged Tensors. Specifically, it throws an error when attempting to cast a ragged column into a `TensorArray` dtype column during `map_batches` calls. These two things in tandem formerly led to errors when attempting to post-process Set output features.

To get this feature working, the post-processing code path now temporarily disables automatic TensorArray casting on the Ray Datasets side. The columns of lists are then instead left as columns of dtype `object` when the predictions DataFrame is returned to the user and/or written out to Parquet.


ludwig/modules/metric_modules.py
```diff
@@ -503,7 +503,6 @@ class JaccardMetric(MeanMetric):
     def __init__(self, threshold: float = 0.5, **kwargs):
         super().__init__(dist_sync_fn=gather_all_tensors)
         self.threshold = threshold
-        self.add_state(name="loss", default=[], dist_reduce_fx="mean")
```
geoffreyangus, line 506: This was removed because it caused an Exception to be raised during distributed training, which then caused the metric to be omitted silently. We catch this with the changes in `test_ray.py::test_ray_outputs` by explicitly asserting the existence of the `jaccard` metric for `Set` output feature metrics.

tests/integration_tests/test_ray.py
```diff
@@ -368,12 +378,16 @@ def test_ray_outputs(dataset_type, ray_cluster_2cpu):
     input_features = [
         binary_feature(),
     ]
+    set_feature_config = set_feature(
+        decoder={"vocab_size": 3},
+        preprocessing={"missing_value_strategy": "fill_with_const", "fill_value": ""},
```
geoffreyangus, line 383: We must fill with empty strings for Set output features, otherwise NaNs will be inserted, and the default behavior for NaNs in output features is to `DROP_ROWS`, which leads to poorly handled non-determinism when comparing the metrics between the local and Ray backends.

arnavgarg1: @geoffreyangus Do you want to add that as a comment for this test? Seems useful for anyone who might change this test in the future

tests/integration_tests/utils.py
```diff
@@ -920,11 +923,26 @@ def filter(stats):
...
                             metric1, metric2, rtol=1e-03, atol=1e-04
-                        ), f"metric {name1}: {metric1} != {metric2}"
+                        ), f"metric {name1} for feature {k1}: {metric1} != {metric2}"
 
         return model
 
 
+def all_required_metrics_exist(
+    feature_to_metrics_dict: Dict[str, Dict[str, Any]], required_metrics: Optional[Dict[str, Set]] = None
+):
```
arnavgarg1, line 933: Can we add a doc string?

ludwig/modules/metric_modules.py
arnavgarg1: OOC, is there a scenario where jaccard wouldn't exist for a set output feature?

arnavgarg1: Also am curious - was an Exception raised when using RayBackend with 1 worker and > 1 worker, or just the latter?

geoffreyangus: I believe that jaccard would always exist for a set output feature (its the main metric measured for sets, other than loss: https://ludwig.ai/latest/configuration/features/set_features/)

Re: the Exception, I've just confirmed that it occurs with 1 worker as well.

arnavgarg1: Got it! i might have misunderstood what you meant by `explicitly asserting the existence of the jaccard metric for Set output feature metrics.`, but I get it now. Essentially, you're removing this line because it's causing errors, but in the tests themselves, it's not causing problems and Jaccard exists as a metric and is being computed correctly.

ludwig/data/dataframe/dask.py
```diff
@@ -155,12 +156,22 @@ def map_partitions(self, series, map_fn, meta=None):
...
 
-    def map_batches(self, series, map_fn):
+    def map_batches(self, series, map_fn, enable_tensor_extension_casting=True):
+        """Map a function over batches of a Dask Series.
+
+        Args:
+            series: Dask Series
+            map_fn: Function to apply to each batch
+            enable_tensor_extension_casting: Whether to enable tensor extension casting at the end of the Ray Datasets
+                map_batches call. This is useful in cases where the output is not supported by the ray Tensor dtype
```
justinxzhao, line 166: nit: "Should be disabled when the output tensors are not supported by the ray Tensor dtype extension, such as when the output consists of ragged tensors."

Or did I get that mixed up?

ludwig/data/dataframe/dask.py
```diff
@@ -224,20 +235,9 @@ def write_predictions(self, df: dd.DataFrame, path: str):
...
-                except TypeError:
-                    # Not a tensor, likely a string which pyarrow can handle natively
-                    pass
-            return pd.DataFrame(data)
-
-        ds = ds.map_batches(to_tensors)
-
-        fs, path = get_fs_and_path(path)
-        ds.write_parquet(path, filesystem=PyFileSystem(FSSpecHandler(fs)))
+        with tensor_extension_casting(False):
```
justinxzhao, line 238: Why is this hard-coded to False?

ludwig/data/postprocessing.py
```diff
@@ -54,7 +54,7 @@ def postprocess_batch(df):
             )
         return df
 
-    predictions = backend.df_engine.map_batches(predictions, postprocess_batch)
+    predictions = backend.df_engine.map_batches(predictions, postprocess_batch, enable_tensor_extension_casting=False)
```
justinxzhao, line 57: Is the intention for this to always be `False`, or are you planning on setting this conditioned on the output feature type?

ludwig/data/dataframe/dask.py
geoffreyangus: Similar to the below comment– I think it makes sense to have a single defined behavior when writing out to Parquet. The rationale here is that, since this function exports the data out of Ray Datasets into Parquet, there is no need to cast the data into TensorDtype before writing out to file. 

Are there particular conditions in which we might want to act otherwise? 

Regardless, it probably makes sense to have a comment here– I'll be sure to add that.

ludwig/data/postprocessing.py
geoffreyangus: Hm, I was thinking of it always being `False` so that we would have predictable/unified behavior across all output features after postprocessing. What do you think?

ludwig/modules/metric_modules.py
geoffreyangus: Right!

To refine that, I would say the issue was the following:

1. Jaccard is a metric defined for Set output features
2. During training with Ray, an exception would be raised caused by the deleted line. The exception would _then be swallowed_ and the metric would be omitted from the metrics dictionary created during evaluation.
3. I fixed this by deleting the line. I ensured the above behavior wouldn't happen again by explicitly checking the metrics dictionary to make sure the metric was present (and not omitted like before).

ludwig/data/dataframe/dask.py
geoffreyangus: That's right! Will add that as a comment.

arnavgarg1 (approved):
LGTM! 🚢

ludwig/data/dataframe/dask.py
```diff
@@ -337,3 +339,16 @@ def parallelism(self):
     @property
     def partitioned(self):
         return True
+
+
+@contextmanager
+def tensor_extension_casting(enforced: bool):
```
justinxzhao, line 345: nit: Add a docstring.

tests/integration_tests/utils.py
```diff
@@ -922,11 +925,55 @@ def filter(stats):
                         assert name1 == name2
                         assert np.isclose(
                             metric1, metric2, rtol=1e-03, atol=1e-04
-                        ), f"metric {name1}: {metric1} != {metric2}"
+                        ), f"metric {name1} for feature {k1}: {metric1} != {metric2}"
 
         return model
 
 
+def all_required_metrics_exist(
```
justinxzhao, line 933: nit: Rename to `assert_all_required_metrics_exist` or `check_all_required_metrics_exist` to make it more obvious that an assert is called.

justinxzhao (commented):
LGTM with a couple of nits, thanks!

PR #2796, opened 2022-11-28
Add new data utility functions for buffers and files, and rename registry

arnavgarg1:
This PR adds two new utility functions:
1. `read_buffer`: Reads data in from a buffer `buf` after writing the data from a buffer to a temporary file, and then processes it based on its format (hdf5, csv, tsv etc)
2. `read_fname`: This function reads data from `fname` using the `df_lib` data processing library (defaults to pandas)

### DIFF
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -40,6 +40,7 @@ from fsspec.config import conf, set_conf_files
 from pandas.errors import ParserError
 from sklearn.model_selection import KFold
 
+from ludwig.api_annotations import DeveloperAPI
 from ludwig.data.cache.types import CacheableDataset
 from ludwig.utils.dataframe_utils import from_numpy_dataset, is_dask_lib, to_numpy_dataset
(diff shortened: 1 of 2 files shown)


ludwig/utils/data_utils.py
```diff
@@ -299,6 +300,26 @@ def read_hdf5(data_fp, **kwargs):
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        temp_name = os.path.join(tmpdir, "dataset")
+        with open(temp_name, "wb") as f:
+            f.write(buf.read())
+        return reader_fn(temp_name, pd)
+
+
+@DeveloperAPI
+@spread
+def read_fname(fname, data_format=None, df_lib=pd, **kwargs):
```
justinxzhao, line 317: Can we add the descriptions from the PR description as docstrings to these functions?

May also be useful to include a comment on why or when these functions would be useful, over existing `read_*` functions.

justinxzhao (approved):

ludwig/utils/data_utils.py
arnavgarg1: Addressed feedback!

PR #2801, opened 2022-11-28
Generate structural change warnings and log_once functionality

arnavgarg1:
This PR adds two new functions to Ludwig:

1. `warn_structure_refactor` creates a warning (once per module) to let users know that the location of a module has changed from location A to location B, and that they should update their imports. 
2. `log_once` enables logging a warning just once based on the key passed in, which is very useful for things like warnings, failed image/data processing reads, etc. 

The motivation to add this functionality to help create neat deprecation warnings for module/structural level changes in the future (such as moving automl utils into the automl top level directory).


ludwig/utils/logging_utils.py
```diff
@@ -0,0 +1,32 @@
+import time
+
+_logged = set()
+_disabled = False
+_periodic_log = False
+_last_logged = 0.0
+
+
+def log_once(key):
```
justinxzhao, line 9: nit: Add bool output type hint.

Also, this isn't really logging once -- it's logging every 60 seconds. Consider changing the name to `log_every_n_seconds`, and take `n` as an input parameter (maybe default `60`), or keep the name `log_once`, but remove the time check. 

Another idea in the name of code re-use would be to have `log_once` call `log_every_n_seconds` with a very large `n`.

arnavgarg1: So, what it's actually doing is that it disables the same warning from being created for up to 60 seconds by that particular key, which actually works fine for our use case since we're rarely going to go back to the same module more than once (the key), and in particular, we will likely not raise the same warning in one model training run. 

In either case, will refactor for clarity and split between log_once and log_every_n_seconds, and also provide better examples of keys that we should use within Ludwig!

justinxzhao (approved):
LGTM! Could you update the PR description?

arnavgarg1:
> LGTM! Could you update the PR description?

Done!

PR #2802, opened 2022-11-28
AutoML libraries that use DatasetProfile instead of DatasetInfo

justinxzhao:
Changes:
- `create_auto_config_with_dataset_profile` (mirrors `create_auto_config`), which uses the `dataset_profile_pb2.DatasetProfile` (whylogs-based) to infer feature types. A few notable API modifications:
  - `dataset` and `dataset_profile` are separate parameters, and it's an error to specify both simultaneously
  - `tune_for_memory` is excluded
  - `include_hyperopt` is a new parameter that is `False` by default.

- Rename `_get_hyperopt_config` to `get_resource_aware_hyperopt_config`
- Add proto-generated go files, which enables `DatasetProfile` objects to be interfaced with in go code.
- Moves the hyperopt section of the `base_automl_config.yaml` to a separate function `get_default_automl_hyperopt`, defined in-line.


ludwig/profiling/proto/dataset_profile.proto
```diff
@@ -3,6 +3,8 @@
 //   > sudo apt install protobuf-compiler
 //   > protoc --proto_path=. --python_out=. \
 //       ludwig/profiling/proto/whylogs_messages.proto ludwig/profiling/proto/dataset_profile.proto
+//   > protoc --proto_path=. --go_out=.\
+// .     ludwig/profiling/proto/whylogs_messages.proto ludwig/profiling/proto/dataset_profile.proto
```
arnavgarg1, line 7: Is the idea here to write the proto files as both python files and go files?

tests/ludwig/automl/test_data_source.py
```diff
@@ -22,15 +22,23 @@
...
+    assert config["input_features"][1]["type"] == TEXT
+
+
+@pytest.mark.distributed
+def test_mixed_csv_data_source_with_profile(ray_cluster_2cpu):
+    temp = tempfile.NamedTemporaryFile(mode="w+")
+    temp.write(CSV_CONTENT)
+    temp.seek(0)
+    ds = read_csv(temp.name, dtype=None)
+    df = dd.from_pandas(ds, npartitions=1)
```
arnavgarg1, line 42: Can we create a common shared utility function for these to be used by both `test_mixed_csv_data_source` and `test_mixed_csv_data_source_with_profile`?

ludwig/automl/base_config.py
```diff
@@ -118,6 +127,33 @@ def _get_stratify_split_config(field_meta: FieldMetadata) -> dict:
...
 
 
+def get_default_automl_hyperopt() -> Dict[str, Any]:
+    """Returns general, default settings for hyperopt.
+
+    For example:
+    - We set a random_state_seed for sample sequence repeatability
+    - We use an increased reduction_factor to get more pruning/exploration.
+
+    TODO: If settings seem reasonable, consider building this into the hyperopt schema, directly.
```
arnavgarg1, line 137: I believe this is almost the default in the schema already, the only different being that the default scheduler is FIFO instead of async_hyperband, but I think async_hyperband is a good choice for automl config generation so we should stick to this for now.

ludwig/automl/base_config.py
```diff
@@ -118,6 +127,33 @@ def _get_stratify_split_config(field_meta: FieldMetadata) -> dict:
...
+    """
+    return yaml.safe_load(
+        """
+  search_alg:
+    type: hyperopt
+  executor:
+    type: ray
+    num_samples: 10
+    time_budget_s: 7200
```
arnavgarg1, line 146: Should we make this 3600 instead of 7200? In most cases, 3600 works well and I also believe this is the default value for async_hyperband (in the schema)

ludwig/automl/automl.py
```diff
@@ -144,6 +157,65 @@ def auto_train(
     return train_with_config(dataset, config, output_directory=output_directory, random_seed=random_seed, **kwargs)
 
 
+def create_auto_config_with_dataset_profile(
```
arnavgarg1, line 160: nit: Can we add the PublicAPI annotation?

ludwig/automl/automl.py
```diff
@@ -144,6 +157,65 @@ def auto_train(
...
+    backend: Union[Backend, str] = None,
+) -> dict:
+    """Returns the best single-shot Ludwig config given a Ludwig dataset or dataset profile.
+
+    If only the dataset is provided, then a new profile is computed.
+    Only one of the dataset or dataset_profile should be specified, not both.
+
+    This function is intended to eventually replace create_auto_config().
+    """
+    if dataset is None and dataset_profile is None:
```
arnavgarg1, line 176: Would it also make sense to add a check for `target` to make sure that it isn't None? or is that handled within the dataset_profile generation functions?

arnavgarg1 (approved):
LGTM! Just left some comments for minor nits

connor-mccorm (approved):

tests/ludwig/automl/test_data_source.py
justinxzhao: Done.

ludwig/profiling/proto/dataset_profile.proto
justinxzhao: Yes.

tests/integration_tests/test_automl.py
```diff
@@ -26,7 +26,7 @@
 
 import dask.dataframe as dd  # noqa
 
-from ludwig.automl.automl import create_auto_config, train_with_config  # noqa
+from ludwig.automl.automl import create_auto_config, create_auto_config_with_dataset_profile, train_with_config  # noqa
```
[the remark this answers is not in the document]
justinxzhao: Done.

ludwig/automl/base_config.py
justinxzhao: If async_hyperband is a clearly better option to use, then that's a good reason to make that the schema default. What are your thoughts on keeping FIFO the schema default?

ludwig/automl/base_config.py
justinxzhao: Done.

ludwig/automl/automl.py
justinxzhao: It probably makes sense to leave this unannotated as I'd like this API to harden a bit before we commit to maintaining it in public.

ludwig/automl/automl.py
justinxzhao: Done.

ludwig/automl/base_config.py
arnavgarg1: I think it's a safe change to make

ludwig/automl/automl.py
arnavgarg1: Got it! Maybe in that case, we can flag it as DeveloperAPI for now and then make it PublicAPI later?

justinxzhao: SGTM!

PR #2805, opened 2022-11-29
Add InterQuartileTransformer normalization strategy for Number Features

arnavgarg1:
This PR adds a new normalization method for numeric features called `InterQuartileTransformer`, which removes the median and scales the data according to the interquartile range. The resulting data has a zero mean and median and a standard deviation of 1, and it is not skewed by outliers (the outliers are still present with the same relative relationships to other values).

This effectively implements https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.RobustScaler.html, which is useful in cases where numeric features have large outliers that skew the distribution that emerges post min-max normalization.

This can be set in Ludwig by specifying `iq` as the normalization method in number feature preprocessing

### DIFF
--- a/ludwig/features/number_feature.py
+++ b/ludwig/features/number_feature.py
@@ -103,35 +103,37 @@ class MinMaxTransformer(nn.Module):
 
 
 class InterQuartileTransformer(nn.Module):
-    def __init__(self, q1: float = None, q3: float = None, **kwargs: dict):
+    def __init__(self, q1: float = None, q2: float = None, q3: float = None, **kwargs: dict):
         super().__init__()
         self.q1 = float(q1) if q1 is not None else q1
+        self.q2 = float(q2) if q2 is not None else q2
         self.q3 = float(q3) if q3 is not None else q3
         self.interquartile_range = self.q3 - self.q1
         self.feature_name = kwargs.get(NAME, "")
         if self.interquartile_range == 0:
             raise RuntimeError(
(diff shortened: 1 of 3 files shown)


ludwig/features/number_feature.py
```diff
@@ -102,6 +102,45 @@ def fit_transform_params(column: np.ndarray, backend: "Backend") -> dict:  # noq
...
+    def inverse_transform_inference(self, x: torch.Tensor) -> torch.Tensor:
+        return x * self.interquartile_range + self.q2
+
+    @staticmethod
+    def fit_transform_params(column: np.ndarray, backend: "Backend") -> dict:  # noqa
+        compute = backend.df_engine.compute
+        return {
+            "q1": compute(np.percentile(column.astype(np.float32), 25)),
+            "q2": compute(np.percentile(column.astype(np.float32), 50)),
+            "q3": compute(np.percentile(column.astype(np.float32), 75)),
```
justinxzhao, line 140: Just an idea for a future improvement -- we may be able to get this information from the dataset profile directly instead of having to calculate it anew during preprocessing.

arnavgarg1: We may want to calculate all of these statistics in advance then (even for the other transformers (e.g., mean, stddev, min, max, log1p etc)

arnavgarg1:
Tested this with Titanic with @justinxzhao. I commented out the `parch` feature since it has an interquartile range of 0 which would cause a DivisionByZero error. Results look reasonable, will merge.

PR #2806, opened 2022-11-29
LIghtGBMRayTrainer repartition datasets with fewer blocks than Ray actors

jeffkinnison:
This PR adds a `repartition` method to `RayDataset` and updates `LightGBMRayTrainer` to use `RayDataset.repartition` rather than accessing the underling dataset object. Additionally, it adds tests to ensure `LightGBMRayTrainer` repartitions data only if the number of blocks in the RayDataset is less than the number of actors. This addresses an error that can occur during distributed training (`AttributeError: 'Dataset' object has no attribute 'ds' (type: RayTaskError(AttributeError), retryable: true`).

### DIFF
--- a/tests/integration_tests/test_trainer.py
+++ b/tests/integration_tests/test_trainer.py
@@ -8,7 +8,7 @@ import pytest
 
 from ludwig.api import LudwigModel
 from ludwig.callbacks import Callback
-from ludwig.constants import DEFAULT_BATCH_SIZE, LIGHTGBM_TRAINER, TRAINER
+from ludwig.constants import DEFAULT_BATCH_SIZE, TRAINER
 from tests.integration_tests.utils import (
     binary_feature,
     category_feature,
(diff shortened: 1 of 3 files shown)


tests/integration_tests/test_trainer.py
```diff
@@ -173,3 +183,58 @@ def test_changing_parameters_on_plateau(tmpdir):
...
+
+
+@pytest.mark.distributed
+def test_lightgbm_dataset_partition(ray_cluster_2cpu):
+    # Create a LightGBM model with a Ray backend
+    config = {
+        "input_features": [{"name": "in_column", "type": "binary"}],
+        "output_features": [{"name": "out_column", "type": "binary"}],
+        "model_type": "gbm",
+        LIGHTGBM_TRAINER: {"epochs": 1, "batch_size": 128},
```
jppgks, line 195:
```suggestion
```

Suggesting to remove this from the config here since it doesn't seem to be used. 

Also heads up that parameters for any trainer type live on the top-level `trainer` section of the config (not `LIGHTGBM_TRAINER`). For GBM models, `epochs` and `batch_size` are not defined. Length of training can be controlled with `num_boost_round`.

jppgks (approved):
Looks good, thanks @jeffkinnison ! Minor comment

tests/integration_tests/test_trainer.py
jeffkinnison: Ah, good catch!

tgaddair (approved):

PR #2807, opened 2022-12-01
WIP AVMnist dataset

skanjila:
# Code Pull Requests

Please provide the following:

- The initial iteration of the av_mnist dataset using the torch datasetLoader like structures
- https://github.com/ludwig-ai/ludwig/issues/2638
- a test is included but very much a WIP

# Documentation Pull Requests
Will be added once this PR is complete


ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+            self.modal = modal
+            self.audio_data = None
+            self.mnist_data = None
+            self.labels = None
+            self.data = None
+            self.file_names = {
+                "train_data": "train-images-idx3-ubyte.gz",
+                "train_labels": "train-labels-idx1-ubyte.gz",
+                "test_data": "t10k-images-idx3-ubyte.gz",
+                "test_labels": "t10k-labels-idx1-ubyte.gz",
```
arnavgarg1, line 54: Should we grab these from the `config` object so that if there are any changes in the files names within the config, the loader will also use the updated names?

ludwig/datasets/configs/av_mnist.yaml
```diff
@@ -0,0 +1,18 @@
+version: 1.0
+name: avmnist
+download_urls:
+  - http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
+  - http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz
+  - http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz
+  - http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
```
arnavgarg1, line 7: nit: Add comments to indicate that these two archive files are the test set files

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+            }
+            self.transform = transform
+            self.modal_separate = modal_separate
+            self.modal = modal
```
arnavgarg1, line 58: I think these have already been defined in lines 43-45

arnavgarg1: Should we also save the cache_dir passed into the constructor?

```
self.cache_dir = cache_dir
```

arnavgarg1: Alternatively, you can skip saving them in this constructor completely and just pass them into `super().__init__()` just like you have with the cache_dir

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+        :args:
+            raw_dataset_dir (str) : the directory where all the raw data lives
+        :returns:
+            None"""
+        file_names = {
+            "train_data": "train-images-idx3-ubyte.gz",
+            "train_labels": "train-labels-idx1-ubyte.gz",
+            "test_data": "t10k-images-idx3-ubyte.gz",
+            "test_labels": "t10k-labels-idx1-ubyte.gz",
+        }
```
arnavgarg1, line 127: nit: use `self.file_names`

arnavgarg1: Actually, you can remove this completely since you're already using `self.file_names.items()` in like 130

ludwig/datasets/configs/av_mnist.yaml
```diff
@@ -0,0 +1,18 @@
+version: 1.0
+name: avmnist
+download_urls:
+  - http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
```
arnavgarg1, line 4: Seems like these are password protected?

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+        print("Raw dataset working directory: %s" % raw_dataset_dir)
+
+        for key, file_name in self.file_names.items():
+            file_path = os.path.join(raw_dataset_dir, file_name)
+            print("file: %s" % key)
+            f = open(file_path)
+            # read the definition of idx1-ubyte and idx3-ubyte
+            f.seek(4)
+            num = f.read(4)
+            num = int().from_bytes(num, "big")
```
arnavgarg1, line 137: nit: maybe rename this to `num_items` for clarity?

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+            print("size of %s : %d" % (key, num))
+            if re.match(r".*data.*", key) is not None:
```
arnavgarg1, line 139: nit: perhaps, it may be cleaner to use something like:

```
if 'data' in key:
    ....
```

Either is fine!

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+                if not os.path.exists(saved_path):
+                    os.makedirs(saved_path)
+                saved_name = key + ".npy"
+                np.save(os.path.join(saved_path, saved_name), rec)
+            else:
+                data = np.frombuffer(f.read(), np.uint8)
+
+                saved_path = os.path.join(raw_dataset_dir, "labels")
+                if not os.path.exists(saved_path):
+                    os.makedirs(saved_path)
```
arnavgarg1, line 164: nit: since you've imported `makedirs` from `fs_utils.py`, should we use that everywhere instead of `os.makedirs`? That should also account for other filesystems like S3 etc if users want to use them / if we want to allow passing in non local filesystem paths in the future!

ludwig/datasets/loaders/av_mnist.py
arnavgarg1: I also think that `self.transform` is defined but not used. Should we remove it from both the constructor args and from member variable definition?

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+    def __init__(
+        self, config: DatasetConfig, cache_dir: Optional[str] = None, transform=None, modal_separate=None, modal=None
+    ):
+        try:
+            from torchvision.io import write_png
+
+            self.write_png = write_png
+            self.transform = transform
+            self.modal_separate = modal_separate
+            self.modal = modal
```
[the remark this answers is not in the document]
arnavgarg1: Would it also be worth adding a check here to make sure that only one of the two is specified/passed into the constructor? The doc string can reflect that only one should be specified as well.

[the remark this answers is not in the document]
arnavgarg1: Would it also be worth adding a check here to make sure that only one of the two is specified/passed into the constructor? The doc string can reflect that only one should be specified as well.

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+                else:
+                    raise ValueError("the value of modal should be given")
+
+        return super().transform_files(file_paths)
+
+    def load_unprocessed_dataframe(self, file_paths: List[str]) -> pd.DataFrame:
+        """Load dataset files into a dataframe."""
+        return self.output_training_and_test_data()
+
+    def process_source_dataset(self, raw_dataset_dir="."):
```
arnavgarg1, line 116: From my understanding, it seems like this doesn't take into account either `self.modal` or `self.modal_separate` while saving the files as npy files.  So, this will always create the following directory structure:

```
raw_dataset_dir
    |-- images
        |-- train_data.npy
        |-- test_data.npy
    |-- labels
        |-- train_labels.npy
        |-- test_labels.npy
```

This by itself is great, but will break if `self.modal == 'audio'` since the file to be loaded in `transform_files` won't exist

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+                        self.labels = np.load(os.path.join(self.raw_dataset_dir, "train_labels.npy"))
+                    else:
+                        self.data = np.load(os.path.join(self.raw_dataset_dir, self.modal, "test_data.npy"))
+                        self.labels = np.load(os.path.join(self.raw_dataset_dir, "test_labels.npy"))
+
+                    if self.modal == "audio":
+                        self.data = self.data[:, np.newaxis, :, :]
+                    elif self.modal == "image":
+                        self.data = self.data.reshape(self.data.shape[0], 1, 28, 28)
+                    self.write_output_dataset(self.labels, self.data, os.path.join(self.raw_dataset_dir, dataset))
```
arnavgarg1, line 106: To avoid code duplication, would it be worth creating two private functions `_transform_and_load_image_files` and `_transform_and_load_audio_files` that look something like this:

```
def _transform_and_load_image_files(self, dataset: str) -> None:
    if dataset == "training":
        self.image_data = np.load(os.path.join(self.raw_dataset_dir, "image", "train_data.npy"))
    ....
       self.mnist_data = self.mnist_data.reshape(self.mnist_data.shape[0], 1, 28, 28)
       self.write_output_dataset(self.labels, self.mnist_data, os.path.join(self.raw_dataset_dir, dataset))
```

That way, we can just call these functions and greatly simplify `transform_files`

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+            else:
+                data = np.frombuffer(f.read(), np.uint8)
+
+                saved_path = os.path.join(raw_dataset_dir, "labels")
+                if not os.path.exists(saved_path):
+                    os.makedirs(saved_path)
+                saved_name = key + ".npy"
+                np.save(os.path.join(saved_path, saved_name), data)
+
+    def write_output_dataset(self, labels, images, output_dir):
```
arnavgarg1, line 168: nit: add type hints

ludwig/datasets/configs/av_mnist.yaml
skanjila: As I mentioned these were the same files used in the mnist dataloader

ludwig/datasets/configs/av_mnist.yaml
skanjila: added

ludwig/datasets/loaders/av_mnist.py
skanjila: Lets get the image workflow tested and working end to end first, then we'll focus on audio, just trying to get past smaller bits of functionality

ludwig/datasets/loaders/av_mnist.py
```diff
@@ -0,0 +1,214 @@
...
+                width = int().from_bytes(width, "big")
+
+                data = np.frombuffer(f.read(), np.uint8).reshape(num, height, width)
+
+                # PCA projecting with 75% energy removing
+                n_comp = int(height * width)
+                pca = PCA(n_components=n_comp)
+                projected = pca.fit_transform(data.reshape(num, height * width))
+                n_comp = ((np.cumsum(pca.explained_variance_ratio_) > 0.25) != 0).argmax()
+                rec = np.matmul(projected[:, :n_comp], pca.components_[:n_comp])
```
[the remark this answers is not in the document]
skanjila: I will dig into this a bit more and let you know my thoughts on path forward

ludwig/datasets/loaders/av_mnist.py
skanjila: Updated

ludwig/datasets/loaders/av_mnist.py
skanjila: removed , we're using self.config object to retrieve this dictionary from the yaml file

ludwig/datasets/loaders/av_mnist.py
skanjila: Done

ludwig/datasets/loaders/av_mnist.py
skanjila: Changed

ludwig/datasets/loaders/av_mnist.py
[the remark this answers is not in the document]
skanjila: Fixed

ludwig/datasets/loaders/av_mnist.py
skanjila: Added

ludwig/datasets/loaders/av_mnist.py
skanjila: Added

ludwig/datasets/loaders/av_mnist.py
skanjila: I think we need to keep this as the transforms seem to be part of the feature transformations that get done before the dataset gets ready to be plugged into the model, I will dig into the other repo to figure out where this is used and after we get images working I will add these transforms in with the audio code

ludwig/datasets/loaders/av_mnist.py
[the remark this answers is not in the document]
arnavgarg1: Hey @skanjila, I had a chance to chat with @justinxzhao and @dantreiman and we collectively felt that it's not worth keeping PCA in the dataloader module. Here's some reasons why:

1. It seems like this preprocessing step for MNIST is from [one paper](https://github.com/slyviacassell/_MFAS) only, so it feels very specific to that particular paper.
2. It's not clear whether this preprocessing strategy is state-of-the-art
3. Doing PCA to only keep the top 75% of pixels that explain the image is a preprocessing step, so it shouldn't be part of the data loading step. One thing we could do is add it within Ludwig as a preprocessing step that users can do downstream, independent of the data loader itself. 
4. As a Ludwig user, you can do the same PCA step after loading the dataset manually (when using Ludwig in a programmatic way), and it might be better to leave this choice to the end user. 

What do you think?

tests/integration_tests/test_avmnist.py
```diff
@@ -0,0 +1,271 @@
...
+    )
+    parser.add_argument("--datadir", type=str, help="data directory", default="/workspace/mfas/datasets/avmnist/")
+    parser.add_argument(
+        "--audio_cp",
+        type=str,
+        help="Audio net checkpoint (assuming is contained in checkpointdir)",
+        default="audio_model.pkl",
+    )
+    parser.add_argument(
+        "--rgb_cp",
```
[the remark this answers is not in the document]
arnavgarg1: Another thing to note is that we don't actually have formal tests in our test suite for dataset downloads or model training as such. We currently just run our tests for datasets manually. 

I'm thinking - what if just make an example script like `test_titanic_workflow` or `test_mnist_workflow`?

cc: @dantreiman can provide more context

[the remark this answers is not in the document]
skanjila: @arnavgarg1 I added the unit test, can  you please comment on that, will work on integration test next

ludwig/datasets/configs/av_mnist.yaml
```diff
@@ -0,0 +1,25 @@
...
+  t10k-labels-idx1-ubyte.gz: f7ae60f92e00ec6debd23a6088c31dbd2371eca3ffa0defaefb259924204aec6
+preserve_paths:
+  - training
+  - testing
+loader: av_mnist.AVMNISTLoader
+file_names:
+  train_data: train-images-idx3-ubyte.gz
+  train_labels: train-labels-idx1-ubyte.gz
+  test_data: t10k-images-idx3-ubyte.gz
+  test_labels: t10k-labels-idx1-ubyte.gz
```
arnavgarg1, line 22: @skanjila Don't need this anymore

ludwig/datasets/configs/av_mnist.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
+sha256:
+  t10k-images-idx3-ubyte.gz: 8d422c7b0a1c1c79245a5bcf07fe86e33eeafee792b84584aec276f5a2dbc4e6
+  train-images-idx3-ubyte.gz: 440fcabf73cc546fa21475e81ea370265605f56be210a4024d2ca8f203523609
+  train-labels-idx1-ubyte.gz: 3552534a0a558bbed6aed32b30c495cca23d567ec52cac8be1a0730e8010255c
+  t10k-labels-idx1-ubyte.gz: f7ae60f92e00ec6debd23a6088c31dbd2371eca3ffa0defaefb259924204aec6
+preserve_paths:
+  - training
+  - testing
+loader: av_mnist.AVMNISTLoader
```
arnavgarg1, line 17: Should be `av_mnist.AV_MNISTLoader`

examples/av_mnist/model_training.py
```diff
@@ -0,0 +1,57 @@
...
+
+# clean out old results
+shutil.rmtree("./results", ignore_errors=True)
+shutil.rmtree("./visualizations", ignore_errors=True)
+
+# list models to train
+list_of_model_ids = ["standard_model"]
+list_of_eval_stats = []
+
+training_set, val_set, test_set = av_mnist.load()
```
arnavgarg1, line 23: `av_mnist.load(split=True)`

PR #2815, opened 2022-12-01
refactor: Add filtering based on `model_type` for feature, combiner, and model type schemas

ksbrar:


ludwig/schema/features/utils.py
```diff
@@ -99,6 +117,12 @@ def get_output_feature_jsonschema():
         },
     }
 
+    if model_type == MODEL_GBM:
+        prune_gbm_features(schema)
+        schema["maxItems"] = 1
```
ksbrar, line 122: @justinxzhao This particular line may mean you can delete the extra validation check you're writing for gbm models to have one output feature

ludwig/schema/features/utils.py
```diff
@@ -18,15 +21,25 @@ def get_output_feature_cls(name: str):
     return output_config_registry[name]
 
 
+def prune_gbm_features(schema: Dict):
```
justinxzhao, line 24: Nit: Add a docstring.

ludwig/schema/features/utils.py
```diff
@@ -70,14 +88,14 @@ def get_input_feature_conds():
 
 
 @DeveloperAPI
-def get_output_feature_jsonschema():
+def get_output_feature_jsonschema(model_type: str = MODEL_ECD):
```
justinxzhao, line 91: nit: Remove the default argument and make users pass the model type explicitly.

Side reading: https://quuxplusone.github.io/blog/2020/04/18/default-function-arguments-are-the-devil/#conclusion

ludwig/schema/features/utils.py
justinxzhao: Cool. What would the error message from this be? If the error message is cryptic, perhaps there's an argument to keeping both around.

justinxzhao (approved):

connor-mccorm (approved):

ludwig/schema/features/utils.py
ksbrar: fixed and read

ludwig/schema/features/utils.py
ksbrar: added

ksbrar:
@justinxzhao Good question. It would be a standard JSON validation error message that would depend on the particular validator package (and version) in use. But it would read something like
```
output_features array too long / only supports length of 1
```

Not cryptic but not as human readable as it could be. So yeah, might as well keep it around for now. We might consider in the future exploring a workstream where we override JSON validation error messages within Ludwig - I believe the `jsonschema` package we use has support for this.

ludwig/schema/features/utils.py
ksbrar: @justinxzhao Good question. It would be a standard JSON validation error message that would depend on the particular validator package (and version) in use. But it would read something like
```
output_features array too long / only supports length of 1
```

Neither too cryptic nor as human readable as it could be. So yeah, might as well keep it around for now. We might consider in the future exploring a workstream where we override JSON validation error messages within Ludwig - I believe the `jsonschema` package we use has support for this.

ksbrar: @justinxzhao my bad I deleted your comment :(

PR #2817, opened 2022-12-01
Config Object Bug Fix

connor-mccorm:


ludwig/schema/utils.py
```diff
@@ -17,6 +17,8 @@
 from ludwig.schema.metadata.parameter_metadata import convert_metadata_to_json, ParameterMetadata
 from ludwig.utils.torch_utils import activations, initializer_registry
 
+RECURSION_STOP_ENUM = {"weights_initializer", "bias_initializer"}
```
justinxzhao, line 20: Let's add `norm_params` to this list as well.

justinxzhao (approved):
Thanks for the fast fix!

ludwig/schema/utils.py
connor-mccorm: Good call!

PR #2819, opened 2022-12-02
[Annotations] Encoder and Decoder Registries

arnavgarg1:
This PR creates accessor methods for encoder and decoder registries and uses the DeveloperAPI annotation. This will be the best way to retrieve these registries in the future. 

I've made the original variable paths private because there's no good way to remove them without breaking the logic where we instantiate a registry and incrementally register things.

### DIFF
--- a/ludwig/decoders/registry.py
+++ b/ludwig/decoders/registry.py
@@ -1,28 +1,37 @@
 from typing import Dict, List, Type, Union
 
+from ludwig.api_annotations import DeveloperAPI
 from ludwig.decoders.base import Decoder
 from ludwig.utils.registry import Registry
 
 decoder_registry = Registry()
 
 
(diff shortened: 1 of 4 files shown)


ludwig/decoders/registry.py
```diff
@@ -1,28 +1,37 @@
 from typing import Dict, List, Type, Union
 
+from ludwig.api_annotations import DeveloperAPI
 from ludwig.decoders.base import Decoder
 from ludwig.utils.registry import Registry
 
-decoder_registry = Registry()
+_decoder_registry = Registry()
```
justinxzhao, line 7: Will this break anything for dependent codebases?

justinxzhao (approved):

connor-mccorm (approved):

PR #2811, opened 2022-12-01
[IN PROGRESS] By default, set validation_metric to the output feature type's default validation metric.

justinxzhao:
Changes:
- Deprecate `validate_config` in `schema/__init__.py`, and move config validation to `ludwig/schema/model_config.ModelConfig`, and perform validation on top of a comprehensive config.
- Introduce auxiliary validations that checks config errors that are not easily covered by marshmallow schemas like parameter interdependencies. As these are built into schemas, auxiliary validations can be gradually removed.
- By default, set `validation_metric` to the output feature type's `default_validation_metric`. If there are multiple output features, use `validation_field=combined`, `validation_metric=loss`.
- Update docstrings for `validation_field` and `validation metric` in the schema

PR #2820, opened 2022-12-02
Set default validation metrics based on the output feature type.

justinxzhao:
Reduced version of #2811

By default, set `validation_metric` to the output feature type's `default_validation_metric`. If there are multiple output features, use `validation_field=combined, `validation_metric=loss`.

Update schema docstrings for `validation_field` and `validation metric`.


ksbrar (approved):

ludwig/schema/model_config.py
```diff
@@ -447,6 +456,15 @@ def _set_gbm_attributes(self, config_dict: ModelConfigDict) -> None:
             else:
                 raise ValidationError("GBM Models currently only support Binary, Category, and Number " "features")
 
+    def _set_validation_parameters(self):
```
tgaddair, line 459: What if `trainer.validation_field` or `traner.validation_metric` is already set in the config?

We should make sure we have a test for that scenario as well.

justinxzhao: Done.

PR #2825, opened 2022-12-06
Validate top level config keys

abidwael:
Add check that all top level keys in the config are valid


connor-mccorm:
So I believe we can actually achieve this functionality by adding `additionalProperties: False` to the top level of the json schema and then we don't need to add all this extra code.

abidwael:
Solved by https://github.com/ludwig-ai/ludwig/pull/2826

PR #2826, opened 2022-12-06
fix: Restrict allowed top-level config keys

ksbrar:
By setting `additionalProperties` to `False` at the top level of the schema, a user is prevented from entering erroneous values like `type` instead of `model_type` (or from otherwise making typos for top-level keys).

### DIFF
--- a/ludwig/schema/model_config.py
+++ b/ludwig/schema/model_config.py
@@ -124,12 +124,9 @@ class ModelConfig(BaseMarshmallowConfig):
     """
 
     def __init__(self, config_dict: ModelConfigDict):
-        print("dog")
-        print(config_dict.keys())
 
         # ===== Backwards Compatibility =====
         upgraded_config_dict = self._upgrade_config(config_dict)
-        print(config_dict.keys())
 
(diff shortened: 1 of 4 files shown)


abidwael (commented):
Could you add an error message explaining the error? This currently prints the whole schema which is very big

abidwael (approved):

ludwig/schema/model_config.py
```diff
@@ -198,6 +198,9 @@ def __init__(self, config_dict: ModelConfigDict):
         self._set_hyperopt_defaults()
 
         # ===== Validate Config =====
+        if self.model_type == MODEL_GBM:
```
ksbrar, line 201: @connor-mccorm In order to get this to work with existing tests, I have to prevent `self.combiner` from getting set/serialized in the config object when the model type is GBM. Otherwise in some tests there are (presumably empty/ignored) `combiner` configs that are not allowed by this PR. Let me know if there is a better way to do this.

connor-mccorm: This sounds good to me!

connor-mccorm (approved):

PR #2827, opened 2022-12-06
Explicitly track all metrics related to the best evaluation in the progress tracker.

justinxzhao:
Progress tracker changes:

Added:
- `checkpoint_number`
- `best_eval_metric_epoch`
- `best_eval_metric_checkpoint_number`
- `best_eval_train_metrics`
- `best_eval_validation_metrics`
- `best_eval_test_metrics`
- `best_eval_metric_steps`

Renamed:
- `best_eval_metric` -> `best_eval_metric_value`

All fields in the ProgressTracker:

```
epoch: The current epoch number.
steps: The current step of training.
batch_size: The current batch size.
tune_checkpoint_num: The hyperopt checkpoint number (Ray Tune).
checkpoint_number: The current checkpoint number.

best_eval_metric_steps: The step of training that has the best evaluation so far.
best_eval_metric_epoch: The epoch of training that has the best evaluation so far.
best_eval_metric_checkpoint_number: The checkpoint number that has the best evaluation so far.

last_improvement_steps: The number of steps since the last improvement.
last_learning_rate_reduction_steps: The training step of the last learning rate reduction.
last_increase_batch_size_steps: The training_step of the the last batch size increase.

best_eval_metric_value: The metric value of the best evaluation so far.
best_reduce_learning_rate_eval_metric:
    The metric value of the best evaluation so far, for reducing the learning rate.
best_increase_batch_size_eval_metric:
    The metric value of the best evaluation so far, for increasing the batch size.

last_learning_rate_reduction: The number of steps since the last learning rate reduction.
last_increase_batch_size: The number of steps since the last batch size increase.

last_reduce_learning_rate_eval_metric_improvement:
    The number of checkpoints since the last learning rate reduction.
last_increase_batch_size_eval_metric_improvement:
    The number of checkpoints since the last batch size increase.

num_reductions_learning_rate: The number of total reductions in learning rate.
num_increases_batch_size: The number of total increases in batch size.

train_metrics: Training metrics. <output feature name> -> <metric name> -> History of metrics.
validation_metrics: Validation metrics. <output feature name> -> <metric name> -> History of metrics.
test_metrics: Test metrics. <output feature name> -> <metric name> -> History of metrics.

best_eval_train_metrics:
    Best eval train metrics: <output feature name> -> <metric name> -> <metric value>.
best_eval_validation_metrics:
    Best eval validation metrics: <output feature name> -> <metric name> -> <metric value>.
best_eval_test_metrics:
    Best eval test metrics: <output feature name> -> <metric name> -> <metric value>.
```


tests/ludwig/utils/test_backward_compatibility.py
```diff
@@ -597,22 +599,56 @@ def test_upgrade_model_progress():
...
-
-    assert "vali_metrics" not in new_model_progress
-    assert "validation_metrics" in new_model_progress
-
-    metric = new_model_progress["validation_metrics"]["combined"]["loss"][0]
-    assert len(metric) == 3
-    assert metric[-1] == 0.59
+    assert new_model_progress == {
+        "batch_size": 64,
+        "best_eval_metric_value": 0.5,
```
tgaddair, line 604: Do we track what the best eval metric name was anywhere?

justinxzhao: The best eval metric name comes from the config -- `trainer.validation_field` and `trainer.validation_metric`.

If we can get that information from the config, that would be great since these aren't things that change over the course of a training run. cc: @hungcs

tgaddair: Hmm, seems there is a bug with this `validation_metric` field, actually. The docs claim it is set to the feature's default validation metric, but the schema code shows it defaults to LOSS:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/schema/trainer.py#L155

Can you fix this to use the correct default validation metric as part of this PR?

ludwig/api.py
```diff
@@ -523,7 +523,10 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                         logger.info(tabulate(experiment_description, tablefmt="fancy_grid"))
 
                         print_boxed("LUDWIG CONFIG")
-                        logger.info(pformat(self.config_obj.to_dict(), indent=4))
+                        logger.info(
+                            "Full config saved to: "
+                            f"{output_directory}/{experiment_name}/model/model_hyperparameters.json"
+                        )
```
arnavgarg1, line 529: Very welcome change :)

arnavgarg1:
Generally looks great to me! Thanks for being so thorough with the tests 🙏

ludwig/api.py
justinxzhao: This is a good idea, however, I believe we only have the full config object at this point in the code. I'll add a TODO for introducing the notion of explicitly set fields in the model config object so that we can reconstruct the "user's config".

tests/ludwig/utils/test_backward_compatibility.py
justinxzhao: Fixed the documentation in the schema. Changing the default validation metric to be the output feature's default_validation_metric will be addressed in https://github.com/ludwig-ai/ludwig/pull/2820.

ludwig/api.py
justinxzhao: Looks like we are tracking the user config after all! I added a small change to the ModelConfig object to include the user config as an additional attribute.

tests/ludwig/utils/test_backward_compatibility.py
justinxzhao: Resolved.

ludwig/api.py
justinxzhao: This has been landed as part of #2820

hungcs (approved):

tgaddair (approved):

PR #2828, opened 2022-12-07
Speed Up Ray Image Tests

geoffreyangus:


arnavgarg1 (approved):
LGTM! Should we wait for the test to finish running?

PR #2829, opened 2022-12-07
Use native LightGBM for intermittent eval during training

jeffkinnison:
During GBM training, we intermittently run an eval step for metrics reporting and early stopping detection. Based on a benchmarking sweep over eval batch sizes using two 1GB datasets (Criteo 1GB and HIggs) it was determined that native LightGBM eval outperformed Hummingbird-compiled pytorch models. This update refactors the in-training eval step to use LightGBM's sklearn interface and updates the default eval batch size to 1,048,576 (which should in many cases encompass the validation and test sets). Additionally, eval on the training set is now disabled by default.

### DIFF
--- a/ludwig/models/gbm.py
+++ b/ludwig/models/gbm.py
@@ -98,24 +98,24 @@ class GBM(BaseModel):
             raise ValueError("Model has not been trained yet.")
 
         if isinstance(inputs, tuple):
-            inputs, targets = inputs
+            inputs, _ = inputs
             # Convert targets to tensors.
-            for target_feature_name, target_value in targets.items():
-                if not isinstance(target_value, torch.Tensor):
-                    targets[target_feature_name] = torch.from_numpy(target_value)
-                else:
-                    targets[target_feature_name] = target_value
-        else:
-            targets = None
+            # for target_feature_name, target_value in targets.items():
+            #     if not isinstance(target_value, torch.Tensor):
+            #         targets[target_feature_name] = torch.from_numpy(target_value)
+            #     else:
+            #         targets[target_feature_name] = target_value
+        # else:
+        #     targets = None
 
         assert list(inputs.keys()) == self.input_features.keys()
 
         # Convert inputs to tensors of type float as expected by hummingbird GEMMTreeImpl.
-        for input_feature_name, input_values in inputs.items():
-            if not isinstance(input_values, torch.Tensor):
-                inputs[input_feature_name] = torch.from_numpy(input_values).float()
-            else:
-                inputs[input_feature_name] = input_values.view(-1, 1).float()
+        # for input_feature_name, input_values in inputs.items():
+        #     if not isinstance(input_values, torch.Tensor):
+        #         inputs[input_feature_name] = torch.from_numpy(input_values).float()
+        #     else:
+        #         inputs[input_feature_name] = input_values.view(-1, 1).float()
 
         # TODO(travis): include encoder and decoder steps during inference
         # encoder_outputs = {}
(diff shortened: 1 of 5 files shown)


ludwig/models/gbm.py
```diff
@@ -94,75 +94,26 @@ def forward(
...
+        if isinstance(inputs, tuple):
+            inputs, _ = inputs
+
+        assert list(inputs.keys()) == self.input_features.keys()
+
+        in_array = np.stack(list(inputs.values()), axis=0).T
 
         if output_feature.type() == NUMBER:
-            # regression
+            preds = torch.from_numpy(self.lgbm_model.predict(in_array))
```
arnavgarg1, line 110: Quick thing I've seen in the past: Torch sometimes creates warnings when we create a Tensor from the array directly rather than a copy of the array, so you may want to change these to:

``` 
preds = torch.from_numpy(np.array(self.lgbm_model.predict(in_array)), copy=True)
```

jeffkinnison: Ah, good catch! I don't think I encountered that in the benchmark, but will run a quick test to see if copying has any significant overhead.

ShreyaR, line 110: nit: Can you add some comments that talk about the expected input/output shape dimensions of preds & probs?

justinxzhao (commented):
@jeffkinnison I'm curious what was the overall speed difference you observed?

ludwig/schema/trainer.py
```diff
@@ -347,13 +347,13 @@ class GBMTrainerConfig(BaseTrainerConfig):
     # NOTE: Overwritten here to provide a default value. In many places, we fall back to eval_batch_size if batch_size
     # is not specified. GBM does not have a value for batch_size, so we need to specify eval_batch_size here.
     eval_batch_size: Union[None, int, str] = schema_utils.PositiveInteger(
-        default=1024,
+        default=1048576,
```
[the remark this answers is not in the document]
jeffkinnison: I can certainly do that, but need to run another benchmark sweep over different column sizes. The schema doesn't seem like the right place to set a varying parameter like that, though. Is there a better place to handle setting eval batch size?

jeffkinnison:
@justinxzhao Depending on the dataset size and batch size, we were seeing 5-10x speedups in some cases.

jppgks (commented):
Looks like this removes all usages of the compiled Hummingbird model. This would move all GBM inference to using native LightGBM, not just intermittent evaluation during training. I suspect that this wouldn't be compatible with TorchScript (TorchScript tests seem to be failing) which we rely on for serving.

One possibility would be to only use native LightGBM for eval during training, and to keep the Torch model for every other inference call.

ShreyaR (approved):

ludwig/models/gbm.py
```diff
@@ -94,75 +98,106 @@ def forward(
...
+                    probs = torch.logit(probs)
+                logits = probs
 
-        if output_feature.type() == NUMBER:
-            # regression
-            logits = preds.view(-1)
         else:
-            # classification
-            _, probs = preds
+            if isinstance(inputs, tuple):
```
tgaddair, line 138: I can see us wanting to refactor this to use more interfaces and less if-else statements, but this is fine for now, I think.

ludwig/models/gbm.py
```diff
@@ -94,75 +98,106 @@ def forward(
...
+                    probs = probs.transpose(2, 1)  # shape (batch_size, num_classes, 2)
 
-                # probabilities for belonging to each class
-                probs = probs[:, :, 1]  # shape (batch_size, num_classes)
+                    # probabilities for belonging to each class
+                    probs = torch.softmax(torch.logit(probs[:, :, 1]), -1)  # shape (batch_size, num_classes)
 
-            # invert sigmoid to get back logits and use Ludwig's output feature prediction functionality
-            logits = torch.logit(probs)
+                # invert sigmoid to get back logits and use Ludwig's output feature prediction functionality
```
tgaddair, line 199: Comment here applies to the line above.

ludwig/schema/trainer.py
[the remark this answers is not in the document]
tgaddair: I think we can address this with `eval_batch_size=auto` in a follow-up PR. I have a PR in progress that will allow us to find the optimal batch size for an arbitrary model, including a GBM:

https://github.com/ludwig-ai/ludwig/pull/2867/files#diff-0eb1b9b38c118d5e05a8fbeab4cd51e89eb7061794fa04333bd122a4970c45dcR15

ludwig/models/gbm.py
```diff
@@ -86,6 +88,8 @@ def compile(self):
             else {}
         )
         self.compiled_model = convert(self.lgbm_model, "torch", extra_config=extra_config)
+        yield
+        self.compiled_model = None
```
tgaddair, line 92: This should be wrapped in a try-finally (in case an exception is raised, we still set back to `None`).

ludwig/models/gbm.py
```diff
@@ -185,18 +220,15 @@ def load(self, save_path):
...
     def to_torchscript(self, device: Optional[TorchDevice] = None):
         """Converts the ECD model as a TorchScript model."""
-
-        # Disable gradient calculation for hummingbird Parameter nodes.
-        self.compiled_model.model.requires_grad_(False)
-
-        return super().to_torchscript(device)
+        with self.compile():
+            # Disable gradient calculation for hummingbird Parameter nodes.
+            device = torch.device(get_torch_device())
```
tgaddair, line 227: I think `torch.device` is redundant with `get_torch_device()` here, we typically just do `model.to(get_torch_device())`.

tgaddair (approved):
LGTM, just minor comments. I'm also curious how the affects the model save format, if it all. Does it still save to disk in the HB format?

jeffkinnison:
@tgaddair It doesn't look like we are saving the GBMs with Hummingbird: both `GBM.save` and `GBM.load` use joblib to serialize/deserialize the LightGBM models. Based on the commit log, it looks like we have been doing it that way for months now. The only update this PR makes on that front is to remove mandatory compilation from `GBM.load` for compatibility with the compile context manager.

jppgks (approved):
Really neat solution with the context manager. Thanks @jeffkinnison!

PR #2834, opened 2022-12-12
Add go module for dataset profile protos.

justinxzhao:
## Overview

Getting to this PR involved a fair amount of trial and error with `protoc`, proto `package` names, `go_package` names, `go get` in a 3rd party repository (`XYZ@latest`, `XYZ@commit`, `XYZ /dataset_profile@commit`, etc.), different placements of `go.mod` in the ludwig repo, and different go module names (thanks @jeffreyftang for pairing with me on it!)

I still don't fully understand all of the restrictions or "best practices" despite looking through the `buf` and google developer documentation, so still open to suggestions or revisting after the initial prototype, but the good news is that what we have here certainly works.

## Reading

- https://www.digitalocean.com/community/tutorials/how-to-use-go-modules
- https://github.com/protocolbuffers/protobuf/blob/main/examples/go/cmd/add_person/add_person.go
- https://developers.google.com/protocol-buffers/docs/gotutorial
- https://developers.google.com/protocol-buffers/docs/reference/go-generated#package
- https://stackoverflow.com/questions/74259855/import-protobuf-file-from-github-repository
- https://github.com/bufbuild/buf

## Empirical observations
- `go.mod` and `go` source files must be in the same directory.
- For a 3rd-party repo to successfully fetch a ludwig-defined go object with `go get XYZ`, `XYZ` must also be the path to the directory in the source repo where `go.mod` exists.
- The module name specified in Ludwig's `go.mod` file must also be consistent with the directory path where `go.mod` lives within the Ludwig repo, in order for users to fetch the module properly with `go get`.
- It seems that this restriction can be mitigated if users specifies a `replace` statement in their own go mod, but this would be a nice complexity to avoid.

## Sample usage (in a 3rd party repo)

Run `go get`.

```
> go get github.com/ludwig-ai/ludwig/ludwig/profiling/go
```

Or if there are changes not at HEAD, specify the commit:

```
> go get github.com/ludwig-ai/ludwig/ludwig/profiling@commit
```

i.e. for this PR:

```
> go get go get github.com/ludwig-ai/ludwig/ludwig/profiling/go@076845a
```

Sample go code:

```go
package main

import (
	"fmt"

	dataset_profile "github.com/ludwig-ai/ludwig/ludwig/profiling/go"
)

func main() {
	// A protocol buffer can be created like any struct.
	p := &dataset_profile.DatasetProfile{}
	p.NumExamples = 3

	fmt.Println(p)
}
```


ludwig/profiling/proto/gen-proto.sh
```diff
@@ -0,0 +1,6 @@
+protoc --go_out=. --python_out=. \
```
brightsparc, line 1: May want to consider using [betterproto](https://github.com/danielgtaylor/python-betterproto) to produce cleaner dataclasses as output instead of the default python_out - eg:
```
protoc --python_betterproto_out=. -I ./ludwig/profiling/proto/ whylogs_messages.proto dataset_profile.proto
```

pyproject.toml
```diff
@@ -7,4 +7,4 @@ order_by_type = "False"
 
 [tool.black]
 line-length = 120
-exclude = "./proto/"
+exclude = "./python/"
```
brightsparc, line 10: May not be required if using betterproto.

brightsparc (changes requested):
Suggest adding [betterproto](https://github.com/danielgtaylor/python-betterproto) to dev requirements for producing cleaner proto output.

ludwig/profiling/proto/gen-proto.sh
justinxzhao: I really like betterproto, but unfortunately it doesn't support [`Any` message types](https://developers.google.com/protocol-buffers/docs/proto3#any).

pyproject.toml
justinxzhao: Ack.

brightsparc (approved):

ludwig/profiling/proto/dataset_profile_pb2.py
```diff
@@ -18,9 +18,9 @@
   name='ludwig/profiling/proto/dataset_profile.proto',
```
jeffreyftang, line 18: Seems like this file ought to be in a `python` folder for consistency with having a `go` folder.

jeffreyftang: Same with the `whylogs_messages_pb2.py` file.

jeffreyftang (commented):
Looking forward, and assuming the plan is to transition more Ludwig data classes into proto form, would it make sense to have a single place to look for Ludwig protos, rather than splitting them by type? For example, something like `/ludwig/proto` for `*.proto` files, and `/ludwig/proto/go` for generated go files?

PR #2838, opened 2022-12-13
Fix TorchText version in tokenizers ahead of torch 1.13.0 upgrade

geoffreyangus:


justinxzhao (approved):

arnavgarg1 (approved):

geoffreyangus:
it seems like the error had to do with modifications to `test_triton.py`... After removing them, the tests seem to pass. Will have to revisit that at some point. Will merge this in the meantime.

PR #2839, opened 2022-12-13
Catch and surface Runtime exceptions during preprocessing

arnavgarg1:
In the existing implementation of `model.preprocess`, if there is an error or exception thrown, we never get to see it because  of the way our try-except-finally block is written. 

In particular, when preprocessing fails, the variables returned from `preprocess_for_training` aren't set/don't exist in scope during the finally block, which causes an `UnboundLocalError` for referencing a variable before definition. 

This PR updates the implementation to catch the exception and issue a RuntimeError since model training can never succeed if preprocessing the data itself fails.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -1335,48 +1335,50 @@ class LudwigModel:
     ) -> PreprocessedDataset:
         """This function is used to preprocess data.
 
-        # Inputs
+        # Args:
+            :param dataset: (Union[str, dict, pandas.DataFrame], default: `None`)
+                source containing the entire dataset to be used in the experiment.
+                If it has a split column, it will be used for splitting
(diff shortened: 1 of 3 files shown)


ludwig/api.py
```diff
@@ -1406,6 +1407,8 @@ def preprocess(
             (proc_training_set, proc_validation_set, proc_test_set, training_set_metadata) = preprocessed_data
 
             return PreprocessedDataset(proc_training_set, proc_validation_set, proc_test_set, training_set_metadata)
+        except Exception as e:
```
justinxzhao, line 1410: nit: Add a raises section to the docstring: https://google.github.io/styleguide/pyguide.html#doc-function-raises

justinxzhao (approved):

PR #2841, opened 2022-12-14
Allow early stop based on walltime since improvement

jppgks:
This PR:
- stores the Posix timestamp of the last improvement in ProgressTracker
- adds a convenience callback implementing `should_early_stop` that returns True if the given timeout has been reached and no improvement has been made for the given early stopping steps


tgaddair (approved):

justinxzhao (approved):
LGTM!

ludwig/utils/trainer_utils.py
```diff
@@ -163,6 +171,39 @@ def log_metrics(self):
...
+        self.early_stopping_steps = early_stopping_steps
+
+    def should_early_stop(self, trainer, progress_tracker, is_coordinator):
+        """
+        Returns:
+            True if the timeout has been reached and no improvement has been made for the given early stopping steps.
+        """
+        # Time since last improvement > timeout
+        timed_out = (
+            datetime.now() - datetime.fromtimestamp(progress_tracker.last_improvement_timestamp) > self.timeout_delta
```
tgaddair, line 196: Actually, something important here: how does this work with resume from previous training run? We may want to make sure this value is reset explicitly when the user is resuming from a previous training run.

tgaddair (changes requested):
Can we also wire this into the config alongside `early_stop`, something like:

```
trainer:
    early_stop_timeout_s = 3600
```

ludwig/utils/trainer_utils.py
jppgks: @tgaddair where does the ProgressTracker get loaded/initialized when resuming from a previous training run?

justinxzhao: `ProgressTracker` gets loaded in the trainer, here:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/trainers/trainer.py#L683

ludwig/utils/trainer_utils.py
```diff
@@ -244,6 +255,40 @@ def log_metrics(self):
...
+        self.timeout_delta = timedelta(seconds=timeout_s)
+        self.early_stopping_steps = early_stopping_steps
+
+    def should_early_stop(self, trainer, progress_tracker, is_coordinator):
+        """
+        Returns:
+            True if the timeout has been reached and no improvement has been made for the given early stopping steps.
+        """
+        # Time since last improvement > timeout
+        timed_out = (
```
tgaddair, line 280: @jppgks are you testing this comparison? I don't think you can compare the difference between two datetimes against a float:

```
a = datetime.now()
b = datetime.now()
b-a > 10
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'datetime.timedelta' and 'int'
```

I think you need `total_seconds()`. Also, let's make sure we're testing this.

tgaddair: Ah, nevermind, see you're using a `timedelta` here.

PR #2843, opened 2022-12-15
Select best batch size to maximize training throughput

tgaddair:


abidwael (approved):

ludwig/trainers/trainer.py
```diff
@@ -342,7 +347,8 @@ def train_for_tuning(
                 for output_feature_name, output_feature in self.model.output_features.items()
             }
             self.train_step(inputs, targets)
-        return self.model
+        duration_s = time.time() - start_ts
+        return batch_size / duration_s / total_steps
```
geoffreyangus, line 351: isn't this supposed to be `batch_size * total_steps / duration_s`?

For example, if you have `batch_size=30` and `total_steps=30` and `duration_s=900`, wouldn't you want to say `total_samples=batch_size*total_steps=30*30=900`, making `samples_per_sec=total_samples/duration_s=900/900`, which is 1 sample per second? Could be missing something, but that was my impression of samples per second.

ludwig/trainers/trainer.py
```diff
@@ -467,70 +473,66 @@ def tune_batch_size(
...
+                logger.info(
+                    f"Batch size {batch_size} is invalid, must be smaller than training set size "
+                    f"{len(training_set)} and less than or equal to max batch size {self.max_batch_size}"
+                )
+            return is_valid
+
+        # TODO (ASN) : Circle back on how we want to set default placeholder value
+        # Currently, since self.batch_size is originally set to auto, we provide a
+        # placeholder starting value
+        batch_size = 2
```
geoffreyangus, line 491: nit: is there anyway to support `batch_size=1`?

ludwig/trainers/trainer.py
abidwael: The way I looked at it was

- `average_duration_per_step` = `duration_s` / `total_steps`
- `time_per_one_training_example` = `average_duration_per_step` / `batch_size`
- `throughput` = 1 / `time_per_one_training_example` = `batch_size` / `duration_s` / `total_steps`

tgaddair: Yes, I think you're right. I forgot about order of operations. My original version was meant to be `batch_size / (duration_s / total_steps)`, which would be equivalent to your form.

ludwig/trainers/trainer.py
tgaddair: No, because of batch norm. We skip all batches with batch_size<2 during training.

geoffreyangus: Ah good to know, glad I asked. I'll have to update my batch size halving PR to avoid this case then. Thanks!

geoffreyangus (approved):

PR #2845, opened 2022-12-15
Make batch_size=auto the default

tgaddair:


arnavgarg1 (approved):

jppgks (approved):

ludwig/api.py
```diff
@@ -797,8 +783,29 @@ def train_online(
                 config=self.config_obj.trainer, model=self.model, random_seed=random_seed
             )
 
+            if (
+                self.config_obj.trainer.to_dict().get(BATCH_SIZE, None) == AUTO
+                or self.config_obj.trainer.to_dict().get(EVAL_BATCH_SIZE, None) == AUTO
+            ):
+                self._tune_batch_size(self._online_trainer, dataset, random_seed=random_seed)
```
geoffreyangus, line 790: I've enabled batch size tuning for online training. This seems like it would be an unexpected behavior from the user's POV (especially if they are continuously training on very small datasets). Let me know if there is a different behavior we should implement for this use case.

PR #2849, opened 2022-12-16
Extend parameter guidance documentation for regularization, and add explicit maxes to Non-Negative floats

justinxzhao:


ludwig/schema/trainer.py
```diff
@@ -165,12 +165,15 @@ class ECDTrainerConfig(BaseTrainerConfig):
...
-        default="l2", description="Type of regularization."
+        default="l2",
+        description="Type of regularization.",
+        parameter_metadata=TRAINER_METADATA["regularization_type"],
     )
 
-    regularization_lambda: float = schema_utils.FloatRange(
+    regularization_lambda: float = schema_utils.NonNegativeFloat(
         default=0.0,
         min=0,
```
ksbrar, line 175: I don't think `min` is a property of `NonNegativeFloat`

ksbrar (approved):
Could be quibbling over triviality, but adding `max` to `NonNegativeFloat` encroaches on the purpose of `FloatRange`. IMO if you know your float is non-negative (i.e. a min) and had a max in mind for it then it would seem more natural to use `FloatRange`

ludwig/schema/trainer.py
justinxzhao: Done.

ksbrar (approved):

tgaddair (approved):

PR #2853, opened 2022-12-17
Set RayBackend Config to use single worker for tests

arnavgarg1:
This PR sets the number of training workers to 1 (from 2) to speed up our Ray tests.

### DIFF
--- a/tests/integration_tests/test_ray.py
+++ b/tests/integration_tests/test_ray.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import copy
 import os
 import tempfile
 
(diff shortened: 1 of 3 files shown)


tgaddair (changes requested):
It would be good to have at least on test use 2 worker to verify distributed training is working.

justinxzhao (approved):

tests/integration_tests/test_ray.py
```diff
@@ -847,13 +847,14 @@ def test_ray_distributed_predict(tmpdir, ray_cluster_2cpu):
 
     with tempfile.TemporaryDirectory() as tmpdir:
         backend_config = {**RAY_BACKEND_CONFIG}
+        # Manually override num workers to 2 for distributed training and distributed predict
+        backend_config["trainer"]["num_workers"] = 2
```
tgaddair, line 851: This actually overrides the value in `RAY)BACKEND_CONFIG`, since you're modifying a nested dict and only performed a shallow copy on line 849. So any tests that run after this will use 2 workers.

To fix, you can change line 849 to use `copy.deepcopy` instead.

arnavgarg1: @tgaddair Ah, good catch, updated!

tgaddair (approved):

PR #2854, opened 2022-12-17
Replace `learning rate: auto` with feature type and encoder-based heuristics

abidwael:
Introduces heuristics for `learning_rate: auto`
```
if feature with large encoders (text, image, audio, sequence, timeseries):
    if HF encoder: 
        if trainable: false: 0.00002
        else 0.00001
    else: 0.0001
else: 0.001
```

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -437,10 +437,6 @@ class RayTrainerV2(BaseTrainer):
             )
         )
 
-    def tune_learning_rate(self, config) -> float:
-        trainer = RemoteTrainer(model=self.model, **self.executable_kwargs)
-        return trainer.tune_learning_rate(config)
-
     @property
     def validation_field(self):
         return self._validation_field
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -286,13 +286,9 @@ def tune_batch_size_fn(
 
 @ray.remote(max_calls=1)
 def tune_learning_rate_fn(
-    dataset: RayDataset,
     config: ModelConfigDict,
-    data_loader_kwargs: Dict[str, Any] = None,
     executable_kwargs: Dict[str, Any] = None,
     model: ECD = None,  # noqa: F821
-    training_set_metadata: TrainingSetMetadataDict = None,
-    features: Dict[str, Dict] = None,
     **kwargs,
 ) -> float:
     # Pin GPU before loading the model to prevent memory leaking onto other devices
@@ -300,18 +296,11 @@ def tune_learning_rate_fn(
     try:
         initialize_pytorch(horovod=hvd)
 
-        pipe = dataset.pipeline(shuffle=False, **data_loader_kwargs)
-        train_shard = RayDatasetShard(
-            pipe,
-            features,
-            training_set_metadata,
-        )
-
         device = get_torch_device()
         model = model.to(device)
 
         trainer = RemoteTrainer(model=model, horovod=hvd, **executable_kwargs)
-        return trainer.tune_learning_rate(config, train_shard, **kwargs)
+        return trainer.tune_learning_rate(config)
     finally:
         torch.cuda.empty_cache()
         hvd.shutdown()
@@ -470,16 +459,12 @@ class RayTrainerV2(BaseTrainer):
             )
         )
 
-    def tune_learning_rate(self, config, training_set: RayDataset, **kwargs) -> float:
+    def tune_learning_rate(self, config, **kwargs) -> float:
         return ray.get(
             tune_learning_rate_fn.options(num_cpus=self.num_cpus, num_gpus=self.num_gpus).remote(
-                dataset=training_set,
                 config=config,
-                data_loader_kwargs=self.data_loader_kwargs,
                 executable_kwargs=self.executable_kwargs,
                 model=ray.put(self.model),
-                training_set_metadata=training_set.training_set_metadata,
-                features=training_set.features,
                 **kwargs,
             )
         )
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -284,28 +284,6 @@ def tune_batch_size_fn(
         hvd.shutdown()
 
 
-@ray.remote(max_calls=1)
-def tune_learning_rate_fn(
-    config: ModelConfigDict,
-    executable_kwargs: Dict[str, Any] = None,
-    model: ECD = None,  # noqa: F821
-    **kwargs,
-) -> float:
-    # Pin GPU before loading the model to prevent memory leaking onto other devices
-    hvd = initialize_horovod()
-    try:
-        initialize_pytorch(horovod=hvd)
-
-        device = get_torch_device()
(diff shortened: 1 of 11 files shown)


ludwig/trainers/trainer.py
```diff
@@ -353,115 +365,73 @@ def train_for_tuning(
...
+            encoder_type = encoder_section.get("type")
+            encoder_is_trainable = encoder_section.get("trainable")
+            trainable_in_config.add(encoder_is_trainable)
+            encoders_in_config.add(encoder_type)
+
+        is_feature_with_large_encoder = len(features_in_config & features_with_large_encoders) > 0
+        has_huggingface_encoder = len(encoders_in_config & huggingface_encoders) > 0
+        has_trainable_encoder = any(trainable_in_config)
+
+        if is_feature_with_large_encoder:
```
tgaddair, line 425: Can you add some comments explaining the intuition or data backing these heuristics? At a glance they feel arbitrary, but I assume there's a reason these specific values were chosen.

abidwael: Comment added

ludwig/trainers/trainer.py
```diff
@@ -353,115 +365,78 @@ def train_for_tuning(
...
+        param:
+            config: Ludwig config used to train the model.
+        """
+        logger.info("Tuning learning rate...")
 
-        optimal_lr = get_optimal_lr(losses, learning_rates)
-        if optimal_lr:
-            learning_rate = optimal_lr
+        features_with_large_encoders = {TEXT, IMAGE, SEQUENCE, AUTO, TIMESERIES}
+        huggingface_encoders = {
```
tgaddair, line 377: I worry this list is going to get out of sync very quickly as we add new pretrained encoders. Maybe what we can do instead is have a set of `pretrained_encoders` in both `text_encoders.py` and `image_encoders.py` and then merge them here. That way at least it's all in one place and easier to update when a new encoder gets added.

ludwig/backend/ray.py
```diff
@@ -470,16 +459,12 @@ def tune_batch_size(
             )
         )
 
-    def tune_learning_rate(self, config, training_set: RayDataset, **kwargs) -> float:
+    def tune_learning_rate(self, config, **kwargs) -> float:
```
tgaddair, line 462: No need to run this as a `ray.remote` function anymore, since it's not doing any actual training. Maybe also rename from `tune_learning_rate` to `detect_learning_rate`.

tgaddair (approved):

ludwig/trainers/trainer.py
```diff
@@ -353,115 +366,62 @@ def train_for_tuning(
...
+            feature_defaults = defaults.get(feature_type, {})
+            encoder_section = feature_defaults.get("encoder", {})
+            encoder_type = encoder_section.get("type")
+            encoder_is_trainable = encoder_section.get("trainable")
+            trainable_in_config.add(encoder_is_trainable)
+            encoders_in_config.add(encoder_type)
+
+        is_feature_with_large_encoder = len(features_in_config & features_with_large_encoders) > 0
+        has_huggingface_encoder = len(encoders_in_config & huggingface_encoders) > 0
+        has_trainable_encoder = any(trainable_in_config)
```
justinxzhao, line 408: I would suggest changing this code a bit:

First, move these to separate helper functions in config_utils.py

```python
def has_trainable_encoder(config_dict: ModelConfigDict) -> bool:
    for feature in config_dict["input_features"]:
        feature_encoder = feature.get("encoder", {})
        encoder_type = feature_encoder.get("type")
        if encoder_section.get("trainable"):
            return True
    return False

def has_unstructured_input_feature(config_dict: ModelConfigDict) -> bool:
    for feature in config_dict["input_features"]:
        if feature.get("type", {}) in {TEXT, IMAGE, SEQUENCE, AUTO, TIMESERIES}:
            return True
    return False

def has_pretrained_encoder(config_dict: ModelConfigDict) -> bool:
    for feature in config_dict["input_features"]:
        feature_encoder = feature.get("encoder", {})
        encoder_type = feature_encoder.get("type")
        if encoder_type in get_huggingface_encoder_registry():
             return True
    return False
```

Then, in `learning_rate_utils.py` (or `learning_rate.py`), use these helpers to decide the learning rate, which I refactored a bit to [minimize nesting](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html):

```python
def get_auto_learning_rate(config_dict: ModelConfigDict) -> float:
    """Uses config heuristics to determine an appropriate learning rate.

    The main idea behind the following heuristics is that smaller learning rates are more suitable for features with larger encoders, which are typically used with unstructured features.

    Note that these are meant to be rough heuristics that are solely based on feature types and the type of the corresponding encoder.

    More factors could be taken into consideration such as model size, dataset size, batch size, number of features, etc.
    """
    if not has_unstructured_input_feature(config_dict):
        return return 0.001

    if not has_pretrained_encoder(config_dict):
        return 0.0001

    if has_trainable_encoder(config_dict):
        return 0.00001
    return 0.00002
```

ludwig/encoders/registry.py
```diff
@@ -27,15 +32,20 @@ def wrap(cls):
     return wrap
 
 
-def register_encoder(name: str, features: Union[str, List[str]]):
+def register_encoder(name: str, features: Union[str, List[str]], is_huggingface=False):
```
justinxzhao, line 35: Rename `is_huggingface` -> `is_pretrained`

tests/integration_tests/test_ray.py
```diff
@@ -786,7 +789,7 @@ def test_tune_batch_size_lr_cpu(tmpdir, ray_cluster_2cpu, max_batch_size, expect
     dataset_parquet = create_data_set_to_use("parquet", dataset_csv)
     model = run_api_experiment(config, dataset=dataset_parquet, backend_config=backend_config)
     assert model.config[TRAINER]["batch_size"] == expected_final_batch_size
-    assert model.config[TRAINER]["learning_rate"] != "auto"
+    assert model.config[TRAINER]["learning_rate"] == 0.001
```
justinxzhao, line 792: `expected_final_learning_rate`?

ludwig/utils/heuristics.py
```diff
@@ -0,0 +1,25 @@
...
+def get_auto_learning_rate(config_dict: ModelConfigDict) -> float:
+    """Uses config heuristics to determine an appropriate learning rate.
+
+    The main idea behind the following heuristics is that smaller learning rates are more
+    suitable for features with larger encoders, which are typically used with unstructured features.
+    Note that these are meant to be rough heuristics that are solely based on feature types and the
+    type of the corresponding encoder. More factors could be taken into consideration such as model
+    size, dataset size, batch size, number of features, etc.
+
+    param:
```
justinxzhao, line 14: nit: Call this "Args": https://google.github.io/styleguide/pyguide.html#244-decision

justinxzhao (approved):

justinxzhao:
Thank you!

PR #2855, opened 2022-12-17
Fix `trainable=False` to freeze all params for HF encoders

tgaddair:


arnavgarg1 (approved):
Nice, looks great!

justinxzhao (approved):
Lgtm

PR #2856, opened 2022-12-17
Evaluate training set in the training loop

tgaddair:
This PR dramatically speeds up training for large datasets that take hours to train. In the previous implementation, `evaluate_training_set` defaulted to `True`, which meant that at the start of every evaluation phase (e.g., at the end of each epoch) there would be a pass over the training set again to compute the train metrics. In practice, this ends up being about 30 - 35% of the total training wall clock time. In other words, if your training process took 3 hours to run, you could save an entire hour by disabling this evaluation phase.

Instead of computing training metrics in a separate pass, we can do it during the training loop that updates the model weights, at the end of the forward / backward pass, which is what this PR now does when `evaluate_training_set` is `False` (the new default). The only downside to this approach is that the model weights are changing each step, so the reported training metrics are actually an accumulation over different batch + model weight combinations during the epoch. So if your goal is to know how well the model performed using the weights at the end of the epoch, then this approach is an approximation.

In practice, this noisiness in the training metrics is only an issue in the first few epochs when the model weights are very random. As training progresses, the weights change less dramatically, so the difference in performance between the first step in the epoch and the last is minor, and the train metrics end up looking very close to what they would be if evaluated at the end of the epoch, but without having to pay the cost of computing them (which is dominated by the I/O).

Need to implement similar logic for GBMs. This PR only addresses ECD.

### DIFF
--- a/ludwig/models/base.py
+++ b/ludwig/models/base.py
@@ -170,6 +170,10 @@ class BaseModel(LudwigModule, metaclass=ABCMeta):
     def predictions(self, inputs):
         """Returns the model's predictions for the given inputs."""
         outputs = self(inputs)
+        return self.outputs_to_predictions(outputs)
+
+    def outputs_to_predictions(self, outputs):
+        """Returns the model's predictions given the raw model outputs."""
         predictions = {}
         for of_name in self.output_features:
             predictions[of_name] = self.output_features[of_name].predictions(outputs, of_name)
--- a/ludwig/schema/trainer.py
+++ b/ludwig/schema/trainer.py
@@ -139,8 +139,15 @@ class ECDTrainerConfig(BaseTrainerConfig):
     )
 
     evaluate_training_set: bool = schema_utils.Boolean(
-        default=True,
-        description="Whether to include the entire training set during evaluation.",
+        default=False,
+        description=(
+            "Whether to evaluate on the entire training set during evaluation. By default, training metrics will be "
+            "computed at the end of each training step, and accumulated up to the evaluation phase. In practice, "
+            "computing training set metrics during training is up to 30% faster than running a separate evaluation "
+            "pass over the training set, but results in more noisy training metrics, particularly during the earlier "
+            "epochs. It's recommended to only set this to True if you need very exact training set metrics, and are "
+            "willing to pay a significant performance penalty for them."
+        ),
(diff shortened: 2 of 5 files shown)


geoffreyangus:
Nice change. A couple of questions:

- What's the aggregation policy at the end of the epoch? I.e. are we taking a raw average, the "last"metric computed in the epoch, or a moving average? Option 1 would be a (possibly severe) underestimate of performance given the model will likely do worse at the beginning of the epoch than at the end. Option 2 would be more true to model performance in expected value, but likely very noisy, especially with small batch sizes. Option 3 would probably be best, but a little more difficult to implement.
- What happens to metrics that only make sense when evaluated against the entire dataset, i.e. AUROC and PR-curves? Right now it seems like we're not doing anything special to handle those cases. Do we compute it on every batch and then average the results at the end? If so, I'd be concerned about small batch sizes and div-by-0 errors, particularly on imbalanced datasets.

tgaddair:
@geoffreyangus good questions.

1. We are aggregating all the metrics collected during the training loop. I call this out in the docs and the PR comment, but it's acknowledged that the metrics in the earlier epochs are going to be very noisy. I personally think this is an acceptable tradeoff, as by the third epoch or so the noise will be virtually unnoticeable. It would be better to use a moving average, I agree, but that's a bit more work. And possibly we'd rather move to reporting metrics every step if we were to make a more substantive change.
2. It works the same way as if you were to do it in a separate pass over the dataset at the end of the epoch. For every batch we update the sufficient statistics of the metric, then at the end of the epoch we aggregate the statistics across workers and compute the final metric value. So it's not an "average" of batchwise metrics, it's a global metric computed piecemeal across the entire dataset.

geoffreyangus:
Got it, thanks for the clarification!

geoffreyangus (approved):

justinxzhao:
This is a great change, thanks for putting it together.

This seems like a solid improvement, and if the new training loss (and other metrics) curves seem overly noisy, we can look at last-batch or moving-average to see if that improves noise in earlier epochs, as @geoffreyangus points out.

justinxzhao (approved):

arnavgarg1 (approved):
LGTM! Great change 🚀

PR #2857, opened 2022-12-18
Add support for automatic mixed precision (AMP) training

tgaddair:
Usage:

```
trainer:
    use_mixed_precision: true
```

In our tests fine-tuning BERT with 8 Ampere GPUs, we found a 2.2x speedup using mixed precision training with similar convergence properties.

### DIFF
--- a/tests/integration_tests/test_trainer.py
+++ b/tests/integration_tests/test_trainer.py
@@ -6,6 +6,7 @@ from unittest import mock
 import numpy as np
 import pandas as pd
 import pytest
+import torch
 
 from ludwig.api import LudwigModel
 from ludwig.callbacks import Callback
@@ -18,6 +19,7 @@ from tests.integration_tests.utils import (
     number_feature,
     RAY_BACKEND_CONFIG,
     sequence_feature,
+    text_feature,
     vector_feature,
 )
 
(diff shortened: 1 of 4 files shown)


tests/integration_tests/test_trainer.py
```diff
@@ -239,3 +241,33 @@ def create_dataset(model: LudwigModel, size: int) -> RayDataset:
     assert train_ds.ds.num_blocks() == 2
     assert val_ds.ds.num_blocks() == 2
     assert test_ds.ds.num_blocks() == 2
+
+
+@pytest.mark.skipif(torch.cuda.device_count() == 0, reason="test requires at least 1 gpu")
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="test requires gpu support")
```
justinxzhao, line 247: nit: It could be worth adding this test condition to `conftest.py` if we think we will re-use this, i.e.

```python
def gpu_available():
    return torch.cuda.device_count() == 0 and torch.cuda.is_available()

@pytest.mark.skipif(not gpu_available(), reason="test requires gpus")
```

justinxzhao (approved):

tests/integration_tests/test_trainer.py
tgaddair: Agree with this, I think it would be right to address this as part of the re-enablement of GPU tests.

arnavgarg1 (commented):
Nice, it's great that we're now going to support mixed precision training!

PR #2862, opened 2022-12-20
Auto resize images for ViTEncoder when use_pretrained is True or False

arnavgarg1:
The ViTEncoder requires images of the same dimension. However, if `width` and `height` are not specified, we auto-infer the image dimensions. In situations where the images are not already in square shape (say the images are of dimensions 480 x 720), we may infer different widths and heights (for e.g., width = 205 and height = 227). This causes issues when using the ViTEncoder since it expects square images. Additionally, when  using a pretrained ViTEncoder (where `use_pretrained=True`),we require images with an exact shape of 224 x 224 since this is the image dimension the model was pretrained on.

To ensure we always get square-shaped images when using ViT, this PR adds 3 new attributes to the image encoder schema:
- `requires_equal_dimensions`: Whether the encoder requires images with equal dimensions or not
- `required_width`: If the encoder requires images to have a specific width
- `required_height`: If the encoder requires images to have a specific height

These are used during preprocessing to shape images into the right dimensions if needed. They are currently marked as internal only parameters since the users shouldn't have to set these for their models to work correctly.


arnavgarg1:
> Am I understanding correctly that these new requirements are not meant to be changed after they're set on the respective schemas? 

I don't see a scenario where these parameters would change for our existing image encoders.

> And that they're exclusively used for coercing image sizes?

That's right, exclusively used to coerce image sizes because some encoders may require strict restrictions on input dimensions.

ksbrar (approved):
@arnavgarg1 Gotcha. I'll approve this PR from a schema perspective, but it spawned some thoughts:

So, I realize that these internal parameters are a pattern we've established and that we are eventually going to move away from, but it still feels quite odd to me to add static parameters onto these configs that will never change and are _only_ used for coercion.

Two questions:
1. (Unrelated to this PR) Does the current config object filter out internal only parameters if a user accidentally sets them? (CC @connor-mccorm )
2. I would prefer to further distinguish these parameters somehow. Easiest way would be to leave a comment. Second way might be to create auxiliary field helpers that only allow the parameter to be set to one value (e.g. like `ProtectedString`). Third option is to experiment with `ClassVar` and let these live as class variables on the schemas but not as actual parameters (though this would require more code changes).

ShreyaR (approved):

arnavgarg1:
> @arnavgarg1 Gotcha. I'll approve this PR from a schema perspective, but it spawned some thoughts:
> 
> So, I realize that these internal parameters are a pattern we've established and that we are eventually going to move away from, but it still feels quite odd to me to add static parameters onto these configs that will never change and are _only_ used for coercion.
> 
> Two questions:
> 
> 1. (Unrelated to this PR) Does the current config object filter out internal only parameters if a user accidentally sets them? (CC @connor-mccorm )
> 2. I would prefer to further distinguish these parameters somehow. Easiest way would be to leave a comment. Second way might be to create auxiliary field helpers that only allow the parameter to be set to one value (e.g. like `ProtectedString`). Third option is to experiment with `ClassVar` and let these live as class variables on the schemas but not as actual parameters (though this would require more code changes).

All of these are great points! 

I like the idea of creating auxiliary field helpers that only allow the parameter to be set to one value (particularly for `required_width` and `required_height`). 

What if we create a similar field to ProtectedString called ProtectedInteger?

```
@DeveloperAPI
def ProtectedInteger(
    pint: int,
    description: str = "",
    parameter_metadata: ParameterMetadata = None,
):
    """Alias for a `IntegerOptions` field with only one option."""
    return IntegerOptions(
        options=[pint],
        default=pint,
        allow_none=False,
        description=description,
        parameter_metadata=parameter_metadata,
    )
```

Would that suffice?

tgaddair (changes requested):
Putting it in the schema feels odd since it's then exposed to the user. Wouldn't we just want to do this automatically?

There is already precedent for these encoder-dependent preprocessing params with HF tokenizers / encoders.

Check this out: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1292

In other words, if the encoder requires certain preprocessing values, override user selections.

The one limitation of this approach is that it is not dependent on encoder params, but it can easily be extended to be so!

Imagine this API instead:

```
class VitEncoder:
    @classmethod
    def get_fixed_preprocessing_params(cls, encoder_params):
        if not encoder_params.get("trainable", False):
            return {"image_width": 224, "image_height": 224}
        ...
```

What do you think?

ksbrar:
@arnavgarg1 That sounds fine to me. I would also prefer Travis's suggestion :)

justinxzhao:
Hi @arnavgarg1, thanks for raising the issue!

Agreed with @tgaddair that the schema doesn't seem like the right place for this information. 

One additional suggestion is that if it seems like there are many image encoders that share needing to set these fixed preprocessing parameters, it could be worth defining a new base class:

```python
class ImageEncoder(Encoder):

    @classmethod
    def requires_equal_dimensions(cls) -> bool:
        raise NotImplementedError

    @classmethod
    def required_width(cls) -> Optional[int]:
        raise NotImplementedError

    @classmethod
    def required_height(cls) -> Optional[int]:
        raise NotImplementedError

    @classmethod
    def get_fixed_preprocessing_params(cls, encoder_params):
        if cls.requires_equal_dimensions() and cls.required_width() != cls.required_height():
            raise ValueError("Invalid definition.")
        if not encoder_params.get("trainable", False):
            return {"image_width": cls.required_width(), "image_height": cls.required_height()}
        return {"requires_equal_dimensions": cls.requires_equal_dimensions()}
```

arnavgarg1:
@tgaddair and @justinxzhao great suggestions! 

I've modified this implementation to include:
1. A new `ImageEncoder` base class like Justin suggested
2. A new condition during preprocessing to check if the encoder class has the `get_fixed_preprocessing_params` class method and use it to update the preprocessing_parameters dictionary.
3. Simplified implementation in `image_feature.py` to only resize images if `requires_equal_dimensions` is True and it is trainable/not using a pretrained model

ludwig/data/preprocessing.py
```diff
@@ -1293,11 +1293,18 @@ def build_preprocessing_parameters(
             if TYPE in feature_config[ENCODER]:
                 encoder_class = get_encoder_cls(feature_config[TYPE], feature_config[ENCODER][TYPE])
                 if hasattr(encoder_class, "fixed_preprocessing_parameters"):
+                    # Update preprocessing parameters for text encoders using fixed_preprocessing_parameters
                     encoder_fpp = encoder_class.fixed_preprocessing_parameters
 
                     preprocessing_parameters = merge_dict(
                         preprocessing_parameters, resolve_pointers(encoder_fpp, feature_config, "feature.")
                     )  # TODO(Connor): Temporary fix, refactor this during preproc refactor
+                elif callable(getattr(encoder_class, "get_fixed_preprocessing_params", None)):
```
tgaddair, line 1302: Instead of doing this, we should use good object-oriented principles and have a function called `get_fixed_preprocessing_params` that exists in the base Encoder class. This can return `{}`. Then we override this function in subclasses.

Similarly, this new function would subsume the previous `fixed_preprocessing_parameters` attribute above, so the code would be simplified to:

```
preprocessing_parameters = merge_dict(
                        preprocessing_parameters, encoder_class.get_fixed_preprocessing_params(feature_config[ENCODER])
                    )
```

And we can remove these two `if-else` blocks. Make sense?

connor-mccorm:
I don't think it does @ksbrar!

ksbrar:
@connor-mccorm I feel like we should add that, even if its only a short-term (potential) problem

ludwig/data/preprocessing.py
arnavgarg1: @tgaddair Totally agree, thanks for pointing this out. This is the cleanest way forward! Will update

ludwig/encoders/image_encoders.py
```diff
@@ -36,9 +36,40 @@
 logger = logging.getLogger(__name__)
 
 
+@DeveloperAPI
+class ImageEncoder(Encoder):
+    @classmethod
+    def requires_equal_dimensions(cls) -> bool:
```
tgaddair, line 42: This is great! To reduce boilerplate for non-ViT, you could have these methods default to `False`, `None`, and `None`. That way you only need to override for the ones that require it.

ludwig/encoders/text_encoders.py
```diff
@@ -297,19 +286,17 @@ def output_shape(self) -> torch.Size:
     def input_dtype(self):
         return torch.int32
 
+    @classmethod
+    def get_fixed_preprocessing_params(cls, encoder_params: Optional[Dict]) -> Dict[str, Union[None, int]]:
+        return {
+            "tokenizer": "hf_tokenizer",
+            "pretrained_model_name_or_path": "google/mt5-base",
```
tgaddair, line 293: It looks like we dropped the `feature.pretrain_model_name_or_path` part here. I'm not sure how that was being used before. Can you explain why it was okay to drop that part?

tgaddair (commented):
Looks great! Just a couple small things left.

tgaddair:
@arnavgarg1 one other thing I forgot about here: since the encode params can now affect the preprocessing, we also need to update the `calculate_checksum` function to take this information into consideration when determining whether we can reuse previously computed preprocessed data or not.

Specifically, you can update it like so:

```
fixed_preprocessing_params = [
    get_fixed_preprocessing_params(feature)
    for feature in features
]
info = {
    ...
    "fixed_preprocessing_params": fixed_preprocessing_params,
}
```

Where `get_fixed_preprocessing_params` is a utility function that gets the fixed preprocessing params from a feature dictionary.

ludwig/encoders/text_encoders.py
tgaddair: Ah, I see, it's needed for `build_sequence_matrix`. So this is actually a place where your improved interface will be very useful.

tgaddair:
@arnavgarg1 this ended up being useful for a PR I'm working on, so went ahead and commandeered it to implement the suggested changes.

arnavgarg1:
@tgaddair Awesome, thanks for doing that! Is this PR good to merge? I see one failing test where the assert fails related to learning rate when using HF trainable encoder that I assume is related to some of Wael's recent work that was merged

arnavgarg1:
> @arnavgarg1 one other thing I forgot about here: since the encode params can now affect the preprocessing, we also need to update the `calculate_checksum` function to take this information into consideration when determining whether we can reuse previously computed preprocessed data or not.
> 
> Specifically, you can update it like so:
> 
> ```
> fixed_preprocessing_params = [
>     get_fixed_preprocessing_params(feature)
>     for feature in features
> ]
> info = {
>     ...
>     "fixed_preprocessing_params": fixed_preprocessing_params,
> }
> ```
> 
> Where `get_fixed_preprocessing_params` is a utility function that gets the fixed preprocessing params from a feature dictionary.

Great call out, would've definitely missed that

tgaddair:
@arnavgarg1 @connor-mccorm I believe this is actually a bug in the config object. When a default encoder type specified, it's not overriding the default encoder type, from what I can tell.

tgaddair (approved):

PR #2863, opened 2022-12-20
support stratified split with low cardinality features

abidwael:
This PR enables using stratified split even when we have a training example with cardinality = 1 in the column we're splitting on. It achieves this by separating out the examples with cardinality = 1, splitting as usual (stratified) then appending these items back to the dataframe.

### DIFF
--- a/ludwig/data/split.py
+++ b/ludwig/data/split.py
@@ -40,7 +40,6 @@ split_registry = Registry()
 default_random_seed = 42
 logger = logging.getLogger(__name__)
 
-
 TMP_SPLIT_COL = "__SPLIT__"
 DEFAULT_PROBABILITIES = (0.7, 0.1, 0.2)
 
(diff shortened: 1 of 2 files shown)


ludwig/data/split.py
```diff
@@ -152,31 +151,36 @@ def get_schema_cls():
...
-        relative_frac_test = frac_test / (frac_val + frac_test)
-        df_val, df_test, _, _ = train_test_split(
-            df_temp, y_temp, stratify=y_temp, test_size=relative_frac_test, random_state=random_seed
-        )
-    except ValueError:
-        raise ValueError(
-            "Stratified splitting cannot be performed because the least populated class has too few examples. "
-            "Consider adding more data, consolidating rare classes by setting vocab_size, or using a random split."
-        )
+    def _safe_stratify(df, column, test_size):
```
justinxzhao, line 164: Let's add a test to verify this new functionality.

abidwael: Added

justinxzhao (approved):

PR #2866, opened 2022-12-21
Remove `_to_tensors_fn` from Ray Datasets

geoffreyangus:
Since Ludwig is now on Ray 2.0, we no longer need to call `to_tensors` on the Ray Datasets objects, which should speed up overall training throughput.


tgaddair (approved):

geoffreyangus:
The failing test had to do with some weird behaviors with the `assign` op here: https://github.com/ludwig-ai/ludwig/blob/a0461a84c9a690001fb8c25a2dbbc0f8b557a8a3/ludwig/features/binary_feature.py#L365

Seems like it was fixed by using a `map` call. Timed it in order to confirm there is no regression in performance. It actually seems like it improves performance. Results:

`remove_to_tensors_fn@b823d26479bfe8794d9162253bbef9c39cfddf3f`:
```
(_map_block_nosplit pid=6254) time of assign operation: 0.0016100406646728516
```

`master@3fe41ad7fae57456d6303066dc2ac96a36da428b`
```
(_map_block_nosplit pid=74821) time of assign operation: 0.0029959678649902344
```

PR #2868, opened 2022-12-21
fix: load spacy model for lemmatization in `EnglishLemmatizeFilterTokenizer` to work

abidwael:
`nlp_pipeline.tokenizer(text)` only loads the spacy tokenizer, not the full model. This causes `EnglishLemmatizeFilterTokenizer` to return a list of empty strings since inferring the lemmas requires the model not just the tokenizer.


arnavgarg1 (approved):
LGTM!

PR #2870, opened 2022-12-21
Fix CLIPTokenizer initialization for torchtext 0.12.0

arnavgarg1:
In torchtext 0.12.0, the CLIPTokenizer class takes `vocab_bpe_path` instead of `merges_path` as the input parameter. I found this when running `tests/ludwig/features/test_sequence_features.py::test_sequence_preproc_module_clip_tokenizer` locally.

Torchtext 0.12: https://github.com/pytorch/text/blob/orig/release/0.12/torchtext/transforms.py#L312
Torchtext 0.13: https://github.com/pytorch/text/blob/orig/release/0.13/torchtext/transforms.py#L385

### DIFF
--- a/ludwig/utils/tokenizers.py
+++ b/ludwig/utils/tokenizers.py
@@ -920,7 +920,7 @@ try:
     import torchtext
 
     torchtext_version = torch.torch_version.TorchVersion(torchtext.__version__)
-    torchtext_112 = torchtext_version == (0, 12, 0)
+    torchtext_113 = torchtext_version < (0, 13, 0)
 
     if torchtext_version >= (0, 12, 0):
         """torchtext 0.12.0 tokenizers.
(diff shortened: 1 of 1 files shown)


ludwig/utils/tokenizers.py
```diff
@@ -996,9 +999,14 @@ def __init__(
...
             def _init_tokenizer(self, pretrained_model_name_or_path: str, vocab_file: str):
-                return torchtext.transforms.CLIPTokenizer(
-                    encoder_json_path=vocab_file, merges_path=pretrained_model_name_or_path
-                )
+                if torchtext_112:
+                    return torchtext.transforms.CLIPTokenizer(
+                        encoder_json_path=vocab_file, vocab_bpe_path=pretrained_model_name_or_path
+                    )
+                else:
+                    return torchtext.transforms.CLIPTokenizer(
```
tgaddair, line 1007: If I understand correctly, this param was changed in `0.12.1`?

arnavgarg1: @tgaddair Afaik, torchtext only has 0.12 and 0.13 with no minor versions in between. So this changed in 0.13. What's interesting is that the docs show that these parameters are the same, but the actual underlying implementation has different parameters.

In either case, idk why the two tests are failing since I have torch 1.11 and torchtext 0.12 locally and the tests pass with this change (but were failing on my local before)

tgaddair: Something to consider: they could go back and add a patch release for a security vulnerability, etc. So safest thing to do is to have a separate branches for `< 0.13` and `>= 0.13`.

tgaddair: Looks like the failing test is for Python 3.7 only.

arnavgarg1: Agreed, changed to branch on 0.13. It's weird that it's failing for Py37 since I have the exact same dependencies locally (torch 1.11 with torchtext 0.12) and it works

PR #2871, opened 2022-12-21
Add `preprocessing.most_common_percentile`

abidwael:
Add `most_common_percentile` parameter in addition to `most_common`. Specifying `most_common_percentile: 0.9` will keep the top 90% most frequent tokens in the vocab. `most_common` is set to `None` by default. If both `most_common` and `most_common_percentile` are set, `most_common` overrides.


ludwig/schema/features/preprocessing/bag.py
```diff
@@ -54,10 +54,14 @@ class BagPreprocessingConfig(BasePreprocessingConfig):
         parameter_metadata=FEATURE_METADATA[BAG][PREPROCESSING]["lowercase"],
     )
 
-    most_common: int = schema_utils.PositiveInteger(
+    most_common: int = schema_utils.OneOfOptionsField(
```
ksbrar, line 57: Make sure to make the type annotation `Union[int, float]`

But if the value can be an integer or a float, I think it would be simpler to just let the type be a float and if the user enters a whole number to just treat it as an integer. WDYT?

ksbrar (changes requested):

ludwig/schema/features/preprocessing/bag.py
abidwael: Good catch! I think `Union[int, float]` is clearer than just having `float` in this case.

ksbrar (approved):

ludwig/schema/features/preprocessing/bag.py
```diff
@@ -54,10 +56,14 @@ class BagPreprocessingConfig(BasePreprocessingConfig):
         parameter_metadata=FEATURE_METADATA[BAG][PREPROCESSING]["lowercase"],
     )
 
-    most_common: int = schema_utils.PositiveInteger(
+    most_common: Union[int, float] = schema_utils.OneOfOptionsField(
```
justinxzhao, line 59: I would add this as a separate parameter instead of overloading this existing one, as to not change the semantics of the pre-existing field. It should be rather straightforward to add a validation check (perhaps as a follow-up PR) to make sure that both parameters aren't explicitly specified at the same time. 

Using a separate parameter also disambiguates what `1` means -- does it mean 100% of the vocabulary or call all other tokens besides the most common as "rare"? And each parameter can be more clearly documented.

justinxzhao (changes requested):
Suggestion: use a second parameter, `most_common_percentile`

ludwig/schema/features/preprocessing/bag.py
tgaddair: I agree with making it a separate param, though it will complicate things a little bit, since the default for `most_common` would need to be changed to `None` to allow for specifying `most_common_fraction` (or whatever it's called) without explicitly specifying `most_common=null`.

tgaddair: It also means the meaning of `most_common=null` changes from "use all tokens" to "use the default". If the user wanted to use all tokens, they would then need to set `most_common_fraction=1` instead.

connor-mccorm (commented):
I agree that a second parameter here is probably the best idea!

abidwael:
Added a new parameter `most_common_percentile`. I set the defaults to
- `most_common_percentile`: 0.95
- `most_common`: None

Previously, we set a hard coded number that will truncate the vocab. But this is suboptimal especially that as datasets grow, this default number will be too small. Therefore, it makes more sense to set this as a fraction and make `most_common_percentile` should be the defacto way of truncating the vocab size.

`most_common_percentile` can't be null but `most_common` will override `most_common_precentile` if specified: if a user specifies this, then it's reasonable to assume that they know the original vocab size and they'd like to truncate it to a smaller number.

Thoughts?

ludwig/schema/metadata/configs/features.yaml
```diff
@@ -166,7 +166,30 @@ bag:
                 important rare tokens are, 90%, 95% or 99% of the number of tokens
                 will leave out only very rare tokens that should not influence performance
                 substantially
-            ui_display_name: Most common (vocabulary size)
+        most_common_percentile:
```
justinxzhao, line 169: I'm not a fan of duplicating code or documentation as it drastically increases the maintenance difficulty and the chance of drift.

Suggestion: Rewrite the metadata to be feature-type generic, decide on once place where the metadata should live (perhaps here, or in a new file i.e. `shared.yaml`), and remove all other copies.

justinxzhao (commented):
LGTM, with one comment about how to organize the parameter metadata.

ludwig/schema/features/preprocessing/category.py
```diff
@@ -46,13 +46,24 @@ class CategoryPreprocessingConfig(BasePreprocessingConfig):
...
-        default=10000,
+        default=None,
         allow_none=True,
         description="The maximum number of most common tokens to be considered. if the data contains more than this "
         "amount, the most infrequent tokens will be treated as unknown.",
         parameter_metadata=FEATURE_METADATA[CATEGORY][PREPROCESSING]["most_common"],
     )
 
+    most_common_percentile: int = schema_utils.FloatRange(
+        default=1,
```
tgaddair, line 57: Should this be 0.95?

ludwig/schema/features/preprocessing/bag.py
```diff
@@ -55,9 +55,20 @@ class BagPreprocessingConfig(BasePreprocessingConfig):
...
-        default=10000,
+        default=None,
         allow_none=True,
         description="The maximum number of most common tokens to be considered. If the data contains more than this "
         "amount, the most infrequent tokens will be treated as unknown.",
         parameter_metadata=FEATURE_METADATA[BAG][PREPROCESSING]["most_common"],
     )
+
+    most_common_percentile: int = schema_utils.FloatRange(
+        default=1,
```
tgaddair, line 66: 0.95?

ludwig/schema/features/preprocessing/bag.py
```diff
@@ -55,9 +55,20 @@ class BagPreprocessingConfig(BasePreprocessingConfig):
...
+        min=0.0,
+        max=1.0,
+        min_inclusive=False,
+        allow_none=False,
```
tgaddair, line 70: How does this interact with `most_common` if we can't set to null?

ludwig/schema/features/preprocessing/set.py
```diff
@@ -53,13 +53,24 @@ class SetPreprocessingConfig(BasePreprocessingConfig):
...
-        default=10000,
+        default=None,
         allow_none=True,
         description="The maximum number of most common tokens to be considered. If the data contains more than this "
         "amount, the most infrequent tokens will be treated as unknown.",
         parameter_metadata=FEATURE_METADATA[SET][PREPROCESSING]["most_common"],
     )
 
+    most_common_percentile: int = schema_utils.FloatRange(
+        default=1,
```
tgaddair, line 64: 0.95?

ludwig/schema/features/preprocessing/category.py
abidwael: My rationale is that since the vocab size of category, bag, set features is small, I set these to 1.

ludwig/schema/features/preprocessing/bag.py
abidwael: Add `most_common_percentile` parameter in addition to `most_common`. Specifying `most_common_percentile: 0.9` will keep the top 90% most frequent tokens in the vocab. `most_common` is set to `None` by default. If both `most_common` and `most_common_percentile` are set, `most_common` overrides.

ludwig/schema/metadata/configs/features.yaml
abidwael: Made the change

ludwig/schema/features/preprocessing/category.py
abidwael: Updated

justinxzhao (approved):

PR #2880, opened 2022-12-24
Save GBM after training if not previously saved

jeffkinnison:
GBM models save during training if one of two conditions is met:

1. [No validation set was provided](https://github.com/ludwig-ai/ludwig/blob/gbm-not-saved/ludwig/trainers/trainer_lightgbm.py#L317-L329)
2. [The model improved since the last evaluation](https://github.com/ludwig-ai/ludwig/blob/gbm-not-saved/ludwig/trainers/trainer_lightgbm.py#L421-L429)

If neither condition is met--no validation set and the model never improved--the GBM will never be saved, and a `FileNotFoundError` is thrown [when loading the best model after training](https://github.com/ludwig-ai/ludwig/blob/gbm-not-saved/ludwig/trainers/trainer_lightgbm.py#L688-L690).

This update adds a method to check whether or not the GBM has been saved, and a post-training check that saves the model if it was never saved.

### DIFF
--- a/ludwig/models/gbm.py
+++ b/ludwig/models/gbm.py
@@ -15,6 +15,7 @@ from ludwig.globals import MODEL_WEIGHTS_FILE_NAME
 from ludwig.models.base import BaseModel
 from ludwig.schema.model_config import ModelConfig, OutputFeaturesContainer
 from ludwig.utils import output_feature_utils
+from ludwig.utils.fs_utils import path_exists
 from ludwig.utils.torch_utils import get_torch_device
 from ludwig.utils.types import TorchDevice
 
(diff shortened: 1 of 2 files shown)


ludwig/models/gbm.py
```diff
@@ -224,6 +224,9 @@ def to_torchscript(self, device: Optional[TorchDevice] = None):
             trace = super().to_torchscript(device)
         return trace
 
+    def has_saved(self, save_path):
+        return os.path.exists(os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME))
```
tgaddair, line 228: Can you use path_exists so it works with remote file systems as well?

tgaddair (approved):
Looks good, one comment

PR #2890, opened 2022-12-29
Added DistributedStrategy interface with support for DDP

### LINKED ISSUE #2886: Possible using torch DDP(DistributedDataParallel)?
How could we train model with torch DDP?

tgaddair:
Closes #2886.

Usage:

```
backend:
  trainer:
    strategy: ddp
```

This will allow us to use PyTorch 2.0 model compilation with distributed training:

https://pytorch.org/get-started/pytorch-2.0/

Caveats:
- Will hang if an exception raised in only one of the training workers due to [this](https://discuss.ray.io/t/torchtrainer-hangs-when-only-1-worker-raises-error/7447/11) issue in Ray Train. Should be addressed by the upgrade to Ray 2.1 / 2.2 in #2709.
- Requires that all checkpoints are readable from every worker, as DDP does not support Horovod's broadcast state operations, which allow the files to only exist on the coordinator (should not be an issue for Ray users, as they already have to work with the assumption that the coordinator could land on any node).


ludwig/distributed/base.py
```diff
@@ -0,0 +1,159 @@
+import contextlib
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Optional
+
+import torch
+from torch import nn
+from torch.optim import Optimizer
+
+
+class DistributedStrategy(ABC):
```
justinxzhao, line 10: Please add docstrings to this new ABC. What is a DistributedStrategy?

ludwig/backend/horovod.py
```diff
@@ -37,26 +37,28 @@ class HorovodBackend(LocalPreprocessingMixin, Backend):
 
     def __init__(self, **kwargs):
         super().__init__(dataset_manager=PandasDatasetManager(self), **kwargs)
-        self._horovod = None
+        self._distributed = None
```
justinxzhao, line 40: Should `self._distributed` be an attribute of the `Backend` class?

justinxzhao (approved):
LGTM. Strategy seems like a nice abstraction to add support for pytorch DDP

ludwig/backend/horovod.py
tgaddair: Only for Horovod. In the case of Ray, we are not in the right context from the backend to have access to the distributed api.

PR #2894, opened 2022-12-31
Fix dataset synthesizer

connor-mccorm:
Turns out this was just an issue with the dataset synthesizer. The dataset synthesizer groups input and output features together as one group of features and then checks whether there is an encoder or decoder present on each feature. If there is not, then by default it adds an encoder. This causes a problem when there is an output feature without a decoder specified since it creates an encoder entry on the decoder with a structure like this:
`{
     'name': 'product', 
     'type': 'category', 
     'decoder': {'encoder': {'idx2str': ['oXpJUe', 'fmzl', 'xZFbGsbW', 'YlUXGspxB', 'lru', 'wtQppfKBo', 'XgIhvyQg', 'wcvA', 'oVWamB', 'zZJjV'], 'vocab_size': 10}},
      'column': 'product', 
     'proc_column': 'product_mZFLky'
}`

I created a quick check in the dataset synthesizer that adds a decoder to output features without a decoder parameter which fixes the failure.

### DIFF
--- a/ludwig/data/dataset_synthesizer.py
+++ b/ludwig/data/dataset_synthesizer.py
@@ -164,6 +164,9 @@ parameters_builders_registry = {
 
 @DeveloperAPI
 def build_synthetic_dataset_df(dataset_size: int, config: ModelConfigDict) -> pd.DataFrame:
+    for feature in config[OUTPUT_FEATURES]:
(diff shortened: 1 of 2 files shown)


tgaddair (approved):
Nice catch!

PR #2896, opened 2023-01-03
Added support for using embedding features in GBMs (one-hot, BERT, etc.)

tgaddair:


ludwig/features/category_feature.py
```diff
@@ -198,13 +200,50 @@ def add_feature_data(
 class CategoryInputFeature(CategoryFeatureMixin, InputFeature):
     def __init__(self, input_feature_config: CategoryInputFeatureConfig, encoder_obj=None, **kwargs):
         super().__init__(input_feature_config, **kwargs)
+        if encoder_obj is None:
+            encoder_obj = self.initialize_encoder(input_feature_config.encoder)
+        self.encoder_obj = encoder_obj
 
-        if encoder_obj:
-            self.encoder_obj = encoder_obj
+        if input_feature_config.encoder.skip:
```
tgaddair, line 207: We can abstract this stuff behind the `InputFeature` interface. It will be a big change, though, so I want to do it in a follow-up PR.

ludwig/schema/model_config.py
```diff
@@ -172,17 +172,6 @@ def __init__(self, config_dict: ModelConfigDict):
                 ):
                     raise ValidationError("GBM Model trainer must be of type: 'lightgbm_trainer'")
 
-                for feature in self.input_features.to_dict().keys():
```
tgaddair, line 175: We should have the feature encoder tell us via its interface what model types it accepts.

ksbrar: @tgaddair I just saw your latest comments on my PR and this is what I was thinking as well. As you said, it'll be a relatively large refactor. CC @connor-mccorm

PR #2898, opened 2023-01-05
Move `on_batch_end` callback to omit eval from batch duration during benchmarking

geoffreyangus:
This PR moves the `on_batch_end` callback to before `run_evaluation` is called. This is useful in getting us more accurate batch throughput measurements during benchmarking. Prior to this change, any measurements of average batch duration would incorporate the evaluation step, particularly if the average batch duration was computed across multiple epochs, leading to a (potentially significant) underestimation of average batch duration.

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -948,6 +948,8 @@ class Trainer(BaseTrainer):
                     f"{psutil.Process(os.getpid()).memory_info()[0] / 1e6:0.2f}MB"
                 )
 
+            self.callback(lambda c: c.on_batch_end(self, progress_tracker, save_path))
+
             if progress_tracker.steps % final_steps_per_checkpoint == 0:
                 # Checkpoint the model.
                 if self.is_coordinator() and not self.skip_save_progress:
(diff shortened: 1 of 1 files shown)


tgaddair (approved):

justinxzhao (approved):

ludwig/trainers/trainer.py
```diff
@@ -948,6 +948,8 @@ def _train_loop(
                     f"{psutil.Process(os.getpid()).memory_info()[0] / 1e6:0.2f}MB"
                 )
 
+            self.callback(lambda c: c.on_batch_end(self, progress_tracker, save_path))
```
justinxzhao, line 951: nit: Consider adding a note:

The `on_batch_end` callback happens before `run_evaluation` to get more accurate batch throughput measurements which exclude the evaluation step.

justinxzhao:
For clarity -- prior to this change, would including `run_evaluation` mean that we've been "over" estimating the average batch duration? As in, after this change I'd expect average batch duration to decrease.

geoffreyangus:
Ah yes @justinxzhao, good catch. We would be underestimating training throughput, and overestimating batch duration.

ludwig/trainers/trainer.py
geoffreyangus: Good call– done 👍

PR #2899, opened 2023-01-06
Annotate Confusion Matrix with updated cmap

arnavgarg1:
Before:

Screen Shot 2023-01-05 at 4 13 34 PM

Now:

Screen Shot 2023-01-05 at 4 28 58 PM

Changes:

- Changed cmap from viridis to pastel1 to ensure that annotated text (cm numbers) can always be seen. Pastel has lighter colors only so ensures that this happens.
- Rotate y-axis labels to be consistent with the x-axis
- Adds confusion matrix number annotations inside the confusion matrix plot with a medium font-weight (can change this to normal if it's too much)

### DIFF
--- a/ludwig/utils/visualization_utils.py
+++ b/ludwig/utils/visualization_utils.py
@@ -1074,12 +1074,24 @@ def confusion_matrix_plot(
     ax.xaxis.tick_top()
     ax.xaxis.set_label_position("top")
 
-    cax = ax.matshow(confusion_matrix, cmap="viridis")
+    cax = ax.matshow(confusion_matrix, cmap="Pastel1")
+    # Annotate confusion matrix plot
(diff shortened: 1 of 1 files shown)


hungcs:
nice! maybe out of scope for this pr, but could we do something about the case where labels overflow the matrix with tons of categories because the image size is too small?

hungcs (approved):

arnavgarg1:
> nice! maybe out of scope for this pr, but could we do something about the case where labels overflow the matrix with tons of categories because the image size is too small?

@hungcs Do you have an example plot? At the moment, I've added wrap for the numbers inside the confusion matrix so that should be taken care of, but I can also figure out if there's a way to dynamically adjust the plot size so that the x-axis and y-axis labels don't get cut and are always seen

justinxzhao (approved):

PR #2902, opened 2023-01-06
fix: Remove extra properties from `input_features`, `output_features`, `defaults` schemas for GBM models

ksbrar:
Removes extra properties from the features and defaults schemas that are otherwise unsupported by GBM.

In particular
* `encoder` and `decoder` sub schemas are removed
* Only GBM-supported feature types are left behind after pruning the defaults schema
* Extra fields with references to encoders/decoders are removed: `input_size`, 'top_k`, `num_classes` - can I get a double check on this?

### DIFF
--- a/ludwig/schema/defaults/defaults.py
+++ b/ludwig/schema/defaults/defaults.py
@@ -1,3 +1,5 @@
+from typing import Dict
+
 from marshmallow_dataclass import dataclass
 
 from ludwig.api_annotations import DeveloperAPI
@@ -9,6 +11,8 @@ from ludwig.constants import (
     DATE,
     H3,
     IMAGE,
+    MODEL_ECD,
+    MODEL_GBM,
     NUMBER,
     SEQUENCE,
     SET,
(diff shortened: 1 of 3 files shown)


arnavgarg1:
@ksbrar Looks good! I think `num_classes` and `top_k` might be ok though since we can do multiclass classification using GBMs?

tgaddair (changes requested):
I feel like GBMs have a lot of special-cases in the schema. Is there not an object-oriented way to achieve this by creating custom subschemes for ECD and GBM?

ksbrar:
@tgaddair Sure, and I agree. As I made these changes I also felt they were suboptimal. This was initially scoped as an "on-call"-sized fix so I just tried to make a minimal change, but I can reformulate this PR.

tgaddair:
Hmm, I suppose if this is breaking something at the moment, we can land it as-is, but I want to think of a more longterm fix as well. We can sync up about it early next week. Thanks for looking into it!

ludwig/schema/features/utils.py
```diff
@@ -28,10 +28,19 @@ def prune_gbm_features(schema: Dict):
...
     gbm_feature_types = ["binary", "category", "number"]
     pruned_all_of = []
+
     for cond in schema["items"]["allOf"]:
         if_type = cond["if"]["properties"]["type"]["const"]
         if if_type in gbm_feature_types:
+            pruned_then_properties = {
+                k: v
+                for k, v in cond["then"]["properties"].items()
+                if k not in ["encoder", "decoder", "tied", "input_size", "num_classes", "top_k"]
```
tgaddair, line 38: GBMs will support encoders and decoders in #2896.

ludwig/schema/defaults/defaults.py
```diff
@@ -51,15 +55,36 @@ class DefaultsConfig(schema_utils.BaseMarshmallowConfig):
     vector: BaseFeatureConfig = DefaultsDataclassField(feature_type=VECTOR)
 
 
+def prune_gbm_props(schema: Dict):
+    """Removes unsupported props from the given JSON schema.
+
+    Designed for use with `get_defaults_jsonschema`.
+    """
+    gbm_feature_types = ["binary", "category", "number"]
```
tgaddair, line 63: GBMs will support text features in #2896.

ksbrar:
@tgaddair What is the timeline for #2896 getting merged in? If it's soon I might just cancel this PR. Then we can sync on the refactoring necessary to enable a more robust/many-to-many encoder interface.

ksbrar:
Closed in favor of #2896

PR #2906, opened 2023-01-09
Refactored ModelConfig object into a Marshmallow schema

tgaddair:


ludwig/utils/config_utils.py
```diff
@@ -69,17 +68,7 @@ def merge_config_preprocessing_with_feature_specific_defaults(
...
-) -> Dict[str, Any]:
-    """Update preprocessing parameters if encoders require fixed preprocessing parameters."""
-    encoder_type = encoder_params.get(TYPE, get_default_encoder_type(feature_type))
-    encoder_class = get_encoder_cls(feature_type, encoder_type)
-    return merge_dict(preprocessing_params, encoder_class.get_fixed_preprocessing_params(encoder_params))
-
-
-@DeveloperAPI
-def get_default_encoder_type(feature_type: str) -> str:
+def get_default_encoder_type(model_type: str, feature_type: str) -> str:
```
arnavgarg1, line 71: Might be missing something here, but where is `model_type` being used?

ludwig/schema/model_config.py
```diff
@@ -128,7 +129,7 @@ class OutputFeaturesContainer(BaseFeatureContainer):
 
 @DeveloperAPI
 @ludwig_dataclass
-class ModelConfig(BaseMarshmallowConfig):
+class _ModelConfig(BaseMarshmallowConfig):
```
connor-mccorm, line 132: Looks like we're not using this anymore, are we just keeping it around as a legacy reference?

ludwig/schema/combiners/base.py
```diff
@@ -1,8 +1,11 @@
+from marshmallow_dataclass import dataclass
+
 from ludwig.api_annotations import DeveloperAPI
 from ludwig.schema import utils as schema_utils
 
 
 @DeveloperAPI
+@dataclass(repr=False, order=True)
```
connor-mccorm, line 8: Can we import and use ludwig_dataclass here instead?

```
from ludwig.schema.utils import ludwig_dataclass

@ludwig_dataclass
```

connor-mccorm (approved):
This is great, definitely a lot of terrific changes in here. Thanks for putting this together Travis. I just had a few nits I noticed while going through this but at a high level the logic all makes sense and doesn't have any major errors. The one thing I'll callout is that we may need to double check that there are no changes needed on the Predibase side to reconcile these changes. It looks like the api has been preserved so I don't expect anything here, but I do remember running into issues last time making these changes so just wanted to mention it. Overall, LGTM!

ludwig/utils/config_utils.py
tgaddair: Good catch! Seems we don't actually use this function anymore, so removed.

ludwig/schema/features/utils.py
```diff
@@ -42,44 +53,36 @@ def get_input_feature_jsonschema(model_type: str):
...
-            "properties": {
-                "name": {"type": "string", "title": "name", "description": "Name of the input feature."},
-                "type": {
-                    "type": "string",
-                    "enum": input_feature_types,
-                    "title": "type",
-                    "description": "Type of the input feature",
-                },
-                "column": {"type": "string", "title": "column", "description": "Name of the column."},
+        "type": "object",
```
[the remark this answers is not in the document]
tgaddair: Good question, it should get it from the `FeaturesTypeSelection` which has `min_length=1` set by defaults. But let me check.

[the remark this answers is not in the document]
tgaddair: Yes, looks like it works. Added tests to verify.

ludwig/schema/features/preprocessing/image.py
```diff
@@ -133,3 +133,8 @@ class ImagePreprocessingConfig(BasePreprocessingConfig):
         description="Specifies the number of processes to run for preprocessing images.",
         parameter_metadata=FEATURE_METADATA[IMAGE][PREPROCESSING]["num_processes"],
     )
+
+    requires_equal_dimensions: bool = schema_utils.Boolean(
+        default=False,
+        description="If true, then width and height must be equal.",
+    )
```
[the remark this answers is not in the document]
tgaddair: Good call. Done.

ludwig/schema/model_config.py
tgaddair: Yes, I'll add a comment. Idea is to remove in a follow-up PR, as this PR is already very big.

ludwig/schema/combiners/base.py
tgaddair: Done. What about `BaseFeatureConfig` and the other classes in that file? Does it matter that `order=True` isn't specified on those?

ludwig/schema/features/utils.py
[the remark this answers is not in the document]
ksbrar: Awesome that it works, also verified it locally. I'm a little confused about how this works though... the custom schema here is no longer even an array so how does it end up being one in the final assembly (let alone how does the `allOf` end up in the right place)?

ksbrar (approved):
Super, super clean. a few nits... and questions 🤔

ludwig/schema/combiners/base.py
connor-mccorm: Ya we should use `@ludwig_dataclass` on those as well. Currently the it's not a huge deal since a lot of the parameters on those classes are filtered out or not shown for one reason or another. However, we will need it eventually and having `order=True` doesn't hurt so I think adding it in is a great idea.

ludwig/config_validation/validation.py
```diff
@@ -1,30 +1,22 @@
...
-from ludwig.schema.combiners.utils import get_combiner_jsonschema
-from ludwig.schema.defaults.defaults import get_defaults_jsonschema
-from ludwig.schema.features.utils import get_input_feature_jsonschema, get_output_feature_jsonschema
-from ludwig.schema.hyperopt import get_hyperopt_jsonschema
-from ludwig.schema.preprocessing import get_preprocessing_jsonschema
-from ludwig.schema.trainer import get_model_type_jsonschema, get_trainer_jsonschema
+from ludwig.constants import MODEL_ECD, MODEL_TYPE, PREPROCESSING, SPLIT
+from ludwig.schema import utils as schema_utils
+
+# TODO(travis): figure out why we need these imports to avoid circular import error
```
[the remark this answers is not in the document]
tgaddair: Yeah, unfortunately so. I think the right fix here is to remove the circular import structure so schema has no deps on the encoders, etc. But this is a larger refctor.

ludwig/schema/features/binary_feature.py
```diff
@@ -30,19 +32,35 @@ class BinaryInputFeatureConfigMixin(BaseMarshmallowConfig):
 
     preprocessing: BasePreprocessingConfig = PreprocessingDataclassField(feature_type=BINARY)
 
+
+@DeveloperAPI
+@ludwig_dataclass
+class BinaryInputFeatureConfig(BaseInputFeatureConfig, BinaryInputFeatureConfigMixin):
+    """BinaryInputFeatureConfig is a dataclass that configures the parameters used for a binary input feature."""
+
+    encoder: BaseEncoderConfig = None
```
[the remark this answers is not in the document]
tgaddair: Right, I think it has to do with the "can't have a default param before a require param" stuff.

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -109,12 +105,14 @@ class CommonSchedulerOptions:
         ),
     )
 
+    max_t: int = max_t_alias()
```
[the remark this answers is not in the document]
arnavgarg1: +1

ludwig/schema/optimizers.py
```diff
@@ -441,18 +441,18 @@ def _deserialize(self, value, attr, data, **kwargs):
...
                     try:
                         return opt.Schema().load(value)
                     except (TypeError, ValidationError) as e:
-                        raise ValidationError(
-                            f"Invalid params for optimizer: {value}, see `{opt}` definition. Error: {e}"
-                        )
+                        raise ValidationError(f"Invalid params for optimizer: {value}, see `{opt}` definition") from e
                 raise ValidationError(
-                    f"Invalid params for optimizer: {value}, expect dict with at least a valid `type` attribute."
+                    f"Invalid optimizer type: '{opt_type}', expected one of: {list(optimizer_registry.keys())}."
```
arnavgarg1, line 453: Much more useful error -> resolution messages, thanks for adding these in

ludwig/schema/optimizers.py
```diff
@@ -441,18 +441,18 @@ def _deserialize(self, value, attr, data, **kwargs):
...
                 )
-            raise ValidationError("Field should be None or dict")
+            raise ValidationError(f"Invalud optimizer param {value}, expected `None` or `dict`")
```
arnavgarg1, line 455: nit: `invalid`

ludwig/schema/utils.py
```diff
@@ -1070,3 +1082,106 @@ def is_primitive(value):
...
+    def _deserialize(self, value, attr, data, **kwargs):
+        if value is None:
+            return value
+        if isinstance(value, dict):
+            try:
+                return self.cls.Schema().load(value)
+            except (TypeError, ValidationError) as e:
+                # TODO(travis): this seems much too verbose, does the validation error not show the specific error?
+                raise ValidationError(f"Invalid params: {value}, see `{self.cls}` definition. Error: {e}")
+        raise ValidationError(f"Invalud param {value}, expected `None` or `dict`")
```
arnavgarg1, line 1177: nit: `invalid` typo

ludwig/utils/config_utils.py
```diff
@@ -68,37 +66,6 @@ def merge_config_preprocessing_with_feature_specific_defaults(
...
-@DeveloperAPI
-def get_default_encoder_or_decoder(feature: FeatureConfigDict, config_feature_group: str) -> str:
-    """Returns the default encoder or decoder for a feature."""
-    if config_feature_group == INPUT_FEATURES:
-        return get_default_encoder_type(feature[TYPE])
-    else:
-        return get_default_decoder_type(feature[TYPE])
-
-
 def has_trainable_encoder(config: ModelConfig) -> bool:
```
arnavgarg1, line 69: nit: would it make sense to also move `has_trainable_encoder` and `has_pretrained_encoder` into `ludwig/schema/model_types/utils.py`?

tests/ludwig/utils/test_config_utils.py
```diff
@@ -1,50 +1,50 @@
-from typing import Any, Dict, Optional
-
-import pytest
-
-from ludwig.utils.config_utils import merge_fixed_preprocessing_params
-
-
```
arnavgarg1, line 7: Should we just delete this test module? I'm assuming you've commented out these tests because these functions are only used within the ModelConfig object, and they seem to work?

arnavgarg1 (approved):
This cleans up all the schema implementations we've had by a lot, thanks for putting this together 👏 

Only minor nits, but looks good otherwise!

ludwig/schema/features/utils.py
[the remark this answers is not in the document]
tgaddair: I think it's `marshmallow-jsonschema` magic.

ludwig/schema/hyperopt/scheduler.py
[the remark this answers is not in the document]
tgaddair: Reason why I had to move it here is because we run some checks against it for hyperopt defaults, but I can rework.

ludwig/schema/features/utils.py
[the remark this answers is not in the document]
ksbrar: @tgaddair I'm not so sure. Whatever is set in `_json_type_mapping` should override any kind of automatic logic so at minimum while I might expect it to perhaps fix the `type` to be an `array` I would not expect it to insert `items` or know to cherry-pick `allOf` to one level higher than it is in the final schema. But I actually wanted to talk about this in our next sync

ludwig/schema/hyperopt/scheduler.py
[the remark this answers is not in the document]
tgaddair: Fixed.

ludwig/utils/config_utils.py
tgaddair: Yes, we can do it in a follow-up.

tests/ludwig/utils/test_config_utils.py
tgaddair: I want to revisit in a follow-up to make it work with the new structure.

ludwig/schema/features/utils.py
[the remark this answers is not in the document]
ksbrar: I just figured this out. Despite the fact that `json_type_mapping` is _supposed_ to be the ultimate override method, `marshmallow_jsonschema` _will_ change the type of the field if the class in question derives a native marshmallow type. Since `FeaturesTypeSelection` ultimately derives `marshmallow.fields.List`, the package _will_ force it to be a list - in particular by taking the output of `json_type_mapping` and shoving it inside of a standard list's JSON schema (that is where the surrounding `type: array : { items: { ...` comes from).

We are lucky that all of the feature's schema attributes and conditional logic indeed are supposed to be nested inside of `items`, otherwise this would have caused a major error!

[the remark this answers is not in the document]
tgaddair: Very interesting! Nice discovery.

PR #2910, opened 2023-01-11
Ray 2.2 compatibility

arnavgarg1:
This PR modifies Ludwig code paths to make Ludwig compatible with Ray 2.2. There are 3 changes:

1. The `ray.train.local_rank()` API was deprecated in Ray 2.2. To work around this, we had to migrate to the new Ray AIR equivalent, `ray.air.session.get_local_rank()`. This is required for setting the right distributed training strategy (DDP or Horovod). _Note_: To use `ray.air.session.get_local_rank()`, you need to be inside a train session or a tune session. In Ludwig's current code implementation, batch size tuning doesn't get instantiated inside of a RayTrainer class, so we manually set the local_rank to 0 so that it picks up the right device to tune batch size on.
3. The `Checkpoint.from_object_ref()` API was deprecated in favour of `Checkpoint(obj_ref=)`. Strangely enough, without this change, it was causing hyperopt tests to stall because of resource scheduling issues. I've flagged this way the Ray team for further review.
4. Image and audio test preprocessing was failing for a variety of tests. This is because Ray 2.2 does some additional checks when reading data from binary files, specifically that it needs a list of string paths instead of list of tuples of (str_path,idx). The reason we need idx is because this is the global dataframe index that we want to preserve so we don't need to repartition the dataset globally since this is expensive. This was fixed by @geoffreyangus by modifying our custom Datasource class to get past the check by modifying the function args to use `paths_and_idx` instead of `paths`.


tgaddair (approved):
LGTM! Nice :)

justinxzhao (approved):
The size of this PR is pleasantly manageable :)

jppgks (approved):
Thanks for the changes & clear PR description!

geoffreyangus (approved):

PR #2911, opened 2023-01-11
Adds quick fix for pretrained models not loading by modifying state_dict keys on load.

dantreiman:
Quick fix for loading modules with pre-trained encoders which were saved prior to the addition of FreezeModule (https://github.com/ludwig-ai/ludwig/pull/2855).

### DIFF
--- a/ludwig/models/ecd.py
+++ b/ludwig/models/ecd.py
@@ -14,6 +14,7 @@ from ludwig.schema.model_config import ModelConfig
 from ludwig.utils import output_feature_utils
 from ludwig.utils.data_utils import clear_data_cache
 from ludwig.utils.fs_utils import open_file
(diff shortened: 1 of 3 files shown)


connor-mccorm (commented):
Looks good to me, just a couple of concerns, I was seeing this error while running predict queries on a model with a text feature that used a `parallel_cnn` instead of a huggingface encoder. Is that expected/does this fix account for that? Secondly, if we're going to be adding another backwards compatibility file, I think it may be worth while to put all backwards compatibility files under their own backwards_compatibility directory under `ludwig/utils/` since there may be more in the future. WDYT?

tgaddair (approved):

connor-mccorm:
Okay looks like I actually hadn't set the default model for queries to use the parallel_cnn. model so false alarm there.

PR #2916, opened 2023-01-12
Adds backwards compatibility for legacy image encoders

tgaddair:
Followup on #2408 that does the following:

- Adds backwards compatibility layer for `resnet_legacy` and `vit_legacy` encoders with deprecation warning.
- Renames legacy encoders to include `_` prefix to mark them as internal.
- Reorganizes image encoders into a common module.
- Adds back vit legacy test.


ludwig/utils/backward_compatibility.py
```diff
@@ -750,6 +752,57 @@ def learning_rate_scheduler(trainer: TrainerConfigDict) -> TrainerConfigDict:
...
+    legacy_encoder_cls = get_encoder_cls(feature[TYPE], encoder_mapping[encoder_type])
+    legacy_encoder_fields = legacy_encoder_cls.get_valid_field_names()
+
+    user_fields = set(encoder.keys())
+    user_fields.remove(TYPE)
+
+    removed_fields = legacy_encoder_fields.difference(new_encoder_fields)
+    added_fields = new_encoder_fields.difference(legacy_encoder_fields)
+
+    user_legacy_fields = user_fields.intersection(removed_fields)
```
jimthompson5802, line 782: @tgaddair   I took a look at the PR and have a question. 

What is the expected behavior if the user specifies `vit` or `resnet`?   Given earlier conversations, I believe for these encoders the actual encoder to use are `_vit_legacy` or `_resnet_legacy`, respectively.

Is this a correct understanding?

I pulled down the branch and ran a small image classification training with `vit` and `resnet` specified.  From what I can tell, the torchvision encoders were used, not the legacy encoders.

From what I can see, the cited statement may be the cause.  If there is no overlap in the encoder parameters between the legacy and torchvision encoder, then this intersection comes up to the empty set.  So when
```
if let(user_legacy_files) >0:
```
at line 784 is evaluted, the condition is `False`  and the current torcvision encoder is used and not the legacy one the deprecation warning messages are not produced.

This is the test program I used to arrive at the above observation
```
import logging
import os
import warnings
import shutil
import pandas as pd
import sys

from ludwig.encoders.registry import get_encoder_registry, get_encoder_cls
import pprint
from ludwig.data.dataset_synthesizer import cli_synthesize_dataset
warnings.simplefilter(action="ignore", category=pd.errors.PerformanceWarning)


FEATURES_LIST = [
    {"name": "binary_output_feature", "type": "binary"},
    {"name": "category_output_feature", "type": "category"},
    {
        "name": "my_image", "type": "image",
        "destination_folder": os.path.join(os.getcwd(), "data2/images"),
        "preprocessing": {"height": 600, "width": 600, "num_channels": 3}
    }
]

shutil.rmtree("data2", ignore_errors=True)
os.makedirs("data2", exist_ok=True)
cli_synthesize_dataset(50, FEATURES_LIST, "data2/syn_train.csv")


config = {
    "input_features": [
        {"name": "category_output_feature", "type": "category"},
        {
            "name": "my_image",
            "type": "image",
            "encoder": {
                "type": "vit",
            },
            "preprocessing": {"standardize_image": "imagenet1k", "width": 300, "height": 300}
        },
    ],
    "output_features": [
        {
            "name": "binary_output_feature", "type": "binary",
        }
    ],
    "trainer" : {"epochs": 2, "batch_size": 7},
}

model = LudwigModel(config, logging_level=logging.INFO)
_, _, output_dir = model.train(dataset="data2/syn_train.csv", skip_save_processed_input=True)

```

tgaddair: Yes, this was my intent with this PR. The logic I wanted to implement was:

- If user has legacy params, then use `_legacy` encoder
- If user has new params, use tochvision encoder
- If user has both legacy and new params, raise an error
- If user has no params, use torchvision encoder (because the other encoder is deprecated and we want people to use the new one).

Does sound reasonable to you?

jimthompson5802: LGTM

jimthompson5802: LGTM

jimthompson5802 (approved):

PR #2917, opened 2023-01-12
fix: Schema split conditions should pass in [TYPE] and not string

hungcs:


ksbrar (approved):

tgaddair (commented):
Test?

justinxzhao (approved):

tests/ludwig/schema/test_utils.py
```diff
@@ -0,0 +1,8 @@
+from ludwig.constants import TYPE
+from ludwig.schema import utils as schema_utils
+
+
+def test_remove_duplicate_fields():
+    props = {TYPE: "random"}
+    schema_utils.remove_duplicate_fields(props, [TYPE])
+    assert TYPE not in props
```
martindavis, line 8: We should test the positive case as well: a property not listed in the array continues to remain in the `props` object.

martindavis (changes requested):

martindavis (approved):

PR #2919, opened 2023-01-12
Upgrade benchmarking and automl configs to be compliant with torch LR scheduler schema.

justinxzhao:


jppgks (commented):
Can we add a test that validates these configs against the schema?

justinxzhao:
Closed in favor of #2921

PR #2922, opened 2023-01-13
Update automl heuristics for fine-tuning and multi-modal tasks

tgaddair:
For text tasks:

- Set `trainable=True`
- `use_mixed_precision=True`
- Enable learning rate warmup and linear decay
- Increase training epochs to 10

Also fixes multi-modal to only use `tabnet` as a combiner when all feature types are tabular.


ludwig/automl/defaults/text/bert_config.yaml
```diff
@@ -1,9 +1,17 @@
...
+    decay: linear
   optimizer:
     type: adamw
+  use_mixed_precision: true
+
+defaults:
+  text:
+    encoder:
+      type: bert
+      trainable: true
```
jppgks, line 14: Why do we default to `trainable: true`? Thinking about an iterative approach where you would:

1. Train default model (parallel CNN)
2. Train a model with pretrained BERT
3. Train a model to fine-tune BERT

tests/integration_tests/test_automl.py
```diff
@@ -112,10 +137,39 @@ def test_data_multimodal():
...
+        ("test_data_tabular_small", {"combiner": {"type": "concat"}}),
+        ("test_data_image", {"combiner": {"type": "concat"}}),
+        (
+            "test_data_text",
+            {
+                "input_features": [{"type": "text", "encoder": {"type": "bert"}}],
+                "combiner": {"type": "concat"},
+                "trainer": {
+                    "batch_size": "auto",
+                    "learning_rate": "auto",
```
justinxzhao, line 152: I'd vouch for `0.00002` instead of `auto` for LR.

ludwig/automl/defaults/text/bert_config.yaml
justinxzhao: Agreed in the iterative case. In the "best single shot model" case, I would say that setting `trainable: true` right away makes sense.

justinxzhao: Agreed in the iterative case. In the "best single shot model" case, I would say that setting `trainable: true` right away makes sense.

tests/integration_tests/test_automl.py
tgaddair: Done.

justinxzhao (approved):
LGTM!

jppgks (approved):

PR #2923, opened 2023-01-13
Raise RuntimeError only for category output features with vocab size 1

arnavgarg1:
Ludwig raises RuntimeErrors complaining about category features when they have a vocab size of 1. However, this is only a problem if the feature is a category output feature, but isn't an issue if the feature is a category input feature. 

This PR adds a new kwarg is_input_feature to `get_feature_meta`, which is used to set the right warning when the vocab size is 1 for category output features depending on whether the feature is an input feature or output feature.


ludwig/data/preprocessing.py
```diff
@@ -1305,6 +1305,14 @@ def build_preprocessing_parameters(
     return feature_name_to_preprocessing_parameters
 
 
+def is_input_feature(feature_config: FeatureConfigDict) -> bool:
+    """Utility function to check for the presence of encoder in the feature config to determine if the feature is
+    an input feature or output feature."""
+    if ENCODER in feature_config:
```
jppgks, line 1311: is this always present on the input feature?

jppgks (approved):

ludwig/data/preprocessing.py
arnavgarg1: @jppgks I believe it is by this point because we only reach preprocessing after the config object is created and validated

tests/integration_tests/test_preprocessing.py
```diff
@@ -485,6 +486,26 @@ def test_non_conventional_bool_without_fallback_logs_warning(binary_as_input, ca
...
+        input_feature = output_feature
+        output_feature = [category_feature(decoder={"vocab_size": 1})]
+
+    config = {"input_features": input_feature, "output_features": output_feature}
+
+    training_data_csv_path = generate_data(config["input_features"], config["output_features"], data_csv_path)
+
+    ludwig_model = LudwigModel(config)
+    with pytest.raises(RuntimeError) if feature_type == "output_feature" else contextlib.nullcontext():
+        _, _, _, _ = ludwig_model.preprocess(dataset=training_data_csv_path)
```
jppgks, line 506: can we test that it trains successfully too?

jppgks (changes requested):

tests/integration_tests/test_preprocessing.py
arnavgarg1: Just updated

jppgks (approved):

ludwig/data/preprocessing.py
justinxzhao, line 1311: nit: `return ENCODER in feature_config`

ludwig/features/audio_feature.py
```diff
@@ -92,7 +92,9 @@ def cast_column(column, backend):
         return column
 
     @staticmethod
-    def get_feature_meta(column, preprocessing_parameters: PreprocessingConfigDict, backend):
+    def get_feature_meta(
+        column, preprocessing_parameters: PreprocessingConfigDict, backend, **kwargs
```
justinxzhao, line 96: Use `is_input_feature` explicitly instead of `kwargs`, which requires higher cognitive load when trying to understand the code.

ludwig/features/category_feature.py
```diff
@@ -121,16 +121,32 @@ def cast_column(column, backend):
...
             raise InputDataError(
-                column.name, CATEGORY, f"At least 2 distinct values are required, column only contains {str(idx2str)}"
+                column.name,
+                CATEGORY,
+                f"""
+                At least 2 distinct values are required for category output features, but column
+                only contains {str(idx2str)}.
+                """,
+            )
+        elif vocab_size <= 1:
```
justinxzhao, line 144: nit: Since you're raising an error in the previous block, use a plain `if` here.

justinxzhao (approved):
LGTM with minor comments.

ludwig/data/preprocessing.py
```diff
@@ -1305,6 +1305,12 @@ def build_preprocessing_parameters(
     return feature_name_to_preprocessing_parameters
 
 
+def is_input_feature(feature_config: FeatureConfigDict) -> bool:
+    """Utility function to check for the presence of encoder in the feature config to determine if the feature is
+    an input feature or output feature."""
+    return ENCODER in feature_config
```
abidwael, line 1311: Pretty minor but could you add a test for this? Could come in handy in case we change what's (not) in the config

abidwael (approved):
Thanks for the change @arnavgarg1!

ludwig/data/preprocessing.py
arnavgarg1: Done!

Issue #490, opened 2019-08-13
Data Augmentation in Image Classification


vsvanshi:
Hi,
How can we enable Data augmentation in Image classification?

w4nderlust:
There is no data augmentation feature at the moment.
We definitely want to add it in the future.
Contributions are welcome.

w4nderlust:
@manesioz the example you provide is useful. The problem with it is that it looks like it assumes that the inputs to your model are only images and that the outputs to your model are only categories judging from this snippet: https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/preprocessing/image.py#L376-L403
In Ludwig that's not the case, so if someone wants to add the same functionality in Ludwig, one would have to deal with other input features and potentially multiple output features.

manesioz:
@w4nderlust thanks for the advice, I agree that this feature should be generally available for different inputs/outputs. However, considering how most data augmentation techniques are highly domain-specific (image, text, timeseries, etc), I was wondering if we should initially approach this feature as being configurable depending on the input data type (text, image, numerical, timeseries, etc)?   

Some brief [literature review](https://arxiv.org/pdf/1702.05538.pdf) indicates that some domain-agnostic techniques exist, but I am not sure of their accuracy/performance/general acceptance within the ML community. 

Any advice would be much appreciated, thanks!

w4nderlust:
@manesioz  sorry I did not mean that the data augmentation should work for all feature types, what I meant is that the example works with only one feature input, in Ludwig there could be several, so for instance if you apply 4 transformations to your image input feature, all the other input features have to be replicated 4 more times, which the example does not consider.

carlogrisetti:
Just for a reference, I am very proficiently using ImgAug (https://github.com/aleju/imgaug) for image augmentation.
Has *lots* of augmentation possibilities, and correctly maps points\vectors\shapes\segmentations\etc on the destination image\array.
It takes account of any chained modifications on the image (multiple transformations do not require multiple adaptations of the other features, but are passed in one unique transformation sequence) and has many many transformations built-in (see the project homepage for a visual demo)

I agree with the fact that augmentation should not be domain agnostic... In the last 2 or 3 projects I used data augmentation on (always and only talking about images) I had to perform very different image augmentations based on the dataset I'm working on.
[...]

PR #2925, opened 2023-01-13
Feature: Data Augmentation for Image Input Features

jimthompson5802:
# Code Pull Requests

Add data augmentation capability for image input features

Addresses #490


ludwig/schema/features/image_feature.py
```diff
@@ -25,6 +27,8 @@ class ImageInputFeatureConfigMixin(BaseMarshmallowConfig):
         default="stacked_cnn",
     )
 
+    augmentation: BaseAugmentationConfig = AugmentationContainerDataclassField(default=[])
```
ksbrar, line 30: Let's switch `BaseAugmentationConfig` to `List[BaseAugmentationConfig]`.

ludwig/schema/features/augmentation/base.py
```diff
@@ -0,0 +1,16 @@
...
+
+@DeveloperAPI
+class BaseAugmentationConfig(schema_utils.BaseMarshmallowConfig):
+    """Base class for augmentation."""
+
+    pass
+
+
+@DeveloperAPI
+class BaseAugmentationContainerConfig(schema_utils.BaseMarshmallowConfig):
```
ksbrar, line 13: I think this class will end up being unnecessary, can be removed

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Returns: Initialized dataclass field that converts an untyped dict with params to an augmentation config.
+    """
+
+    class AugmentationContainerMarshmallowField(fields.Field):
+        """Custom marshmallow field that deserializes a list for a valid augmentation config from the
+        augmentation_registry and creates a corresponding JSON schema for external usage."""
+
+        def _deserialize(self, value, attr, data, **kwargs):
+            if value is None:
```
ksbrar, line 70: Let's get rid of this line. Fundamentally what it allows is for the user to manually set `augmentation: None` in their Ludwig config. I don't think this should be allowed - either the user should not set `augmentation` in their config at all or they need to provide a list.

Let's instead replace this with an assert `isinstance(value, list)` or some kind of equivalent type checking.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,157 @@
...
+from ludwig.api_annotations import DeveloperAPI
+from ludwig.constants import TYPE
+from ludwig.schema import utils as schema_utils
+from ludwig.schema.features.augmentation.base import BaseAugmentationContainerConfig
+from ludwig.utils.registry import Registry
+
+
+@DeveloperAPI
+@dataclass(repr=False)
+class ImageAugmentationContainerConfig(BaseAugmentationContainerConfig):
```
ksbrar, line 15: Also unnecessary atm, can be removed

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    """Augmentation container for image features."""
+
+    pass
+
+
+# TODO: Is all of this needed?
```
ksbrar, line 21: Doesn't look like we need it!

ludwig/schema/defaults/defaults.py
```diff
@@ -57,6 +58,11 @@ def get_defaults_jsonschema():
     combiner's field constraints."""
     preproc_schema = schema_utils.unload_jsonschema_from_marshmallow_class(DefaultsConfig)
     props = preproc_schema["properties"]
+
+    # TODO:  Is this the best approach?  Discuss with ksbrar
+    # Special handling for image augmentation, overwrite default with list version
+    props[IMAGE]["properties"]["augmentation"] = get_augmentation_jsonschema()
```
ksbrar, line 64: I don't think is necessary because you've already added augmentation onto `ImageInputFeatureConfigMixin`

ludwig/schema/utils.py
```diff
@@ -1034,3 +1034,44 @@ def is_primitive(value):
         },
         **default_kwarg,
     )
+
+
+@DeveloperAPI
```
ksbrar, line 1039: I don't think either of these classes are necessary.

tests/ludwig/augmentation/test_augmentation_pipeline.py
```diff
@@ -0,0 +1,38 @@
...
+@pytest.fixture(scope="module")
+def test_image():
+    # return random normal batch of images of size 2x3x32x32 [batch_size, channels, height, width]
+    return torch.randn(2, 3, 32, 32)
+
+
+@pytest.mark.parametrize(
+    "augmentation_pipeline_ops",
+    [
+        [],
```
ksbrar, line 17: Is this empty list the one that causes test failures?

ludwig/schema/features/utils.py
```diff
@@ -86,6 +87,13 @@ def get_input_feature_conds():
         feature_schema = schema_utils.unload_jsonschema_from_marshmallow_class(schema_cls)
         feature_props = feature_schema["properties"]
         schema_utils.remove_duplicate_fields(feature_props)
+
+        # TODO: Is this the best way?  Discuss with ksbrar
+        # special handling for image augmentation
+        if feature_type == IMAGE:
```
ksbrar, line 93: I don't think this is necessary, because by including `AugmentationContainerDataclassField` directly on the image feature schema it should already have the augmentation props.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+    return wrap
+
+
+def get_augmentation_cls(name: str):
+    return get_augmentation_config_registry()[name]
+
+
+@DeveloperAPI
+def AugmentationContainerDataclassField(default=None, description=""):
```
ksbrar, line 54: The `default` should be `[]` (or some kind of list).

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,157 @@
...
+                        f"Invalid augmentation params: {value}, see `{pre}` definition. Error: {error}"
+                    )
+            return augmentation_list
+
+        @staticmethod
+        def _jsonschema_type_mapping():
+            return get_augmentation_jsonschema()
+
+    try:
+        augmentation_container = ImageAugmentationContainerConfig
```
ksbrar, line 90: These three lines should copy what you are doing above in `_deserialize` except you do everything to `default` instead of `value`. 

Some context on this pattern:
The point of defining `dump_default` and `load_default` using a "`Schema.load`" is so that the `default` argument (which is defined at schema declaration) _is also validated_. Hypothetically, you could delete these lines entirely and set `dump_default=default` directly and `marshmallow` would accept that - we _don't_ want that to happen (marshmallow only allows it because it is designed to be as flexible as possible), so we duplicate the validation logic we use for user-specified values.

I bet this might be causing some of your test weirdness. Because what is currently happening is you use `ImageAugmentationContainerConfig` to `load` the `default` argument provided to this field - and `ImageAugmentationContainerConfig`'s schema is empty so by using `Schema().load` on empty dictionaries `load_default` gets set to `ImageAugmentationConfig()` (an instance of that empty class instead of what you want, which is a list of augmentation instances) and `dump_default` gets set to `{}`.

ksbrar (changes requested):
@jimthompson5802 I just did a quick once-over of your PR. Assuming my understanding is correct, I think you can delete a good amount of unnecessary code. We can discuss some issues here and use it as a framing for our sync next week.

ludwig/schema/features/image_feature.py
jimthompson5802: doen

ludwig/schema/features/augmentation/base.py
jimthompson5802: commented out, added TODO to clean before merge

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

ludwig/schema/features/augmentation/utils.py
jimthompson5802: commented out, TODO to clean up before merge

ludwig/schema/defaults/defaults.py
jimthompson5802: commented out, TODO to clean up before merge

ludwig/schema/utils.py
jimthompson5802: commented out, TODO to clean u[p before merge

tests/ludwig/augmentation/test_augmentation_pipeline.py
jimthompson5802: The empty list does not cause an error.  

I assuming that if **user does not** specify `augmentation` in the config, then after all the default setting and backward compatiblity  processing the generated configuration dictionary will have an `augmentation` section added with an empty list.

Let me know if I have the wrong understanding.

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

ludwig/schema/features/augmentation/utils.py
jimthompson5802: removed

ludwig/schema/features/augmentation/utils.py
jimthompson5802: @ksbrar   I'd appreciate if you take a look at how I interepreted your comments, especially how I assigned values to `load_default` and `dump_default`and set `default_factory` parameter in the `field()` call.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,145 @@
...
+                augmentation_op = augmentation[TYPE]
+                augmentation_cls = get_augmentation_cls(augmentation_op)
+                pre = augmentation_cls()
+                try:
+                    augmentation_list.append(pre.Schema().load(augmentation))
+                except (TypeError, ValidationError) as error:
+                    raise ValidationError(
+                        f"Invalid augmentation params: {default}, see `{pre}` definition. Error: {error}"
+                    )
+            load_default = dump_default = augmentation_list
```
ksbrar, line 79: My last comment was a bit unclear - I referred to both `dump_default` and `load_default` using `Schema.load` but the former should use `Schema.dump`. So in this loop make two lists - one that is populated using `Schema.load` and one that is populated using `Schema.dump` and set the to `_default` variables accordingly.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
```
ksbrar, line 80: if `default` is `None` we should just raise an error rather than doing any kind of fallback logic.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,145 @@
...
+                        f"Invalid augmentation params: {value}, see `{pre}` definition. Error: {error}"
+                    )
+            return augmentation_list
+
+        @staticmethod
+        def _jsonschema_type_mapping():
+            return get_augmentation_jsonschema()
+
+    try:
+        if default:
```
ksbrar, line 67: I would add the assert `isinstance(default, list)` here as well

tests/ludwig/augmentation/test_augmentation_pipeline.py
ksbrar: Gotcha. Yes that is the correct understanding!

To close this loop, I would like if you could add tests that explicitly check if an error is raised if a user manually sets `augmentation: None` (note that this is subtly different from the user choosing _not_ to specify `augmentation` at all) or to some erroneous (non-list) type. You could put maybe one here to see that the error is surfaced through to the pipeline logic layer and then the rest under `tests/schema` where you can use `validate_config` to do some basic schema validation tests.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+_augmentation_config_registry = Registry()
+
+
+@DeveloperAPI
+def get_augmentation_config_registry() -> Registry:
+    return _augmentation_config_registry
+
+
+def register_augmentation_config(name: str):
```
ksbrar, line 18: One further refinement of this registry: since `BaseAugmentationConfig` will eventually be the base of non-image augmentations as well, we want to have a one-to-many mapping of augmentation types to related feature types (rather than having one registry where everything is dumped). See `register_encoder_config` for an example of how to do this.

ksbrar (changes requested):
Awesome work! 🚀 🟢🚀 🟢 

Added a couple more comments related to tests and refinements.

ludwig/data/dataset/ray.py
```diff
@@ -336,6 +351,11 @@ def _prepare_batch(self, batch: pd.DataFrame) -> Dict[str, np.ndarray]:
             else:
                 res[c] = batch[c].to_numpy()
 
+        if self.augmentation_pipeline:
```
tgaddair, line 354: Nice! The only issue with this approach is it does the augmentation on the "consumer" process (the one doing the forward / backward pass), when it would be more efficient to offload this to the data ingest processes (the "producers"). Looks like we removed the code that showed and example of this in practice, but I found it by going back in the history a bit:

https://github.com/ludwig-ai/ludwig/blob/bec48803f859b64ca439d6b964dde766c197861c/ludwig/data/dataset/ray.py#L413

Notice the `pipeline.map_batches()` call. What this is doing is lazily defining a set of transformations to perform on the input data before sending it to the local process for consumption.

In this case, we can define a similar transformation for data augmentation like so:

```
    def _augment_batch_fn(self):
        pipeline = self.augmentation_pipeline

        def augment_batch(df: pd.DataFrame) -> pd.DataFrame:
            for c, augmentations in self.augmentation_pipeline.items():
                logger.info(f"RayDatasetBatcher applying augmentation pipeline to batch for feature {c}")
                df[c] = augmentations(torch.tensor(df[c]))
            return df

...

augment_batch = self._augment_batch_fn()

for batch in pipeline.map_batches(augment_batch, batch_size=batch_size, batch_format="pandas").iter_batches( 
    prefetch_blocks=0, batch_size=batch_size, batch_format="pandas"
):
    res = self._prepare_batch(batch)
```

jimthompson5802: @tgaddair   Got it.  I need to move the execution of augmentation from the "consumer" task to the "producer" task.  I'll take a look at the cited code and if I have questions, I'll reach back out.

thank you for the guidance.

jimthompson5802: @tgaddair   Made the change as requested:  moved executing augumentation from "consumer" to "producer" process.

Turns out `_augment_batch_fn()` was bit more involved.  Since `map_batches()` uses `"pandas"` format, the data seen in the function is a `pd.Series`.  I had to extract out the underlying numpy arrays, consolidate into a batch of np.arrays to feed into the augmentation pipeline.  Once this was done, then split the batch into the individual arrays to make it conform to the incoming `pandas` dataframe shape.  

At this point, no run-time exceptions and my simple end-to-end test program with synthetic data runs to completion.

Here is an example of the log message when augmentation is executed.  As you can see the augmentation is occurring id three diffferent processes.
```
(_map_block_nosplit pid=1723) RayDatasetBatcher applying augmentation pipeline to batch for feature my_image_P6DoV_
(_map_block_nosplit pid=1723) RayDatasetBatcher applying augmentation pipeline to batch for feature my_image_P6DoV_
(_map_block_nosplit pid=1720) RayDatasetBatcher applying augmentation pipeline to batch for feature my_image_P6DoV_
(_map_block_nosplit pid=1720) RayDatasetBatcher applying augmentation pipeline to batch for feature my_image_P6DoV_
(_map_block_nosplit pid=1721) RayDatasetBatcher applying augmentation pipeline to batch for feature my_image_P6DoV_
```

I'm attaching the full log file if you want to see all the messages. 
[ray_with_augmentation.txt](https://github.com/ludwig-ai/ludwig/files/10429484/ray_with_augmentation.txt)



Re: the numpy array manipulations described above, there is an alternative way that could avoid the batching and splitting operations.  I decided to do the above to make it consistent with how the Local augmentation operation runs.  If you want me to try this alternative way let me know.

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

ludwig/data/dataset/ray.py
```diff
@@ -342,6 +357,31 @@ def _prepare_batch(self, batch: pd.DataFrame) -> Dict[str, np.ndarray]:
...
+
+        def augment_batch(df: pd.DataFrame) -> pd.DataFrame:
+            if augmentation_pipeline:
+                for c, augmentations in augmentation_pipeline.items():
+                    # TODO: convert to debug message when done with development
+                    logger.info(f"RayDatasetBatcher applying augmentation pipeline to batch for feature {c}")
+
+                    # df[c] is a pd.Series, so we need to convert it to a np.array
+                    # consolidate the individual np.array into a single np.array for the batch
+                    data_to_augment = np.stack(df[c].values, axis=0)
```
tgaddair, line 371: Nice! You may be able to simplify by using `to_numpy_dataset` and `from_numpy_dataset.

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/dataframe_utils.py#L64-L94

Like:

```
data_to_augment = to_numpy_dataset(df)
augmented_data = {}
for c, augmentations in augmentation_pipeline.items():
    augmented_data[c] = augmentations(torch.tensor(data_to_augment[c])).numpy()
return from_numpy_dataset(augmented_data)    
```

tgaddair: This assumes there is a column in the augmentation pipeline for every column in the input df. If not, we could rearrange a little bit to ensure we also include the columns that aren't augmented.

ludwig/data/dataset/ray.py
```diff
@@ -352,9 +392,12 @@ def sync_read():
     def _create_async_reader(self, pipeline: DatasetPipeline):
         q = queue.Queue(maxsize=100)
         batch_size = self.batch_size
+        augment_batch = self._augment_batch_fn()
 
         def producer():
-            for batch in pipeline.iter_batches(prefetch_blocks=0, batch_size=batch_size, batch_format="pandas"):
+            for batch in pipeline.map_batches(augment_batch, batch_size=batch_size, batch_format="pandas").iter_batches(
```
tgaddair, line 398: Nice!

Nit: if the augmentation pipeline is empty, we can skip calling `pipeline.map_batches`, which will save us an extra step in the pipeline. Something like:

```
def producer:
    if not augmentation_pipeline.empty():
        pipeline = pipeline.map_batches(augment_batch, batch_size=batch_size, batch_format="pandas")
```

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

ludwig/schema/features/augmentation/utils.py
jimthompson5802: @ksbrar   I think the change for this comment https://github.com/ludwig-ai/ludwig/pull/2925/files#r1070381132  may have addressed this.

What do you think?

jimthompson5802: @ksbrar    Nevermind...I misunderstood your commet.  I re-read it and see where I went astray.  I'll make the suggested fix.

ludwig/data/dataset/ray.py
jimthompson5802: done

ludwig/data/dataset/ray.py
jimthompson5802: @tgaddair   made the change.  Take a look at the use of nonlocal.  I believe it is correct use.  Let me know.

tests/ludwig/augmentation/test_augmentation_pipeline.py
jimthompson5802: with `augmentation` set to `None` in the config file, this exception is raised
```
File "/usr/local/lib/python3.9/site-packages/jsonschema/validators.py", line 1058, in validate
    raise error
jsonschema.exceptions.ValidationError: None is not of type 'array'
```

this should address the comment.

ksbrar: great!

ludwig/schema/features/augmentation/utils.py
jimthompson5802: With last set of changes, I've addressed this.

jimthompson5802:
Depedent on PR #2992 to support the augmentation pipeline validation

ludwig/features/image_feature.py
```diff
@@ -637,6 +806,34 @@ def __init__(self, input_feature_config: ImageInputFeatureConfig, encoder_obj=No
         else:
             self.encoder_obj = self.initialize_encoder(input_feature_config.encoder)
 
+        # set up for augmentation if specified
+        if input_feature_config.augmentation:
```
ksbrar, line 810: So based on this comment and how you declared the `augmentation` parameter on the image feature schema you will explicitly allowing a user to set `augmentation` to `None` (though the default if they specify nothing will be `False`). Want to confirm if that is what you intended.

ludwig/schema/features/image_feature.py
```diff
@@ -23,6 +28,21 @@ class ImageInputFeatureConfigMixin(BaseMarshmallowConfig):
         default="stacked_cnn",
     )
 
+    augmentation: Union[bool, List[BaseAugmentationConfig]] = schema_utils.OneOfOptionsField(
```
ksbrar, line 31: As mentioned in an above comment, without explicitly setting `allow_none=False` here the parameter actually can be set to `None`. If that's intentional then also add it to the type annotation.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,156 @@
...
+@DeveloperAPI
+def AugmentationContainerDataclassField(feature_type: str, default=[], description=""):
+    """Custom dataclass field that when used inside a dataclass will allow the user to specify an augmentation
+    config.
+
+    Args:
+        default: The default augmentation config to use.
+        description: The description of the augmentation config.
+
+    Returns: Initialized dataclass field that converts an untyped dict with params to an augmentation config.
```
ksbrar, line 53: description nit: this converts a *list* to an augmentation

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,156 @@
...
+    return get_augmentation_config_registry()[feature][name]
+
+
+@DeveloperAPI
+def get_augmentation_classes(feature: str):
+    return get_augmentation_config_registry()[feature]
+
+
+@DeveloperAPI
+def AugmentationContainerDataclassField(feature_type: str, default=[], description=""):
```
ksbrar, line 45: nit: let's be super explicit this and call this <pre>Augmentation<i>**List**</i>DataclassField</pre>. If this class actually handled the boolean case for augmentations then "container" would be appropriate (since it would be handling everything), but this class _only_ handles _lists_.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,156 @@
...
+                )
+            },
+            default_factory=lambda: load_default,
+        )
+    except Exception as e:
+        raise ValidationError(f"Unsupported augmentation type. See augmentation_registry. " f"Details: {e}")
+
+
+@DeveloperAPI
+def get_augmentation_jsonschema(feature_type: str):
```
ksbrar, line 111: nit: similarly: <pre>get_augmentation_<i>list</i>_jsonschema</pre>

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,156 @@
...
+            "title": "augmentation",
+        },
+        # "uniqueItemProperties": ["name"],
+    }
+
+    return schema
+
+
+@DeveloperAPI
+def get_augmentation_conds(feature_type: str):
```
ksbrar, line 142: nit: similarly: <pre>get_augmentation_<i>list</i>_conds</pre>

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,156 @@
...
+        """Custom marshmallow field that deserializes a list for a valid augmentation config from the
+        augmentation_registry and creates a corresponding JSON schema for external usage."""
+
+        def _deserialize(self, value, attr, data, **kwargs):
+            assert isinstance(value, list), "Augmentation config must be a list."
+
+            augmentation_list = []
+            for augmentation in value:
+                augmentation_op = augmentation[TYPE]
+                augmentation_cls = get_augmentation_cls(augmentation_op)
```
ksbrar, line 66: I think you are missing the `feature_type` argument here?

tests/ludwig/augmentation/test_image_augmentation.py
```diff
@@ -0,0 +1,29 @@
+import pytest
```
ksbrar, line 1: Let's add a test to this file that checks whether unexpected types supplied to the `augmentation` parameter cause a validation error.

ksbrar (changes requested):
Hi @jimthompson5802 , thanks for all the great work! I have a few final nits and questions.

ludwig/features/image_feature.py
jimthompson5802: @ksbrar 
Not the intent.   `None` should be flagged as an error.  The only valid values should be
* False
* True
* List of augmentation operations.

That code was written before the decision was made to allow boolean values. 

What is the best way to resolve this?  check explictly for `None` and raise an exception if it is `None`.

ludwig/schema/features/image_feature.py
jimthompson5802: Since the intent is not to support `None`, I'll leave this alone and look for guidance on the earlier comment.

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

jimthompson5802: Though it is curious this error was not flagged in any unit test. 

When does `_deseralize()` method invoked?  What would be a reasonable way to unit test this code path?

tests/ludwig/augmentation/test_image_augmentation.py
jimthompson5802: @ksbrar 
Took  a stab at the test.  https://github.com/ludwig-ai/ludwig/blob/d6ff7e3db9194866e477069481070a1f55aad98a/tests/ludwig/augmentation/test_augmentation_pipeline.py#L267-L289

If this looks reasonable, how extensive should the tests be for invalid parameters.  For example, `random_rotate` parameters needs to be a positive odd integer.  If we have to test for invalid values for this augmentation, I may need some help to define the schema conditions.

ludwig/features/image_feature.py
ksbrar: @jimthompson5802 You don't need to do anything here. I would just change the comment to # set up for augmentation if its enabled. Validation will handle any unsupported parameters before this code path gets hit, and then when it finally does this if statement should logically only be checking if `augmentation` is `False` (and if its `True` or a `List` then as it already does it'll set up the pipeline).

I'll add the details for the fix to the other thread.

ludwig/schema/features/image_feature.py
ksbrar: To fix this, simply add `None` to the `Union` and set `allow_none=False`

ludwig/schema/features/augmentation/utils.py
ksbrar: I think it passed because `get_augmentation_cls` will use `image` by default if nothing else is provided, which makes sense.

`_deserialize` can only be invoked at runtime when Ludwig tries to convert a user config to a config object. If the user specifies nothing for `augmentation` then its actually skipped because the compile-time `default` will be used. Any test that attempts training with a user-specified `augmentation` will hit this code path.

tests/ludwig/augmentation/test_image_augmentation.py
ksbrar: This is a good one. I would add these cases:
* `None`
* Empty list `[]`
* Any random value like a string or something

ludwig/features/image_feature.py
jimthompson5802: got it.  revised comment.

ludwig/schema/features/image_feature.py
jimthompson5802: done.

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done.

ludwig/schema/features/augmentation/utils.py
jimthompson5802: done

tests/ludwig/augmentation/test_image_augmentation.py
jimthompson5802: @ksbrar   Here are the test cases for the failure test
```
@pytest.mark.parametrize(
    "augmentation_pipeline_ops",
    [
        None,
        [
            {"type": "invalid_string"}
        ],
        [
            "random_horizontal_flip"
        ],
        "random_horizontal_flip",
        [
            {"type": "random_rotate", "degree": "45"},
        ],
    ],
)
```

All of these generate the `ValidationError`.

Interestingly enough, an empty list "`[]`" does not generate an error.  I'm still researching this.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -0,0 +1,156 @@
...
+@DeveloperAPI
+def get_augmentation_list_jsonschema(feature_type: str):
+    """This function returns a JSON augmentation schema.
+
+    Returns: JSON Schema
+    """
+    augmentation_types = sorted(list(get_augmentation_config_registry()[feature_type].keys()))
+    schema = {
+        "type": "array",
+        "minItems": 0,
```
ksbrar, line 119: @jimthompson5802 This is why `[]` does not throw a validation error. You should also add a check to the `_deserialize` that explicitly checks the length of the list is >= 1 to match.

jimthompson5802: I added the check to `_deserialize` and still do not generate `ValidationError` with empty list.

If I set `minItems`  to 1 and failure test catches the `ValidationError` with empthy list but the other tests that should work now fail.

Should I have set `minItems` to 1?
 
If I set `minItems` back to 0, I believe I can still catch the empty list situation but the code will need to be in `ECD.get_augmentation_pipelines()` method.  This means the "validation" logic not contained completely in the schema validation code base.

jimthompson5802: Setting `minItems` back to zero and adding a test in `ECD.get_augmentation_pipelines()` work.  

All augmentation pipeline tests now pass include the test to confirm failure when augmentation pipeline is an empty list.

ksbrar: @jimthompson5802 If that is the case, something a little funky is still going on. You should set `minItems` to 1 because requirement is what should throw the validation error for the empty list. Can you trace where precisely in the code the error is thrown for that case?

> added the check to _deserialize and still do not generate ValidationError with empty list

That is fine. The checks in `_deserialize` are the last line of defense - the error for the empty list should be thrown before reaching it.

> but the code will need to be in ECD.get_augmentation_pipelines() method. This means the "validation" logic not contained completely in the schema validation code base.

That makes sense, though we should be able to avoid this.

jimthompson5802: OK.  I'll set minItems back to zero and trace where the error is thrown.

ludwig/schema/model_config.py
```diff
@@ -216,10 +216,6 @@ def __init__(self, config_dict: ModelConfigDict):
         if self.model_type == MODEL_GBM:
             self.combiner = None
 
-        # TODO: alternative approaches
-        # special handling for image augmentation default
-        # self.defaults.image.augmentation = []
```
jimthompson5802: @ksbrar   Found issue with not flagging empty list.   this code used to be active and was placed here when I first tried doing the schema validation modifications and before you helped me with working through the issues.

By remove the code, `minItems: 1` is now set and the test for invalid parameters flags an empyt list.

So I think this takes care of all the feedback as of now.

ludwig/schema/features/augmentation/utils.py
jimthompson5802: @ksbrar   this commit https://github.com/ludwig-ai/ludwig/pull/2925/commits/4039e55777f25a86ad98655d213b8eef4159ce1c takes care of the issue.  this code was first put there when I first tried modifying the schema validation and before you helped me work through the issues.   

With the removal of the code and setting `minItems: 1`, the validation test now correctly flags an empty list as an error.

ksbrar: Wonderful!

ksbrar (approved):
@jimthompson5802 From a schema perspective, I believe we are complete. Once again, thanks for the major effort in getting this all tightly implemented 🚀 🚀 

@tgaddair Not sure about your feelings about when to land which PR, but if your marshmallow refactor goes in first we can make a few more small changes here as well and the major benefit will of course be that we can use the `TypeSelection` interface.

ksbrar:
@jimthompson5802 Now that https://github.com/ludwig-ai/ludwig/pull/2906 has gone through we can make a few more changes. I'll post some more details later

tgaddair (approved):

PR #2927, opened 2023-01-13
Store mlflow tracking URI to ensure consistency across processes

tgaddair:
In the previous implementation, we pull the tracking URI from the environment every time we change context, which can result in inconsistent behavior if, for example, the current working directory changes and no external MLFlow service is being used.


geoffreyangus (approved):
LGTM!

arnavgarg1 (approved):

PR #2934, opened 2023-01-14
Tune batch size using distributed training to catch edge case CUDA OOMs

tgaddair:
In some cases, CUDA OOMs can occur only when the additional overhead of distributed training (from the GPU fusion buffer in Horovod) are introduced. This PR runs batch size eval in a full distributed context to ensure these edge cases are accounted for when selecting the best batch size.


arnavgarg1 (approved):

ludwig/backend/ray.py
```diff
@@ -340,19 +341,26 @@ def run(
...
+        callbacks = callbacks or []
         trainer_cls = HorovodTrainerRay210 if not _ray220 else HorovodTrainer
         trainer = trainer_cls(
             train_loop_per_worker=train_loop_per_worker,
             train_loop_config=config,
             horovod_config=self.backend_config,
             datasets=dataset,
             scaling_config=self.scaling_config,
-            dataset_config=self._get_dataset_configs(dataset, stream_window_size, data_loader_kwargs),
+            dataset_config=dataset_config,
```
geoffreyangus, line 363: In what settings might the `dataset_config` be `None`?

tgaddair: This one, we don't use the dataset for batch size tuning.

geoffreyangus: Ah I see– just took a closer look at the tune batch size logic. Thanks!

ludwig/backend/ray.py
```diff
@@ -358,22 +394,35 @@ def run(
...
         trainer_cls, kwargs = get_dist_strategy(self.strategy).get_trainer_cls(self.backend_config)
         trainer = trainer_cls(
             train_loop_per_worker=train_loop_per_worker,
             train_loop_config=config,
             datasets=dataset,
             scaling_config=self.scaling_config,
-            dataset_config=self._get_dataset_configs(dataset, stream_window_size, data_loader_kwargs),
-            run_config=RunConfig(callbacks=callbacks, verbose=0),
+            dataset_config=dataset_config,
+            run_config=RunConfig(callbacks=callbacks, verbose=1),
```
arnavgarg1, line 418: I'd revert back to 0, because this creates too much logging when we run hyperopt.

ludwig/trainers/trainer.py
```diff
@@ -349,6 +360,34 @@ def write_step_summary(cls, train_summary_writer, combined_loss, all_losses, ste
 
         train_summary_writer.flush()
 
+    def train_for_tuning(
```
tgaddair, line 363: This function is no longer used. There's a new API called `BatchSizeEvaluator` that we use instead.

tgaddair (commented):
LGTM

PR #2941, opened 2023-01-15
[automl] Exclude text fields with low avg words

tgaddair:
Many tabular datasets have text features that aren't particularly informative like "name", which are really just unique identifiers, but tricky to identify as such (as they can contain more than one word).

This PR adds heuristics to filter out such fields under the assumption that they occur primarily when there are many other fields and when the number of average words per field is low.


arnavgarg1:
Nice, this is a good addition

PR #2942, opened 2023-01-16
Remove safe_move_directory

arnavgarg1:
Removes this function since it doesn't work as intended


tgaddair (changes requested):
Copy is not atomic, had to be a move.

PR #2945, opened 2023-01-17
Integrated Gradients: reset sample_ratio to 1.0 if set

jppgks:
If `model.config_obj.preprocessing.sample_ratio` was set for a model, it was also applied to the explanation input. For small input sizes, this could result in empty explanation inputs.

Resolution: reset sample_ratio to 1.0 before preprocessing explanation inputs.


tgaddair (commented):
Should we set it back when we're done?

PR #2951, opened 2023-01-18
Retry HuggingFace pretrained model download on failure

jeffkinnison:
This adds a utility function to download pretrained HuggingFace models with retry on download failure. Calls to from_pretrained in text encoder and tokenizer classes have been replaced with calls to the utility.


tgaddair (approved):
Nice!

PR #2957, opened 2023-01-18
fix: enable binary dense encoder

abidwael:
This PR fixes binary feature dense encoder


ludwig/encoders/binary_encoders.py
```diff
@@ -35,15 +35,15 @@ def __init__(self, encoder_config=None, **kwargs):
 
         logger.debug(f" {self.name}")
 
-    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+    def forward(self, inputs: torch.Tensor) -> dict:
```
justinxzhao, line 38: Are there other encoders that aren't forwarding Dict outputs?

justinxzhao (approved):

ludwig/encoders/binary_encoders.py
tgaddair: Can you make this `Dict[str, torch.Tensor]`?

arnavgarg1 (approved):

ludwig/modules/fully_connected_modules.py
```diff
@@ -93,8 +92,10 @@ def __init__(
             self.layers.append(Dropout(dropout))
 
     def forward(self, inputs, mask=None):
-        hidden = inputs
+        if inputs.dtype == torch.bool:
+            inputs = inputs.to(torch.float32)
```
justinxzhao, line 96: Why is this necessary?

abidwael: Binary input gets passed with dtype `torch.bool` and the linear layer weights are `torch.float32` and they need to be of the same dtype.

justinxzhao: I suggest adding a comment to that effect to this code block

abidwael: added!

tgaddair: Should we do this case at the feature / encoder level instead of down here?

ludwig/encoders/binary_encoders.py
abidwael: @justinxzhao both the `BinaryPassthroughEncoder` and the generic `DenseEncoder` which are the two available encoders for binary features return dictionaries.

ludwig/modules/fully_connected_modules.py
tgaddair: Specifically, Binary inputs only support two encoders: dense and passthrough. Both of them require a cast from bool to float32 as the first step. If we introduce a new encoder, we'll probably need to do the same thing. If we were to make this change of casting at the feature level, we could remove `BinaryPassthroughEncoder` and have binary features just use the generic `passthrough` and `dense` encoders.

abidwael: Agree that it's a lot more elegant to do it at the feature level. I made the change which required adding an `is_input_feature` argument to `add_feature_data`.

tgaddair (changes requested):
Hey @abidwael, I actually meant we can do it in the input feature's forward function. Just before this line:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/features/binary_feature.py#L215

This would be consistent with how we handle other such type conversions today. For example, here's where we do the same thing for category features:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/features/category_feature.py#L236

ludwig/modules/fully_connected_modules.py
tgaddair: I would rather not do it in preprocessing, but during the feature's forward function. See comment below.

tgaddair (approved):
LGTM!

PR #2959, opened 2023-01-19
Move all config validation to its own standalone module, `config_validation`.

justinxzhao:
Motivation:
- Config validation is becoming its own key service in its own right. `schema/__init__.py` is imported everywhere in Ludwig, and is ripe for circular dependency issues as validation grows.
- There are several cases where it's useful to expand config validation to non-schema auxiliary checks, like those that [aren't easily covered by marshmallow schemas like parameter interdependencies](https://github.com/ludwig-ai/ludwig/pull/2811/files#diff-274da73fbe753bf48e11f0f22e07e861c6b022837d28a33439ae7fd00bd1de8fR1), which are scattered throughout the codebase, as well as for recent changes like #2956 and #2954.


ksbrar (approved):
YES!

PR #2961, opened 2023-01-19
Add `sacremoses` as a dependency for transformer_xl encoder

arnavgarg1:


tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -107,7 +107,6 @@ def test_distil_bert(use_pretrained: bool, reduce_output: str, max_sequence_leng
     assert outputs["encoder_output"].shape[1:] == distil_bert_encoder.output_shape
 
 
-@slow
```
abidwael, line 110: Why did we remove the `slow` marker?

abidwael (approved):

tests/ludwig/encoders/test_text_encoders.py
arnavgarg1: @abidwael Want to make sure we start running tests for this text encoder. It's not actually that slow to run to be honest!

jeffkinnison (approved):

justinxzhao (approved):

PR #2968, opened 2023-01-19
Populate hyperopt defaults using schema

arnavgarg1:
This PR adds a new step to ensure that all of the defaults for search algorithm, split, executor etc. are correctly populated using the schema object before accessing them to prevent key errors. All of these parameters are required by hyperopt and expected to be present downstream for config validation, executor creation, and trial execution.

The reason we've not noticed this before in our tests is because all of our tests specify `search_alg` in the hyperopt config. 

Co-authored-by: Kabir Brar <kabir@predibase.com>
Co-authored-by: Travis Addair <tgaddair@gmail.com>

### DIFF
--- a/ludwig/hyperopt/run.py
+++ b/ludwig/hyperopt/run.py
@@ -41,7 +41,6 @@ from ludwig.hyperopt.utils import (
     should_tune_preprocessing,
     update_hyperopt_params_with_defaults,
 )
-from ludwig.schema.hyperopt import HyperoptConfig
 from ludwig.schema.model_config import ModelConfig
 from ludwig.utils.backward_compatibility import upgrade_config_dict_to_latest_version
 from ludwig.utils.dataset_utils import generate_dataset_statistics
@@ -223,11 +222,6 @@ def hyperopt(
     full_config = config_obj.to_dict()  # TODO (Connor): Refactor to use config object
 
     hyperopt_config = full_config[HYPEROPT]
-
-    # Convert hyperopt config to hyperopt schema to populate with schema defaults
-    # This fills in missing splits, executor config, search_alg, etc.
-    hyperopt_config_from_schema = HyperoptConfig.Schema().load(hyperopt_config)
-    hyperopt_config = hyperopt_config_from_schema.to_dict()
 
     # Explicitly default to a local backend to avoid picking up Ray or Horovod
     # backend from the environment.
--- a/ludwig/hyperopt/run.py
+++ b/ludwig/hyperopt/run.py
@@ -41,6 +41,7 @@ from ludwig.hyperopt.utils import (
     should_tune_preprocessing,
     update_hyperopt_params_with_defaults,
 )
+from ludwig.schema.hyperopt import HyperoptConfig
 from ludwig.schema.model_config import ModelConfig
 from ludwig.utils.backward_compatibility import upgrade_config_dict_to_latest_version
 from ludwig.utils.dataset_utils import generate_dataset_statistics
@@ -222,6 +223,11 @@ def hyperopt(
     full_config = config_obj.to_dict()  # TODO (Connor): Refactor to use config object
 
     hyperopt_config = full_config[HYPEROPT]
+
+    # Convert hyperopt config to hyperopt schema to populate with schema defaults
+    # This fills in missing splits, executor config, search_alg, etc.
+    hyperopt_config_from_schema = HyperoptConfig.Schema().load(hyperopt_config)
+    hyperopt_config = hyperopt_config_from_schema.to_dict()
 
     # Explicitly default to a local backend to avoid picking up Ray or Horovod
     # backend from the environment.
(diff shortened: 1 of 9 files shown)


ludwig/hyperopt/run.py
```diff
@@ -223,6 +224,11 @@ def hyperopt(
 
     hyperopt_config = full_config[HYPEROPT]
 
+    # Convert hyperopt config to hyperopt schema to populate with schema defaults
+    # This fills in missing splits, executor config, search_alg, etc.
+    hyperopt_config_from_schema = HyperoptConfig.Schema().load(hyperopt_config)
+    hyperopt_config = hyperopt_config_from_schema.to_dict()
```
justinxzhao, line 230: Would the ModelConfig class, e.g. `model_config.py::ModelConfig::_set_hyperopt_defaults()` be a more appropriate place for this code?

arnavgarg1: That's reasonable, I can move it inside there.

tests/integration_tests/test_hyperopt.py
```diff
@@ -557,3 +557,34 @@ def test_hyperopt_nested_parameters(csv_filename, tmpdir, ray_cluster_7cpu):
...
+            "parameters": {
+                "trainer.learning_rate": {
+                    "lower": 0.0001,
+                    "upper": 0.01,
+                    "space": "loguniform",
+                }
+            },
+            "goal": "minimize",
+            "output_feature": output_features[0]["name"],
+            "validation_metrics": "loss",
```
abidwael, line 584: I've also seen `metric` being used in some hyperopt configs (e.g. https://ludwig.ai/latest/configuration/hyperparameter_optimization/). Which one is the official/most recent parameter name?

ksbrar: Interesting. @arnavgarg1 note that `validation_metrics` is not currently on the hyperopt schema.

arnavgarg1: Good catch. The reason the test still works is because of backward compatibility. In either case, I've changed this to be `metric` because that is the latest. Apologies for the confusion (copied the config from another test and then modified it)

tests/integration_tests/test_hyperopt_ray.py
```diff
@@ -207,7 +208,9 @@ def run_hyperopt_executor(
 def test_hyperopt_executor(scenario, csv_filename, tmpdir, ray_cluster_4cpu):
     search_alg = scenario["search_alg"]
     executor = scenario["executor"]
-    run_hyperopt_executor(search_alg, executor, csv_filename, tmpdir)
+    # When using the hb_bohb scheduler, num_epochs must equal max_t (which is 81 by default)
```
justinxzhao, line 211: How did you come to discover this? Is this worth adding as an auxiliary validation?

justinxzhao (approved):

ludwig/schema/model_config.py
```diff
@@ -511,7 +513,11 @@ def _set_hyperopt_defaults(self):
         if not self.hyperopt:
             return
 
-        scheduler = self.hyperopt.get("executor", {}).get("scheduler")
+        # Convert hyperopt config to hyperopt schema to populate with schema defaults
+        # This fills in missing splits, executor config, search_alg, etc.
+        self.hyperopt = HyperoptConfig.Schema().load(self.hyperopt).to_dict()
```
tgaddair, line 518: Should be:

```
HyperoptConfig.from_dict(..).to_dict()
```

Otherwise we circumvent the class cache layer and leak memory.

jeffkinnison (approved):

ludwig/schema/model_config.py
ksbrar: @tgaddair You mean `ModelConfig.from_dict`, right?

tests/integration_tests/test_hyperopt_ray.py
arnavgarg1: I didn't know about this either, but found out that we've added this logic here in our ModelConfig object creation logic: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/schema/model_config.py#L532

ludwig/schema/model_config.py
```diff
@@ -511,7 +511,7 @@ def _set_hyperopt_defaults(self):
         if not self.hyperopt:
             return
 
-        scheduler = self.hyperopt.get("executor", {}).get("scheduler")
+        scheduler = self.hyperopt.get("executor", {}).get("scheduler", {})
```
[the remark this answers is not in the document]
arnavgarg1: I did this initially because I found that it was possible for the `scheduler` key to not be inside the `executor` dict, but that's no longer the case once we populate `self.hyperopt` from the `HyperoptConfig` so I will remove it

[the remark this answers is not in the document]
tgaddair: Even if `scheduler` is not in the dict, it will just return `None`.

PR #2970, opened 2023-01-20
fix: Force populate combiner registry inside of `get_schema` function.

ksbrar:
At the moment, directly calling `get_schema` results in an empty combiner schema.


ludwig/schema/__init__.py
```diff
@@ -39,6 +39,9 @@ def get_ludwig_version_jsonschema():
 @DeveloperAPI
 @lru_cache(maxsize=2)
 def get_schema(model_type: str = MODEL_ECD):
+    # Force populate combiner registry:
+    import ludwig.combiners.combiners  # noqa: F401
```
justinxzhao, line 43: @ksbrar I'm curious is it still the case if you try to call `config_validation.validation.py::get_schema()`?

ksbrar: @justinxzhao Yup, still the case. Shall I add this change there as well?

tgaddair: We should make a point of decoupling these. The combiner schemas should register independently of the actual combiner implementations.

justinxzhao: > @justinxzhao Yup, still the case. Shall I add this change there as well?

Yes please. The plan is to move all callers of schema/__init__.py to use the config validations module.

justinxzhao (approved):

ludwig/schema/__init__.py
justinxzhao: Is there a test we can add to `tests/ludwig/config_validation/test_validate_config_combiner.py` to solidify that the combiner schema isn't empty?

ksbrar: @justinxzhao I just added a simple assertion test. Unfortunately with the way python caches modules it doesn't really test what we want it to test - (e.g. if I comment out my fix here the test will still pass because the combiner registry is populated elsewhere in the cached import chain). I'm not sure if there is an easy way to resolve this - do you have any thoughts?

connor-mccorm (approved):

ludwig/schema/__init__.py
justinxzhao: Yeah, that's ok then. No need for the test.

justinxzhao (approved):

ludwig/schema/__init__.py
ksbrar: Will keep it for now.

PR #2979, opened 2023-01-20
Reorder Advanced Parameters

connor-mccorm:
This PR reorders the schema params to be in a more intuitive for users given the expected impact value of each parameter. Parameters with an expected impact of 2 are considered advanced and will be hidden under an advanced parameter accordion.

### DIFF
--- a/ludwig/schema/decoders/base.py
+++ b/ludwig/schema/decoders/base.py
@@ -11,7 +11,7 @@ from ludwig.schema.metadata import DECODER_METADATA
 
 
 @DeveloperAPI
-@dataclass(repr=False)
+@dataclass(repr=False, order=True)
 class BaseDecoderConfig(schema_utils.BaseMarshmallowConfig, ABC):
     """Base class for decoders."""
 
@@ -105,7 +105,7 @@ class BaseDecoderConfig(schema_utils.BaseMarshmallowConfig, ABC):
 
 @DeveloperAPI
 @register_decoder_config("passthrough", [BINARY, CATEGORY, NUMBER, SET, VECTOR, SEQUENCE, TEXT])
-@dataclass(repr=False)
+@dataclass(repr=False, order=True)
 class PassthroughDecoderConfig(BaseDecoderConfig):
     """PassthroughDecoderConfig is a dataclass that configures the parameters used for a passthrough decoder."""
 
@@ -129,7 +129,7 @@ class PassthroughDecoderConfig(BaseDecoderConfig):
 
 @DeveloperAPI
 @register_decoder_config("regressor", [BINARY, NUMBER])
-@dataclass(repr=False)
+@dataclass(repr=False, order=True)
 class RegressorConfig(BaseDecoderConfig):
     """RegressorConfig is a dataclass that configures the parameters used for a regressor decoder."""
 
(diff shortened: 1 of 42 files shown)


ludwig/schema/lr_scheduler.py
```diff
@@ -48,6 +30,12 @@ class LRSchedulerConfig(schema_utils.BaseMarshmallowConfig, ABC):
         parameter_metadata=TRAINER_METADATA["learning_rate_scheduler"]["decay_steps"],
     )
 
+    decay_steps: int = schema_utils.PositiveInteger(
```
ksbrar, line 33: Did you mean to add these parameters as part of this PR?

ludwig/schema/decoders/base.py
```diff
@@ -11,7 +11,7 @@
 
 
 @DeveloperAPI
-@dataclass(repr=False)
+@dataclass(repr=False, order=True)
```
justinxzhao, line 14: Idea: define this dataclass in `ludwig/schema/utils`.

```python
from marshmallow_dataclass import dataclass

ludwig_schema_dataclass = dataclass(repr=False, order=True)
```

Then in all subsequent schema dataclass definitions, use: `@ludwig_schema_dataclass`

ludwig/schema/features/preprocessing/set.py
```diff
@@ -63,7 +63,7 @@ class SetPreprocessingConfig(BasePreprocessingConfig):
 
 @DeveloperAPI
 @register_preprocessor("set_output")
-@dataclass(repr=False)
+@dataclass(repr=False, order=True)
 class SetOutputPreprocessingConfig(SetPreprocessingConfig):
```
justinxzhao, line 67: Should Set/TextOutputPreprocessingConfig inherit from `SequenceOutputPreprocessingConfig`?

ludwig/schema/lr_scheduler.py
justinxzhao: Looks like they are just re-ordered.

justinxzhao (approved):
One small refactoring idea. Can leave to a follow-up PR if you prefer.

ludwig/schema/decoders/base.py
ksbrar: +1

ludwig/schema/lr_scheduler.py
ksbrar: Obviously... sorry for the brain freeze @connor-mccorm ! Resolving this.

ksbrar (approved):

ludwig/schema/decoders/base.py
connor-mccorm: This is a great idea! I'll create a linear ticket so I can do this in a followup PR!

ludwig/schema/features/preprocessing/set.py
connor-mccorm: We could have `TextOuptutPreprocessingConfig` inherit from `SequenceOutputPreprocessingConfig` (the defaults are a little too different for set for it to make sense IMO. However, I would say that if we do want to do this, we would want the text input feature preprocessing config to inherit from sequence as well and there looks to be one default discrepancy so I think it would warrant it's own PR. Personally, although we can inherit here, I think there's value in a verbose list here for a couple reasons. **A)** it preserves the order of the schema - inheritance ends up tacking any overridden parameters on the end which causes the order to be unideal sometimes. **B)**  I think that for documentation purposes, I find it kind of convenient to be able to project search a specific default without having to redirect to a file where the defaults are inherited. If we think it's valuable enough I'm happy to change, but my vote would be to keep them separated. WDYT?

PR #2981, opened 2023-01-20
MT5 reduce_output can't be cls_pooled - set to sum by default

arnavgarg1:
Same as the PR title

### DIFF
--- a/ludwig/schema/encoders/text_encoders.py
+++ b/ludwig/schema/encoders/text_encoders.py
@@ -254,7 +254,7 @@ class MT5Config(BaseEncoderConfig):
     )
 
     reduce_output: str = schema_utils.String(
-        default="cls_pooled",
+        default="sum",
         description="The method used to reduce a sequence of tensors down to a single tensor.",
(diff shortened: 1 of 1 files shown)


geoffreyangus:
Can we ensure that the default `reduce_output` for each HF encoder is tested here: https://github.com/ludwig-ai/ludwig/blob/62c25edf512a6a81a1adbff63ab67e96d8642056/tests/ludwig/encoders/test_text_encoders.py#L10

You can pull directly from the text encoder config like this: https://github.com/ludwig-ai/ludwig/pull/2962/files#diff-d542019591a4cf8b391f11dd41053cdae9fbc0d6ba534ecd288727b58bdf6b62R44

I would say that testing the default is more important than testing every possible outcome of `reduce_output` (since users are unlikely to change this value). In the future we should incorporate all possible values of `reduce_output` into the schema (probably as an Enum) and then into these unit tests.

You can (and probably should) put that change in a different PR if this is blocking for the release.

geoffreyangus (approved):

arnavgarg1:
@geoffreyangus Agreed, will add a test in a follow-up PR!

PR #2985, opened 2023-01-22
[Hyperopt] Modify _get_best_model_path to grab it from the Checkpoint object with ExperimentAnalysis

arnavgarg1:
This allows us to pipe credentials to _get_best_checkpoint_dir and ensure that we can load/download artifacts from remote URIs where credentials are needed (and don't exist in the environment variables / explicitly passed in through the Ludwig config)

### DIFF
--- a/ludwig/hyperopt/execution.py
+++ b/ludwig/hyperopt/execution.py
@@ -47,6 +47,7 @@ from ludwig.utils import metric_utils
 from ludwig.utils import fs_utils
 from ludwig.utils.data_utils import hash_dict, NumpyEncoder, use_credentials
 from ludwig.utils.defaults import default_random_seed
+from ludwig.utils.error_handling_utils import default_retry
 from ludwig.utils.fs_utils import has_remote_protocol, safe_move_file
 from ludwig.utils.misc_utils import get_from_registry
 
@@ -84,6 +85,14 @@ def identity(x):
 
 def _get_relative_checkpoints_dir_parts(path: Path):
     return path.parts[-2:]
+
+
+@default_retry()
+def _download_local_tmpdir(ckpt_path: str, tmpdir: str, creds: Dict[str, Any]) -> str:
+    local_ckpt_path = os.path.join(tmpdir, uuid.uuid4().hex)
+    with use_credentials(creds):
+        fs_utils.download(ckpt_path, local_ckpt_path)
+    return local_ckpt_path
 
 
 # Follwing disabled at the moment, expect to be re-enabled pending https://github.com/ludwig-ai/ludwig/issues/2039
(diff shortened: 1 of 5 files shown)


ludwig/hyperopt/execution.py
```diff
@@ -358,27 +358,21 @@ def _remove_partial_checkpoints(trial_path: str):
...
 
-        if checkpoint is not None:
-            with checkpoint.as_directory() as path:
-                yield path
+        ckpt_type, ckpt_path = checkpoint.get_internal_representation()
+        if ckpt_type == "uri":
+            # Read remote URIs using Ludwig's internal remote file loading APIs, as
+            # Ray's do not handle custom credentials at the moment.
+            with use_credentials(creds):
+                yield ckpt_path
```
tgaddair, line 372: `ckpt_path` here is a remote URL, so we'll want to convert to a directory before sending it back to the caller:

```
with use_credentials(creds):
    with checkpoint.as_directory() as path:
        yield path
```

arnavgarg1: Fixed

hungcs (approved):

ludwig/hyperopt/execution.py
```diff
@@ -357,28 +367,27 @@ def _remove_partial_checkpoints(trial_path: str):
...
-            logger.warning(
-                f"Cannot get best model path for {trial_path} due to exception below:\n{traceback.format_exc()}"
-            )
-            yield None
-            return
+    def _get_best_model_path(trial_path: str, analysis: ExperimentAnalysis, creds: Dict[str, Any]) -> str:
+        # `trial_dir` returned by RayTune may have a leading slash, but get_best_checkpoint
+        # requires a path without a leading slash since it does a direct key lookup with analysis.trial_dataframes.
+        if isinstance(trial_path, str) and trial_path[-1] == "/":
+            trial_path = trial_path[:-1]
```
jeffkinnison, line 376: This could also be done with `trial_path.rstrip('/')`, which can also let you drop the second clause of the condition.

tests/ludwig/hyperopt/test_hyperopt.py
```diff
@@ -82,10 +82,12 @@ def test_grid_search_more_than_one_sample():
     with pytest.warns(RuntimeWarning):
```
jeffkinnison, line 82: nit: Could you add a short docstring explaining the test?

jeffkinnison (approved):
Great work! Just a couple of minor suggestions.

ludwig/hyperopt/execution.py
arnavgarg1: Done!

tests/ludwig/hyperopt/test_hyperopt.py
arnavgarg1: Done!

PR #2986, opened 2023-01-23
GBM: disable goss boosting type

jppgks:
Leads to `lightgbm.basic.LightGBMError: Number of class for initial score error`. Disabling until we have capacity to support non-default boosting types


arnavgarg1:
Can we create a GitHub issue and add it as a comment with a TODO? Just so that we remember we want to add it back in

arnavgarg1 (approved):

jeffkinnison (approved):

PR #2991, opened 2023-01-23
Add combinatorial tests to the CI

abidwael:
This PR adds combinatorial Ludwig config testing to the CI to automatically generate valid configs and train models using them to test model training success.

TODO: add a README to explain how it works and how to add more combinatorial tests.


abidwael:
The remaining piece is to add the marker `combinatorial` to run along with `not distributed` in the CI.

abidwael:
Looking to get reviews and address them as we fix the failures we're seeing in the GHA jobs.

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,417 @@
...
+from ludwig.constants import SEQUENCE, TEXT, TIMESERIES
+from ludwig.schema.metadata.parameter_metadata import ExpectedImpact
+from ludwig.schema.model_types.base import ModelConfig
+from ludwig.types import ModelConfigDict
+from ludwig.utils.misc_utils import merge_dict
+
+ParameterBaseTypes = Union[str, float, int, bool, None]
+
+
+def explore_properties(
```
tgaddair, line 15: @abidwael how difficult do you think it would be to move this from using the jsonschema version of the config schema to using the schema objects like we do in the ludwig-docs: https://github.com/ludwig-ai/ludwig-docs/blob/master/main.py

I say this because I worry that over time this could become very difficult to maintain in its current form, as it's very removed from the representation of the schema that's tied to Ludwig code. What do you think?

abidwael: I chatted with @ksbrar and our conclusion was that 
1. there isn't a streamlined way to access the fields (the equivalent of `properties` dict) at every level of the config. 
```
cls = model_type_schema_registry["ecd"]
defaults = cls.get_class_schema()().fields["defaults"]
text = defaults.cls.get_class_schema()().fields["text"]

text.fields # fails: no attribute fields
text.cls # fails: no attribute cls
```
2. Even if that wasn't a problem, it doesn't look like there is a way to retrieve fields like minimum, maximum, string options that help us infer the values we suggest for the combinatorial configs. Example of what's returned in `stringOptions`:
```
    return field(
        metadata={
            "marshmallow_field": fields.String(
                validate=validate.OneOf(options),
                allow_none=allow_none,
                load_default=default,
                dump_default=default,
                metadata={
                    "description": description,
                    "parameter_metadata": convert_metadata_to_json(parameter_metadata) if parameter_metadata else None,
                },
            )
        },
        default=default,
    )
```
As you can see, the only available options are defaults. We can add more fields to it, but I don't think we should as it will make the schema bigger than it already is and `ludwig.schema.utils.unload_jsonschema_from_marshmallow_class` already does that.

The benefit of `get_schema` is that it also unpacks the `Validator` class of every field which will contain the rules (min, max, options, etc.). These don't exist in the classes themselves and if we want to make this work with objects, we will have to access the validator attributes. Not all validators are native marshmallow validators, though. We define some of them. So if we wanted to implement them the object-oriented way, we'll also need to have logic that checks whether they're marshmallow-native or defined in Ludwig and then get the validator-specific fields (e.g. validator.choices for oneOf validator.)

Therefore, I think the current approach is the right one. I'm not 100% sure I understood your comment about maintainability correctly. If someone wanted to maintain this, they'd just have to understand what's in the JSON schema. If there's documentation work that I can do to improve this, I'm happy to add that.

justinxzhao (commented):
A handful of minor comments for now. Will review the meat in more detail in a bit.

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+        return explore_boolean()
+    elif property_type == "null":
+        return explore_null()
+    elif property_type == "array":
+        return explore_array(item)
+    else:
+        return []
+
+
+def explore_array(item: Dict[str, Any]) -> List[List[ParameterBaseTypes]]:
```
justinxzhao, line 183: suggestion: Put the `explore_<atomic_type>()` functions in a separate file, i.e. `sample_leaf_property.py` or `explore_leaf_property.py`

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+        config_options: dictionary mapping from ludwig config parameter to all values to be explored.
+            Here's an example of what it could look like:
+
+                trainer.batch_size: ["auto", 2, 43]
+                trainer.learning_rate: ["auto", 0.1, 0.00002, 0.32424]
+                ...
+    """
+    num_configs = 1
+    for key in config_options:
+        if isinstance(config_options[key], list):
```
justinxzhao, line 299: When is `config_options[key]` not a list?

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+
+    Args:
+        config_options: dictionary mapping from ludwig config parameter to all values to be explored.
+            Here's an example of what it could look like:
+
+                trainer.batch_size: ["auto", 2, 43]
+                trainer.learning_rate: ["auto", 0.1, 0.00002, 0.32424]
+                ...
+    """
+    num_configs = 1
```
justinxzhao, line 297: Maybe add a quick comment here.

```
# The number of configs to generate is the max length of the lists of samples over all parameters.
```

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+            num_configs = max(num_configs, len(config_options[key]))
+            config_options[key] = deque(config_options[key])
+
+    for _ in range(num_configs):
+        config = {}
+        for key in config_options:
+            if config_options[key] and not isinstance(config_options[key], str):
+                config[key] = config_options[key].popleft()
+            elif isinstance(config_options[key], str):
+                config[key] = config_options[key]
```
justinxzhao, line 309: Could you add comments with a quick example for both cases?

Also, is there an `else` case that we should be concerned with?

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+        yield config
+
+
+def create_nested_dict(flat_dict: Dict[str, Union[float, str]]) -> Dict[str, Any]:
```
justinxzhao, line 313: nit: Use the `ModelConfigDict` type hint from `from ludwig.types import ModelConfigDict` here and anywhere where you are using a Ludwig model config dictionary.

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+            config_options[key] = deque(config_options[key])
+
+    for _ in range(num_configs):
+        config = {}
+        for key in config_options:
+            if config_options[key] and not isinstance(config_options[key], str):
+                config[key] = config_options[key].popleft()
+            elif isinstance(config_options[key], str):
+                config[key] = config_options[key]
+        yield config
```
justinxzhao, line 310: nit/suggestion: It seems potentially cleaner to simply `yield create_nested_dict(config)` since AFAICT you're not using the intermediate config dict with absolute parameter path keys.

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+    ret = []
+    for config_options, _ in explored:
+        for default_config in generate_possible_configs(config_options=config_options):
+            default_config = create_nested_dict(default_config)
+            merged_config = merge_dict(copy.deepcopy(config), default_config)
+            try:
+                ModelConfig.from_dict(merged_config)
+                ret.append((merged_config, dataset))
+            except Exception:
+                pass
```
justinxzhao, line 361: It could be potentially useful to return a count of the number of Excepted configs.

As part of the combinatorial tests, we could check that the number of excepted configs is below some cap and if the cap were to be exceeded significantly, that would indicate that there's either been a bad schema change, or that the combinatorial sampling might warrant some improving.

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+) -> List[Tuple[ModelConfigDict, pd.DataFrame]]:
+    """Merge base config with explored sections.
+
+    Completes the entity_1 and entity_2 paramters of the comparator combiner.
+
+    Args:
+        explored: deque containing all the config options.
+        config: base Ludwig config to merge the explored configs with.
+    """
+    dataset = build_synthetic_dataset_df(10, config)
```
justinxzhao, line 376: nit/suggestion: Make `10` a constant or an input parameter.

tests/training_success/test_training_success.py
```diff
@@ -0,0 +1,245 @@
...
+        only_include: top-level parameter of the defaults sections that should be included.
+    """
+    assert isinstance(only_include, str)
+    assert only_include in {"preprocessing", "encoder", "decoder", "loss"}
+
+    schema = get_schema()
+    properties = schema["properties"]["defaults"]["properties"][feature_type]["properties"]
+    raw_entry = deque([(dict(), False)])
+    explored = explore_properties(
+        properties, parent_key="defaults." + feature_type, dq=raw_entry, only_include=[only_include]
```
justinxzhao, line 36: nit: `f"defaults.{feature_type}"`

justinxzhao (commented):
@abidwael, my comments are mainly around code structure, variable naming, nits, and questions. Solid work overall.

Higher-level comments:
- The code in `explore_schema` is pretty juicy. It would be great to add unit tests for this code.
- In terms of directory structure, I might suggest putting most of the code in a new directory `ludwig/config_sampling` and putting tests in `tests/ludwig/config_sampling`.
- Tests that actually do config sampling and run training for 1 step can live in `tests/robustness` (or `training_success` as you currently have it, though `robustness` seems more clear to me)
- Move `explore_<atomic type>()` functions to their own file. `explore_schema.py` is really big already.
- Throughout your code, you use `key` and `item`, which aren't particularly descriptive. In one of my comments, I suggest using `jsonschema_property` instead of `item` and `parameter_name/path` instead of `key`, which I would take back if `key` and `item` are formal concepts in jsonschema.
- Prefer to use f"{var}.{var}" strings instead of `var + "." + var`
- nit: Consider defining a dataclass for the `dq` instead of passing around tuple objects.

.github/workflows/pytest.yml
```diff
@@ -188,6 +188,43 @@ jobs:
           name: Unit Test Results (Python ${{ matrix.python-version }} ${{ matrix.test-markers }})
```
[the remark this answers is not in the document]
abidwael: The combinatorial tests will run in their own separate runner and the other jobs can stay intact. I opted for this option as to not add to the current time it takes to run these tests and speed up development.

.github/workflows/pytest.yml
```diff
@@ -188,6 +188,43 @@ jobs:
...
           path: pytest.xml
 
+  robustness-tests:
+    name: Robustness Tests
```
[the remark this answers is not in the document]
abidwael: Will use `combinatorial`

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+
+            if "properties" in item and "allOf" in item:
+                for child_item in item["allOf"]:
+                    expanded_config_options_dq = explore_from_all_of(
+                        config_options=copy.deepcopy(config_options), item=child_item, key_so_far=key_so_far
+                    )
+                    # add returned child config options to the deque to be processed.
+                    dq.extend(expanded_config_options_dq)
+
+            elif "properties" in item and "allOf" not in item:
```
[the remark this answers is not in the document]
abidwael: Added a comment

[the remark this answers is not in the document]
abidwael: This is the case where we don't have a list of properties, just a `properties` dictionary nested inside another.

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+                merged_config_options_dq = merge_dq(config_options, child_config_options_dq)
+                # add returned config options to the deque to be processed.
+                dq.extend(merged_config_options_dq)
+
+            else:
+                # this is the base case.
+                if "oneOf" in item:
+                    temp = []
+                    for elem in item["oneOf"]:
+                        temp += get_potential_values(elem)
```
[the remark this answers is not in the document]
abidwael: changing to list comprehension causes inconsistencies

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+                # a new dequeue to be passed to explore parameters from
+                raw_entry = deque([(copy.deepcopy(config_options), False)])
+                child_config_options_dq = explore_properties(child_properties, key_so_far, raw_entry)
+                merged_config_options_dq = merge_dq(config_options, child_config_options_dq)
+                # add returned config options to the deque to be processed.
+                dq.extend(merged_config_options_dq)
+
+            else:
+                # this is the base case.
+                if "oneOf" in item:
```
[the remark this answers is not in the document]
abidwael: not sure I understand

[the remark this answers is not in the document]
abidwael: good point

tests/training_success/explore_schema.py
```diff
@@ -0,0 +1,422 @@
...
+                    for elem in item["oneOf"]:
+                        temp += get_potential_values(elem)
+                    config_options[key_so_far] = temp
+                else:
+                    config_options[key_so_far] = get_potential_values(item)
+
+                # for config parameters that are internal or for which we can't infer suggested values to explore
+                # e.g. parameters of type array, object, string (in some cases), etc.
+                if len(config_options[key_so_far]) == 0:
+                    del config_options[key_so_far]
```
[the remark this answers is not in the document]
abidwael: good point

tests/training_success/explore_schema.py
abidwael: Added. No else to be concerned with, just wanted to make the conditions clear.

ludwig/config_sampling/explore_schema.py
```diff
@@ -0,0 +1,304 @@
...
+
+                # add config_options back to queue. fully_explored = False because we still didn't finish
+                # exploring all the keys in the properties dictionary.
+                dq.appendleft(ConfigOption(config_options, False))
+
+        # at this point, we finished exploring all keys of the properties dictionary. Add all config options
+        # to the processed queue.
+        while dq:
+            config_options, _ = dq.popleft()
+            processed_dq.append(ConfigOption(config_options, True))
```
justinxzhao, line 98: Can this be simplified to:

```
processed_dq.append(dq.popleft())
```

ludwig/config_sampling/explore_schema.py
```diff
@@ -0,0 +1,304 @@
...
+        config: base Ludwig config to merge the explored configs with.
+    """
+    dataset = build_synthetic_dataset_df(NUM_SYNTHETIC_EXAMPLES, config)
+    ret = []
+    for item in explored:
+        for default_config in generate_possible_configs(config_options=item[0]):
+            merged_config = merge_dict(copy.deepcopy(config), default_config)
+            for i in range(len(merged_config["input_features"])):
+                if merged_config["input_features"][i]["type"] in {SEQUENCE, TEXT, TIMESERIES}:
+                    merged_config["input_features"][0]["encoder"] = {"type": "embed", "reduce_output": None}
```
justinxzhao, line 298: This should be set properly now in the ModelConfig's `__post_init__`. Do we still need this combiner-specific generator?

tests/ludwig/config_sampling/test_config_sampling.py
```diff
@@ -0,0 +1,46 @@
...
+    total_count += full_config_generator(combiner_config_generator, "sequence_concat", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "sequence", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "comparator", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "concat", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "project_aggregate", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "tabnet", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "tabtransformer", static_schema)
+    total_count += full_config_generator(combiner_config_generator, "transformer", static_schema)
+
+    assert total_count == 131
```
justinxzhao, line 46: Instead of checking equality, which means this test would need to be updated every single time someone adds a new parameter of a certain impact, consider asserting:

```python
# In place to check for sudden changes in the number of combinatorially generated configs. Update ranges accordingly if new parameters are added.
assert 130 < total_count < 200
```

tests/training_success/configs.py
```diff
@@ -0,0 +1,154 @@
...
+BINARY_OUTPUT = """
+input_features:
+  - name: binary_1
+    type: binary
+output_features:
+  - name: binary_output_1
+    type: binary
+"""
+
+FEATURE_TYPE_TO_CONFIG_FOR_ENCODER_PREPROCESSING = {
```
justinxzhao, line 119: nit: Add docstrings for these global maps.

tests/training_success/test_training_success.py
```diff
@@ -0,0 +1,263 @@
...
+        config = FEATURE_TYPE_TO_CONFIG_FOR_ENCODER_PREPROCESSING[feature_type]
+        config = yaml.safe_load(config)
+    else:  # decoder and loss
+        config = FEATURE_TYPE_TO_CONFIG_FOR_DECODER_LOSS[feature_type]
+        config = yaml.safe_load(config)
+
+    main_config_keys = list(config.keys())
+    for key in main_config_keys:
+        if key not in ["input_features", "output_features"]:
+            del config[key]
```
justinxzhao, line 52: Why is this necessary if all of the configs in `tests/training_success/configs.py` only have `input_features` and `output_features`?

justinxzhao (commented):
Mostly just a few more nits! Thanks.

ludwig/config_sampling/explore_schema.py
abidwael: I believe we do. This reduces the number of configs that don't pass validation.

justinxzhao (approved):
Thanks!

PR #2996, opened 2023-01-24
LR Scheduler Adjustments

connor-mccorm:


ludwig/schema/lr_scheduler.py
```diff
@@ -132,13 +132,7 @@ def _deserialize(self, value, attr, data, **kwargs):
         @staticmethod
         def _jsonschema_type_mapping():
             return {
-                "oneOf": [
```
ksbrar, line 135: So the scheduler will always be enabled?

connor-mccorm: This is a good point. I don't think we actually want this to always be enabled quite yet - I think if we can figure out a way to set these params automatically then we would. So I actually may revert this part for now.

connor-mccorm: Ok for the time being, I switched decay to have a default of `None` and be a required param. This is in fact supported so we should be good here. Down the line, I think it's a good idea to think about how to automate the setting of lr scheduler params though.

tgaddair (approved):

ludwig/schema/lr_scheduler.py
ksbrar: > I switched decay to have a default of None _and be a required param_
What do you mean?

connor-mccorm: Basically before decay was an optional parameter so we didn't need a default for it. However, decay can take a value of none which just doesn't use any LR decay. So I added that default and removed the `Optional` typing indicator so that `decay` is now a required param from a validation pov but the user doesn't have to specify anything because if they don't it will just default to `None`.

PR #2997, opened 2023-01-24
[Hyperopt] Set default num_samples based on parameter space

arnavgarg1:
This PR sets the default `num_samples` to None, and then infers a value for it (either 1 or 10) based on the parameters search space passed into hyperopt. 

In particular, when there are parameters with grid search, we want to prevent duplication of trials because grid search creates all possible combinations of trials using the discrete values in the space, so we set the num_samples to 1 to prevent this duplication. 

If the parameter space doesn't contain any grid search parameters, there is no chance of duplication and we can safely set the default value to 10 (like we were before). 

This will continue to respect user's choices if they are set in the config.

Note: We don't currently have a `hyperopt.yaml` in `ludwig/schema/metadata` so there's nothing else to update. I will follow this PR up with an update to the docs.

### DIFF
--- a/ludwig/hyperopt/utils.py
+++ b/ludwig/hyperopt/utils.py
@@ -175,16 +175,28 @@ def substitute_parameters(
 
 
 @DeveloperAPI
+def contains_grid_search_parameters(config: ModelConfigDict) -> bool:
+    """Returns True if any hyperopt parameter in the config is using the grid_search space."""
+    for _, param_info in config[PARAMETERS].items():
+        if param_info.get(SPACE, None) == GRID_SEARCH:
+            return True
+    return False
+
+
+@DeveloperAPI
 def get_num_duplicate_trials(hyperopt_config: HyperoptConfigDict) -> int:
+    """Returns the number of duplicate trials that will be created.
+
+    Duplicate trials are only created when there are grid type parameters and num_samples > 1.
+    """
     num_samples = hyperopt_config[EXECUTOR].get(NUM_SAMPLES, 1)
     if num_samples == 1:
         return 0
 
     total_grid_search_trials = 1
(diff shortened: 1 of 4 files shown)


ludwig/schema/model_config.py
```diff
@@ -512,7 +518,20 @@ def _set_hyperopt_defaults(self):
         # This fills in missing splits, executor config, search_alg, etc.
         self.hyperopt = HyperoptConfig.from_dict(self.hyperopt).to_dict()
 
-        scheduler = self.hyperopt.get("executor", {}).get("scheduler")
+        # Set default num_samples based on search space if not set by user
+        if self.hyperopt.get(EXECUTOR).get(NUM_SAMPLES) is None:
```
tgaddair, line 522: Is `EXECUTOR` guaranteed to be populated at this point? If so, we can do:

```
if self.hyperopt[EXECUTOR].get(NUM_SAMPLES) is None:
```

tgaddair (approved):

ludwig/schema/model_config.py
arnavgarg1: Yup, it is guaranteed by this point. Will update

ludwig/hyperopt/utils.py
```diff
@@ -174,17 +174,29 @@ def substitute_parameters(
     return config
 
 
+@DeveloperAPI
+def contains_grid_search_parameters(config: ModelConfigDict) -> bool:
```
tgaddair, line 178: Is this `ModelConfigDict` or `HyperoptConfigDict`? I think it should be the latter.

arnavgarg1: Ah copy pasta + GitHub copilot, yeah should be ModelConfigDict.

tgaddair (approved):

PR #2999, opened 2023-01-25
Transformer Encoder - Representation Parameter Fix

connor-mccorm:
This PR addresses this training failure:
```
Exception: Embedding representation None not supported. (type: RayTaskError, retryable: true)
```


tests/conftest.py
```diff
@@ -27,12 +27,15 @@
...
     INPUT_FEATURES,
+    MODEL_ECD,
+    MODEL_GBM,
     NAME,
     OUTPUT_FEATURES,
     TRAINER,
     TYPE,
 )
 from ludwig.hyperopt.run import hyperopt
+from ludwig.schema import get_schema
```
justinxzhao, line 38: Can you import `from ludwig.config_validation.validations import get_schema`? Want to try to remove `ludwig.schema.__init__.py`

tests/ludwig/schema/test_schema_fields.py
```diff
@@ -0,0 +1,27 @@
+from ludwig.constants import ALLOF, ENCODER, ENUM, INPUT_FEATURES, ITEMS, PROPERTIES, THEN
```
justinxzhao, line 1: If we keep these tests, add them tests to `tests/ludwig/config_validation/`

tests/ludwig/schema/test_schema_fields.py
```diff
@@ -0,0 +1,27 @@
...
+
+def parse_schema(input_dict: dict, key_list: list):
+    """Parse a nested dictionary by a list of keys."""
+    parsed_dict = input_dict
+    for key in key_list:
+        parsed_dict = parsed_dict[key]
+    return parsed_dict
+
+
+def test_transformer_encoder_representations(ecd_schema):
```
justinxzhao, line 12: I'm not convinced that we absolutely need this test, since it seems like it's testing implementation details, and we already have tests for `StringOptions`.

Guidance on not testing implementation details:
https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html

connor-mccorm: Ya I feel like it's a little pedantic, however Travis asked specifically for this test [here](https://docs.google.com/spreadsheets/d/1By3vMX17dVwpvs0lyiLim5su4gp7pHlA3L7542WtbaI/edit?disco=AAAApOfcHF4)
```
Travis Addair
12:36 PM Yesterday
And add unit test (enumerate all possible values, check does not allow None).
```
So I interpreted that as I should build this test. Happy to remove though if we feel it's unnecessary.

tests/ludwig/schema/test_schema_fields.py
connor-mccorm: Done!

tests/conftest.py
connor-mccorm: Done!

tests/ludwig/schema/test_schema_fields.py
justinxzhao: It looks like there's already a test for `allow_none=False` for `StringOptions`.

https://github.com/ludwig-ai/ludwig/blob/master/tests/ludwig/marshmallow/test_fields_misc.py#L20-L34

I'd say we can do without this test.

connor-mccorm: Sounds good, will remove!

justinxzhao (approved):

PR #3003, opened 2023-01-25
Limit batch size tuning to ≤20% of dataset size

geoffreyangus:


arnavgarg1 (approved):
Worth adding a test?

ludwig/trainers/trainer.py
```diff
@@ -393,8 +393,8 @@ def tune_batch_size(
         logger.info("Tuning batch size...")
 
         def _is_valid_batch_size(batch_size):
-            # make sure that batch size is valid (e.g. less than size of ds)
-            is_smaller_than_training_set = batch_size < len(training_set)
+            # make sure that batch size is valid (e.g. less than 20% of ds size and max_batch_Size)
+            is_smaller_than_training_set = batch_size <= 0.2 * len(training_set)
```
justinxzhao, line 397: nit: Make this a constant, i.e. `MIN_VALID_BATCH_SIZE_AS_PERCENTAGE_OF_DATA`

justinxzhao (approved):

ludwig/trainers/trainer.py
geoffreyangus: Yup was actually just making that change 😅 also realized that some of the unit tests have hard coded expectations for batch size, so tweaking those as well.

PR #3009, opened 2023-01-26
Ray 2.3 Compatibility

arnavgarg1:
Notable changes:

1. Streaming windows are no longer specified as number of bytes, but rather % of object store memory (the value ranges between 0 and 1). This defaults to 20% by default based on best practice.
2. The introduction of a new DatasetIterator object. At the moment, we're using some of the internals of this new object from Ray datasets, but in a follow-up PR, this will be cleaned up to use it in a way that is more in the spirit of the Ray Datasets API.
3. Auto-windowing tests were temporarily disabled. These require a rewrite, and will also be done in a follow-up PR.

Co-authored-by: @geoffreyangus <geoffrey@predibase.com>


.github/workflows/pytest.yml
```diff
@@ -28,15 +28,15 @@ jobs:
           - python-version: "3.8"
             pytorch-version: 1.13.0
             torchscript-version: 1.10.2
-            ray-version: 2.0.0
+            ray-version: 2.1.0
           - python-version: "3.9"
             pytorch-version: 1.13.0
             torchscript-version: 1.10.2
-            ray-version: 2.1.0
+            ray-version: 2.2.0
```
tgaddair, line 35: Can we do: 2.2, 2.3, and nightly?

ludwig/backend/_ray210_compat.py
```diff
@@ -8,15 +8,15 @@
...
-from ray.tune.execution.trial_runner import _ResumeConfig
 from ray.tune.impl.tuner_internal import TunerInternal
 from ray.tune.trainable import Trainable
 from ray.tune.tune_config import TuneConfig
 from ray.tune.tuner import _SELF, _TUNER_INTERNAL, Tuner
-from ray.tune.utils.node import _force_on_current_node
 
 if TYPE_CHECKING:
     from ray.train.trainer import BaseTrainer
+    from ray.tune.execution.trial_runner import _ResumeConfig
```
tgaddair, line 18: This doesn't make sense to me. The `TYPE_CHECKING` guard is there for things that are only used in type hints, as `TYPE_CHECKING=False` is set at runtime (so these things will never get imported, in theory).

tgaddair (commented):
Only one comment, otherwise looks great!

.github/workflows/pytest.yml
```diff
@@ -28,15 +28,15 @@ jobs:
           - python-version: "3.8"
             pytorch-version: 1.13.0
             torchscript-version: 1.10.2
-            ray-version: 2.0.0
+            ray-version: 2.1.0
```
tgaddair, line 31: We should probably increase the minimal Ray version in `requirements_distributed.txt` and `requirements_hyperopt.txt` to 2.1 to be safe. Altneratively, I would keep the min tested version as 2.0 here and not bother testing 2.1 (since we test 2.2 and 2.3).

ludwig/backend/_ray210_compat.py
arnavgarg1: Agreed, just fixed!

.github/workflows/pytest.yml
arnavgarg1: @tgaddair Going to do 2.0, 2.2 and 2.3

tgaddair (approved):
LGTM if tests pass!

justinxzhao (approved):
Thanks for your diligence on this!

PR #3010, opened 2023-01-26
Fix checkpoint loading for HuggingFace encoders

geoffreyangus:


tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -1,55 +1,133 @@
...
         configs.ELECTRAConfig,
         configs.FlauBERTConfig,
         configs.T5Config,
         configs.XLNetConfig,
-        configs.DistilBERTConfig,
     ],
 )
-def test_hf_pretrained_default_exists(encoder_config_cls: configs.SequenceEncoderConfig):
-    """Test that the default pretrained model exists on the HuggingFace Hub.
+def test_hf_ludwig_model_e2e(tmpdir, csv_filename, mock_load_encoder_from_hf_hub, encoder_config_cls):
```
geoffreyangus, line 89: With this test, do we still need the these encoder class-level tests: https://github.com/ludwig-ai/ludwig/blob/master/tests/ludwig/encoders/test_text_encoders.py#L82-L308

My understanding is that combiner class instantiation relies on input feature output shape, and that a full forward pass would require the input feature output shape to match with the combiner's dimensions. Since this is an end-to-end test that tests model training, it seems like we would achieve equal coverage with the end-to-end test alone.

If we can remove the other tests, it would greatly reduce the CI run time. Happy to leave it if people think there's a reason to though.

ludwig/api.py
```diff
@@ -585,6 +585,7 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                 update_config_with_metadata(self.config_obj, training_set_metadata)
                 logger.info("Warnings and other logs:")
                 self.model = LudwigModel.create_model(self.config_obj, random_seed=random_seed)
+                update_config_with_model(self.config_obj, self.model)
```
w4nderlust, line 588: before this line i would add a comment explaining why we need to update the config after the model has been created

ludwig/api.py
```diff
@@ -776,6 +777,7 @@ def train_online(
         if not self.model:
             update_config_with_metadata(self.config_obj, training_set_metadata)
             self.model = LudwigModel.create_model(self.config_obj, random_seed=random_seed)
+            update_config_with_model(self.config_obj, self.model)
```
w4nderlust, line 780: a comment here too

ludwig/encoders/text_encoders.py
```diff
@@ -59,6 +60,48 @@ def _cls_pooled_error_message(encoder: str):
...
+        pretrained models will have an embeddings table that is a different size than the vocab size. Examples:
+
+        CamemBERT:  https://github.com/huggingface/tokenizers/issues/900#issue-1122256698
+        T5:         https://github.com/huggingface/transformers/issues/4875#issue-635471552
+
+        Args:
+            transformer: The transformer model.
+            vocab_size: The vocab size of the dataset.
+        """
+        if vocab_size != transformer.config.vocab_size:
```
w4nderlust, line 102: shouldn't we be doing the opposite? use wathever vocab the transformer is telling us to use?

geoffreyangus: Yup, this method is only called if we are instantiating the model from scratch (i.e. `use_pretrained==False`). I've included this as part of the docstring of this method [here](https://github.com/ludwig-ai/ludwig/pull/3010/files#diff-f34cc840e21b57eebb191a5198da3fce8ce7dc5278fa4ea404d8d960a23f303dR88-R90), and you can see an example of this in one of the text encoders [here](https://github.com/ludwig-ai/ludwig/pull/3010/files#diff-f34cc840e21b57eebb191a5198da3fce8ce7dc5278fa4ea404d8d960a23f303dR415-R421). 

If `use_pretrained==True`, then this function is not called and instead we directly use the token embeddings table determined by the transformer.

tests/ludwig/encoders/test_text_encoders.py
justinxzhao: Let's remove the other tests since the text encoders are covered by this test, and reducing the CI runtime would be a big win!

ludwig/features/feature_registries.py
```diff
@@ -46,6 +46,11 @@
 from ludwig.features.text_feature import TextFeatureMixin, TextInputFeature, TextOutputFeature
 from ludwig.features.timeseries_feature import TimeseriesFeatureMixin, TimeseriesInputFeature
 from ludwig.features.vector_feature import VectorFeatureMixin, VectorInputFeature, VectorOutputFeature
+
+if TYPE_CHECKING:
```
justinxzhao, line 50: Interesting. Didn't realize this was an option for bypassing what I presume are flake8 errors?

justinxzhao (approved):

ludwig/features/feature_registries.py
geoffreyangus: Yup, it prevents circular import errors, which can happen more frequently when typing with custom modules

tests/ludwig/encoders/test_text_encoders.py
geoffreyangus: Sounds good! I also found these tests: https://github.com/ludwig-ai/ludwig/blob/master/tests/integration_tests/test_experiment.py#L88-L96

I'm going to update these new tests to incorporate the functionality in those tests and delete them as well (i.e. using `LudwigModel.experiment` instead of `LudwigModel.train`). I believe the tests are semantically equivalent.

jeffkinnison (approved):
LGTM! Particularly liked breaking encoder and tokenizer load into different functions, very good call.

PR #3011, opened 2023-01-26
Refactored norm layer and added additional norm at the start of the FCStack

tgaddair:


w4nderlust:
Looks good to me, the only minidoubt is if FC stack works with num layers 0 (in that case before we were just skipping it by making it None). If that's the case, than this looks good to me ;)

tgaddair:
@w4nderlust I had thought that as well, but it turns out that because we default `num_fc_layers` to `0`, we end up always creating the FCStack, so the code as written was misleading since in practice FCStack always exists, even if there are no layers (unless the user explicitly set `num_fc_layers=None`).

PR #3012, opened 2023-01-26
Fix assignment that undoes tensor move to CPU

jeffkinnison:
A quick update that stops an assignment from overwriting a tensor move from GPU to CPU.


arnavgarg1 (approved):

arnavgarg1:
🚢

tgaddair (approved):

PR #3015, opened 2023-01-27
Handle CUDA OOMs in explanations with retry and batch size halving

### LINKED ISSUE #3004: Fix explanations with cache embeddings
Remaining issue from #2867. Batch size selected by `batch_size=auto` is likely too large for explanations, which require using the pretrained model during the forward / backward pass.

tgaddair:
Fixes #3004.

### DIFF
--- a/ludwig/explain/captum.py
+++ b/ludwig/explain/captum.py
@@ -43,7 +43,7 @@ class ExplanationRunConfig:
 
 def retry_on_cuda_oom(run_config: ExplanationRunConfig):
     def retry_on_cuda_oom_fn(fn):
-        def wrapper(*args, **kwargs):
+        def retry_on_cuda_oom_wrapper(*args, **kwargs):
             while run_config.batch_size >= 2:
                 try:
                     return fn(*args, **kwargs)
@@ -59,7 +59,7 @@ def retry_on_cuda_oom(run_config: ExplanationRunConfig):
 
             raise RuntimeError("Minimum support batch size 2 still resulted in CUDA OOM during explanation")
 
-        return wrapper
+        return retry_on_cuda_oom_wrapper
 
     return retry_on_cuda_oom_fn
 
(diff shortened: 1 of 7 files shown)


ludwig/explain/captum.py
```diff
@@ -215,6 +187,80 @@ def explain(self) -> Tuple[List[Explanation], List[float]]:
...
+
+    config = model.config_obj.to_dict()
+    training_set_metadata = copy.deepcopy(model.training_set_metadata)
+    for feature in config[INPUT_FEATURES]:
+        preprocessing = training_set_metadata[feature[NAME]][PREPROCESSING]
+        if preprocessing.get("cache_encoder_embeddings"):
+            preprocessing["cache_encoder_embeddings"] = False
+
+    # Convert raw input data into preprocessed tensor data
+    dataset, _ = preprocess_for_prediction(
```
jppgks, line 216: Do we want to use `run_config.batch_size` also for preprocessing? Or will preprocessing not require PyTorch since we set `cache_encoder_embeddings` to False?

ludwig/explain/captum.py
```diff
@@ -215,6 +187,80 @@ def explain(self) -> Tuple[List[Explanation], List[float]]:
...
+    # Dict of lists to list of dicts
+    input_batches = [dict(zip(inputs, t)) for t in zip(*inputs.values())]
+
+    # List of dicts to dict of lists
+    preproc_inputs = {k: torch.cat([d[k] for d in input_batches]) for k in input_batches[0]}
+
+    data_to_predict = [v for _, v in preproc_inputs.items()]
+    tensors = []
+    for t in data_to_predict:
+        if t.dtype == torch.int8 or t.dtype == torch.int16 or t.dtype == torch.int32 or t.dtype == torch.int64:
```
jeffkinnison, line 250: Could this be simplified to `if not torch.is_floating_point(t.dtype)`? The body of the block could then be reorganized to handle the bool case.

jeffkinnison (approved):
Just a minor comment, it shouldn't stop us from merging.

ludwig/explain/captum.py
tgaddair: Correct, we don't need the separate batch size here since we don't do the embedding during preprocessing.

ludwig/explain/captum.py
tgaddair: Good point, I can leave a TODO for now.

PR #3022, opened 2023-01-30
[schema] Mark skip internal only

jppgks:


ksbrar (approved):

ludwig/schema/metadata/configs/encoders.yaml
```diff
@@ -1,3 +1,10 @@
+BaseEncoder:
+    skip:
+        internal_only: true
+        other_information: Internal Only
+        related_parameters:
```
connor-mccorm, line 5: I would actually just leave `related_parameters` out since by default it will fill it with `None` which will be easier to deal with in the future!

connor-mccorm (approved):

PR #3024, opened 2023-01-30
Add a registry of additional config checks to check inter-parameter incompatibilities.

justinxzhao:
## Overview

There are many interdependent parameters in a ludwig configuration, and generating a marshmallow schema for incompatible interdependent parameters can be much more awkward/difficult than a straightforward traversal through a python dictionary.

There are also auxiliary checks scattered throughout the codebase that raise errors upon discovering incompatible parameter values.

This PR attempts to consolidate the lion's share of them in a single registry, and each check gets called as part of `validate_config`.

## Validation API

After this PR, it might be a good idea to fold all validation into the `from_dict()` method, so that there's one API for getting a comprehensive and validated config. 

```python
comprehensive_and_validated_config = ModelConfig.from_dict(user_config)
```

## List of consolidated and new checks

These now exist as part of the `config_check_registry`:
- `check_feature_names_unique`: Checks that all feature names are unique.
- `check_tied_features_are_valid`: Checks that all tied features are valid.
- `check_training_runway`: Checks that checkpoints_per_epoch and steps_per_checkpoint aren't simultaneously defined.
- `check_gbm_horovod_incompatibility`: Checks that GBM model type isn't being used with the horovod backend.
- `check_gbm_single_output_feature`: GBM models only support a single output feature.
- `check_gbm_feature_types`: Checks that all input features for GBMs are of supported types.
- `check_ray_backend_in_memory_preprocessing`: Checks that in memory preprocessing is used with Ray backend.
- `check_sequence_concat_combiner_requirements`: Checks that sequence concat combiner has at least one input feature that's sequential.
- `check_comparator_combiner_requirements`: Checks ComparatorCombiner requirements. All of the feature names for entity_1 and entity_2 are valid features.
- `check_class_balance_preprocessing`: Class balancing is only available for datasets with a single output feature.
- `check_sampling_exclusivity`: Oversample minority and undersample majority are mutually exclusive.
- `check_validation_metric_exists`: Checks that validation fields in config.trainer are valid.
- `check_splitter`: Checks the validity of the splitter configuration.


ludwig/schema/model_config.py
```diff
@@ -264,13 +257,30 @@ def _upgrade_config(config_dict: ModelConfigDict) -> ModelConfigDict:
...
 
         Args:
             config_dict: Config Dictionary
         """
-        validate_config(config_dict)
+        # Schema validation.
+        check_schema(comprehensive_config)
+
+        # Additional checks.
+        auxiliary_checks.check_feature_names_unique(comprehensive_config)
```
ksbrar, line 270: Nit: for brevity, you could create a registry of auxiliary checks

justinxzhao: Done.

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,260 @@
...
+    TYPE,
+    VECTOR,
+)
+from ludwig.error import ConfigValidationError
+from ludwig.types import ModelConfigDict
+from ludwig.utils.metric_utils import get_feature_to_metric_names_map
+from ludwig.utils.registry import Registry
+
+# Set of all sequence feature types.
+SEQUENCE_OUTPUT_FEATURE_TYPES = {SEQUENCE, TEXT, SET, VECTOR}
```
ksbrar, line 40: worth moving to `ludwig.constants`?

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,260 @@
...
+    @staticmethod
+    def check(config: ModelConfigDict) -> None:
+        from ludwig.data.split import get_splitter
+
+        splitter = get_splitter(**config[PREPROCESSING][SPLIT])
+        splitter.validate(config)
+
+
+@register_config_check("Checks the schema.")
+class CheckSchema(ConfigCheck):
```
ksbrar, line 257: Hmm, jw - does this cause the schema to be validated twice?

justinxzhao: Instantiating a model config will run schema validation twice -- once at the beginning of the constructor to check required fields (the rest of the config class code was written assuming that required fields are all present), and once at the end when we iterate through the registry.

ludwig/config_validation/checks.py
justinxzhao: I don't see any immediate use for this being used outside of this file. Happy to do that refactoring in the future when the need arises.

ludwig/config_validation/checks.py
ksbrar: Why do it the second time?

tgaddair:
It might make sense to hold off on some of these changes until #2906 lands, which makes some major changes to config object.

ludwig/config_validation/checks.py
justinxzhao: Removed the second schema check.

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,327 @@
...
+    """Checks instances of comprehensive (all parameters and defaults filled in) schema-validated config."""
+
+    @staticmethod
+    @abstractmethod
+    def check(config: ModelConfigDict) -> None:
+        """Checks config for validity."""
+        raise NotImplementedError
+
+
+def check_basic_required_parameters(config: ModelConfigDict) -> None:
```
ksbrar, line 86: I still wonder if we should consider exposing the auxiliary checks through a separate API so they can be used in the FE as well. Flagging here to not lose track of that thread, but this is great.

ksbrar (approved):
clean 👌🏽

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,327 @@
...
+            raise ConfigValidationError("Trainer.optimizer cannot be None.")
+        if TYPE in config[TRAINER]["optimizer"]:
+            if config[TRAINER]["optimizer"][TYPE] not in optimizer_registry:
+                raise ConfigValidationError(
+                    f"Trainer.optimizer.type must be one of: {list(optimizer_registry.keys())}."
+                )
+
+
+@register_config_check
+def check_feature_names_unique(config: ModelConfigDict) -> None:
```
tgaddair, line 155: I would remove this one since it's in the schema already and I would prefer to keep a single source of turth ofr this.

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,327 @@
...
+        if config[TRAINER]["checkpoints_per_epoch"] != 0 and config[TRAINER]["steps_per_checkpoint"] != 0:
+            raise ConfigValidationError(
+                "It is invalid to specify both trainer.checkpoints_per_epoch AND "
+                "trainer.steps_per_checkpoint. Please specify one or the other, or specify neither to "
+                "checkpoint/eval the model every epoch."
+            )
+
+
+@register_config_check
+def check_gbm_horovod_incompatibility(config: ModelConfigDict) -> None:
```
tgaddair, line 195: This is fine for now because we don't validate on the backend, but can be removed in the future since we have separate schemas for ECD and GBM.

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,327 @@
...
+    if BACKEND not in config:
+        return
+    if config[BACKEND] is None:
+        return
+    if config[MODEL_TYPE] == MODEL_GBM and config[BACKEND][TYPE] == "horovod":
+        raise ConfigValidationError("Horovod backend does not support GBM models.")
+
+
+@register_config_check
+def check_gbm_single_output_feature(config: ModelConfigDict) -> None:
```
tgaddair, line 206: This is already in jsonschema / marshmallow, so let's remove.

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,327 @@
...
+    """GBM models only support a single output feature."""
+    if config[MODEL_TYPE] == MODEL_GBM:
+        if len(config[OUTPUT_FEATURES]) != 1:
+            raise ConfigValidationError("GBM models only support a single output feature.")
+
+
+@register_config_check
+def check_gbm_feature_types(config: ModelConfigDict) -> None:
```
tgaddair, line 214: This is also in the schema already. Remove.

ludwig/config_validation/checks.py
ksbrar: @tgaddair Where in the schema? If you are referring to `uniqueItemProperties` remember that that's not an official JSON keyword - merely an extra one that AJV supports. So we would get the validation for unique feature names in the FE but not on the Ludwig side if we removed this check.

ludwig/config_validation/checks.py
ksbrar: +1

ludwig/config_validation/checks.py
ksbrar: +1

ludwig/config_validation/checks.py
tgaddair: Yep, good call. I think it's fine to keep this in, then.

ludwig/config_validation/checks.py
```diff
@@ -0,0 +1,327 @@
...
+
+    if validation_metric_name not in all_valid_metrics:
+        raise ConfigValidationError(
+            f"User-specified trainer.validation_metric '{validation_metric_name}' is not valid. "
+            f"Available metrics are: {all_valid_metrics}"
+        )
+
+
+@register_config_check
+def check_splitter(config: ModelConfigDict) -> None:
```
tgaddair, line 322: All these split checks are run after we have generated the `ModelConfig` object, but for some reason we're running them against untyped dicts instead of the config object? We should change these to operate on the `ModelConfig` object for better type safety.

ludwig/config_validation/validation.py
```diff
@@ -86,14 +83,16 @@ def validate_upgraded_config(updated_config):
...
     from ludwig.utils.backward_compatibility import upgrade_config_dict_to_latest_version
 
     # Update config from previous versions to check that backwards compatibility will enable a valid config
     updated_config = upgrade_config_dict_to_latest_version(config)
-    validate_upgraded_config(updated_config)
+    check_schema(updated_config)
+
+    model_config = ModelConfig.from_dict(updated_config)
+
+    get_config_check_registry().check_config(model_config.to_dict())
```
tgaddair, line 98: As described above, should change this to:

```
get_config_check_registry().check_config(model_config)
```

So validate on the config object, not on a dict.

Furthermore, since this is useful to check anytime we're creating a `ModelConfig`, this should really be run as part of the `ModelConfig.from_dict()` function at the end. That way, any user that wants to create a `ModelConfig` will benefit from these checks.

ludwig/error.py
```diff
@@ -47,3 +47,18 @@ def __str__(self):
 
     def __reduce__(self):
         return type(self), (self.column_name, self.feature_type, self.message)
+
+
+@PublicAPI
+class ConfigValidationError(LudwigError, ValueError):
```
tgaddair, line 53: You need to make this error type serializable. See https://github.com/ludwig-ai/ludwig/pull/2695.

In doing so, please add a unit test similar to the one from the linked PR to verify that serialization works.

tests/integration_tests/test_gbm.py
```diff
@@ -84,48 +84,6 @@ def _train_and_predict_gbm(input_features, output_features, tmpdir, backend_conf
     return preds, model
 
 
-def run_test_gbm_output_not_supported(tmpdir, backend_config):
```
tgaddair, line 87: Are these tests all replicated elsewhere?

tgaddair (changes requested):
Changes requested:

- Use `ModelConfig` in these validation hooks, not a dict
- Run validation hooks at the end of `ModelConfig.from_dict()` instead of within the standalone validation function
- Make `ConfigValidationError` serializable and add a unit test
- Wrap mashmallow ValidationErrors with `ConfigValidationError` to make the exception type consistent.

tests/ludwig/config_validation/test_validate_config_combiner.py
```diff
@@ -113,16 +112,16 @@ def test_config_bad_combiner_types_enums():
 
     # Test weights initializer:
     config["combiner"]["weights_initializer"] = {"test": "fail"}
-    with pytest.raises(ValidationError, match=r"'type' is a required property"):
+    with pytest.raises(ConfigValidationError):
```
tgaddair, line 115: Why drop the match? Is there a way to get this back in? Otherwise, we lose confidence that we're actually testing the right thing.

tgaddair (commented):
As discussed, would still be cool to move this into the ModelConfig, otherwise it's not teribly useful.

tests/ludwig/config_validation/test_validate_config_combiner.py
justinxzhao: I removed the error strings at first because some of them had changed slightly while adding aux checks, and I wanted to minimize noise during development.

After removing the regex strings, however, the tests felt easier to maintain and cleaner overall.

Testing for specific error string regex'es gives us some additional fidelity, but on the whole to me it feels like testing an implementation detail, where the maintenance cost might not outweigh the increased specificity. Happy to add these matches back though if you feel strongly about it.

justinxzhao:
> As discussed, would still be cool to move this into the ModelConfig, otherwise it's not teribly useful.

I'd prefer to make that change in a separate PR if that's ok. I anticipate that folding schema validation and aux checks into the model config will affect several more tests, as every config in ludwig tests will be validated, which we've never tested explicitly for before.

abidwael:
@justinxzhao you can merge with latest master to fix https://github.com/ludwig-ai/ludwig/actions/runs/4178266592/jobs/7236853564#step:12:4290

tests/ludwig/config_validation/test_validate_config_features.py
```diff
@@ -116,3 +116,33 @@ def test_too_many_features_config():
...
+    with pytest.raises(ConfigValidationError):
+        validate_config(
+            {
+                "input_features": [binary_feature()],
+                "output_features": [text_feature()],
+                "model_type": "gbm",
+            }
+        )
+
+    # Multi-output is fine for ECD
```
tgaddair, line 141: Comment is incorrect.

tests/ludwig/utils/test_errors.py
```diff
@@ -14,3 +14,14 @@ def test_input_data_error_serializeable():
     assert loaded_err.feature_type == err.feature_type
     assert loaded_err.message == err.message
     assert str(err) == str(loaded_err)
+
+
+def test_config_validation_error_serializeable():
+    err = ConfigValidationError(message="At least 2 distinct values are required, column only contains ['here']")
+
+    print(f"pickle.dumps(err): {pickle.dumps(err)}")
```
tgaddair, line 22: Remove print.

tgaddair (approved):
LGTM, please address nits before landing.

tests/ludwig/config_validation/test_validate_config_features.py
justinxzhao: Done.

tests/ludwig/utils/test_errors.py
justinxzhao: Done.

PR #3029, opened 2023-01-30
Enables a new GitHub Action for `slow` tests

geoffreyangus:
This PR enables tests with the pytest marker `pytest.mark.slow`. It will run on every commit made to `master`.

This PR removes the `@slow` decorator we used in the passed in favor of a vanilla `pytest.mark` to enable more fine-grained control of the tests run. Prior to this change, `RUN_SLOW=1` was required to run the slow tests _in addition to_ the normal tests. Using `pytest.mark.slow` enables us to run the slow tests exclusively.

The slow test runner downloads the complete Ludwig package (including distributed requirements), so `ray`-enabled tests marked `slow` should be runnable with this action as well.


justinxzhao (approved):
Nice! Thanks for the change, and some cleanup as well.

PR #3034, opened 2023-01-31
Hyperopt Quick Fix

connor-mccorm:


jeffkinnison (approved):

ludwig/schema/model_config.py
```diff
@@ -187,7 +187,7 @@ def __init__(self, config_dict: ModelConfigDict):
 
         # ===== Combiner =====
         if COMBINER in upgraded_config_dict:
-            if self.combiner.type != upgraded_config_dict[COMBINER][TYPE]:
+            if self.combiner.type != upgraded_config_dict.get(COMBINER, {}).get(TYPE, None):
                 self.combiner = combiner_registry.get(upgraded_config_dict[COMBINER][TYPE]).get_schema_cls()()
```
ShreyaR, line 191: Would you also need to use the `dict.get` method in line 191?

connor-mccorm: We could add it there, though if the check on line 190 passes, then we know for sure that the `combiner` and `type` exist so this won't throw a key error.

PR #3038, opened 2023-01-31
Expected Impact and Ordering for GBM Params

connor-mccorm:


connor-mccorm:
Ya I think it works for the time being. They are logically grouped so related params are all together so I think it looks good for now!

PR #3039, opened 2023-02-01
Fix gradient clipping typo

geoffreyangus:


arnavgarg1 (approved):
Good catch!

justinxzhao (approved):

justinxzhao (commented):
Is there a simple test that'd be worth adding to catch this typo? Alternatively, if we're thinking about turning on gradient clipping by default, this might be a moot point.

geoffreyangus:
Hm, @justinxzhao it seems like more trouble than its probably worth– to test this specifically, since the bug was in the trainer, we would likely have to:

1. Move the function out of the trainer into `utils` somewhere

Then in a test, we would have to:

1. instantiate a torch module
2. artificially get its gradient to be something high (haven't done this before, but I'm sure I could figure it out)
3. pass in the modules parameters into the new function along with a `GradientClippingConfig` with some test values
4. verify that the torch module's parameters have an overall norm that coincides with the values we would expect

I haven't worked with gradients this closely before, and since it's a one-line change and the test is very specific to these few lines of code, it might not be worth doing. Curious what you think though or if you think there's an easier way to do this.

justinxzhao:
> Hm, @justinxzhao it seems like more trouble than its probably worth– to test this specifically, since the bug was in the trainer, we would likely have to:
> 
> 1. Move the function out of the trainer into `utils` somewhere
> 
> Then in a test, we would have to:
> 
> 1. instantiate a torch module
> 2. artificially get its gradient to be something high (haven't done this before, but I'm sure I could figure it out)
> 3. pass in the modules parameters into the new function along with a `GradientClippingConfig` with some test values
> 4. verify that the torch module's parameters have an overall norm that coincides with the values we would expect (not sure what each of the gradnorm values in `GradientClippingConfig` does– would have to read up / learn how they interact)
> 
> I haven't worked with gradients this closely before, and since it's a one-line change and the test is very specific to these few lines of code, it might not be worth doing. Curious what you think though or if you think there's an easier way to do this.

Ah, I mistook the bug for a misspelled field accessor instead of using the wrong field accessor.

Let's go forward without a test, as a test that's not more trouble than it's worth would probably just be testing implementation details.

PR #3042, opened 2023-02-02
Removed log spam from distributed loader

tgaddair:
Because this function gets called many times on the critical path for each metric, it produces a lot of spam, and is not a very informative message.

### DIFF
--- a/ludwig/distributed/__init__.py
+++ b/ludwig/distributed/__init__.py
@@ -1,4 +1,3 @@
-import logging
 from typing import Type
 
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):

arnavgarg1 (approved):
Alternative is to use the `log_once` function in `ludwig/utils/logging_utils`

tgaddair:
@arnavgarg1 Nice! Didn't know about this. I think this log might be kinda useless in general, but good to know about that!

PR #3046, opened 2023-02-03
Fix TLE: safe divide by zero + normalize at sequence level

jppgks:
This PR fixes
1. a bug where token attributions contained NaN values
2. a bug where token attributions were normalized by the norm across the whole batch instead of the per-sequence norm 

and adds a test to avoid these from happening again going forward.

For 1., the cause was a division by zero silently returning NaNs. 
The norm of token attributions is 0 when all attributions are 0. In that case, we now return the original 0 attributions instead of NaNs.

For 2., the cause was a missing `dim` argument in the call to `torch.linalg.norm`.

### DIFF
--- a/ludwig/explain/captum.py
+++ b/ludwig/explain/captum.py
@@ -429,7 +429,10 @@ def get_token_attributions(
     )
 
     # Normalize token-level attributions to visualize the relative importance of each token.
-    token_attributions = token_attributions / torch.norm(token_attributions)
+    norm = torch.linalg.norm(token_attributions, dim=1)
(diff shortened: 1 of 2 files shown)


connor-mccorm (approved):
LGTM! Thanks for the fix Joppe!

martindavis:
Thanks for updating the BE @jppgks ! Hey @connor-mccorm , is there any FE code we can simplify now that this is merged?

connor-mccorm:
I think so actually. Hung-wei added some logic to control for nans but if this prevents them to begin with we can remove. I'll add a linear issue to track this.

PR #3053, opened 2023-02-07
Update schema to correctly reflect supported missing value strategies for different feature types

arnavgarg1:
This PR ensures that Ludwig's schema correctly reflects the missing value strategies supported for input and output feature types in Ludwig.

Notable changes:
- `drop_row` is not supported for input features, but only for output features. This is now removed as an option for input feature types.
- `fill_with_mean` is only supported for number features, so this is removed from the schema options for all other feature types.
- `date` input features only support `fill_with_const`, `bfill` and `ffill`.

Associated documentation update PR: https://github.com/ludwig-ai/ludwig-docs/pull/258


justinxzhao:
I can see reasons to keep `drop_row` around for input features, for example, as a convenience to automatically drop examples that are missing them.

Curious if other people have thoughts on this.

arnavgarg1:
@justinxzhao That makes sense, but I think that can cause situations where all rows get dropped in the training set after preprocessing which I don't think, is a good outcome. Wdyt?

justinxzhao:
@arnavgarg1 I would expect that to be an odd, but desirable outcome actually, for what's probably a mismatch between the configuration and the user's understanding of their dataset.

arnavgarg1:
@justinxzhao makes sense, added drop row back :)

justinxzhao (approved):
LGTM

PR #3054, opened 2023-02-07
Fix `fill_with_mode` when using Dask

arnavgarg1:
Dask returns a Dask index structure because of lazy evaluation, so we need to force computation of the `value_counts()` method so we can grab the mode of the column to use as the fill_value.


justinxzhao (approved):
Thanks for the quick fix!

PR #3058, opened 2023-02-08
Improve observability when using cached datasets during preprocessing

arnavgarg1:


ludwig/data/cache/manager.py
```diff
@@ -81,6 +87,11 @@ def delete(self):
                 # Parquet entries in the cache_ma can be pointers to directories.
                 delete(fname, recursive=True)
 
+    def get_cached_obj_path(self, cached_obj_name: str) -> str:
+        if cached_obj_name in self.cache_map:
```
tgaddair, line 91: Simplify:

```
return self.cache_map.get(cached_obj_name)
```

ludwig/data/cache/manager.py
```diff
@@ -25,20 +25,26 @@ def get(self):
             return None
 
         try:
-            cache_training_set_metadata = data_utils.load_json(training_set_metadata_fp)
+            cached_training_set_metadata = data_utils.load_json(training_set_metadata_fp)
         except Exception as e:
-            logger.error(f"failed to load cached training set metadata at {training_set_metadata_fp}", exc_info=e)
+            logger.error(f"Failed to load cached training set metadata at {training_set_metadata_fp}", exc_info=e)
```
tgaddair, line 30: Should probably be:

```
logger.exception(f"Failed to load cached training set metadata at {training_set_metadata_fp}")
```

Then we don't need the exc_info shennigans.

tgaddair (approved):
LGTM, just nits.

PR #3061, opened 2023-02-08
Remove passthrough encoder from sequence and text features encoder registry

jeffkinnison:
Using the passthrough encoder with sequence or text features leads to the following error:

```
File "/home/ray/anaconda3/lib/python3.8/site-packages/ludwig/utils/torch_utils.py", line 211, in _computed_output_shape
    dummy_input = torch.rand(2, *self.input_shape, device=self.device)
TypeError: rand() argument after * must be an iterable, not NoneType
```
This update removes the passthrough encoder from the sequence and text features' encoder registries.

### DIFF
--- a/ludwig/schema/encoders/sequence_encoders.py
+++ b/ludwig/schema/encoders/sequence_encoders.py
@@ -19,7 +19,7 @@ class SequenceEncoderConfig(BaseEncoderConfig):
 
 
 @DeveloperAPI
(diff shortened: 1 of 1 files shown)


connor-mccorm:
We're just removing this temporarily until it's fixed right?

jeffkinnison:
The underlying issue--`self.input_shape` not being set--seems fixable. I'll look into a fix this week and re-enable if possible.

connor-mccorm (approved):

PR #3062, opened 2023-02-08
Remove RNN invalid cell types from the schema

jeffkinnison:
The schema currently supports `StackedRNN` and `StackedCNNRNN` cell types that are not in the registry, including `lstm_block`, `ln`, `lstm_cudnn`, `gru_block`, `gru_cudnn`, and `None`. This update removes the invalid cell types from the schema and updates docstrings.


justinxzhao (approved):
Thanks!

PR #3063, opened 2023-02-08
fix: Set `hyperopt.parameters` default to `{}` and reject `None`

ksbrar:


ludwig/schema/model_types/utils.py
```diff
@@ -181,7 +181,9 @@ def set_hyperopt_defaults_(config: "ModelConfig"):
 
     # Set default num_samples based on search space if not set by user
     if config.hyperopt.executor.num_samples is None:
-        _contains_grid_search_params = contains_grid_search_parameters(config.hyperopt.to_dict())
+        from ludwig.schema.hyperopt import HyperoptConfig
+
+        _contains_grid_search_params = contains_grid_search_parameters(HyperoptConfig.Schema().dump(config.hyperopt))
```
tgaddair, line 186: Hmm, not sure I understand this. Shouldn't the hyperopt config object already be fully materialized in this way?

Also, need a test that repros the issue.

tgaddair:
Would prefer to remove the `hyperopt: {}` section since it's essentially deprecated. See #3064.

ksbrar:
Closed in favor of #3064

PR #3064, opened 2023-02-09
Fixed handling of {} hyperopt config section

tgaddair:


arnavgarg1 (approved):

ludwig/schema/hyperopt/__init__.py
```diff
@@ -65,7 +65,7 @@ class HyperoptConfig(schema_utils.BaseMarshmallowConfig, ABC):
         )
     )
 
-    parameters: Dict = schema_utils.Dict()
+    parameters: Dict = schema_utils.Dict(allow_none=False)
```
ksbrar, line 68: @tgaddair It won't throw an error but technically this is a contradiction since the `default` value for `Dict` fields is still `None`. I suggest either changing that or manually setting the default here to be `{}`

ksbrar (changes requested):

ludwig/schema/hyperopt/__init__.py
tgaddair: I changed the behavior to make the default `{}` when `None` is provided. Does that address the issue?

tgaddair: I think in general we should prefer not to allow things to be null when materialized, since it means more null-checking. Defaulting to an empty dict feels more correct to me. Let me know what you think!

ksbrar: Ah, sorry glanced over that. I guess that works. IMO it still looks odd but I'll approve if you're comfortable with it.

ksbrar (approved):

ludwig/schema/hyperopt/__init__.py
ksbrar: @tgaddair I agree - I would prefer to just change the default at the field level to be an empty dict rather than manually handling the case with the if statement. Maybe in the future there will be a dict field we really do want to be set to `None` by default.

tgaddair: Problem with doing something like `def Dict(default={})` is that the dict there is mutable, which means any modifications will break subsequent callers. That's why in Python the convention is to set default params to `None` and then override to `{}` in the function. That may be what you're seeing as odd.

ksbrar: OHH. That makes sense, totally forgot about that.

PR #3068, opened 2023-02-09
Increase default early stopping to 15.

justinxzhao:


arnavgarg1 (approved):
I would see if we can prevent the indentation changes for `trainer.yaml` since the old version is far more readable because of the tabbed indenting

justinxzhao:
> I would see if we can prevent the indentation changes for `trainer.yaml` since the old version is far more readable because of the tabbed indenting

Good call. Done.

tests/ludwig/utils/test_defaults.py
```diff
@@ -243,7 +242,6 @@ def test_merge_with_defaults():
     assert DECODER in merged_config[DEFAULTS][CATEGORY]
     assert LOSS in merged_config[DEFAULTS][CATEGORY]
     assert COMBINER in merged_config
-    assert merged_config[TRAINER][EARLY_STOP] == 5
```
arnavgarg1, line 246: Set this to 15 instead of removing this check entirely?

PR #3069, opened 2023-02-09
Generate proc_column only after all preprocessing parameters are merged in to prevent incorrect cached dataset reads

arnavgarg1:
This PR fixes situations where two configs using the same dataset would result in the same checksum, but the underlying feature proc_cols would be different because of different feature-specific preprocessing params.

Namely, it makes two changes:

1. Moves generation of the proc_col at the end of the instantiation of the ModelConfig object.
2. Modifies checksum calculation to account for the proc column name rather than the raw feature column name.

### DIFF
--- a/ludwig/data/cache/util.py
+++ b/ludwig/data/cache/util.py
@@ -1,5 +1,5 @@
 import ludwig
-from ludwig.constants import DEFAULTS, INPUT_FEATURES, OUTPUT_FEATURES, PREPROCESSING, PROC_COLUMN, TYPE
+from ludwig.constants import DEFAULTS, INPUT_FEATURES, NAME, OUTPUT_FEATURES, PREPROCESSING, TYPE
 from ludwig.data.cache.types import CacheableDataset
 from ludwig.types import ModelConfigDict
 from ludwig.utils.data_utils import hash_dict
@@ -12,7 +12,7 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "dataset_checksum": original_dataset.checksum,
         "global_preprocessing": config.get(PREPROCESSING, {}),
         "global_defaults": config.get(DEFAULTS, {}),
-        "feature_proc_columns": [feature[PROC_COLUMN] for feature in features],
+        "feature_names": [feature[NAME] for feature in features],
         "feature_types": [feature[TYPE] for feature in features],
         "feature_preprocessing": [feature.get(PREPROCESSING, {}) for feature in features],
     }
--- a/ludwig/data/cache/util.py
+++ b/ludwig/data/cache/util.py
@@ -12,7 +12,7 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "dataset_checksum": original_dataset.checksum,
         "global_preprocessing": config.get(PREPROCESSING, {}),
         "global_defaults": config.get(DEFAULTS, {}),
-        "feature_names": [feature[PROC_COLUMN] for feature in features],
+        "feature_proc_columns": [feature[PROC_COLUMN] for feature in features],
         "feature_types": [feature[TYPE] for feature in features],
         "feature_preprocessing": [feature.get(PREPROCESSING, {}) for feature in features],
     }
--- a/ludwig/data/cache/util.py
+++ b/ludwig/data/cache/util.py
@@ -1,5 +1,5 @@
 import ludwig
-from ludwig.constants import DEFAULTS, INPUT_FEATURES, NAME, OUTPUT_FEATURES, PREPROCESSING, TYPE
+from ludwig.constants import DEFAULTS, INPUT_FEATURES, OUTPUT_FEATURES, PREPROCESSING, PROC_COLUMN, TYPE
 from ludwig.data.cache.types import CacheableDataset
 from ludwig.types import ModelConfigDict
 from ludwig.utils.data_utils import hash_dict
@@ -12,7 +12,7 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "dataset_checksum": original_dataset.checksum,
         "global_preprocessing": config.get(PREPROCESSING, {}),
         "global_defaults": config.get(DEFAULTS, {}),
-        "feature_names": [feature[NAME] for feature in features],
+        "feature_names": [feature[PROC_COLUMN] for feature in features],
         "feature_types": [feature[TYPE] for feature in features],
         "feature_preprocessing": [feature.get(PREPROCESSING, {}) for feature in features],
     }
--- a/ludwig/schema/model_types/utils.py
+++ b/ludwig/schema/model_types/utils.py
@@ -164,15 +164,15 @@ def set_derived_feature_columns_(config_obj: "ModelConfig"):
     Proc_column is set to a hash of the feature's preprocessing configuration.
     """
     for feature in config_obj.input_features:
-        if not feature.column:
+        if feature.column is None:
             feature.column = feature.name
-        if not feature.proc_column:
+        if feature.proc_column is None:
             feature.proc_column = compute_feature_hash(feature.to_dict())
 
     for feature in config_obj.output_features:
-        if not feature.column:
+        if feature.column is None:
             feature.column = feature.name
-        if not feature.proc_column:
+        if feature.proc_column is None:
             feature.proc_column = compute_feature_hash(feature.to_dict())
 
 
--- a/ludwig/schema/model_types/utils.py
+++ b/ludwig/schema/model_types/utils.py
@@ -163,7 +163,9 @@ def set_validation_parameters(config: "ModelConfig"):
 
 def set_derived_feature_columns_(config_obj: "ModelConfig"):
     """Assigns column and proc_column values to features that do not have them set.
-    Proc_column is set to a hash of the feature's preprocessing configuration."""
+
+    Proc_column is set to a hash of the feature's preprocessing configuration.
+    """
     for feature in config_obj.input_features:
         if not feature.column:
             feature.column = feature.name
--- a/ludwig/schema/model_types/utils.py
+++ b/ludwig/schema/model_types/utils.py
@@ -161,12 +161,20 @@ def set_validation_parameters(config: "ModelConfig"):
         config.trainer.validation_metric = output_config_registry[out_type].default_validation_metric
 
 
-def set_derived_feature_columns_(config: ModelConfigDict):
-    for feature in config.get(INPUT_FEATURES, []) + config.get(OUTPUT_FEATURES, []):
-        if COLUMN not in feature:
-            feature[COLUMN] = feature[NAME]
-        if PROC_COLUMN not in feature:
-            feature[PROC_COLUMN] = compute_feature_hash(feature)
+def set_derived_feature_columns_(config_obj: "ModelConfig"):
+    """Assigns column and proc_column values to features that do not have them set.
+    Proc_column is set to a hash of the feature's preprocessing configuration."""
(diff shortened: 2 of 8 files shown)


w4nderlust (approved):

ludwig/data/cache/util.py
```diff
@@ -12,7 +12,7 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "dataset_checksum": original_dataset.checksum,
         "global_preprocessing": config.get(PREPROCESSING, {}),
         "global_defaults": config.get(DEFAULTS, {}),
-        "feature_names": [feature[NAME] for feature in features],
+        "feature_names": [feature[PROC_COLUMN] for feature in features],
```
tgaddair, line 15: Something to note: there is not necessarily a 1:1 map between feature name and proc col. It's possible that for a single proc col, there are multiple features that use it.

That said, proc cols is what we care about, so this should be fine (though slightly more aggressive at biasing towards cache misses than is strictly necessary). The one thing I would do is change the key from `feature_names` to `feature_proc_columns`.

tgaddair (commented):
Should we remove all the places in the code we're setting `PROC_COLUMN` that isn't in generation of the ModelConfig? I think so.

ludwig/data/cache/util.py
arnavgarg1: @tgaddair Ah yeah, good callout. 

>It's possible that for a single proc col, there are multiple features that use it.

This would happen if two features have the same preprocessing parameters right?

If we want to be less aggressive about biasing towards cache misses, what would you recommend? Keeping feature_names and not using `proc_column`? Also ok to do that since we're only computing this at the end of ModelConfig creation

ludwig/schema/model_types/utils.py
```diff
@@ -161,12 +158,22 @@ def set_validation_parameters(config: "ModelConfig"):
...
-            feature[COLUMN] = feature[NAME]
-        if PROC_COLUMN not in feature:
-            feature[PROC_COLUMN] = compute_feature_hash(feature)
+def set_derived_feature_columns_(config_obj: "ModelConfig"):
+    """Assigns column and proc_column values to features that do not have them set.
+
+    Proc_column is set to a hash of the feature's preprocessing configuration.
+    """
+    for feature in config_obj.input_features:
+        if not feature.column:
```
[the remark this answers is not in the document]
arnavgarg1: Agreed

arnavgarg1:
> Should we remove all the places in the code we're setting `PROC_COLUMN` that isn't in generation of the ModelConfig? I think so.

Will take a look at this, I do agree.

ludwig/data/cache/util.py
tgaddair: So, I'll start off by saying it's a little risky because if we get it wrong, a false positive cache hit is obviously worse than false negative cache miss. That said, the idea is that technically so long as two configs (a) have the same columns (not features), and (b) those columns are preprocessed the same, then they can share the same preprocessed dataset. That's independent of how many features there are, their names, etc.

So, to do this the "right" way we would want to:

- Change the `proc_column` calculation to consider the fully rendered preprocessing of the feature after all fixed preprocessing params have been merged in (should already be doing this) AND include the feature type (right now we're not doing this for some reason, so this seems like a bit of an oversight that could get us into trouble, unless I;'m forgetting something).
- Change the checksum calculation to exclude the per-feature metadata like feature types, etc. Only consider the unique SET of proc_column hashes (order and number of duplicates doesn't matter), ludwig_version, and dataset_info.

I think this should be a follow-up PR with A LOT of tests to go with it, though, haha.

ludwig/data/cache/util.py
```diff
@@ -12,7 +12,7 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "dataset_checksum": original_dataset.checksum,
         "global_preprocessing": config.get(PREPROCESSING, {}),
         "global_defaults": config.get(DEFAULTS, {}),
-        "feature_names": [feature[NAME] for feature in features],
+        "feature_proc_columns": [feature[PROC_COLUMN] for feature in features],
```
tgaddair, line 15: Just to be safe for now, I'd suggest adding back `feature_names`. Just to be 100% sure we're net decreasing the number of false positive cache hits.

arnavgarg1: Sounds good!

arnavgarg1:
Seems like we use `proc_col` pretty flexibly to either indicate:

1. the hash of the feature name that gets appended (so the feature name after the preprocessed dataset is loaded in)
2. to refer to "processed" columns during preprocessing, not the hash

Kinda confusing, I will rename 2. to be `processed_col` or `processed_cols` instead to delineate this in our code base

w4nderlust:
> Seems like we use `proc_col` pretty flexibly to either indicate:
> 
> 1. the hash of the feature name that gets appended (so the feature name after the preprocessed dataset is loaded in)
> 2. to refer to "processed" columns during preprocessing, not the hash
> 
> Kinda confusing, I will rename 2. to be `processed_col` or `processed_cols` instead to delineate this in our code base

Can you point to the specific parts of the code where we do either of these? There may be a reason but without concrete examples I can't fully say.

arnavgarg1:
@w4nderlust Here's some links.

Everything makes sense in context, but I think longer naming might just be better to clearly express what this represents. I think that we should called items in (2) like `processed_columns`, `processed_column` and `processed_column_name` to just make this all very clear I guess. Maybe I'm being very picky here.

For 1, pretty much the references in this PR.

For 2:
    - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1186
    - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1413
    - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/dask.py#L87 (also true with Pandas backend in similar fashion)

ludwig/data/cache/util.py
arnavgarg1: Makes sense. I will address this in a follow-up PR!

For now, this PR addresses:

>Change the proc_column calculation to consider the fully rendered preprocessing of the feature after all fixed preprocessing params have been merged in (should already be doing this)

Do we feel good with this change for now?

w4nderlust: The `type` thing is a good point, ideally there's only one type associated with a column and users always use it, but there is nothing enforcing it, so two entries of the input feature list can have same column and different types. A possible case I can imagine is having two features given the same column that contains ordinal information. You may want to both have a category feature for it and a numerical one (as the numerical will already know about the relative ordering of the values). I think we never found issues with it for a couple reasons though. One is that I doubt anyone has done it. Second is that even if they did it, if the two features have different fully rendered preprocessing sections (which I beleive is true for all features, maybe with the exception of text / sequence and set/bag) then there wouldn't be any collision anyway because the hash is computed on the preproc dict. Anyway, I endorse the idea od including explicitly the type when computing the hash.

Regarding the proc column point. I can imagine the case of two different configs leading to the same set of proc columns. that's obvious when 2 configs only differ by training parameters or encoder parameters. But it could be the case also if they are identical and only differ by name of features that use the same column, or also number of features that use the same column. In all those cases we should keep the same cache. i believe that the implementation of:
```
        "feature_names": [feature[PROC_COLUMN] for feature in features],
```
Is actually incorrect, as it may cover the case of different names, but notthe case of different number of features.
I believe that a correct implementation would be:
```
        "feature_proc_cols": set(feature[PROC_COLUMN] for feature in features),
```
To avoid repeated entries of `PROC_COLUMN` coming from different (and haigher nuebr of) features.

A concrete example would be these configs:
```{input_features: [{name: a, column: x}, {name: b, column: x}]}```
and
``````{input_features: [{name: a, column: x}, {name: c, column: x}]}```
and
``````{input_features: [{name: a, column: x}, {name: b, column: x}, {name: c, column: x}]}```
They should all produce the same cache and the cache that works for one should be reusable for the others.

It would be different if the `c` feature in the last example had a different type from the others.

Anyway, I agree all these corner cases should be tested out with a separate PR, containing only this modification, and a very comprehensive set of tests to determine expected set of proc cols given a config and if the checksum match or do not. We would need both many positive adn negative cases.

So maybe the safest thing to do for now is to land the pr by reverting just this line to what it was before, and have a follow up PR to have full confidence on the logic I proposed above.

w4nderlust:
> @w4nderlust Here's some links.
> 
> Everything makes sense in context, but I think longer naming might just be better to clearly express what this represents. I think that we should call items in (2) like `processed_columns`, `processed_column` and `processed_column_name` to just make this all very clear I guess. Maybe I'm being very picky here.
> 
> For 1, pretty much the references in this PR.
> 
> For 2: - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1186 - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1413 - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/dask.py#L87 (also true with Pandas backend in similar fashion)

I see. I actually feel ok with it. It's kinda the same thing with `column`, it isn't the column really, it's the column name, but we call it column for brevity in the cofig.
Maybe one thing we can do about it is whenever we set strings that are external facing we say `PROC_COLUMN` and the intermediate internal variables could be called `proc_cols_names/keys` or something like that. Although if we decide to do it for clarity, we should do it in a followup PR.

ludwig/data/cache/util.py
arnavgarg1: @w4nderlust Thanks for the thorough write up on the edge cases.

It seems like between the three of us, we're pretty aligned on what changes should we made going forward:

1. `proc_col` hash should not just consider preprocessing parameters, but both preprocessing parameters along with the feature type. This is also going to require a pretty interesting backwards compatibility layer because we want to ensure that models that just used preprocessing parameters (and have those `proc_cols` set) now are updated to reflect preprocessing + feature type combined hash.
3.  Instead of using just `feature_names` in the info dict to compute the checksum hash, we should use the set of proc col hashes since it has higher guarantees about ensuring caches are only used when they should be.

Does that summarize it reasonably? @w4nderlust @tgaddair?

arnavgarg1:
> > @w4nderlust Here's some links.
> > Everything makes sense in context, but I think longer naming might just be better to clearly express what this represents. I think that we should call items in (2) like `processed_columns`, `processed_column` and `processed_column_name` to just make this all very clear I guess. Maybe I'm being very picky here.
> > For 1, pretty much the references in this PR.
> > For 2: - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1186 - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1413 - https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/dask.py#L87 (also true with Pandas backend in similar fashion)
> 
> I see. I actually feel ok with the way it is right now. It's kinda the same thing with `column`, it isn't the column really, it's the column name, but we call it column for brevity in the cofig.
> 
> Maybe if we really wanted to make a change, one thing we can do about it is whenever we set strings that are external facing we say `PROC_COLUMN` and the intermediate internal variables could be called `proc_cols_names/keys` or something like that. Although if we decide to do it for clarity, we should do it in a followup PR.

Agreed! We can think of a good convention to delineate these, but that will be a separate PR as well. 

Going to keep the scope of this PR to what's currently in the PR.

ludwig/data/cache/util.py
tgaddair: Yep, I think we're all aligned that this is a good next step. Thanks everyone :)

tgaddair (approved):
Nice change!

PR #3073, opened 2023-02-10
Fix default image on image read failure

geoffreyangus:
This PR fixes the default image used on image read failure. The default image is inserted _after_ images are read, meaning that it bypasses the cast to `np.float32` the default image was set to be `np.uint8`. This change ensures that it is also `np.float32`, and more importantly, that we have tests around default image to prevent this from happening again. The test mocks the image read function to simulate image read failure.


jeffkinnison (approved):
LGTM! That was a subtle issue, great job tracking it down!

tests/integration_tests/utils.py
```diff
@@ -961,6 +965,54 @@ def assert_all_required_metrics_exist(
...
+    if_config = if_configs[0]
+
+    if_config_proc_column = if_config.proc_column
+    for result in [
+        preprocessed_dataset.training_set,
+        preprocessed_dataset.validation_set,
+        preprocessed_dataset.test_set,
+    ]:
+        result_df = result.to_df()
+        result_df_image_col = result_df[if_config_proc_column]
```
justinxzhao, line 1001: Nit: it looks like this function is named generically for any feature type, but image is used here.

justinxzhao (approved):
Just one nit!

tests/integration_tests/utils.py
geoffreyangus: Good catch! Fixed in latest.

PR #3072, opened 2023-02-09
Add MPS device support

tgaddair:

PR #3074, opened 2023-02-10
Require env var LUDWIG_USE_MPS to enable MPS

tgaddair:
Follow-up to #3072. 

It's been observed in testing that MPS can sometimes hurt performance for small tabular models, and can lead to runtime errors when `PYTORCH_ENABLE_MPS_FALLBACK` is not set.


arnavgarg1 (approved):
Nice, looks good!

PR #3075, opened 2023-02-10
Schemify initializers

tgaddair:


ludwig/schema/utils.py
```diff
@@ -803,83 +791,6 @@ def _jsonschema_type_mapping(self):
     )
 
 
-@DeveloperAPI
-def InitializerOrDict(
```
ksbrar, line 807: Jotting it down (as discussed in our sync @justinxzhao ) - this change will remove the ability for users to pass arbitrary dictionaries of values as initialization parameters.

PR #3076, opened 2023-02-10
Use `window_size_bytes: auto` to specify automatic windowing

jeffkinnison:
This updates `RayDataset`, `RayDatasetManager`, and `RayBackend` to accept `"auto"` as a valid input for `window_size_bytes`, and consolidates `RayDataset` creation logic to handle auto-windowing.

### DIFF
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -89,7 +89,7 @@ class RayDataset(Dataset):
         if isinstance(window_size_bytes, int):
             window_size = window_size_bytes
 
-        # If the user does not supply a window size and the dataset is large,
+        # If the user requests auto window sizing and the dataset is large,
         # set the window size to `<available memory> // 5`.
         elif window_size_bytes == "auto":
             ds_memory_size = self.in_memory_size_bytes
@@ -168,7 +168,7 @@ class RayDatasetManager(DatasetManager):
         training_set_metadata: TrainingSetMetadataDict,
     ) -> "RayDataset":
         """Create a new Ray dataset with config."""
-        window_size_bytes = config.get("backend", {}).get("loader", None).get("window_size_bytes", None)
+        window_size_bytes = self.backend._data_loader_kwargs.get("window_size_bytes", None)
         return RayDataset(
             dataset, get_proc_features(config), training_set_metadata, self.backend, window_size_bytes=window_size_bytes
         )
--- a/ludwig/data/dataset/ray.py
+++ b/ludwig/data/dataset/ray.py
@@ -68,6 +68,7 @@ class RayDataset(Dataset):
         features: Dict[str, FeatureConfigDict],
         training_set_metadata: TrainingSetMetadataDict,
         backend: Backend,
+        window_size_bytes: Optional[Union[int, Literal["auto"]]] = None,
     ):
         self.df_engine = backend.df_engine
         self.ds = self.df_engine.to_ray_dataset(df) if not isinstance(df, str) else read_remote_parquet(df)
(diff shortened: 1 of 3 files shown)


ludwig/data/dataset/ray.py
```diff
@@ -60,20 +60,15 @@ def read_remote_parquet(path: str):
 
 @DeveloperAPI
 class RayDataset(Dataset):
-    """Wrapper around ray.data.Dataset.
-
-    Attributes:
-        auto_window: If True and the dataset is larger than available memory,
-            automatically set window size to `<available memory> // 5`.
-    """
+    """Wrapper around ray.data.Dataset."""
```
arnavgarg1, line 63: nit: While we're updating the doc strings, should we update it to all of the inputs and their types for this class? toally ok to skip as well

jeffkinnison: Ah, good catch. I meant to add `window_size_bytes` back in and fill in the rest of the docstring.

ludwig/data/dataset/ray.py
```diff
@@ -82,16 +77,21 @@ def __init__(
...
+    def get_window_size_bytes(self, window_size_bytes: Optional[Union[int, Literal["auto"]]] = None) -> int:
+        """Return this dataset's window size in bytes, or translate auto-windowing into bytes."""
+        # By default, set to -1 so that an infinite window size
+        # will be used which effectively results in bulk data ingestion
+        window_size = -1
 
-        # If the user does not supply a window size and the dataset is large,
+        # If user has specified a window size, use it as-is.
+        if isinstance(window_size_bytes, int):
+            window_size = window_size_bytes
```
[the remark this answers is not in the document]
jeffkinnison: I'm generally a fan of a single exit point for a function unless some serious computation happens, but happy to refactor.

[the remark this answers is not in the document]
arnavgarg1: That's fair, can leave this as is

tests/integration_tests/test_ray.py
```diff
@@ -1015,15 +1014,14 @@ def create_dataset_pipeline(
             "output_features": [{"name": "out_column", "type": "binary"}],
             TRAINER: {"epochs": 1, BATCH_SIZE: 128},
         }
-        backend_config = {**RAY_BACKEND_CONFIG}
+        backend_config = copy.deepcopy(RAY_BACKEND_CONFIG)
+        backend_config["loader"] = {"window_size_bytes": window_size_bytes}
```
arnavgarg1, line 1018: Just curious - are we always guaranteed to have the `loader` key in the backend config? If not, do we need to add a default value for this in the schema (possibly `None`)?

ludwig/data/dataset/ray.py
jeffkinnison: Added in [db4218b](https://github.com/ludwig-ai/ludwig/pull/3076/commits/db4218bbb5404fb45e3a7795b6177eea0f05484e)

ludwig/data/dataset/ray.py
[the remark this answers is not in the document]
jeffkinnison: Revised in [db4218b](https://github.com/ludwig-ai/ludwig/pull/3076/commits/db4218bbb5404fb45e3a7795b6177eea0f05484e)

tests/integration_tests/test_ray.py
jeffkinnison: Under the hood, this update uses `backend._data_loader_kwargs` which the `RayBackend` initializer sets with `self._data_loader_kwargs = loader or {}`. We don't actually rely on the existence of `loader`, but rather the backend's handling of its config.

tgaddair (approved):

ludwig/data/dataset/ray.py
```diff
@@ -82,16 +86,17 @@ def __init__(
         self.data_hdf5_fp = training_set_metadata.get(DATA_TRAIN_HDF5_FP)
         self.data_parquet_fp = training_set_metadata.get(DATA_TRAIN_PARQUET_FP)
         self._processed_data_fp = df if isinstance(df, str) else None
-        self.auto_window = auto_window
+        self.window_size_bytes = self.get_window_size_bytes(window_size_bytes)
 
-    def get_window_size_bytes(self, window_size_bytes: Optional[int] = None) -> int:
-        # If user has specified a window size, use it as is
-        if window_size_bytes:
+    def get_window_size_bytes(self, window_size_bytes: Optional[Union[int, Literal["auto"]]] = None) -> int:
```
arnavgarg1, line 91: Really nice use of literal, we should use that more in Ludwig

arnavgarg1 (approved):
Great work! Thanks for cleaning up the implementation/API

PR #3079, opened 2023-02-11
Disallow certain config parameters from accepting `null` as a value

abidwael:


ludwig/schema/hyperopt/executor.py
```diff
@@ -27,7 +27,7 @@ class ExecutorConfig(schema_utils.BaseMarshmallowConfig):
     )
 
     time_budget_s: int = schema_utils.PositiveInteger(
-        default=3600, description="The number of seconds for the entire hyperopt run."
+        default=3600, allow_none=True, description="The number of seconds for the entire hyperopt run."
```
tgaddair, line 30: Hmm, shouldn't `None` here correspond to "no time limit"?

ludwig/schema/utils.py
```diff
@@ -242,7 +242,7 @@ def ReductionOptions(default: Union[None, str] = None, description="", parameter
 @DeveloperAPI
 def RegularizerOptions(
     default: Union[None, str] = None,
-    allow_none: bool = True,
+    allow_none: bool = False,
```
tgaddair, line 245: Having default of `None` is.contradictory. If we remove `None`, we should choose a different default.

ludwig/schema/utils.py
```diff
@@ -260,10 +260,11 @@ def RegularizerOptions(
 def String(
     description: str,
     default: Union[None, str] = None,
-    allow_none: bool = True,
+    allow_none: bool = False,
```
tgaddair, line 263: Same as above.

ludwig/schema/utils.py
```diff
@@ -260,10 +260,11 @@ def RegularizerOptions(
...
     pattern: str = None,
     parameter_metadata: ParameterMetadata = None,
 ):
+    allow_none = allow_none or default is None
```
tgaddair, line 267: This is too implicit. I prefer making it explcit on the caller side. If `allow_none` is `False` they must choose a non-null default.

ludwig/schema/utils.py
```diff
@@ -354,7 +355,7 @@ def ProtectedString(
 def IntegerOptions(
     options: TList[int],
     default: Union[None, int] = None,
-    allow_none: bool = True,
+    allow_none: bool = False,
```
tgaddair, line 358: Same as above.

ludwig/schema/utils.py
```diff
@@ -484,7 +485,7 @@ def PositiveInteger(
 def NonNegativeInteger(
     description: str,
     default: Union[None, int] = None,
-    allow_none: bool = True,
+    allow_none: bool = False,
```
tgaddair, line 488: Same as above. Note that you can make `default` a required param if needed.

tgaddair (changes requested):
General feedback is that it's confusing to both have `None` be the default and `allow_none=False`. I would suggest either:

- Making `allow_none=True` the default
- Making `default` a required param

abidwael:
Makes sense @tgaddair.

I made `default` a required param and kept `allow_none=False` so we don't run into the same problem when for other parameters in the config or when we add new parameters. I did this for most of the config parameter types (all except `Dict`, `List`, and `DictList`). I kept the rest as they are because it seems like lots of them (if not all) have `default=None` in the schema. These could be schemafied and that way the default values could live in the schema instead of the model initialization code.

justinxzhao (approved):

ludwig/schema/utils.py
```diff
@@ -558,15 +556,15 @@ def IntegerRange(
...
+    default: Union[None, float],
+    allow_none: bool = False,
     description: str = "",
     max: Optional[float] = None,
     parameter_metadata: ParameterMetadata = None,
 ):
     """Returns a dataclass field with marshmallow metadata enforcing numeric inputs must be nonnegative."""
     val = validate.Range(min=0.0, max=max)
-    allow_none = allow_none or default is None
+    # allow_none = allow_none or default is None
```
tgaddair, line 567: Please remove commented code.

ludwig/schema/utils.py
```diff
@@ -605,7 +603,7 @@ def FloatRange(
     """Returns a dataclass field with marshmallow metadata enforcing numeric inputs must be in range set by
     relevant keyword args."""
     val = validate.Range(min=min, max=max, min_inclusive=min_inclusive, max_inclusive=max_inclusive)
-    allow_none = allow_none or default is None
+    # allow_none = allow_none or default is None
```
tgaddair, line 606: Remove.

tests/integration_tests/test_model_training_options.py
```diff
@@ -274,8 +274,9 @@ def test_optimizers(optimizer_type, tmp_path):
     train_losses = train_stats[TRAINING]["combined"]["loss"]
     last_entry = len(train_losses)
 
-    # ensure train loss for last entry is less than or equal to the first entry.
-    assert train_losses[last_entry - 1] <= train_losses[0]
+    # ensure train loss for last entry is less than to the first entry.
+    np.testing.assert_array_less(train_losses[last_entry - 1], train_losses[0])
+    # assert train_losses[last_entry - 1] <= train_losses[0]
```
tgaddair, line 279: Remove.

tgaddair (approved):
LGTM, only nit is to remove the dead code you commented out.

Issue #3078, opened 2023-02-10
Feature preprocessing support


Jeffwan:
**Is your feature request related to a problem? Please describe.**
I checked nunmber features preprocessing page https://ludwig.ai/latest/configuration/features/number_features/  and can not find following features.

- Remove Outliers 
- Remove Collinear Features

**Describe the use case**
Removing outliers increases the accuracy of the model. It is also common topics of feature engineering and should be supported. 

**Describe the solution you'd like**
I expect more options could be exposed in the configuration so user could configure it. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

tgaddair:
Hey @Jeffwan, thanks for these suggestions! I agree with both of them.

For outlier removal, are you imagining that we would essentially treat them as a "missing value" that gets replaced with the mean of the dataset, etc.? Or would you want to drop the entire row from the training data if it has outliers?

w4nderlust:
I think this could be implemented as a preprocessing strategy, like we do for missing values, so both replacing with mean or dropping could be valid options for the user to choose.

An open question is if one wats to do it at the individual feature level or not. For instance, if a dataset has 100 features and only one of these features has values that are considered outliers (for instance outside of the 99% of the probability mass), would we want to drop the datapoint? Or would at least a certain percentage of the features be needed for dropping the datapoint?

w4nderlust:
@jimthompson5802 another potential option as per or previous conversation about what to work on next

tgaddair:
@Jeffwan for the colinearity part, I was imagining this could be done as part of the AutoML / type inference part of Ludwig. Is that what you were thinking as well?

One way we might try doing this would be something like:

1. Calculate [VIF](https://towardsdatascience.com/statistics-in-python-collinearity-and-multicollinearity-4cc4dcd82b3f) score for every feature in the dataset.
2. Remove feature with highest VIF > 10
3. Repeat (1) with remaining features until there are no features with VIF > 10

We can also show the VIF computed as part of the result returned in the `DatasetInfo`.

Does that sound reasonable to you?

tgaddair:
For outliers, I'm imagining the config API could look something like:

```
preprocessing:
  # defaults to null which means "use missing value strategy", can override with any missing value strategy
  outlier_strategy: null  

  # defaults to 3 standard deviations from the mean, can be set to null which means don't replace outliers
  outlier_threshold: 3.0 
```

[...]

PR #3080, opened 2023-02-11
Adds config parameters to replace outliers via a missing_value_strategy

tgaddair:
Related to #3078.

This PR allows users to replace outliers in `number` features similar to how they would replace missing values:

```
preprocessing:
  outlier_strategy: fill_with_mean
  outlier_threshold: 3.0
```

The `outlier_strategy` can be any valid `missing_value_strategy`. By default, it will be `null` which means we will not perform outlier replacement.

`outlier_threshold` is how many standard deviations from the mean we allow input values to be before considering them outliers.


tgaddair:
cc @geoffreyangus to help review torchscript changes.

Jeffwan:
The user interaction part looks good to me.

ludwig/data/preprocessing.py
```diff
@@ -1414,6 +1431,11 @@ def build_data(
         # Need to run this again here as cast_columns may have introduced new missing values
         handle_missing_values(input_cols, feature_config, preprocessing_parameters, backend)
 
+        # For features that support it, we perform outlier removal here using metadata computed on the full dataset
+        handle_outliers(
+            input_cols, feature_config, preprocessing_parameters, training_set_metadata[feature_config[NAME]], backend
+        )
+
```
arnavgarg1, line 1438: QQ - when you say `full dataset`, do you mean each of the splits individually or just the train set?

ludwig/data/preprocessing.py
```diff
@@ -1530,10 +1553,29 @@ def precompute_fill_value(dataset_cols, feature, preprocessing_parameters: Prepr
...
+def handle_outliers(dataset_cols, feature, preprocessing_parameters: PreprocessingConfigDict, metadata, backend):
+    outlier_strategy = preprocessing_parameters.get("outlier_strategy")
+    if outlier_strategy is None:
+        return
+
+    outlier_threshold = preprocessing_parameters["outlier_threshold"]
+    computed_fill_value = preprocessing_parameters.get("computed_outlier_fill_value")
+
+    # Identify all outliers and set them to NA so they can be removed
+    s = dataset_cols[feature[COLUMN]]
```
arnavgarg1, line 1570: nit: can we use a better variable name

ludwig/data/preprocessing.py
arnavgarg1: nvm silly question, it's training set metadata

ludwig/data/preprocessing.py
arnavgarg1: Also, for observability reasons, can we add the number of rows in the column we found/replaced outliters for?

ludwig/data/preprocessing.py
tgaddair: At this point it's always the combined train + val + test dataset. Splitting happens at the end.

ludwig/data/preprocessing.py
tgaddair: `s` is for "series", yes. `column` usually refers to a column name in Pandas terminology. Happy to change it to `series` though.

tgaddair: Problem with printing the number of rows removed is that, because of Dask's lazy evaluation, it will be very slow, since we'll need to recompute some things twice (once to print the count, once to apply the replacement). So I would prefer not to, which is consistent without how we handle missing values at the moment.

ludwig/features/number_feature.py
```diff
@@ -266,12 +314,19 @@ def cast_column(column, backend):
...
             numeric_transformation_registry,
         )
 
-        return numeric_transformer.fit_transform_params(column, backend)
+        params = numeric_transformer.fit_transform_params(column, backend)
+
+        # Ensure mean and std are computed if we're removing outliers
+        outlier_strategy = preprocessing_parameters.get("outlier_strategy")
+        if outlier_strategy is not None and ("mean" not in params or "std" not in params):
+            params.update(ZScoreTransformer.fit_transform_params(column, backend))
```
arnavgarg1, line 327: Nice!

tests/integration_tests/test_missing_value_strategy.py
```diff
@@ -125,3 +125,76 @@ def test_missing_values_drop_rows(csv_filename, tmpdir):
...
+        }
+    )
+
+    dataset_fp = os.path.join(tmpdir, "dataset.csv")
+    dataset_df.to_csv(dataset_fp)
+
+    config = {
+        "input_features": input_features,
+        "output_features": output_features,
+        TRAINER: {"epochs": 2, BATCH_SIZE: 128},
```
arnavgarg1, line 177: nit: Don't really need this since we're just checking preprocessing

arnavgarg1 (approved):

ludwig/data/preprocessing.py
arnavgarg1: That's completely fair - I do think we should figure out a mechanism for better observability - maybe using the verbosity level flag we pass around in Ludwig. Some users may be willing to wait longer for their preprocessing/training so that they can see these kinds of things that happen internally within Ludwig. 

Perfectly ok to skip for now!

tests/integration_tests/test_missing_value_strategy.py
tgaddair: True!

justinxzhao (approved):

PR #3081, opened 2023-02-11
test: tv encoders with large memory footprint

jimthompson5802:
# Code Pull Requests

Add unit test for TorchVision pretrained models that have a large memory footprint.  These large memory models are excluded from the existing unit tests.  This new test accounts for the excluded models and allows running the test on a server with a suitable configuration.

This test has the custom pytest marker `large_memory`


tests/ludwig/encoders/test_image_encoders.py
```diff
@@ -677,3 +679,45 @@ def test_tv_wide_resnet_encoder(
     inputs = torch.rand(2, *pretrained_model.input_shape)
     outputs = pretrained_model(inputs)
     assert outputs["encoder_output"].shape[1:] == pretrained_model.output_shape
+
+
+# test TorchVision encoders that were excluded from the earlier tests due to large memory requirements
+@pytest.mark.large_memory
```
tgaddair, line 685: It might make sense to make use of the more general `@slow` test decorator. We now have a custom action that runs these tests on dedicated hardware following https://github.com/ludwig-ai/ludwig/pull/3029

@geoffreyangus can comment on whether it would be appropriate to use that decorator in this case.

jimthompson5802: converted to @pytest.mark.slow decorator.

jimthompson5802:
@tgaddair   @geoffreyangus   I think it might make sense to pass this PR to someone in Predibase to finish up because it requires running it on a larger server to confirm all the encoders covered will pass.  Right now I don't know how much memory is needed for these encoders.

PR #3082, opened 2023-02-12
Allow providing a Ludwig dataset as a URI of the form `ludwig://<dataset>`

tgaddair:
Example:

```
ludwig train --dataset ludwig://titanic --config config.yaml
```

To use an individual split:

```
ludwig train --training_set ludwig://titanic --test_set ludwig://titanic
```

The latter form will only use the `train` and `test` splits from the dataset and discard the validation set.

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1954,6 +1954,9 @@ def preprocess_for_prediction(
     if isinstance(dataset, Dataset):
         return dataset, training_set_metadata
 
+    # preload ludwig datasets
+    dataset, _, _, _ = load_dataset_uris(dataset, None, None, None, backend)
+
     # determine data format if not provided or auto
     if not data_format or data_format == "auto":
         data_format = figure_data_format(dataset)
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -59,6 +59,7 @@ from ludwig.data.concatenate_datasets import concatenate_df, concatenate_files, 
 from ludwig.data.dataset.base import Dataset
 from ludwig.data.split import get_splitter, split_dataset
 from ludwig.data.utils import set_fixed_split
+from ludwig.datasets import load_dataset_uris
 from ludwig.features.feature_registries import get_base_type_registry
 from ludwig.models.embedder import create_embed_batch_size_evaluator, create_embed_transform_fn
 from ludwig.schema.encoders.utils import get_encoder_cls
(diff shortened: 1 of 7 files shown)


ludwig/datasets/__init__.py
```diff
@@ -80,6 +63,71 @@ def get_dataset(dataset_name, cache_dir=None) -> Any:
...
+            else:
+                test_set_out = _load_cacheable_dataset(test_set, backend)
+
+    return dataset_out, training_set_out, validation_set_out, test_set_out
+
+
+def _load_cacheable_dataset(dataset: str, backend: Backend) -> CacheableDataframe:
+    dataset_name = dataset[len(URI_PREFIX) :]
+    loader = get_dataset(dataset_name)
+    df = loader.load(split=False)
```
arnavgarg1, line 126: In the case that `dataset` is provided, would we want `loader.load(split=True)`? I understand that in the instance that individual sets are passed in this makes sense, but wanted to understand this better

ludwig/datasets/__init__.py
```diff
@@ -80,6 +63,71 @@ def get_dataset(dataset_name, cache_dir=None) -> Any:
...
+            loader = get_dataset(dataset_name)
+            train_df, test_df, val_df = loader.load(split=True)
+            training_set_checksum = str(loader.get_mtime())
+            train_df = backend.df_engine.from_pandas(train_df)
+            training_set_out = CacheableDataframe(df=train_df, name=training_set, checksum=training_set_checksum)
+
+        if isinstance(validation_set, str) and validation_set.startswith(URI_PREFIX):
+            if validation_set == training_set:
+                # Reuse the loaded DF from the training split
+                val_df = backend.df_engine.from_pandas(val_df)
```
arnavgarg1, line 107: I might be missing something here, but won't `val_df` be None since we set it to that on line 93?

arnavgarg1: Oh nvm I see it gets set in line 99

ludwig/data/preprocessing.py
```diff
@@ -1613,6 +1614,11 @@ def preprocess_for_training(
     if dataset is None and training_set is None:
         raise ValueError("No training data is provided!")
 
+    # preload ludwig datasets
+    dataset, training_set, validation_set, test_set = load_dataset_uris(
+        dataset, training_set, validation_set, test_set, backend
+    )
```
arnavgarg1, line 1620: Should we moe this one level above, i.e., at the `LudwigModel` level?

ludwig/datasets/__init__.py
tgaddair: Calling `split=True` will break the DF into 3 individual DFs, so when using the full dataset we want to keep it unsplit (the splitter can handle this downstream).

ludwig/data/preprocessing.py
tgaddair: Interesting thought. Looking at the code, `preprocess_for_training` is called from a few different places, so seems more general to keep it here to avoid being redundant.

ludwig/datasets/__init__.py
arnavgarg1: Ah, got it, makes sense

ludwig/data/preprocessing.py
arnavgarg1: Sure, that seems fair

arnavgarg1 (approved):
Good catch on adding in all of the places where datasets are loaded

PR #3084, opened 2023-02-13
Add Precision Recall curves to Ludwig

arnavgarg1:
Normal PR Curve API
Screen Shot 2023-02-13 at 1 30 14 AM

PR Curve API using test statistics
Screen Shot 2023-02-13 at 2 56 01 AM


w4nderlust (approved):
I like having the two versions, as one could be used to calculate dynamically from live results of an evaluation run and the other one can be used by loading the results of an evaluation run that is part of an experiment

PR #3091, opened 2023-02-14
Raise `ValueError` if no paths are successfully read for a given image feature column

geoffreyangus:


tests/integration_tests/test_preprocessing.py
```diff
@@ -240,14 +240,47 @@ def test_read_image_from_numpy_array(tmpdir, csv_filename):
...
+    }
+
+    data_csv = generate_data(
+        input_features, output_features, os.path.join(tmpdir, csv_filename), num_examples=NUM_EXAMPLES, nan_percent=0.2
+    )
+
+    model = LudwigModel(config)
+
+    with pytest.raises(RuntimeError):
+        model.preprocess(data_csv)
```
arnavgarg1, line 271: Are we guaranteed that the train set will have all NaNs? Wondering if it would be safer to bump up the nan_percent

arnavgarg1: Ah nvm 

```
    def mock_read_binary_files_no_success(self, column, map_fn, file_size):
        """Mock read_binary_files to return None (failed image read) to test error handling."""
        return column.map(lambda x: None)
```

is clever

arnavgarg1: Wondering if we can make do without the monkey patching and just setting nan_percent to something higher. Alternatively, maybe just using made up paths for images which can't be read to simulate this a bit further. 
Lmk what you think!

geoffreyangus: Interesting point– setting `nan_percent` higher wouldn't necessarily work because the missing value strategy is applied before the image read step, so it may end up testing something totally different.

Using made up paths actually throws an earlier, similar error here: https://github.com/ludwig-ai/ludwig/blob/e13df72fa499c878ebef92636886e16ee0af5b83/ludwig/features/image_feature.py#L603-L606

Given the context, it seems like preprocessing for image features will actually almost always fail at that point, rather than the new code path in this PR. Curious what you think about that

tests/integration_tests/test_preprocessing.py
```diff
@@ -240,14 +240,47 @@ def test_read_image_from_numpy_array(tmpdir, csv_filename):
...
+    output_features = [category_feature(decoder={"vocab_size": 5}, reduce_input="sum")]
+
+    config = {
+        "input_features": input_features,
+        "output_features": output_features,
+        TRAINER: {"epochs": 2, BATCH_SIZE: 128},
+    }
+
+    data_csv = generate_data(
+        input_features, output_features, os.path.join(tmpdir, csv_filename), num_examples=NUM_EXAMPLES, nan_percent=0.2
```
arnavgarg1, line 265: Do we need `nan_percent=0.2`?

tests/integration_tests/test_preprocessing.py
arnavgarg1: Gotcha, I didn't realize the order of steps in preprocessing images.

Let's keep it this way for now!

arnavgarg1 (approved):
Nice work, especially on the test :)

arnavgarg1:
@geoffreyangus should we land this?

PR #3095, opened 2023-02-14
Use proc col hash for checksum computation

arnavgarg1:
This PR makes two fundamental changes:

1. `compute_feature_hash`, i.e., the function used to compute the hash for each feature that we refer to in `proc_column` is now computed over a dict of preprocessing + feature type, rather than just preprocessing parameters.
2. `calculate_checksum`, i.e., the function that is used to check if two configs map to the same cached dataset now uses a set of the proc column values for a config rather than a list of feature names. This is because it ensures that we only use the same cache if the feature types, feature names, and feature preprocessing values are exactly what they should be. Any deviations from this will (and should) cause a cache miss.

### DIFF
--- a/ludwig/data/cache/util.py
+++ b/ludwig/data/cache/util.py
@@ -6,12 +6,11 @@ from ludwig.utils.data_utils import hash_dict
 
 
 def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDict):
-    """
-    Calculates a checksum for a dataset and model config.
+    """Calculates a checksum for a dataset and model config.
 
-    The checksum is used to determine if the dataset and model config have changed since the last time the
-    model was trained. If either has changed, a different checksum will be produced which will lead to a cache
-    miss and force preprocessing to be performed again.
+    The checksum is used to determine if the dataset and model config have changed since the last time the model was
+    trained. If either has changed, a different checksum will be produced which will lead to a cache miss and force
+    preprocessing to be performed again.
     """
     features = config.get(INPUT_FEATURES, []) + config.get(OUTPUT_FEATURES, []) + config.get("features", [])
     info = {
@@ -21,7 +20,7 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "global_defaults": config.get(DEFAULTS, {}),
         # PROC_COLUMN contains both the feature name and the feature hash that is computed
         # based on each feature's preprocessing parameters and the feature's type.
-        "feature_proc_columns": set([feature[PROC_COLUMN] for feature in features]),
+        "feature_proc_columns": {feature[PROC_COLUMN] for feature in features},
         "feature_types": [feature[TYPE] for feature in features],
         "feature_preprocessing": [feature.get(PREPROCESSING, {}) for feature in features],
     }
--- a/ludwig/data/cache/util.py
+++ b/ludwig/data/cache/util.py
@@ -1,18 +1,27 @@
 import ludwig
-from ludwig.constants import DEFAULTS, INPUT_FEATURES, NAME, OUTPUT_FEATURES, PREPROCESSING, TYPE
+from ludwig.constants import DEFAULTS, INPUT_FEATURES, OUTPUT_FEATURES, PREPROCESSING, PROC_COLUMN, TYPE
 from ludwig.data.cache.types import CacheableDataset
 from ludwig.types import ModelConfigDict
 from ludwig.utils.data_utils import hash_dict
 
 
 def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDict):
+    """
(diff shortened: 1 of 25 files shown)


ludwig/features/feature_utils.py
```diff
@@ -100,23 +100,19 @@ def sanitize(name):
 
 
 def compute_feature_hash(feature: dict) -> str:
-    """
-    This function computes a hash for each feature based on the preprocessing dictionary associated with each feature.
-    The input is always the feature dict, however sometimes, this is called from BaseFeature which dumps the feature
-    dict from a ConfigObject and the preprocessing key corresponds to a nested Preprocessing config. This is why it
-    includes the if/else statement.
+    """This function computes a hash for each feature based on the preprocessing dictionary associated with each
```
tgaddair, line 103: This function is only called once after the `ModelConfig` is constructed. As such, instead of calling `feature.to_dict()` you can pass the feature object directly in, which makes it more type safe and avoids the unnecessary safety check when getting the PREPROCESSING section (with the ModelConfig, the preprocessing section is always populated).

ludwig/features/feature_utils.py
```diff
@@ -100,23 +100,19 @@ def sanitize(name):
...
     Returns: Feature hash name
-
     """
-    preproc = feature.get(PREPROCESSING, {})
-    if isinstance(preproc, dict):
-        preproc_hash = hash_dict(preproc)
-    else:
-        preproc_hash = hash_dict(preproc.to_dict())
-    return sanitize(feature[NAME]) + "_" + preproc_hash.decode("ascii")
+    feature_data = dict(
```
tgaddair, line 111: I would add some simple unit tests to check that features with identical preprocessing but different types have different proc cols.

tgaddair (commented):
Minor nits.

ludwig/features/feature_utils.py
arnavgarg1: I started making this change, but realized that our tests have a deep relationship with `compute_feature_hash` and are super sensitive to this. Too many test failures downstream locally, so I think we need to refactor our tests to not use dicts.

For e.g., even something as simple as `text_feature()` and `number_feature()` initialization in our tests compute the proc hashes in their implementations, and getting rid of them causes other errors because many of our tests directly use/reference the proc hash columns as dicts rather than attributes.

I'll skip this for now, but maybe this is something we should also follow up with.

ludwig/features/feature_utils.py
arnavgarg1: Agreed, going to add tests.

ludwig/data/cache/util.py
```diff
@@ -1,18 +1,26 @@
...
     features = config.get(INPUT_FEATURES, []) + config.get(OUTPUT_FEATURES, []) + config.get("features", [])
     info = {
         "ludwig_version": ludwig.globals.LUDWIG_VERSION,
         "dataset_checksum": original_dataset.checksum,
         "global_preprocessing": config.get(PREPROCESSING, {}),
         "global_defaults": config.get(DEFAULTS, {}),
-        "feature_names": [feature[NAME] for feature in features],
+        # PROC_COLUMN contains both the feature name and the feature hash that is computed
+        # based on each feature's preprocessing parameters and the feature's type.
+        "feature_proc_columns": {feature[PROC_COLUMN] for feature in features},
```
arnavgarg1, line 23: This is such a fundamental change because PROC_COLUMN is a combination of `feature_name`_`hash of dict of type + preprocessing parameters`

So a change to either the feature's name, type or preprocessing parameters will cause a change in name to both proc col and well as compute checksum. 

The second thing to note is that we're more or less guaranteed different preprocessing hashes for two different features anyway since the preprocessing for category vs binary (internal representation after all internal parameters are merged) are completely different!

ludwig/features/feature_utils.py
tgaddair: Ah, I see! Yes, this is a bigger refactor we'll need to make.

tgaddair (approved):

tests/ludwig/data/test_cache_util.py
```diff
@@ -92,3 +92,20 @@ def test_proc_col_checksum_consistency():
     for of1, of2 in zip(config1.output_features, config2.output_features):
         assert of1.name == of2.name
         assert of1.proc_column == of2.proc_column
+
+
+def test_proc_col_checksum_consistency_same_preprocessing_different_types():
+    """Tests that proc_col is equal if checksum are equal."""
```
tgaddair, line 98: Nit: this tests that proc_cols are different when types are different.

arnavgarg1: Ahh co-pilot 🤦  Just fixed.

justinxzhao (approved):

PR #3096, opened 2023-02-14
Fixed serialization and deserialization of augmentation configuration

jimthompson5802:
# Code Pull Requests

Add test case for saving and loading a model with augmentation

### DIFF
--- a/ludwig/schema/features/augmentation/utils.py
+++ b/ludwig/schema/features/augmentation/utils.py
@@ -43,22 +43,30 @@ def get_augmentation_classes(feature: str):
 
 
 @DeveloperAPI
-def AugmentationDataclassField(feature_type: str, default=[], description=""):
+def AugmentationDataclassField(feature_type: str, default=False, default_augmentations=None, description=""):
     """Custom dataclass field that when used inside a dataclass will allow the user to specify an augmentation
     config.
 
     Args:
         default: The default augmentation config to use.
+        default_augmentations: The default list of augmentations to use when param value is set to `True`.
         description: The description of the augmentation config.
 
     Returns: Initialized dataclass field that converts a list with params to an augmentation config.
     """
+
+    default_augmentations = default_augmentations or []
+    if isinstance(default, bool):
+        default = default_augmentations if default else []
 
     class AugmentationContainerMarshmallowField(fields.Field):
         """Custom marshmallow field that deserializes a list for a valid augmentation config from the
         augmentation_registry and creates a corresponding JSON schema for external usage."""
 
         def _deserialize(self, value, attr, data, **kwargs):
+            if isinstance(value, bool):
+                value = default_augmentations if value else []
+
             assert isinstance(value, list), "Augmentation config must be a list."
             if len(value) == 0:
                 raise ValidationError("Augmentation config list must not be empty.")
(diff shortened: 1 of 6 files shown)


jimthompson5802:
@tgaddair I added the fix for bypassing the execution of augmentation.  This was switching the order of superclasses for `ImageInputFeatureConfig` class.  Local testing confirms that augmentations are now executed.

I also confirmed with local testing that the `model.load()` schema validation issue is resolved.

One thing I noticed was the removal of the empty list test case for raising `ValidationErrro`.  @ksbrar thought testing for this situation was useful.

From my perspective, this PR fixes the blocking issues with schema validation and augmentation operation.  LGTM.

tests/ludwig/schema/test_model_config.py
```diff
@@ -634,3 +636,50 @@ def test_number_feature_zscore_preprocessing_default():
     config_obj = ModelConfig.from_dict(config)
 
     assert config_obj.input_features.number_input_feature1.preprocessing.normalization == "zscore"
+
+
+@pytest.mark.parametrize(
+    "augmentation,expected",
+    [
+        (None, []),
```
ksbrar, line 644: @tgaddair This is fine, though are you sure we wanna support `None` as an option?

tests/ludwig/schema/test_model_config.py
```diff
@@ -634,3 +636,50 @@ def test_number_feature_zscore_preprocessing_default():
...
+        (False, []),
+        (True, AUGMENTATION_DEFAULT_OPERATIONS),
+        (
+            [{"type": "random_blur"}, {"type": "random_rotate", "degree": 30}],
+            [RandomBlurConfig(), RandomRotateConfig(degree=30)],
```
ksbrar, line 649: Slick tests.

ludwig/schema/features/augmentation/utils.py
```diff
@@ -112,30 +134,37 @@ def _jsonschema_type_mapping():
...
-        "minItems": 1,
-        "items": {
-            "type": "object",
-            "properties": {
-                "type": {
-                    "type": "string",
-                    "enum": augmentation_types,
-                    "title": "type",
-                    "description": "Type of augmentation to apply.",
+        "oneOf": [
```
ksbrar, line 144: So this is the primary change. It matches our other patterns, but ideally we shouldn't need to do this - the `OneOfOptionsField` should handle it, so perhaps we should take another look at that field and if its having issues.

(Imagine a contributor defining a new field type and then wanting to combine it with an existing field type. It would suck to have to then redefine your new field or otherwise define another field just to get the switching to work.)

tests/ludwig/schema/test_model_config.py
tgaddair: This is a bit misleading. If you look further down, you'll see that `None` here actually means "don't include in the config".

ksbrar: ah, my bad glanced over that `del`. thanks!

ludwig/schema/features/augmentation/utils.py
ksbrar: I see, thanks for the explanation.

I think that is a tenable problem we can solve by introducing meta-attributes for rendering complex fields. But worth a separate PR!

ksbrar (approved):

ludwig/schema/features/augmentation/utils.py
tgaddair: Yes, I definitely want to find a better, more general solution!

PR #3100, opened 2023-02-14
Deflake the lbgfs optimizer test

justinxzhao:
[Example of flakiness](https://github.com/ludwig-ai/ludwig/actions/runs/4177152055/jobs/7234368909)

Note that I'm not able to reproduce the `nan <= nan` error locally with --count=1000, but empirically the loss values that come from the lower LR seem more tractable. Hopefully the lower LR helps with flakiness.

I'm also not opposed to removing the test altogether for the lbfgs optimizer as we're still testing that we're functionally registering and referencing the torch implementation correctly, which is covered by tests for other optimizers.


arnavgarg1 (approved):
Looks good to me!

tgaddair (approved):

PR #3101, opened 2023-02-15
Add -1 as a valid negative class for binary type inference

arnavgarg1:
While playing around with automl using the `ivory` dataset in the dataset zoo, I noticed that automl infers the `label` to be `number` since the labels are [-1, 1]. However, this is actually a classification problem, and this should be inferred as binary.

This PR updates the conventional bools to also allow -1 to be a valid negative class so we correctly infer the type as binary.

### DIFF
--- a/ludwig/utils/strings_utils.py
+++ b/ludwig/utils/strings_utils.py
@@ -34,7 +34,7 @@ PANDAS_TRUE_STRS = {"true"}
 PANDAS_FALSE_STRS = {"false"}
 
 BOOL_TRUE_STRS = {"yes", "y", "true", "t", "1", "1.0"}
-BOOL_FALSE_STRS = {"no", "n", "false", "f", "0", "0.0", "-1"}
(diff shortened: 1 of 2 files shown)


abidwael (approved):
Nice!

justinxzhao (approved):

ludwig/utils/strings_utils.py
```diff
@@ -34,7 +34,7 @@
 PANDAS_FALSE_STRS = {"false"}
 
 BOOL_TRUE_STRS = {"yes", "y", "true", "t", "1", "1.0"}
-BOOL_FALSE_STRS = {"no", "n", "false", "f", "0", "0.0"}
+BOOL_FALSE_STRS = {"no", "n", "false", "f", "0", "0.0", "-1"}
```
tgaddair, line 37: Should we also throw in `-1.0` for consistency with 0 and 1?

tgaddair (approved):

ludwig/utils/strings_utils.py
arnavgarg1: Added

PR #3104, opened 2023-02-15
Fold all validation into ModelConfig.

justinxzhao:
API for fully rendered and validated configs.

```python
comprehensive_and_validated_config = ModelConfig.from_dict(user_config)
```

Config initialization breakdown for what happens when `ModelConfig.from_dict(user_config)` is called:

    1. Upgrade to latest version (syntactical changes, based on the `config_transformation_registry`, i.e. `numerical` → `number`)
    2. Check basic required parameters
        - All features have name and type and they are valid.
        - All types are valid (encoders, decoders, optimizer)
        - Trainer is not None
        - (idea) Most of this can go away if we set `additionalProperties: False` for JSON schemas everywhere.
    3. Merge with defaults
        - Applies the config’s `defaults` section (type-specific default parameters) to all features with unspecified parameters.
    5. Update preprocessing parameters based on selected encoders.
    6. Set the backend type. (Can go away when we schemify `backend`)
    7. JSON schema validation (draft7)
    8. JSON deserialization into a type object (based on the model type).
    9. (post-init) Set validation parameters
    10. (post-init) Set hyperopt defaults
    11. (post-init) Set derived feature columns (hashes for feature cache)
    12. (post-init) Auxiliary checks.


ludwig/schema/model_types/base.py
```diff
@@ -88,12 +91,17 @@ def from_dict(config: ModelConfigDict) -> "ModelConfig":
...
+        check_schema(config)
+
         cls = model_type_schema_registry[model_type]
         schema = cls.get_class_schema()()
         try:
             config_obj: ModelConfig = schema.load(config)
         except ValidationError as e:
             raise ConfigValidationError(f"Config validation error raised during config deserialization: {e}") from e
+
+        get_config_check_registry().check_config(config_obj)
```
tgaddair, line 104: Please move this to `__post_init__` above.

ludwig/schema/model_types/base.py
```diff
@@ -88,12 +91,17 @@ def from_dict(config: ModelConfigDict) -> "ModelConfig":
         if isinstance(backend, str):
             config[BACKEND] = {"type": backend}
 
+        # JSON schema validation.
+        check_schema(config)
```
tgaddair, line 95: May no longer be necessary to call this when loading the `ModelConfig` object.

tests/ludwig/config_validation/test_validate_config_misc.py
```diff
@@ -105,7 +100,7 @@ def test_config_features():
         }
 
         with pytest.raises(ConfigValidationError):
-            validate_config(config)
+            ModelConfig.from_dict(config)
```
tgaddair, line 103: Something important I want to call out here: the `validate_config` (JSONSchema validation) and `ModelConfig.from_dict` serve different purposes in practice.

It's important that we test that `validate_config` works independent of any non-JSONSchema validation checks, because this tests that the JSONSchema validation itself is working correctly.

Ideally, the JSONSchema and `ModelConfig` checks behave identically (though the ModelConfig checks should be faster since it doesn't rely on compiling the JSONSchema, in theory). In practice, ModelConfig checks will be more extensive, and that's okay. But we shouldn't be losing coverage of the JSONSchema checks, because they are useful in cases where the JSONSchema is all we have.

Does that make sense?

So what I would like to see is either keeping these `validate_config` checks in place to verify correctness in the JSONSchema, or we add a new set of tests that check parity between JSONSchema and ModelConfig checks.

ludwig/schema/model_types/base.py
justinxzhao: Done.

ludwig/schema/model_types/base.py
justinxzhao: Marshmallow deserialization permits additional properties by default, while JSON schema validation doesn't when `additionalProperties=False`, which is the case for at least some of our schema (e.g `trainer`). I added a comment in the code to give a bit more context. 

That said, if we are able to set `additionalProperties=False` for all Ludwig schema, which seemed desirable based on an offline discussion with @ksbrar (cc: @connor-mccorm), then it might be possible to get to behavior parity marshmallow deserialization by using `schema.load(unknown='RAISE')`.

tests/ludwig/config_validation/test_validate_config_misc.py
justinxzhao: That makes sense.

It sounds like we want to have separate tests, 1) specifically for the behavior of `check_schema` (that, notably, can be used in an environment without marshmallow) and 2) for any additional behavior that `ModelConfig.from_dict` would check.

We aspire to have 1) and 2) to have perfect parity, which is true for the most part with the notable exception of checks for interdependent parameters.

tgaddair (approved):
LGTM!

PR #3066, opened 2023-02-09
Update transformer `hidden_size / num_heads` error message

jeffkinnison:
A constraint of the `MultiHeadSelfAttention` module is that `hidden_size` must be divisible by `num_heads`. The error message was somewhat unclear on how to address a mismatch, so this updates the message with more concrete instructions.

PR #3105, opened 2023-02-15
Transformer divisibility error validation

jeffkinnison:
Followup to #3066 that adds a validation check for `encoder.hidden_size % encoder.num_heads == 0` for transformer encoders.

### DIFF
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -354,13 +354,6 @@ def check_stacked_transformer_requirements(config: "ModelConfig") -> None:  # no
         return hidden_size % num_heads == 0
 
     sequence_types = [SEQUENCE, TEXT, TIMESERIES]
-    for sequence_type in sequence_types:
-        encoder = config.defaults.__getattribute__(sequence_type).encoder
-        if encoder.type == "transformer" and not is_divisible(encoder.hidden_size, encoder.num_heads):
-            raise ConfigValidationError(
-                f"Default {sequence_type} transformer encoder requires encoder.hidden_size to be divisible by "
-                f"encoder.num_heads. Found hidden_size {encoder.hidden_size} and num_heads {encoder.num_heads}."
-            )
 
     for input_feature in config.input_features:
         if_type = input_feature.type
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -23,6 +23,7 @@ from ludwig.constants import (
     SEQUENCE,
     SET,
     TEXT,
+    TIMESERIES,
     TRAINER,
     TYPE,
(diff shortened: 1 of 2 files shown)


ludwig/config_validation/checks.py
```diff
@@ -339,3 +340,34 @@ def check_hf_encoder_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+        return hidden_size % num_heads == 0
+
+    sequence_types = [SEQUENCE, TEXT, TIMESERIES]
+    for sequence_type in sequence_types:
+        encoder = config.defaults.__getattribute__(sequence_type).encoder
+        if encoder.type == "transformer" and not is_divisible(encoder.hidden_size, encoder.num_heads):
+            raise ConfigValidationError(
+                f"Default {sequence_type} transformer encoder requires encoder.hidden_size to be divisible by "
+                f"encoder.num_heads. Found hidden_size {encoder.hidden_size} and num_heads {encoder.num_heads}."
+            )
```
justinxzhao, line 360: I think we might be able to get away without checking `config.defaults` as all of these auxiliary checks are run after `defaults` have been resolved into feature configs. WDYT?

tests/ludwig/config_validation/test_validate_config_encoder.py
```diff
@@ -0,0 +1,58 @@
...
+    the values from the original error.
+    """
+    config = {
+        INPUT_FEATURES: [number_feature(), sequence_feature(), text_feature(), timeseries_feature()],
+        OUTPUT_FEATURES: [binary_feature()],
+        DEFAULTS: {feature_type: {ENCODER: {TYPE: "transformer", "hidden_size": 9, "num_heads": 18}}},
+    }
+
+    with pytest.raises(ConfigValidationError):
+        validate_config(config)
```
justinxzhao, line 29: Use `ModelConfig.from_dict(config)` instead of `validate_config`, which I'm planning on removing in https://github.com/ludwig-ai/ludwig/pull/3104

ludwig/config_validation/checks.py
jeffkinnison: Oh, cool! I'll remove the defaults block.

tests/ludwig/config_validation/test_validate_config_encoder.py
jeffkinnison: So all of the tests fail when I use `ModelConfig.from_dict`. Do we need to wait for the PR to land before updating?

justinxzhao: @jeffkinnison you're totally right! Let me try to land that PR first.

justinxzhao (approved):

PR #3115, opened 2023-02-17
Update benchmark tests

abidwael:
- Removes `pytest_benchmarks.yaml` and includes the quality regression benchmarks as part of the main `pytest.yaml` workflow. (they take an additional 1.5-2min)
- Improvements to the `ludwig.benchmarking` module for better error surfacing.


.github/workflows/pytest_benchmarks.yml
```diff
@@ -3,9 +3,14 @@
 
 name: benchmark
 
+#on:
+#  push:
+#    branches: ["master", "release-*"]
 on:
   push:
     branches: ["master", "release-*"]
+  pull_request:
```
justinxzhao, line 12: Now that you've verified that the benchmark tests work with your change, is the plan to revert these files before merging the PR?

abidwael: The plan is to add them to the main test suite (`pytest.yml`) and remove `pytest_benchmarks.yml`

tests/regression_tests/benchmark/configs/sarcos.gbm.yaml
```diff
@@ -107,5 +107,3 @@ trainer:
   num_boost_round: 382
   num_leaves: 434
 model_type: gbm
-backend:
-  type: local
```
arnavgarg1, line 111: Is there a reason we're removing backend as local? Should we potentially consider setting it to Ray?

tgaddair (approved):

justinxzhao (approved):

tests/regression_tests/benchmark/configs/sarcos.gbm.yaml
abidwael: I believe default is ray

PR #3118, opened 2023-02-18
feat: Raise deprecation warnings for unknown parameters

ksbrar:
### Description
Uses the `marshmallow.pre_load` decorator to take in the config sent to a particular schema and raise a `DeprecationWarning` for any unknown parameters.

Additional changes:
* Sets `unknown = INCLUDE` on `BaseMarshmallowConfig`. This is necessary so that all parameters are present when it is time to raise the deprecation warning
* Removes phantom `type` from `DefaultsDataclassField` marshmallow defaults. This is unnecessary because no defaults schema has a type field - the right default class is instead chosen at compile time.

### Example

Output in `pytest` with the following test:

```
pytest -vs tests/ludwig/config_validation/test_validate_config_misc.py -k "test_deprecation"  
```
```
================================================================================================================================================================= warnings summary ==================================================================================================================================================================
tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
  /home/ksbrar/work/ludwig/ludwig/utils/backward_compatibility.py:425: DeprecationWarning: encoder specific parameters should now be nested within a dictionary under the 'encoder' parameter. Support for un-nested encoder specific parameters will be removed in v0.7
    warnings.warn(

tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
  /home/ksbrar/work/ludwig/ludwig/schema/utils.py:182: DeprecationWarning: "vocab_size" is not a valid parameter for the "CategoricalEmbedConfig" schema, will be flagged as an error in v0.8
    warnings.warn(

tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
  /home/ksbrar/work/ludwig/ludwig/schema/utils.py:182: DeprecationWarning: "reduce_input" is not a valid parameter for the "CategoricalEmbedConfig" schema, will be flagged as an error in v0.8
    warnings.warn(

tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
  /home/ksbrar/work/ludwig/ludwig/schema/utils.py:182: DeprecationWarning: "unknown_parameter_combiner" is not a valid parameter for the "TabNetCombinerConfig" schema, will be flagged as an error in v0.8
    warnings.warn(

tests/ludwig/config_validation/test_validate_config_misc.py::test_deprecation_warning_raised_for_unknown_parameters
  /home/ksbrar/work/ludwig/ludwig/schema/utils.py:182: DeprecationWarning: "unknown_parameter_trainer" is not a valid parameter for the "ECDTrainerConfig" schema, will be flagged as an error in v0.8
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```
Code_Cfgj4M2jtA

### Further Notes

A brief summary of `marshmallow` order of operations when it comes to deserialization, as observed while writing this PR:
1. `unknown` flag will filter or raise errors for the given config first (this is why I had to manually set `unknown = INCLUDE`).
2. `pre_load decorator`
3. `<schema>._deserialize`

...

One unexpected hiccup was that there are certain fields that are only added dynamically and not declared ahead of time. For example, `type` is not present on all schemas so I avoid throwing the warning in that case. But I was surprised to see that a parameter called `feature_type` also frequently appears on configs themselves (rather than later at deserialization time). I did the same for it.

From the one test above there are also these three:
* `bn_virtual_divider` - is this expected for tabnet? 
* `reduce_input` - is this expected for categorical embeddings?
* `vocab_size` -  is this expected for categorical embeddings?

From a cursory search it looks like all three of these parameters indeed have been removed from the respective complaining schemas, so my guess is we might have a few tests that themselves have outdated parameters.

This spawned two thoughts:
* Until we remove all of them, many tests will throw new deprecation warnings (though currently we are filtering these warnings out via our `pytest.ini` configuration). When we switch over to explicitly rejecting unknown parameters then we will introduce unexpected test failures if we haven't gotten them all by then.
* The internal parameter refactoring that is still planned and would remove them from the schemas IIRC must be implemented carefully. They should be kept totally separate from the configs or otherwise only loaded/add _after_ deserialization - otherwise they will fail against the stricter validation. (OTOH if we introduce a `internal_only_registry` of sorts, then we might be able to tweak marshmallow to manually allow those exceptions, but doing so would overcomplicate the matter IMO.)


tests/ludwig/config_validation/test_validate_config_misc.py
```diff
@@ -417,3 +418,21 @@ def test_decoder_descriptions():
...
+        ],
+        "output_features": [binary_feature()],
+        "combiner": {
+            "type": "tabnet",
+            "unknown_parameter_combiner": False,
+        },
+        TRAINER: {
+            "epochs": 1000,
+        },
+    }
```
connor-mccorm, line 437: Is there any pytest util that can check for a warning raised similar to `with pytest.raises`

connor-mccorm (approved):

tests/ludwig/config_validation/test_validate_config_misc.py
ksbrar: yes there is! forgot to add that 👍🏽

tgaddair (approved):
Very clean!

ludwig/config_validation/validation.py
```diff
@@ -28,7 +28,7 @@ def get_ludwig_version_jsonschema():
     }
 
 
-def get_backend_jsonschema():
+def get_backend_jsonschema():  # TODO: Is this still necessary after setting backend = Dict()?
```
ksbrar, line 31: @tgaddair

PR #3119, opened 2023-02-18
Skip batch norm when chunk size is 1 in GhostBatchNorm

arnavgarg1:
Causes the following error otherwise:

```
ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 6])
```

### DIFF
--- a/ludwig/modules/normalization_modules.py
+++ b/ludwig/modules/normalization_modules.py
@@ -25,8 +25,8 @@ class GhostBatchNormalization(LudwigModule):
             if batch_size == 1:
                 # Skip batch normalization if the batch size is 1.
                 return torch.cat(splits, 0)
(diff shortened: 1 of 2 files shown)


w4nderlust (approved):

tgaddair (approved):
Nice! Only other thing I might consider is logging a warning for each of the two cases where we skip the norm:

- `batch_size <= virtual_batch_size`
- `batch_size % virtual_batch_size == 1`

justinxzhao (approved):

arnavgarg1:
> Nice! Only other thing I might consider is logging a warning for each of the two cases where we skip the norm:
> 
> * `batch_size <= virtual_batch_size`
> * `batch_size % virtual_batch_size == 1`

Added! Let me know if this makes sense

PR #3123, opened 2023-02-20
Update field descriptions for ludwig-docs

tgaddair:


ludwig/schema/metadata/configs/common.yaml
```diff
@@ -0,0 +1,210 @@
...
+    It's often suggested that with normalization, you can use a higher learning
+    rate.
+  example_value:
+    - batch
+  expected_impact: 3
+  literature_references:
+    - https://machinelearningmastery.com/batch-normalization-for-training-of-deep-neural-networks/
+  related_parameters:
+    - norm_params
+  suggested_values: '"batch" or "layer"'
```
connor-mccorm, line 106: Don't know if we should remove these necessarily, but I've always thought it a little unhelpful that the suggested values here are just both of the available values. I get that they are better than None, but I feel like from 85% of user's view they will see this and think that it's not very helpful since it's essentially suggesting that if the user is going to use this parameter they should use it. There are a couple other instances similar to this one throughout the metadata so do you think it would make sense to remove the suggested value here or just leave it in and gloss over it?

connor-mccorm (approved):
This is an awesome enhancement. Not only does this make things more standardized for the repeated parameters, but the new info you've added is very insightful and helpful. I just added in a few thoughts I had while reading through the new metadata but overall love it. Thank you!

ludwig/schema/combiners/tabnet.py
```diff
@@ -22,7 +22,9 @@ def module_name():
     )
 
     size: int = schema_utils.PositiveInteger(
-        default=32, description="`N_a` in the paper.", parameter_metadata=COMBINER_METADATA["TabNetCombiner"]["size"]
+        default=32,
+        description="Size of the hidden layers. `N_a` in the paper.",
```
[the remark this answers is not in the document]
tgaddair: Good point, it's mostly useful for the ludwig docs at the moment, which uses markdown.

ludwig/schema/common_fields.py
```diff
@@ -0,0 +1,123 @@
...
+from ludwig.schema import utils as schema_utils
+from ludwig.schema.metadata import COMMON_METADATA
+from ludwig.schema.metadata.parameter_metadata import ParameterMetadata
+from ludwig.utils.torch_utils import initializer_registry
+
+
+def DropoutField(default: float = 0.0, description: str = None, parameter_metadata: ParameterMetadata = None) -> Field:
+    description = description or (
+        "Default dropout rate applied to fully connected layers. "
+        "Increasing dropout is a common form of regularization to combat overfitting."
```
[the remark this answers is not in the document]
tgaddair: Very good callout. Will put this in the description.

ludwig/schema/metadata/configs/common.yaml
```diff
@@ -0,0 +1,210 @@
+activation:
+  default_value_reasoning: The Rectified Linear Units (ReLU) function is the
+    standard activation function used for adding non-linearity. It is simple,
+    fast, and empirically works well (https://arxiv.org/abs/1803.08375).
+  description_implications: Changing the activation functions has an impact
+    on the computational load of the model and might require further hypterparameter
+    tuning
+  expected_impact: 2
+  suggested_values: The default value will work well in the majority of the
```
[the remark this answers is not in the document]
tgaddair: Good call, done.

ludwig/schema/metadata/configs/common.yaml
tgaddair: Agreed, will remove.

tgaddair:
Thanks for the great comments, @connor-mccorm!

ludwig/schema/common_fields.py
```diff
@@ -0,0 +1,124 @@
...
+        "is missing from the dictionary, the default one provided as a standalone parameter will be used instead."
+    )
+    parameter_metadata = parameter_metadata or COMMON_METADATA["fc_layers"]
+    return schema_utils.DictList(
+        description=description,
+        parameter_metadata=parameter_metadata,
+    )
+
+
+INITIALIZER_SUFFIX = """
```
ksbrar, line 94: Note that this won't be possible after https://github.com/ludwig-ai/ludwig/pull/3075

ludwig/schema/combiners/comparator.py
```diff
@@ -12,6 +14,13 @@
 class ComparatorCombinerConfig(BaseCombinerConfig):
     """Parameters for comparator combiner."""
 
+    def __post_init__(self):
```
ksbrar, line 17: Perhaps better placed in the `config_validation.checks` suite. Or we should decentralize that suite when you have schema-specific requirements (e.g. see `check_sequence_concat_combiner_requirements`).

ludwig/schema/combiners/tabnet.py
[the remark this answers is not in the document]
ksbrar: +1

ludwig/schema/combiners/tabnet.py
```diff
@@ -35,34 +37,34 @@ def module_name():
...
-        description="Output size of a fully connected layer. `N_d` in the paper",
+        description="Output size of a fully connected layer. `N_d` in the paper.",
         parameter_metadata=COMBINER_METADATA["TabNetCombiner"]["output_size"],
     )
 
     num_steps: int = schema_utils.NonNegativeInteger(
         default=3,
         description="Number of steps / repetitions of the the attentive transformer and feature transformer "
-        "computations. `N_steps` in the paper ",
+        "computations. `N_steps` in the paper.",
```
ksbrar, line 47: Nit: I know it's a seminal paper but perhaps we should say `(Arik and Pfister, 2019)` instead of literally `the paper`.

ludwig/schema/common_fields.py
```diff
@@ -0,0 +1,124 @@
...
+
+def ResidualField(
+    default: bool = False, description: str = None, parameter_metadata: ParameterMetadata = None
+) -> Field:
+    description = description or (
+        "Whether to add a residual connection to each fully connected layer block. "
+        "Requires all fully connected layers to have the same `output_size`."
+    )
+    parameter_metadata = parameter_metadata or COMMON_METADATA["residual"]
+    return schema_utils.Boolean(
```
ksbrar, line 34: idk if this is that much more readable/efficient but since we're passing all local variables you could say e.g. `Boolean(**locals())`?

ludwig/schema/optimizers.py
```diff
@@ -502,15 +502,25 @@ class GradientClippingConfig(schema_utils.BaseMarshmallowConfig):
     """Dataclass that holds gradient clipping parameters."""
```
ksbrar, line 502: Would you mind removing `desciption = TODO` from `OptimizerDataclassField` above? :D

ksbrar (approved):
👌🏽

ludwig/schema/common_fields.py
tgaddair: Yes, true, we will need to refactor this.

ludwig/schema/combiners/comparator.py
tgaddair: I originally had it there, but it was very ugly and hacky. I actually much prefer having it here, to better encapsulate the validation logic. In the future, we can look to remove this by combining `num_fc_layers` and `fc_layers` into a single `oneOf` option.

ludwig/schema/common_fields.py
tgaddair: That's true, but a little dangerous if someone ever adds a new local var.

ludwig/schema/combiners/tabnet.py
tgaddair: Good point, I agree that's. good change.

PR #3131, opened 2023-02-21
Fix tests to be compliant with latest version of whylogs

justinxzhao:


geoffreyangus (approved):

geoffreyangus:
Can you add a description as to why these changes were necessary (for posterity)? I don't fully understand why the test needed to be updated, for example.

PR #3132, opened 2023-02-22
[Explain] always return global and row-level explanations

jppgks:
- `Explainer` API now returns both global and row-level explanations via the new `ExplanationsResult` dataclass
- Stores aggregated attributions in `global_explanation`. Token attributions are averaged and the top 100 tokens are included in the `global_explanation`.

### DIFF
--- a/ludwig/explain/captum.py
+++ b/ludwig/explain/captum.py
@@ -20,7 +20,7 @@ from ludwig.api_annotations import PublicAPI
 from ludwig.constants import CATEGORY, DATE, IMAGE, INPUT_FEATURES, NAME, PREPROCESSING, TEXT, UNKNOWN_SYMBOL
 from ludwig.data.preprocessing import preprocess_for_prediction
 from ludwig.explain.explainer import Explainer
-from ludwig.explain.explanation import Explanation
+from ludwig.explain.explanation import ExplanationsResult
 from ludwig.explain.util import get_pred_col
 from ludwig.features.feature_utils import LudwigFeatureDict
 from ludwig.models.ecd import ECD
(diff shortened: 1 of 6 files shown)


justinxzhao:
Broken test now fixed by #3131. Please rebase!

ludwig/explain/captum_ray.py
```diff
@@ -91,21 +93,57 @@ def explain(self) -> Tuple[List[Explanation], List[float]]:
         expected_values = []
         for attrs_ref in tqdm(attrs_refs, desc="Explain"):
             attrs = ray.get(attrs_ref)
-            for (total_attribution, feat_to_token_attribution) in attrs:
-                for i, (feature_attributions, explanation) in enumerate(zip(total_attribution, self.explanations)):
+            for total_attribution, feat_to_token_attributions, total_attribution_global in attrs:
```
connor-mccorm, line 96: Not sure if there's any way getting around these nested for loops but I feel like this might blow up the explanation time by a bit right? If so, do we maybe want to compute the explanations on a sample of the dataset - or is that already happening - I can't remember

ludwig/explain/captum.py
```diff
@@ -142,18 +144,39 @@ def explain(self) -> Tuple[List[Explanation], List[float]]:
...
         for target_idx in tqdm(range(self.vocab_size), desc="Explain"):
-            total_attribution, feat_to_token_attributions = get_total_attribution_with_retry(
+            total_attribution, feat_to_token_attributions, total_attribution_global = get_total_attribution_with_retry(
                 self.model,
                 self.target_feature_name,
                 target_idx if self.is_category_target else None,
                 inputs_encoded,
                 baseline,
-                self.use_global,
                 len(self.inputs_df),
```
connor-mccorm, line 153: Ok looks like we may not be using a sampled df - though I vaguely remember somewhere else in the explanation code we sample it.

connor-mccorm (approved):
LGTM, had some concerns regarding performance but I imagine you've probably considered them so they're more just questions at this point.

Issue #3092, opened 2023-02-14
RunTimeError when running regression example with GBM model type


skunkwerk:
**Describe the bug**
When running on a machine with a GPU, I get the following error:
(from hummingbird/ml/operator_converters/_tree_implementations.py:180):
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat2 in method wrapper_mm)

**To Reproduce**
Code and sample data below:

```
from ludwig.api import LudwigModel
import pandas as pd

url = 'http://archive.ics.uci.edu/ml/machine-learning-databases/auto-mpg/auto-mpg.data'
column_names = ['MPG', 'Cylinders', 'Displacement', 'Horsepower', 'Weight',
                'Acceleration', 'Model Year', 'Origin']

df = pd.read_csv(url, names=column_names,
                          na_values='?', comment='\t',
                          sep=' ', skipinitialspace=True)

ludwig_config = {
    "model_type": "gbm",
    "input_features": [        
        {
            "name": "Cylinders",
            "type": "number",
        },
        {
            "name": "Displacement",
            "type": "number",
        },
        {
            "name": "Horsepower",
            "type": "number",
        },
        {
            "name": "Weight",
            "type": "number",
        },
        {
            "name": "Acceleration",
            "type": "number",
        },
        {
            "name": "Model Year",
            "type": "number",
        },
        {
            "name": "Origin",
            "type": "category",
        },
    ],
    "output_features": [
        {
            "name": "MPG",
            "type": "number",
            "optimizer": {"type": "mean_squared_error"}
        }
    ],
}
model = LudwigModel(config=ludwig_config)
results = model.experiment(dataset=df)
```

**Expected behavior**
Should run without any errors (switching to model_type='ecd' works fine, for example)

**Screenshots**
N/A

**Environment (please complete the following information):**

- OS: Linux
- Version: 4.15.0-163-generic
- Python version: 3.8.5
- Ludwig version: 0.6.3 (from Dockerfile 'ludwigai/ludwig-ray-gpu:0.6.3')

Thanks for the amazing library!

tgaddair:
Thanks for raising this issue and providing and repo script @skunkwerk!

@jeffkinnison can you take a look?

skunkwerk:
I'm also sometimes getting a different error from the same code, and I'm not sure how it surfaces but it seems to be coming from Ray:

[...]

PR #3139, opened 2023-02-24
Disallow GPU tensors in GBM training and eval

jeffkinnison:
Workaround for #3092 that prevents torch tensors from being placed on GPU when training with a GBM. There is an underlying inconsistency with tensor placement, where different functions place prediction and target tensors on  different devices when a GPU is available. This update moves all tensors to CPU for the time being.

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -76,7 +76,8 @@ class Predictor(BasePredictor):
         self._distributed = distributed if distributed is not None else LocalStrategy()
         self.report_tqdm_to_ray = report_tqdm_to_ray
 
-        self.device = get_torch_device()
+        # TODO (jeffkinnison): revert to using the requested device for GBMs when device usage is fixed
+        self.device = get_torch_device() if not isinstance(model, GBM) else "cpu"
         self.model = model.to(self.device)
 
     def batch_predict(self, dataset: Dataset, dataset_name: str = None, collect_logits: bool = False):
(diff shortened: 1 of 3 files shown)


tgaddair (approved):

ludwig/models/predictor.py
```diff
@@ -193,15 +194,16 @@ def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=Fa
                         f"evaluation for {dataset_name}: obtained next batch "
                         f"memory used: {psutil.Process(os.getpid()).memory_info()[0] / 1e6:0.2f}MB"
                     )
+                    # TODO (jeffkinnison): revert to using the requested device for GBMs when device usage is fixed
                     inputs = {
                         i_feat.feature_name: torch.from_numpy(np.array(batch[i_feat.proc_column], copy=True)).to(
-                            self.device
+                            self.device if not isinstance(self.model, GBM) else "cpu"
```
tgaddair, line 200: Would rather have this be a param we could pass in to the batch evaluator, but okay with something like this if it's only meant to be very temporary.

Can you at least pull this out to the init of `self.device` instead, though?

tgaddair (changes requested):
Would actually like to make the change I suggested.

ludwig/models/predictor.py
jeffkinnison: Sure, I can do that. This is only meant to be a temp fix so GBM eval doesn't fail when a GPU is present. I definitely want to clean this up more generally.

tgaddair (approved):
LGTM!

PR #3146, opened 2023-02-27
Not default to adding `None` as accepted tuple for `FloatRangeTupleDataclassField`

abidwael:
Factor `allow_none` in whether to allow `None` value or not.


ksbrar (approved):
nice catch!

PR #3151, opened 2023-02-27
Disable passthrough decoder for all feature types

arnavgarg1:
Ludwig Issue: https://github.com/ludwig-ai/ludwig/issues/3150


tgaddair (approved):

w4nderlust:
so for binary and number it makes sense and we should keep it, for everything else we can remove it

abidwael (changes requested):
Could we disable passthrough decoder altogether for now and re-enable when we include the dot product combiner. 
cc: @w4nderlust

abidwael (approved):
Thanks @arnavgarg1!

PR #3154, opened 2023-02-27
Enable `transformer` encoder and disable `embed` encoder from `SequenceCombiner`

abidwael:
Resolving this failure: https://github.com/ludwig-ai/ludwig/actions/runs/4280114996/jobs/7451509445#step:9:1048

- The schema previously sets incorrectly sets `SEQUENCE` feature as feature type to `EncoderDataclassField` (encoder attribute of the sequence combiner). 
- This PR defines a new feature type `SEQUENCE_COMBINER_HIDDEN_REPRESENTATION` and registers the right encoders to it.
- Another approach to this would be defining another `EncoderForSequenceCombinerDataclassField` and all the associated register functions and registries.
- We could also add a new feature type `embedding` which would accept tensors.


justinxzhao (commented):
This doesn't fully solve the problem of ensuring that the encoder produces the right tensor shape that the sequence combiner expects, as many of the encoders are free to specify `reduce_output != None`, which would squeeze out the sequence length dim.

Excluding the `embed` encoder still seems like the right thing to do as it can never produce a 3D tensor.

However, instead of a registry change, I'd be inclined to do enforcement using an auxiliary check in  `checks.py` that verifies that `config[COMBINER][ENCODER][TYPE] != "embed"` and that `config[COMBINER][ENCODER]["reduce_output"] == None` when `config[COMBINER][TYPE] == "sequence_combiner"`.

abidwael:
@justinxzhao you raise a good point and we should definitely add an auxiliary check for encoder `reduce_output`, but I think that's a separate issue. The advantage of specifying these encoders in the schema an not in the aux checks is that it fails faster in this case. WDYT?

ludwig/schema/encoders/sequence_encoders.py
```diff
@@ -379,7 +379,7 @@ def module_name():
 
 
 @DeveloperAPI
-@register_encoder_config("stacked_parallel_cnn", [AUDIO, SEQUENCE, TEXT, TIMESERIES])
+@register_encoder_config("stacked_parallel_cnn", [AUDIO, SEQUENCE, TEXT, TIMESERIES, SEQUENCE_COMBINER_ENCODERS])
```
justinxzhao, line 382: I'm still not sure it's a good idea to introduce a new "meta-type" concept, just so that we can be more specific about what's allowed in the schema for this specific parameter. New readers will have to learn what the `SEQUENCE_COMBINER_ENCODER` type is to figure out if they should register their new encoder to that meta-type, as it's a remarkable departure from canonical Ludwig types like `AUDIO` and `BINARY`.

Instead, could we maintain a blacklist of forbidden encoders instead?

For example:
```python
_2D_SEQUENCE_ENCODERS = {"embed"}
"""
SEQUENCE encoders that always return 2D [batch_size, hidden_size] tensors, regardless of how they are parameterized.
These should never be used with modules that expect 3D tensors, such as the SequenceCombiner.
"""


# Add a blocklist argument to the EncoderDataclassField
@DeveloperAPI
def EncoderDataclassField(
    model_type: str, feature_type: str, default: str, description: str = "", blocklist: Set[str] = {}
) -> Field:
    """Custom dataclass field that when used inside a dataclass will allow the user to specify an encoder config.

    Returns: Initialized dataclass field that converts an untyped dict with params to an encoder config.
    """
    encoder_registry = get_encoder_classes(model_type, feature_type)

    class EncoderSelection(schema_utils.TypeSelection):
        def __init__(self):
            super().__init__(registry=encoder_registry, default_value=default, description=description)

        def get_schema_from_registry(self, key: str) -> Type[schema_utils.BaseMarshmallowConfig]:
            return encoder_registry[key]

        @staticmethod
        def _jsonschema_type_mapping():
            return {
                "type": "object",
                "properties": {
                    "type": {
                        "type": "string",
                        "enum": list(set(encoder_registry.keys()) - blocklist),
                        "enumDescriptions": get_encoder_descriptions(model_type, feature_type),
                        "default": default,
                    },
                },
                "title": "encoder_options",
                "allOf": get_encoder_conds(encoder_registry),
            }

    return EncoderSelection().get_default_field()


# Provide the blocklist in the parameter's definition:
@DeveloperAPI
@ludwig_dataclass
class SequenceCombinerConfig(BaseCombinerConfig):
    """Parameters for sequence combiner."""

    encoder: BaseEncoderConfig = EncoderDataclassField(
        MODEL_ECD,
        feature_type=SEQUENCE_COMBINER_ENCODERS,
        default="parallel_cnn",
        description="Encoder to apply to `main_sequence_feature`.",
        blocklist=_2D_SEQUENCE_ENCODERS,
    )
```

As an aside -- could text encoders also be used for this particular parameter if the user's feature is a text feature? @w4nderlust @tgaddair ? Maybe something to follow up on in a separate PR.

abidwael: I agree with you comment. Even though I wanted to prevent the case where we'd have to manually deregister some encoders, this seems like a good tradeoff. Will make the change.

justinxzhao (approved):
Nice! Looks good -- please update the PR description before merging.

PR #3166, opened 2023-03-01
Make Horovod an optional dependency when using Ray

### LINKED ISSUE #3161: auto_train shouldn't require horovod
**Describe the bug**
Using the autoML feature in auto_train seems to require horovod, but it should be possible to leverage autoML on a single machine without requiring horovod (which is very difficult to install).  If this is indeed by design, it should be called out as a requirement in the auto_train documentation.

the error is:
ModuleNotFoundError: Horovod isn't installed. To install Horovod with PyTorch support, run 'pip install 'horovod[pytorch]''. To install Horovod with TensorFlow support, run 'pip install 'horovod[tensorflow]''.

**To Reproduce**
Sample code:

```
from ludwig.automl import auto_train
from ludwig.datasets import mushroom_edibility
from ludwig.utils.dataset_utils import get_repeatable_train_val_test_split

mushroom_df = mushroom_edibility.load()
mushroom_edibility_df = get_repeatable_train_val_test_split(mushroom_df, 'class', random_seed=42)

auto_train_results = auto_train(
    dataset=mushroom_edibility_df,
    target='class',
    time_limit_s=120,
    tune_for_memory=False,
    user_config={'preprocessing': {'split': {'column': 'split', 'type': 'fixed'}}},
)

pprint.pprint(auto_train_results)
```

**Expected behavior**
auto_train should run successfully on a machine without Horovod installed.

- OS: Linux
- Python version: 3.8
- Ludwig version: 0.7 installed via pip, along with dask, hummingbird-ml, and lightgbm

thanks

tgaddair:
Closes #3161.

Falls back to DDP if Horovod is not installed. We may choose to make DDP the default in v0.8.

### DIFF
--- a/ludwig/backend/_ray210_compat.py
+++ b/ludwig/backend/_ray210_compat.py
@@ -5,9 +5,6 @@ from typing import Any, Callable, Dict, Optional, Type, TYPE_CHECKING, Union
 
 import ray
 from ray.air.config import RunConfig
-from ray.air.result import Result
-from ray.train.base_trainer import TrainingFailedError
-from ray.train.horovod import HorovodTrainer
 from ray.tune.execution.trial_runner import _ResumeConfig
 from ray.tune.impl.tuner_internal import TunerInternal
 from ray.tune.trainable import Trainable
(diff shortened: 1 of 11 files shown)


ludwig/backend/_ray210_compat.py
```diff
@@ -143,36 +140,3 @@ def _expected_utilization(self, cpus_per_trial, cpus_total):
...
-        tuner = TunerRay210(trainable=trainable, run_config=self.run_config)
-        result_grid = tuner.fit()
-        assert len(result_grid) == 1
-        try:
-            result = result_grid[0]
-            if result.error:
-                raise result.error
-        except TuneError as e:
-            raise TrainingFailedError from e
-        return result
```
arnavgarg1, line 178: Do we officially want to drop support for Ray 2.1?

arnavgarg1: Oh I see, just moved it a different place, all good

arnavgarg1 (approved):
LGTM 🚢

ludwig/distributed/base.py
```diff
@@ -1,8 +1,10 @@
 import contextlib
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Dict, Optional, Tuple, Type
 
 import torch
+from ray.train.backend import BackendConfig
+from ray.train.data_parallel_trainer import DataParallelTrainer
```
arnavgarg1, line 7: Probably need to gate these imports?

ludwig/backend/_ray210_compat.py
tgaddair: Maybe when 2.4 comes out. It would also need to be in v0.8, not 0.7.1.

ludwig/distributed/base.py
tgaddair: Good catch!

PR #3169, opened 2023-03-01
Re-enable benchmark tests on `Sarcos` dataset

abidwael:
Dataset links seem to be back. https://github.com/ludwig-ai/ludwig/pull/3168 will allow us to retrieve the dataset even if they're down again.


arnavgarg1 (approved):
This re-enables Sarcos GBM right?

abidwael:
@arnavgarg1 yes! but not ECD

PR #3175, opened 2023-03-01
Add sequence decoder integration tests

jeffkinnison:
Adds integration tests for SequenceGeneratorDecoder in response to #3152.

### DIFF
--- a/tests/integration_tests/test_sequence_decoders.py
+++ b/tests/integration_tests/test_sequence_decoders.py
@@ -1,0 +1,56 @@
+import os
+
+import pytest
+
+from ludwig.constants import COLUMN, DECODER, DEFAULTS, INPUT_FEATURES, NAME, OUTPUT_FEATURES, SEQUENCE, TEXT, TYPE
+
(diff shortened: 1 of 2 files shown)


jeffkinnison:
@tgaddair The accuracy discrepancy I mentioned earlier caused the distributed test failures here. It looks like it's coming from the eval vs local comparison in `tests.integration_tests.utils.train_with_backend`.

tgaddair:
@jeffkinnison that's what I figured. What this means is the distributed aggregation of the `token_accuracy` metric is not implemented correctly. We should probably fix that! Or if it's too much work, add the metric to the excluded list of metrics to compare in the test utils:

https://github.com/ludwig-ai/ludwig/blob/master/tests/integration_tests/utils.py#L908

tgaddair:
Also, if you're going to cherry-pick this manually with the other fix, I will remove the `release-0.7` label so it doesn't get auto-picked into the branch.

tgaddair (approved):

PR #3177, opened 2023-03-01
[fix] skip resetting index for empty dfs

ShreyaR:


tgaddair (changes requested):
In general we should avoid calling `len(df)` on dask dataframes, as it's super expensive. For the usage of this function in Ludwig, it's only called when we have more than one partition in the source data (which implies more than one row), so it would be just added computation for no benefit.

ShreyaR:
Closing pull request and moving the `empty_df` check into Predibase.

PR #3183, opened 2023-03-02
fix: [REBASE] Hoist uniqueItemProperties to top of feature JSON schema

ksbrar:
Redo of #3159 - this time from a native upstream branch rather than from my own fork.

--

Though this PR is just about moving one line of the schema (which is only relevant for AJV validation) one level up in its tree and the real issue at hand is that I'm working against marshmallow_jsonschema's rules (see this comment: https://github.com/ludwig-ai/ludwig/pull/2906#discussion_r1119576691), this change got me thinking about how we might want to deal with post-generation schema changes in the future.

I thought about tackling this a couple of different ways:

* some kind of post-generation decorator that can be added to a schema
* define a suite of schema transformations (like we already do for backwards compatibility)
* should we maintain separate schema transformations for usage with AJV?
* Ultimately I settled on a lightweight solution for now, but it's not ideal IMO.

### DIFF
--- a/ludwig/schema/features/base.py
+++ b/ludwig/schema/features/base.py
@@ -219,7 +219,12 @@ class FeatureList(fields.List):
 
 class FeaturesTypeSelection(schema_utils.TypeSelection):
     def __init__(
-        self, *args, min_length: Optional[int] = 1, max_length: Optional[int] = 3, supplementary_metadata=None, **kwargs
+        self,
+        *args,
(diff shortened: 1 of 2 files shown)


abidwael (approved):
Glad it worked!

ksbrar:
@abidwael I don't actually know if it did yet, no actions have run ;)

justinxzhao (approved):

PR #3188, opened 2023-03-02
fix: [REBASE] Streamline GBM defaults schema

ksbrar:
Redo of #3145  - this time from a native upstream branch rather than from my own fork

-----

Creates two separate defaults config registries and sets GBM defaults to only include preprocessing

Test failures are happening on benchmarking, idk why but unrelated to any of these changes.


ludwig/schema/features/binary_feature.py
```diff
@@ -24,6 +25,7 @@
 
 
 @DeveloperAPI
+@gbm_defaults_config_registry.register(BINARY)
```
justinxzhao, line 28: From the class name (`BinaryInputFeatureConfigMixin`) alone it was not clear to me why we wouldn't also register this class to the `ecd_defaults_config_registry`. However, upon closer inspection I see that `@ecd_defaults_config_registry.register(BINARY)` decorates the `BinaryDefaultsConfig` class.

Could you add:
- a docstring on `gbm_defaults_config_registry` (in `ludwig/schema/features/utils.py`) that the only parameters that can be specified for the `defaults` section for GBM feature types are `preprocessing` parameters
- a docstring on  `ecd_defaults_config_registry` (in `ludwig/schema/features/utils.py`) explaining the pattern of marshmallow configs that it registers?

jeffkinnison (approved):

justinxzhao (approved):

connor-mccorm (approved):

PR #3190, opened 2023-03-02
Consolidate hyperopt schema in `ludwig.schema.hyperopt`

jeffkinnison:
The hyperopt search algorithm schemas are defined in `ludwig.hyperopt`, and they currently work with dictionary schemas rather than dataclasses. This update will move all hyperopt schema into `ludwig.schema.hyperopt` and update hyperopt to work with object schemas.


ludwig/schema/hyperopt/scheduler.py
```diff
@@ -107,6 +101,19 @@ class BaseSchedulerConfig(schema_utils.BaseMarshmallowConfig, ABC):
         ),
     )
 
+    def dependencies_installed(self):
```
justinxzhao, line 104: Sweet!

ludwig/schema/hyperopt/scheduler.py
```diff
@@ -107,6 +101,19 @@ class BaseSchedulerConfig(schema_utils.BaseMarshmallowConfig, ABC):
...
+        """Some search algorithms require additional packages to be installed, check that they are available."""
+        for package_name in get_scheduler_dependencies(self.type):
+            try:
+                import_module(package_name)
+            except ImportError:
+                raise ImportError(
+                    f"Scheduler {self.type} requires package {package_name}, however package is "
+                    "not installed. Please refer to Ray Tune documentation for packages required for this "
+                    "scheduler."
```
justinxzhao, line 113: Does it make sense to have a ludwig requirements handy that would install all of the schedulers?

`pip install ludwig[hyperopt]`

And/or is there a link to Ray Tune documentation we can provide in the raise message here?

justinxzhao (approved):
Awesome! Great work!

ludwig/config_validation/checks.py
```diff
@@ -291,7 +291,25 @@ def is_divisible(hidden_size: int, num_heads: int) -> bool:
             )
 
 
+def check_hyperopt_search_algorithm_dependencies_installed(config: "ModelConfig") -> None:  # noqa: F821
+    """Check that the hyperopt search algorithm dependencies are installed."""
+    if config.hyperopt is not None:
+        try:
+            config.hyperopt.search_alg.dependencies_installed()
+        except ImportError as e:
+            raise ConfigValidationError(e.msg)
```
arnavgarg1, line 300: Really nice validation checks!

ludwig/hyperopt/execution.py
```diff
@@ -146,7 +146,7 @@ def __init__(
         metric: str,
         goal: str,
         split: str,
-        search_alg: Dict,
+        search_alg: "BaseSearchAlgorithmConfig",  # noqa: F821
```
arnavgarg1, line 149: nit: Can we maybe import at the top? Should help get rid of the noqa

ludwig/hyperopt/execution.py
```diff
@@ -741,21 +740,13 @@ def execute(
...
-                )
+        if self.search_algorithm is not None:
+            self.search_algorithm.set_random_state(random_seed)
+            sa_kwargs = self.search_algorithm.to_dict()
+            del sa_kwargs[TYPE]
+            sa_kwargs["mode"] = mode
+            sa_kwargs["metric"] = metric
+            search_alg = tune.create_searcher(self.search_algorithm.type, **sa_kwargs)
         else:
             search_alg = None
```
arnavgarg1, line 751: Should we log a warning like we were before about the default search algorithm  we're going to be using?

justinxzhao:
Looks like these tests are failing in the CI. Looks like it should be a relatively quick fix @jeffkinnison CC: @arnavgarg1 

```
FAILED tests/integration_tests/test_experiment.py::test_experiment_text_output_feature_with_tagger_decoder[sum] - ValueError: Decoder inputs rank is 2, but should be 3: [batch_size x max_sequence_length x hidden_size] in when using a tagger sequential decoder. Consider setting reduce_output to None if a sequential encoder / combiner is used.
FAILED tests/integration_tests/test_experiment.py::test_experiment_sequence_output_feature_with_tagger_decoder[sum] - ValueError: Decoder inputs rank is 2, but should be 3: [batch_size x max_sequence_length x hidden_size] in when using a tagger sequential decoder. Consider setting reduce_output to None if a sequential encoder / combiner is used.
FAILED tests/integration_tests/test_experiment.py::test_experiment_category_input_feature_with_tagger_decoder - ValueError: Decoder inputs rank is 2, but should be 3: [batch_size x max_sequence_length x hidden_size] in when using a tagger sequential decoder. Consider setting reduce_output to None if a sequential encoder / combiner is used.
```

ludwig/schema/hyperopt/parameter.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def log_base_field(default=10):
+    return schema_utils.Float(default=default, description="Logarithmic base.")
+
+
+@DeveloperAPI
+@ludwig_dataclass
+class BaseParameterConfig(schema_utils.BaseMarshmallowConfig):
+    space: Optional[str] = None
```
arnavgarg1, line 29: Space is required right? Or is it no longer required because the space in each of the subclasses is a predefined ProtectedString?

ludwig/schema/hyperopt/parameter.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+@DeveloperAPI
+@register_parameter_config("choice")
+@ludwig_dataclass
+class CategoricalParameterConfig(BaseParameterConfig):
```
arnavgarg1, line 35: nit: can we potentially rename this to `ChoiceParameterConfig` to match up with the registered name? Will also make it easier if we ever need to search across Ludwig for this Config (I do this a lot with other schema classes and it's really helpful when the names more or less match the registered names)

ludwig/schema/hyperopt/scheduler.py
arnavgarg1: Could also be useful to just say something like, `install this package by running pip install -U {package_name}`

arnavgarg1: But would be great if they could install the full list in one go. That might be out of the scope for this PR

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -54,3 +81,675 @@ def _jsonschema_type_mapping(self):
         },
         default_factory=load_default,
     )
+
+
+@DeveloperAPI
+@register_search_algorithm("random", random_state_field="random_state")
+@register_search_algorithm("variant_generator")
```
arnavgarg1, line 88: Do we also need to add `random_state_field="random_state"` to the `variant_generator` entry in the registry?

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -54,3 +81,675 @@ def _jsonschema_type_mapping(self):
...
+@ludwig_dataclass
+class BasicVariantSAConfig(BaseSearchAlgorithmConfig):
```
arnavgarg1, line 90: Interesting that BasicVariant doesn't require `mode` and `metric` 🤔

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -54,3 +81,675 @@ def _jsonschema_type_mapping(self):
...
+
+    mode: Optional[str] = schema_utils.StringOptions(
+        options=["min", "max", None],
+        default=None,
+        allow_none=True,
+        description=(
+            "One of `{min, max}`. Determines whether objective is minimizing or maximizing the metric attribute. "
+            r"Defaults to \“max\”."
+        ),
+    )
```
arnavgarg1, line 158: nit: To simplify the schema and condense it, you can define these base fields either in the Base Class, or as separate methods like you did with `quantization` in the schedulers and then call these here instead of redefining them multiple times (both mode and metric)

arnavgarg1: Maybe even `points_to_evaluate`?

ludwig/schema/hyperopt/utils.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Args:
+        name: the name to scheduler the parameter class under, does not need to correspond to the value of `type`
+        dependencies: the list of module names that the scheduler requires
+
+    Returns:
+        Wrapper function to decorate a `BaseSchedulerConfig` subclass
+    """
+
+    def wrap(scheduler_config: Type["BaseSchedulerConfig"]) -> Type["BaseSchedulerConfig"]:  # noqa: F821
```
arnavgarg1, line 130: Does it just use FIFO in this case as the default?

ludwig/schema/utils.py
```diff
@@ -577,6 +578,31 @@ def IntegerRange(
     )
 
 
+@DeveloperAPI
+def Float(default: Union[None, float], allow_none=False, description="", parameter_metadata: ParameterMetadata = None):
+    """Returns a dataclass field with marshmallow metadata strictly enforcing (non-float) inputs."""
```
arnavgarg1, line 583: Do you mean strictly enforcing float inputs?

arnavgarg1 (approved):
Really great work @jeffkinnison 🥳  
Couple of nits and ways you can cleanup the schemification to reduce duplication, but all in all it looks great!

arnavgarg1:
For the tests that are failing (the tagger decoder ones), can you merge the latest master into your branch and try again?

ludwig/schema/hyperopt/scheduler.py
jeffkinnison: I think we could register the PyPI package name as well in a tuple of `(module_name, package_name)` and raise the pip install command like that. We would have to do it like that because something like `pip install skopt` won't work. I could update like so and then catch the import error, collect the uninstalled packages, and raise the ConfigValidationError with the full command to install all missing dependencies.

jeffkinnison: I think it would be overkill to have separate requirements for all of these search algs as well as any scheduler deps (only PBT2 for now, but that may change in the future) since most only need one additional package. Is there significant overhead to installing them all?

ludwig/schema/hyperopt/parameter.py
jeffkinnison: I was on the fence about removing `BaseParameterConfig` and just having the base classes of each parameter variant subclass `schema_utils.BaseMarshmallowConfig`. Since this is a parent to non-overlapping config classes, I think it makes sense to get rid of it.

ludwig/schema/hyperopt/search_algorithm.py
jeffkinnison: Good catch!

ludwig/schema/hyperopt/search_algorithm.py
jeffkinnison: I'm actually considering removing `mode` and `metric` from the configs since we overwrite those anyway when passing config to `create_searcher`. [Here's where that happens in the original executor code.](https://github.com/ludwig-ai/ludwig/blob/397e46b44a2e39fc8f6c81b94b398c491ff671c0/ludwig/hyperopt/execution.py#L740-L758). WDYT?

ludwig/schema/hyperopt/utils.py
jeffkinnison: Yes, FIFO is the default.

ludwig/schema/utils.py
jeffkinnison: Ah, thought I fixed that. Thanks for marking it.

ludwig/hyperopt/execution.py
jeffkinnison: Oh, whoops, that should actually be `Dict`, made that change expecting to receive a schema object here. On a separate note, I think it would be worthwhile to load the hyperopt schema from dict in `hyperopt/run.py` and use the OO config throughout. That could be a low-priority followup PR.

ludwig/schema/hyperopt/search_algorithm.py
jeffkinnison: I like the latter idea because the search algs don't have consistent configs.

ludwig/config_validation/checks.py
```diff
@@ -291,7 +291,25 @@ def is_divisible(hidden_size: int, num_heads: int) -> bool:
             )
```
ksbrar, line 291: While `hyperopt.parameters` remains a `Dict`, I might add a validation check here as well that loops through and checks basic stuff like:
1) is every nested parameter a dict itself
2) are the nested parameters valid keywords (e.g. is there something other than `space`, `lower`, `upper`, etc.)
3) if you want you could even try to dynamically apply the parameter space schemas here based on `space`

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -1,22 +1,105 @@
...
+
+
+def mode_field(description: Optional[str] = None) -> fields.Field:
+    return schema_utils.StringOptions(
+        options=["min", "max", None],
+        default=None,
+        allow_none=True,
+        description=description
+        or (
+            "One of `{min, max}`. Determines whether objective is minimizing or maximizing the metric attribute. "
```
ksbrar, line 37: nit: Is None allowed? Is it meaningfully different from fallback to default (max)?

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -54,3 +137,470 @@ def _jsonschema_type_mapping(self):
...
+            "string, `domain` must be set and `optimizer` must be one of [random, bandit, genetic]."
+        ),
+    )
+
+    domain: Optional[str] = schema_utils.StringOptions(
+        options=["cartesian", "euclidean"],
+        default=None,
+        allow_none=True,
+        description=(
+            "Optional domain. Should only be set if you don't pass an optimizer as the `optimizer` argument. Has to "
```
ksbrar, line 345: nit: `*If set*, has to be <...>`

ludwig/schema/hyperopt/search_algorithm.py
```diff
@@ -54,3 +137,470 @@ def _jsonschema_type_mapping(self):
...
+class ZooptSAConfig(BaseSearchAlgorithmConfig):
+    type: str = schema_utils.ProtectedString("zoopt")
+
+    algo: str = schema_utils.ProtectedString(
+        pstring="asracos",
+        description="To specify an algorithm in zoopt you want to use. Only support ASRacos currently.",
+    )
+
+    budget: Optional[int] = schema_utils.PositiveInteger(
+        default=None, allow_none=True, description="Number of samples."
```
ksbrar, line 583: nit: `Optional. Number of samples.`

ludwig/schema/utils.py
```diff
@@ -577,6 +578,31 @@ def IntegerRange(
     )
 
 
+@DeveloperAPI
+def Float(default: Union[None, float], allow_none=False, description="", parameter_metadata: ParameterMetadata = None):
```
ksbrar, line 582: How is this meaningfully different from `FloatRange`? The strictness doesn't [add too much for floats atm](https://github.com/marshmallow-code/marshmallow/issues/1601) - are you seeing a specific use case for it in hyperopt?

ksbrar (commented):
Very nice 💪🏽 . Left some comments.

ludwig/schema/hyperopt/search_algorithm.py
jeffkinnison: `None` is the default in Ray Tune, but I'm actually considering dropping `metric` and `mode`from all of the search algs. We override those with `hyperopt.metric` and `hyperopt.goal` while setting up the search alg in `hyperopt/execution.py`, so I'm not sure it makes sense to expose those parameters in the search alg config.

ludwig/schema/utils.py
jeffkinnison: This is mostly an issue of semantics, and can absolutely be removed if I'm overthinking it. There were a couple of considerations that led to `Float`:

- It's difficult to set generalized ranges for hyperparameter values. Different params will have different viable ranges, so for a general parameter config object, the range would need to be `[-inf, inf]`.
- `lower`/`upper` and `mean`/`sd` act as bounds on the viable range for a given parameter anyway, and each is represented by a single float/int

My thinking was that a singular unbounded float represented the intent of the fields better than `FloatRange`, but it may also be overkill to add an entire field type just for hyperopt.

ksbrar: Gotcha. By default `FloatRange` is a singular unbounded float (leaving the `max` and `min` arguments set to `None` is conceptually the same as `[-inf, inf]`)

jeffkinnison: Should I remove `Float` and replace those fields with `FloatRange`?

ksbrar: You could keep it around as a simplified alias that uses `FloatRange` under the hood, but we shouldn't have two separate, duplicated implementations

ludwig/schema/hyperopt/utils.py
```diff
@@ -0,0 +1,170 @@
...
+    """Register a search algorithm config class by name.
+
+    Args:
+        name: the name to register the search algorithm class under, does not need to correspond to the value of `type`
+        random_state_field: the name of the random state in this search algorithm
+        dependencies: the list of search algorithm dependency package name/install name pairs, e.g.
+                      `("sklearn", "scikit-learn")`
+
+    Returns:
+        Wrapper function to decorate a
```
arnavgarg1, line 161: nit: finish up the doc string?

jeffkinnison:
These most recent tests are failing because of a (possibly) malformed parameter config in `tests/ludwig/utils/test_defaults.py::test_merge_with_defaults_early_stop`. The parameters config references a text input feature named "utterance" in [the config generated by](https://github.com/ludwig-ai/ludwig/blob/21b3bc6cce9a7e17fa891f5a8d7559317f52d272/tests/conftest.py#L221-L236) `conftest._get_sample_config`, however the test does not actually use the config from `_get_sample_config`. Setting the name of the text feature to "utterance" fixes the failure locally.

As a side note, this was actually an instance of the parameter name aux checks introduced in this PR working as intended!

@tgaddair @justinxzhao Is it all right to update this unrelated test to work with the new aux checks?

justinxzhao:
> These most recent tests are failing because of a (possibly) malformed parameter config in `tests/ludwig/utils/test_defaults.py::test_merge_with_defaults_early_stop`. The parameters config references a text input feature named "utterance" in [the config generated by](https://github.com/ludwig-ai/ludwig/blob/21b3bc6cce9a7e17fa891f5a8d7559317f52d272/tests/conftest.py#L221-L236) `conftest._get_sample_config`, however the test does not actually use the config from `_get_sample_config`. Setting the name of the text feature to "utterance" fixes the failure locally.
> 
> As a side note, this was actually an instance of the parameter name aux checks introduced in this PR working as intended!
> 
> @tgaddair @justinxzhao Is it all right to update this unrelated test to work with the new aux checks?

Yes!

ludwig/schema/hyperopt/search_algorithm.py
jeffkinnison: Followup: I dropped metric and mode as described above.

ludwig/schema/utils.py
jeffkinnison: Followup: removed Float and converted those instances to FloatRange.

ludwig/config_validation/checks.py
jeffkinnison: Followup: Added checks for parameter spaces that instantiate the spaces as config objects. Additionally, there are basic checks to ensure that they refer to existing config parameters.

ludwig/schema/hyperopt/parameter.py
```diff
@@ -0,0 +1,177 @@
+from typing import List, Type, Union
```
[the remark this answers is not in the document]
jeffkinnison: Followup from our discussion: I want to do followup work to figure out the best way to handle dynamic fields with marshmallow.

ludwig/config_validation/checks.py
```diff
@@ -322,6 +341,45 @@ def check_tagger_decoder_requirements(config: "ModelConfig") -> None:  # noqa: F
         )
 
 
+@register_config_check
+def check_hyperopt_parameter_dicts(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks for hyperopt parameter dicts against their config objects."""
+    from ludwig.schema.hyperopt.utils import get_parameter_cls, parameter_config_registry
+
+    if config.hyperopt is not None:
```
justinxzhao, line 349: nit: Use less nesting for better code readability.

```python
if config.hyperopt is None:
    return
<rest of code.>
```

ludwig/schema/hyperopt/__init__.py
```diff
@@ -17,7 +18,14 @@
 class HyperoptConfig(schema_utils.BaseMarshmallowConfig, ABC):
     """Basic hyperopt settings."""
 
-    output_feature: str = "combined"  # TODO: make more restrictive
+    output_feature: str = schema_utils.String(  # TODO: make more restrictive
+        default="combined",
+        description=(
```
justinxzhao, line 23: Nice description.

Note, that there's an aux check that you could potentially re-use, for `hyperopt.output_feature` and `hyperopt.metric` below:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/config_validation/checks.py#L217-L232

justinxzhao (commented):
Nice.

justinxzhao (approved):
Looks like there's just a couple of failing tests (hopefully not too crazy to fix?), and I had one nit. Otherwise, LGTM.

jeffkinnison:
This newest issue is a dictionary merge error within automl. During ``, we naively merge the configs of `ludwig/automl/defaults/combiner/concat_config.yaml` and `ludwig/automl/defaults/text/bert_config.yaml`. This leads to the invalid parameter config

```
hyperopt:
  parameters:
    trainer.learning_rate:
      space: loguniform
      lower: 0.00002
      upper: 0.001
      categories: [0.00002, 0.00003]
```

which throws an error during aux validation because `categories` is not a valid field for loguniform spaces. A quick fix would be to edit `bert_config.yaml` to use a loguniform space with lower/upper bounds set to 0.00002/0.00003 respectively, or a even qloguniform space quantized to only ever generate 0.00002 and 0.00003. A more involved fix would update `ludwig.utils.misc_utils.merge_dict` to ignore specific keys (in this case `"parameter"`) by optional arg, then merge the parameter dicts in a separate pass that skips parameter spaces where a merge would be invalid.

@justinxzhao WDYT?

ludwig/config_validation/checks.py
jeffkinnison: What do you think about adding something like a `hyperopt_check` decorator that includes the `if config.hyperopt is not None` logic? That would set us up nicely to encapsulate any future constraints and make the current checks a bit cleaner.

justinxzhao:
Nice catch! @jeffkinnison I'd say for now let's either change concat_config.yaml to also use choice: 0.00001, 0.0001, 0.001, or update the bert config, as you suggest.

It would be good to leave a comment on both files that the search spaces need to be in sync as some tests merge them.

ludwig/config_validation/checks.py
justinxzhao: I think that's a reasonable idea, though one I would suggest doing outside of this PR. There are other potentially reusable aux check preconditions (`if ecd`, `if gbm`, etc.) that would be good to consider in the design of how to extend the existing decorator.

jeffkinnison:
The latest failures required a rewrite of the nested parameter check because of the case where we were hyperopting over specific combiner configs, and checking against the defaults led to a validation error when `combiner.type` changed. Because of the complexity of these checks, going forward we should be mindful to add hyperopt validation check failures to the unit tests as they come up.

ludwig/config_validation/checks.py
```diff
@@ -344,10 +348,14 @@ def check_tagger_decoder_requirements(config: "ModelConfig") -> None:  # noqa: F
...
+    if config.hyperopt is None:
+        return
+
+    from ludwig.schema.hyperopt.utils import get_parameter_cls, parameter_config_registry  # noqa: F401
 
-    if config.hyperopt is not None:
-        for parameter, space in config.hyperopt.parameters.items():
+    for parameter, space in config.hyperopt.parameters.items():
+        # skip nested hyperopt parameters
+        if parameter != ".":
```
ksbrar, line 359: Confused by this. When would parameter be a single period?

ludwig/schema/utils.py
```diff
@@ -714,16 +720,23 @@ def List(
...
 
-    if list_type is str:
-        field_type = fields.String()
-    elif list_type is int:
-        field_type = fields.Integer()
-    elif list_type is float:
-        field_type = fields.Float()
-    elif list_type is list:
-        field_type = fields.List(fields.Float())
+    type_map = {str: fields.String, int: fields.Integer, float: fields.Float}
```
ksbrar, line 723: 👌

tests/ludwig/config_validation/test_validate_config_hyperopt.py
```diff
@@ -0,0 +1,296 @@
...
+        ("foo.bar", True),
+        ("trainer.bar", True),
+        ("foo.learning_rate", True),
+        ("in_feature.encoder.bar", True),
+        ("in_feature.foo.num_fc_layers", True),
+        ("out_feature.encoder.bar", True),
+        ("out_feature.foo.num_fc_layers", True),
+    ],
+)
+def test_parameter_key_check(referenced_parameter, raises_exception):
```
ksbrar, line 133: 👌

ludwig/config_validation/checks.py
```diff
@@ -322,6 +346,49 @@ def check_tagger_decoder_requirements(config: "ModelConfig") -> None:  # noqa: F
...
+def check_hyperopt_parameter_dicts(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks for hyperopt parameter dicts against their config objects."""
+    if config.hyperopt is None:
+        return
+
+    from ludwig.schema.hyperopt.utils import get_parameter_cls, parameter_config_registry  # noqa: F401
+
+    for parameter, space in config.hyperopt.parameters.items():
+        # skip nested hyperopt parameters
+        if parameter != ".":
```
ksbrar, line 359: A bit confused - when would `parameter` be a single dot?

ludwig/config_validation/checks.py
```diff
@@ -409,3 +417,46 @@ def check_concat_combiner_requirements(config: "ModelConfig") -> None:  # noqa:
...
+        """Recursively check config field existence."""
+        for k, v in parameter.items():
+            nested_path = f"{field_path}.{k}"
+            try:
+                if isinstance(v, dict):
+                    check_fields_exist(v, config.__getattribute__(k), field_path=nested_path)
+                else:
+                    config.__getattribute__(k)
+            except AttributeError:
+                invalid_fields.append(nested_path)
```
ksbrar: this is awesome

ludwig/config_validation/checks.py
```diff
@@ -409,3 +417,46 @@ def check_concat_combiner_requirements(config: "ModelConfig") -> None:  # noqa:
             "3D encoder output shapes, or 3) Remove features to ensure that output shapes from all encoders are the "
             "same dimension (all 2D or all 3D)."
         )
+
+
+@register_config_check
+def check_hyperopt_nested_parameter_dicts(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks that all nested parameters in a hyperopt config exist."""
+    if config.hyperopt is None or "." not in config.hyperopt.parameters:
```
ksbrar, line 426: Still confused by the significance of the single dot

ludwig/config_validation/checks.py
```diff
@@ -409,3 +417,46 @@ def check_concat_combiner_requirements(config: "ModelConfig") -> None:  # noqa:
...
+        check_fields_exist(category, config)
+
+    if len(invalid_fields) > 0:
+        raise ConfigValidationError(
+            "The following nested hyperparameters do not correspond to valid config fields: "
+            f"{','.join(invalid_fields)}. Check the Ludwig docs for the list of valid parameters."
+        )
+
+    try:
+        space_cls = get_parameter_cls("choice")
```
ksbrar, line 451: nit/quibble: are we checking parameters here that are 'nested' or just ones that have categorical search spaces?

jeffkinnison: Only nested spaces. The examples only show them as choice spaces, and I think expanding beyond choice or grid search is probably out of scope for this PR, so the check enforces that the space has to be choice.

ludwig/config_validation/checks.py
jeffkinnison: We name nested spaces `.` according to the docs. I think we could revisit this design down the road if we expand on nested spaces.

ludwig/config_validation/checks.py
jeffkinnison: When it's a nested space. According to the docs, we name nested spaces `.` in the config.

ludwig/config_validation/checks.py
ksbrar: Still lost 😓 . Can you give an example?

jeffkinnison:
Looks like the tagger decoder failures were happening because the check was de-registered at some point. That can happen when merging `checks.py` because the registry line is shared between all checks, and so not duplicated in a merge diff. Added it back in here.

ludwig/config_validation/checks.py
justinxzhao: @ksbrar This was functionality that Travis added [a while ago](https://ludwig.ai/latest/user_guide/hyperopt/#nested-ludwig-config-parameters).

ludwig/config_validation/checks.py
justinxzhao: @ksbrar [This](https://ludwig.ai/latest/user_guide/hyperopt/#nested-ludwig-config-parameters) is what Jeff may be referring to.

ludwig/config_validation/checks.py
ksbrar: Ah gotcha! Thanks for the link.

justinxzhao (approved):
LGTM!

ksbrar (approved):
LGTM 🚀

arnavgarg1 (approved):

arnavgarg1:
Amazing work 👏

tgaddair (approved):
LGTM! A lot of great work in this PR :)

PR #3195, opened 2023-03-03
Fixes dict_hash discrepancy

w4nderlust:


tgaddair (approved):

justinxzhao (approved):
Nice!

ludwig/data/cache/util.py
```diff
@@ -20,7 +20,9 @@ def calculate_checksum(original_dataset: CacheableDataset, config: ModelConfigDi
         "global_defaults": config.get(DEFAULTS, {}),
         # PROC_COLUMN contains both the feature name and the feature hash that is computed
         # based on each feature's preprocessing parameters and the feature's type.
-        "feature_proc_columns": {feature[PROC_COLUMN] for feature in features},
+        # creating a sorted list out of the dict because hash_dict requires all values
+        # of the dict to be ordered object to ensure the creation fo the same hash
+        "feature_proc_columns": sorted({feature[PROC_COLUMN] for feature in features}),
```
tgaddair, line 25: I'm wondering if there's a way we can test this. Something that would force the behavior of changing the insertion order of the set.

This looks promising:

By default, the __hash__() values of str, bytes and datetime objects are “salted” with an unpredictable random value. Although they remain constant within an individual Python process, they are not predictable between repeated invocations of Python.

This is intended to provide protection against a denial-of-service caused by carefully-chosen inputs that exploit the worst case performance of a dict insertion, O(n^2) complexity. See http://www.ocert.org/advisories/ocert-2011-003.html for details.

Changing hash values affects the iteration order of dicts, sets and other mappings. Python has never made guarantees about this ordering (and it typically varies between 32-bit and 64-bit builds).

See also PYTHONHASHSEED.

tgaddair: Let me see if I can put a test that exploits this together and add it to this PR.

tgaddair: Okay, added a test and verified it repros the issue without this fix, and with this fix succeeds.

arnavgarg1 (approved):
🚢  this is great!

ludwig/data/cache/util.py
arnavgarg1: Nice, thanks for looking into Python docs @tgaddair. It makes sense that the ordering of sets would cause some problems from the docs, so this is a good change!

Also pretty interesting that the salts are scoped to individual processes rather than across processes, but it makes sense if you think about it

tgaddair: If the salts were the same between processes, then it wouldn't serve its purpose of making the hash function unpredictable (to an attacker). But if they were different within a process, then hash lookups wouldn't work ;). So it makes sense, just never considered that they designed their hash function with that exploit in mind.

tests/ludwig/data/test_cache_util.py
```diff
@@ -109,3 +109,30 @@ def test_proc_col_checksum_consistency_same_preprocessing_different_types():
...
+    @ray.remote(max_calls=1)
+    def calculate_checksum_remote(dataset, config):
+        return calculate_checksum(dataset, config)
+
+    # Run each checksum calculation as a remote function so it gets its own Python interpreter, as
+    # the hash function in Python is deterministic within a process, but not between different processes.
+    # See: https://docs.python.org/3/reference/datamodel.html#object.__hash__
+    checksum1 = ray.get(calculate_checksum_remote.remote(mock_dataset, config.to_dict()))
+    checksum2 = ray.get(calculate_checksum_remote.remote(mock_dataset, config.to_dict()))
+    assert checksum1 == checksum2
```
arnavgarg1, line 138: Will this start a Ray cluster? If yes, should we call `ray.shutdown()` at the end?

arnavgarg1: @tgaddair

arnavgarg1: Nvm we're using the fixture so it'll stop the cluster by itself, all good!

PR #3197, opened 2023-03-03
Use tmpdir for more files generated during tests

tgaddair:
We should try and fine a way to do this at a global level for all tests, probably by mocking `get_output_directory` in some form.

### DIFF
--- a/tests/ludwig/augmentation/test_augmentation_pipeline.py
+++ b/tests/ludwig/augmentation/test_augmentation_pipeline.py
@@ -139,14 +139,14 @@ def run_augmentation_training(
         },
     }
 
-    model = LudwigModel(config, logging_level=logging.INFO)
-    model.experiment(
-        dataset=train_fp,
-        skip_save_processed_input=True,
-        skip_save_model=True,
-    )
-    return model
-
+    with tempfile.TemporaryDirectory() as tmpdir:
+        model = LudwigModel(config, logging_level=logging.INFO)
+        model.experiment(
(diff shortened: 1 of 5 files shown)


tests/ludwig/augmentation/test_augmentation_pipeline.py
```diff
@@ -139,14 +139,14 @@ def run_augmentation_training(
...
-    return model
-
+    with tempfile.TemporaryDirectory() as tmpdir:
+        model = LudwigModel(config, logging_level=logging.INFO)
+        model.experiment(
+            dataset=train_fp,
+            skip_save_processed_input=True,
+            skip_save_model=True,
+            output_directory=os.path.join(tmpdir, "output"),
+        )
```
arnavgarg1, line 149: tmpdir fixture?

justinxzhao: Can the pytest fixture be used in a non-"test_" function?

justinxzhao (approved):

tests/ludwig/augmentation/test_augmentation_pipeline.py
tgaddair: No. We would need to plumb it through, so would just add more complexity, IMO.

arnavgarg1: Yup, I was thinking plumbing it through, but this is also ok for now

PR #3205, opened 2023-03-04
Update sequence/text feature `max_sequence_length` default to `None`

geoffreyangus:
This PR updates `max_sequence_length` behavior to do the following:
1. Default to `None`.
2. If `None`, infer the length from the dataset.
3. If not `None`, use that value and pad and truncate as needed.

Before this change, the `max_sequence_length` behavior was the following:
1. Default to not `None` (256).
2. Always take the minimum value between the (1) inferred length from the dataset and (2) the specified value for `max_sequence_length`.

Before this change, the behavior was unpredictable (dataset dependent) and therefore made it error-prone to attempt to tie weights between two text encoders, since the max sequence length between them could be different. This PR ensures that we can set a firm `max_sequence_length` for a given encoder, which makes the behavior much more clear and less error-prone when using tied text encoders.


justinxzhao (approved):
Clean! Makes sense and LGTM.

PR #3211, opened 2023-03-06
Add the ability to specify a local or S3 HF mirror for more guaranteed loading of pre-trained HF models.

justinxzhao:
## Overview

We've seen that HF servers can go down sometimes. This PR adds the ability to specify a local or S3 HF mirror for more guaranteed loading of pre-trained HF models by setting a `LUDWIG_PRETRAINED_MODELS_DIR` env variable.

## Implementation

If the `LUDWIG_PRETRAINED_MODELS_DIR` environment variable is set, we attempt to load the HF model from this directory, falling back to downloading from the HF hub if the model is not found, downloading fails, or if model initialization fails.

`LUDWIG_PRETRAINED_MODELS_DIR` can be an s3 path. Weights are copied to a local temporary directory, and the model is loaded from there.

The expected structure of the `LUDWIG_PRETRAINED_MODELS_DIR` directory is:

    {LUDWIG_PRETRAINED_MODELS_DIR}/{pretrained_model_name_or_path}/pytorch_model.bin
    {LUDWIG_PRETRAINED_MODELS_DIR}/{pretrained_model_name_or_path}/config.json

For example, if `LUDWIG_PRETRAINED_MODELS_DIR` is set to `s3://my-bucket/pretrained-models`, and `pretrained_model_name_or_path` is set to `bert-base-uncased`, we expect to find the following files:

    s3://my-bucket/bert-base-uncased/
    - pytorch_model.bin
    - config.json

If the `LUDWIG_PRETRAINED_MODELS_DIR` environment variable is not set, we download the model from the HF hub.


tests/ludwig/utils/test_hf_utils.py
```diff
@@ -2,23 +2,38 @@
...
+    # Don't set env var.
+    _, used_fallback = load_pretrained_hf_model_with_hub_fallback(AlbertModel, ALBERTEncoder.DEFAULT_MODEL_NAME)
+    assert used_fallback
+
+    # Download the model, load it from tmpdir, and set env var.
+    load_pretrained_hf_model_from_hub(AlbertModel, "albert-base-v2").save_pretrained(
+        os.path.join(tmpdir, "albert-base-v2")
+    )
+    os.environ["LUDWIG_PRETRAINED_MODELS_DIR"] = str(tmpdir)  # Needs to be an absolute path.
+    _, used_fallback = load_pretrained_hf_model_with_hub_fallback(AlbertModel, ALBERTEncoder.DEFAULT_MODEL_NAME)
```
tgaddair, line 38: Can you also add a test that validates that everything works when the model is not in the `LUDWIG_PRETRAINED_MODELS_DIR`?

tests/ludwig/utils/test_hf_utils.py
```diff
@@ -2,23 +2,38 @@
...
+    """Ensure that the HF models used in ludwig download correctly with S3 or hub fallback."""
+    # Don't set env var.
+    _, used_fallback = load_pretrained_hf_model_with_hub_fallback(AlbertModel, ALBERTEncoder.DEFAULT_MODEL_NAME)
+    assert used_fallback
+
+    # Download the model, load it from tmpdir, and set env var.
+    load_pretrained_hf_model_from_hub(AlbertModel, "albert-base-v2").save_pretrained(
+        os.path.join(tmpdir, "albert-base-v2")
+    )
+    os.environ["LUDWIG_PRETRAINED_MODELS_DIR"] = str(tmpdir)  # Needs to be an absolute path.
```
tgaddair, line 37: Can you change this to `file://{tmpdir}` so we can test treating the directory like a URI?

ludwig/utils/hf_utils.py
```diff
@@ -40,3 +46,64 @@ def load_pretrained_hf_tokenizer(
...
+            try:
+                logger.info(
+                    f"Found existing pretrained model artifact {pretrained_model_name_or_path} in directory "
+                    f"{pretrained_models_dir}. Downloading."
+                )
+                return (
+                    _load_pretrained_hf_model_from_dir(model_class, pretrained_model_path, **pretrained_kwargs),
+                    False,
+                )
+            except Exception as e:
```
tgaddair, line 102: Would be good to figure out what the correct error type is here, so we don't hide other issues that may be causing things to fail.

tgaddair (commented):
Looks good, just a few nits.

ludwig/utils/hf_utils.py
tgaddair: On second thought, that might be a tall order, so this is fine.

arnavgarg1 (approved):
LGTM! Agree with @tgaddair's comments around the tests :) Great addition!

PR #3212, opened 2023-03-06
Add Timeseries forecasting for column-major data, and introduce Timeseries output feature

tgaddair:
Follow-up to the work done in #1131, but for PyTorch instead of TensorFlow.

This PR adds the following:

- Support for windowing over column-major timeseries data. For timeseries input features, this allows users to provide the timeseries column a ordinary numbers and then look back over a sliding window of `preprocessing.window_size` to form the row-major timeseries input.
- Support for timeseries output feature, with similar support for column-major forecasting when `preprocessing.horizon` is set.
- `LudwigModel.forecast` and `ludwig forecast` command that allows forecasting with a trained model.
- Support for `passthrough` and `dense` encoders with timeseries inputs of constant length.
- `huber` loss as an alternative to MAE / MSE (default for timeseries output).
- `mean_absolute_percentage_error` (MAPE) loss and metric
- Optional `multiplier` to apply in the `projector` decoder to the output after activation.

Example:

- Download and unpack hourly weather data from https://www.kaggle.com/selfishgene/historical-hourly-weather-data
- `ludwig train --config config.yaml --dataset temperature.csv`
- `ludwig forecast -n 10 --model_path results/experiment_run/model --dataset temperature.csv`


ludwig/api.py
```diff
@@ -1068,6 +1071,73 @@ def evaluate(
...
+            preds, _ = self.predict(dataset, skip_save_predictions=True, skip_save_unprocessed_output=True)
+
+            next_series = {}
+            for feature in self.config_obj.output_features:
+                if feature.type == TIMESERIES:
+                    key = f"{feature.name}_predictions"
+                    next_series[feature.column] = pd.Series(preds[key].iloc[-1])
+
+            next_preds = pd.DataFrame(next_series)
+            dataset = pd.concat([dataset, next_preds], axis=0).reset_index(drop=True)
```
w4nderlust, line 1119: this appraoch is totally fine, but obviously we are preprocessing over and over again datapoints we have already processed, so there's something to be optimzied.

Also from a model perspective, I belive HF does something smart for caching activations for the next steps of text generation, which is kinda similar here potentially, so we can look into that roo.

ludwig/decoders/generic_decoders.py
```diff
@@ -146,7 +148,7 @@ def input_shape(self):
         return self.dense.input_shape
 
     def forward(self, inputs, **kwargs):
-        values = self.activation(self.dense(inputs))
+        values = self.activation(self.dense(inputs)) * self.multiplier
```
w4nderlust, line 151: I actually forgot we already had the activation! that was probably for doing softmax / sigmoid for vector features :)

ludwig/encoders/sequence_encoders.py
```diff
@@ -92,16 +93,28 @@ def forward(self, input_sequence, mask=None):
         :type mask: Tensor
         """
         input_sequence = input_sequence.type(torch.float32)
+        batch_size = input_sequence.shape[0]
         while len(input_sequence.shape) < 3:
             input_sequence = input_sequence.unsqueeze(-1)
         hidden = self.reduce_sequence(input_sequence)
 
+        # Output may have [batch_size, s, 1] shape, so ensure it is [batch_size, s]
+        hidden = hidden.reshape((batch_size, -1))
```
w4nderlust, line 102: not sure what we decided for sequence and text (probalby to remove the passthrough encoder) but this may not work in some scnarios (happy to elaborate)

ludwig/api.py
tgaddair: Nice, yeah I think we should definitely think about improvements here. But it definitely depends on how far back / forward we want to forecast. If we're looking at +/- 100 samples this should be fine.

ludwig/encoders/sequence_encoders.py
tgaddair: Hmm, seems so from the tests. Our docs are misleading, then. What do you think is a good workaround in this case?

w4nderlust (approved):

MihailMiller:
I'm trying to use the example provided in examples/forecasting which uses 'timeseries' as the output feature type. However, I encounter a ConfigValidationError indicating that 'timeseries' is not a supported output feature type. Here is the error message:

ludwig.error.ConfigValidationError: Output feature Seattle_next uses an invalid/unsupported output type 'timeseries'. Supported output features: ['binary', 'category', 'number', 'set', 'vector', 'sequence', 'text'].

I cloned the Ludwig repository and installed it in a Python 3.8 environment.
I navigated to examples/forecasting and ran the example using the provided command: "ludwig train --config config.yaml --dataset temperature.csv"

As this is one of the provided examples, I expect it to run without any configuration errors. The 'timeseries' output feature should be supported, according to the example.

    Ludwig version: v0.7.4
    Python version: 3.8

I'd appreciate guidance on how to resolve this issue. Could you confirm whether 'timeseries' is a supported output feature in the current Ludwig version? If it is, what am I doing wrong?

Thank you for your assistance.

MihailMiller:
For anyone else experiencing the same issue in the future, the solution was to clone the current repository state (v0.8) instead of the current release.

tgaddair:
Glad you got it working @MihailMiller! v0.8 should be released in the next couple of weeks with this feature officially supported.

PR #3213, opened 2023-03-06
Standardize all initialization parameters on `Bias/Weights/EmbeddingInitializerField`.

justinxzhao:
Replace `InitializerOptions`, `InitializerOrDict` --> `Bias/Weights/EmbeddingInitializerField`.

This PR includes a smaller intermediate change of https://github.com/ludwig-ai/ludwig/pull/3075 that we can check in first.

Notably, `fc_*` initialization parameters are excluded as moving these to use `InitializerOrDict` causes a spike in memory usage.

### DIFF
--- a/ludwig/schema/combiners/common_transformer_options.py
+++ b/ludwig/schema/combiners/common_transformer_options.py
@@ -3,7 +3,7 @@ from typing import Any, Dict, List, Optional, Union
 from ludwig.api_annotations import DeveloperAPI
 from ludwig.schema import common_fields
 from ludwig.schema import utils as schema_utils
-from ludwig.schema.metadata import COMBINER_METADATA, COMMON_METADATA
+from ludwig.schema.metadata import COMBINER_METADATA
 from ludwig.schema.utils import ludwig_dataclass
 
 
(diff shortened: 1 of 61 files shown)


ludwig/schema/combiners/common_transformer_options.py
```diff
@@ -52,9 +52,17 @@ class CommonTransformerConfig:
         parameter_metadata=COMBINER_METADATA["TransformerCombiner"]["use_bias"],
     )
 
-    bias_initializer: Union[str, Dict] = common_fields.BiasInitializerField()
+    bias_initializer: Union[str, Dict] = schema_utils.InitializerOrDict(
+        default="zeros",
```
tgaddair, line 56: Not sure I understand the goal of these changes. The entire purpose for introducing the `BiasInitializerField` was to:

1. Standardize the description of all bias initializers to include all options, give an informative description, etc.
2. Standardize the default params.
3. Standardize the parameter metadata.

This change seems to be a step backward by reverting those changes, but I'm not sure what the benefit is?

justinxzhao: Thanks the call out.

My main goal here is to standardize more of the config against a consistent set of schema structures for our initialization parameters to prime the schema for initializer parameter schemafication, not realizing that `Bias/Weights/EmbeddingInitializerField` was more evolved than `InitializerOrDict`.

I'll standardize on `Bias/Weights/EmbeddingInitializerField` then, for its better descriptions and parameter metadata.

PR #3214, opened 2023-03-06
Fixes Explain step for tied weights

geoffreyangus:
This PR fixes the explain step for tied weights. Prior to this change, the explain step would fail with the following error:

```
One of the differentiated Tensors appears to not have been used in the graph.
```

We tracked this failure down to the fact that Captum doesn't like it when you pass duplicates of the same torch module into the integrated gradients step.

The fix for this was to create a deep copy of the tied torch module and update the computation graph used during the explanation step to use the deep copy so that every target layer passed into Captum is only forward-passed once per batch.


connor-mccorm (approved):
LGTM, side note, great docstrings and comments - made it very easy to review this PR!

PR #3215, opened 2023-03-07
Add `non_zero` to `common_fields.NumFCLayersField`

abidwael:
A dense encoder with 0 `num_layer`s doesn't make sense. This PR adds the option for specifying a `non_zero` argument in `common_fields.NumFCLayersField`.

### DIFF
--- a/ludwig/schema/common_fields.py
+++ b/ludwig/schema/common_fields.py
@@ -48,6 +48,8 @@ def NumFCLayersField(
         " Increasing layers adds capacity to the model, enabling it to learn more complex feature interactions."
     )
     parameter_metadata = parameter_metadata or COMMON_METADATA["num_fc_layers"]
+
+    # When using a dense encoder, the number of fully connected layers must be strictly greater than 0.
     if non_zero:
         return schema_utils.PositiveInteger(
             default=default, allow_none=False, description=full_description, parameter_metadata=parameter_metadata
--- a/ludwig/schema/common_fields.py
+++ b/ludwig/schema/common_fields.py
@@ -41,6 +41,8 @@ def ResidualField(
 def NumFCLayersField(
     default: int = 0, description: str = None, parameter_metadata: ParameterMetadata = None, non_zero=False
 ) -> Field:
+    assert (not non_zero) or (default > 0 and non_zero)
+
     description = description or "Number of stacked fully connected layers to apply."
     full_description = description + (
         " Increasing layers adds capacity to the model, enabling it to learn more complex feature interactions."
--- a/ludwig/schema/common_fields.py
+++ b/ludwig/schema/common_fields.py
@@ -38,12 +38,18 @@ def ResidualField(
     )
 
 
-def NumFCLayersField(default: int = 0, description: str = None, parameter_metadata: ParameterMetadata = None) -> Field:
+def NumFCLayersField(
+    default: int = 0, description: str = None, parameter_metadata: ParameterMetadata = None, non_zero=False
+) -> Field:
     description = description or "Number of stacked fully connected layers to apply."
     full_description = description + (
         " Increasing layers adds capacity to the model, enabling it to learn more complex feature interactions."
     )
     parameter_metadata = parameter_metadata or COMMON_METADATA["num_fc_layers"]
+    if non_zero:
+        return schema_utils.PositiveInteger(
+            default=default, allow_none=False, description=full_description, parameter_metadata=parameter_metadata
+        )
     return schema_utils.NonNegativeInteger(
         default=default,
         allow_none=False,
--- a/tests/ludwig/config_validation/test_checks.py
+++ b/tests/ludwig/config_validation/test_checks.py
@@ -120,7 +120,7 @@ def test_comparator_fc_layer_config(
         ModelConfig.from_dict(config)
 
 
-def test_dense_binary_encoder_1_layer():
+def test_dense_binary_encoder_0_layer():
     config = {
         "defaults": {"binary": {"encoder": {"norm": "ghost", "num_layers": 0, "output_size": 128, "type": "dense"}}},
         "input_features": [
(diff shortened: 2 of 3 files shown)


ludwig/schema/common_fields.py
```diff
@@ -38,12 +38,18 @@ def ResidualField(
...
+) -> Field:
     description = description or "Number of stacked fully connected layers to apply."
     full_description = description + (
         " Increasing layers adds capacity to the model, enabling it to learn more complex feature interactions."
     )
     parameter_metadata = parameter_metadata or COMMON_METADATA["num_fc_layers"]
+    if non_zero:
+        return schema_utils.PositiveInteger(
+            default=default, allow_none=False, description=full_description, parameter_metadata=parameter_metadata
+        )
```
justinxzhao, line 52: At the top of the function, add an assertion that if the `default=0` then `non_zero` must be `False`.

ludwig/schema/common_fields.py
```diff
@@ -38,12 +38,18 @@ def ResidualField(
...
-def NumFCLayersField(default: int = 0, description: str = None, parameter_metadata: ParameterMetadata = None) -> Field:
+def NumFCLayersField(
+    default: int = 0, description: str = None, parameter_metadata: ParameterMetadata = None, non_zero=False
+) -> Field:
     description = description or "Number of stacked fully connected layers to apply."
     full_description = description + (
         " Increasing layers adds capacity to the model, enabling it to learn more complex feature interactions."
     )
     parameter_metadata = parameter_metadata or COMMON_METADATA["num_fc_layers"]
+    if non_zero:
```
justinxzhao, line 49: Could you add a comment about when it's appropriate to support `NumFCLayers=0`?

tests/ludwig/config_validation/test_checks.py
```diff
@@ -117,3 +118,21 @@ def test_comparator_fc_layer_config(
 
     with pytest.raises(ConfigValidationError) if not expect_success else contextlib.nullcontext():
         ModelConfig.from_dict(config)
+
+
+def test_dense_binary_encoder_1_layer():
```
jeffkinnison, line 123: This says it's a test for 1-layer encoders, but `num_layers` is set to 0. Is that correct? If so, could you add a docstring explaining the test and why we expect this config to be invalid?

tests/ludwig/config_validation/test_checks.py
```diff
@@ -11,6 +11,7 @@
 from typing import Any, Dict, Optional
 
 import pytest
+from marshmallow import ValidationError
```
ksbrar, line 14: You should be able to just use `ConfigValidationError`

ludwig/schema/common_fields.py
ksbrar: +1

ludwig/schema/common_fields.py
ksbrar: +1

ksbrar (changes requested):

ludwig/schema/common_fields.py
abidwael: Adding

tests/ludwig/config_validation/test_checks.py
abidwael: It fails with the marshmallow ValidationError

tests/ludwig/config_validation/test_checks.py
```diff
@@ -117,3 +118,21 @@ def test_comparator_fc_layer_config(
 
     with pytest.raises(ConfigValidationError) if not expect_success else contextlib.nullcontext():
         ModelConfig.from_dict(config)
+
+
+def test_dense_binary_encoder_1_layer():
+    config = {
+        "defaults": {"binary": {"encoder": {"norm": "ghost", "num_layers": 0, "output_size": 128, "type": "dense"}}},
```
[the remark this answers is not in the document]
abidwael: Adding

ludwig/schema/common_fields.py
abidwael: @justinxzhao @ksbrar by the way, this would mean that the test I'm adding will fail with an `AssertionError` and not config validation error

tests/ludwig/config_validation/test_checks.py
ksbrar: That's strange. The code should be wrapping it in a `ConfigValidationError`. CC @justinxzhao

ksbrar: Actually glancing through the code we manually raise a `ValidationError` in quite a few locations. (Worth a separate refactoring.)

ksbrar (approved):

tests/ludwig/config_validation/test_checks.py
tgaddair: ConfigValidationError is only raised at the ModelConfig level.

jeffkinnison (approved):

tests/ludwig/config_validation/test_checks.py
justinxzhao: Agreed it might be good to clarify some of our semantics about ConfigValidationError vs. ValidationError from marshmallow, perhaps moving all Ludwig-specific validation code to `ConfigValidationError`.

@abidwael It would be good to understand how the test gets through raises a marshmallow `ValidationError`. Perhaps it's because we don't `except` it in [this block](https://github.com/ludwig-ai/ludwig/blob/f26222046f8415b465944823ad7d0605ac1f40a2/ludwig/config_validation/validation.py#L62-L66).

Otherwise, LGTM.

justinxzhao (approved):

tests/ludwig/config_validation/test_checks.py
ksbrar: Ah @justinxzhao thanks for the link, I totally missed that <imblind.gif>

PR #3216, opened 2023-03-07
Add auxiliary validation for all features to be present in comparator combiner entities

abidwael:
When using a comparator combiner, every input feature needs to be present in either `entity_1` or `entity_2`.
This PR adds an auxiliary check for that.


ludwig/config_validation/checks.py
```diff
@@ -261,10 +261,10 @@ def check_comparator_combiner_requirements(config: "ModelConfig") -> None:  # no
     """Checks that all of the feature names for entity_1 and entity_2 are valid features."""
     if config.model_type != MODEL_ECD:
         return
-    if config.combiner != "comparator":
+    if config.combiner.type != "comparator":
```
justinxzhao, line 264: Nice catch.

justinxzhao (approved):

jeffkinnison (approved):

PR #3217, opened 2023-03-07
Disable number decoder in the decoder config

abidwael:


tests/ludwig/config_validation/test_checks.py
```diff
@@ -17,6 +17,23 @@
 from tests.integration_tests.utils import binary_feature, text_feature
 
 
+def test_passthrough_number_decoder():
+    config = {
+        "defaults": {"number": {"decoder": {"fc_norm": None, "fc_output_size": 10, "type": "passthrough"}}},
```
jeffkinnison, line 22: Do we want to test setting the passthrough decoder in the output feature as well? I know the feature configs get populated with the defaults behind the scenes, but that could be a nice sanity check.

jeffkinnison (approved):
LGTM! Not going to block merging based on the below comment, it was just a thought I had when reviewing the test.

ludwig/schema/decoders/base.py
```diff
@@ -85,7 +85,7 @@ class BaseDecoderConfig(schema_utils.BaseMarshmallowConfig, ABC):
 
 
 @DeveloperAPI
-@register_decoder_config("passthrough", [BINARY, CATEGORY, NUMBER, SET, VECTOR, SEQUENCE, TEXT])
+@register_decoder_config("passthrough", [BINARY, CATEGORY, SET, VECTOR, SEQUENCE, TEXT])
```
tgaddair, line 88: @arnavgarg1 I thought we had disabled the passthrough decoder?

tests/ludwig/config_validation/test_checks.py
abidwael: added!

ludwig/schema/decoders/base.py
abidwael: I think we removed it from the encoder registry, but not from the schema.

PR #3221, opened 2023-03-07
Implements `sequence_length` param

geoffreyangus:
This PR implements a new config param for text and sequence features, `sequence_length`, a nullable positive integer.

`sequence_length` defaults to `None`. This means that the sequence length for a given feature will be inferred from the dataset. The inferred sequence length will be capped at `max_sequence_length`, which defaults to `256`.

If `sequence_length` is not `None`, then the sequence length for a given feature will be the specified value and samples will be padded and truncated as needed. If the specified value for `sequence_length` is greater than `max_sequence_length`, then experiment will fail fast with an error message recommending that you set `max_sequence_length` to a value greater than or equal to `sequence_length`.

### DIFF
--- a/ludwig/features/sequence_feature.py
+++ b/ludwig/features/sequence_feature.py
@@ -212,6 +212,7 @@ class SequenceFeatureMixin(BaseFeatureMixin):
             ngram_size=preprocessing_parameters["ngram_size"],
             processor=backend.df_engine,
         )
+        logger.info(f"Max length of dataset: {max_length} (without start and stop symbols)")
 
         # Use sequence_length if provided, otherwise use max length found in dataset.
         if preprocessing_parameters["sequence_length"] is not None:
--- a/ludwig/features/sequence_feature.py
+++ b/ludwig/features/sequence_feature.py
@@ -212,7 +212,7 @@ class SequenceFeatureMixin(BaseFeatureMixin):
             ngram_size=preprocessing_parameters["ngram_size"],
             processor=backend.df_engine,
         )
-        logger.info(f"Max length of dataset: {max_length} (without start and stop symbols)")
+        logger.info(f"Max length of feature '{column.name}': {max_length} (without start and stop symbols)")
 
         # Use sequence_length if provided, otherwise use max length found in dataset.
         if preprocessing_parameters["sequence_length"] is not None:
--- a/ludwig/features/sequence_feature.py
+++ b/ludwig/features/sequence_feature.py
@@ -213,15 +213,25 @@ class SequenceFeatureMixin(BaseFeatureMixin):
             processor=backend.df_engine,
         )
 
-        # Use max_sequence_length if provided, otherwise use max length found in dataset.
-        if preprocessing_parameters["max_sequence_length"] is not None:
-            logger.info("Using max_sequence_length provided in preprocessing parameters")
-            max_sequence_length = preprocessing_parameters["max_sequence_length"]
+        # Use sequence_length if provided, otherwise use max length found in dataset.
+        if preprocessing_parameters["sequence_length"] is not None:
+            logger.info(
+                f"Setting max length to sequence_length={preprocessing_parameters['sequence_length']} provided in "
+                f"preprocessing parameters"
+            )
+            max_sequence_length = preprocessing_parameters["sequence_length"]
         else:
-            logger.info("Inferring max_sequence_length from dataset")
             max_sequence_length = max_length + 2  # For start and stop symbols.
-        logger.info(f"Using max sequence length of {max_sequence_length} for feature '{column.name}'")
+            logger.info(f"Setting max length using dataset: {max_sequence_length} (including start and stop symbols)")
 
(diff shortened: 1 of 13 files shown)


ludwig/features/text_feature.py
```diff
@@ -106,17 +106,29 @@ def get_feature_meta(
...
+            max_sequence_length = preprocessing_parameters["sequence_length"]
+            max_sequence_length_99ptile = preprocessing_parameters["sequence_length"]
         else:
-            logger.info("Inferring max_sequence_length from dataset")
             max_sequence_length = max_len + 2  # For start and stop symbols.
             max_sequence_length_99ptile = max_len_99ptile + 2  # For start and stop symbols.
-        logger.info(f"Using max sequence length of {max_sequence_length} for feature '{column.name}'")
+            logger.info(f"Setting max length using dataset: {max_sequence_length} (including start and stop symbols)")
+
+            if preprocessing_parameters["max_sequence_length"] < max_sequence_length:
```
abidwael, line 123: I believe that this will fail `if preprocessing_parameters["sequence_length"] is None and preprocessing_parameters["max_sequence_length"] is None`. Perhaps we could add an aux check for that too.

geoffreyangus: Good point! Conceptually, the case where they are both None just means that we infer from the dataset entirely (without setting a hard cap on the text length). I'll be sure this is reflected in the code (you are right that this would fail as-is).

geoffreyangus:
Good point @justinxzhao – updated the PR!

justinxzhao (approved):
Nice!

w4nderlust:
The lgoic looks fine to me.
One thing that I don't see (please point me to it if I just missed it) is the actual solution of the original issue.
In my understanding this was needed because of tied encoders, and:
1. I d like to understand how this does it. i believe it boils down to the user specifying sequence length explicitly for all tied features to the same value, but maybe I'm wrong
2. I see no test that actually tests that this solution solves that problem

abidwael (approved):

geoffreyangus:
@w4nderlust, thanks for the comments– I've added a new test that confirms this fix addresses [this GitHub issue](https://github.com/ludwig-ai/ludwig/issues/3220), as well as updated the schema so that the description of `tied` recommends setting `sequence_length` if using sequence and text features with a `sequence` combiner.

The unit test that works on the `explain` workflow is still in progress, since there seems to be nested errors having to do with tied weights. Using `sequence_length` resolves on of the errors, but not the next one, so some more investigation is needed there. Let me know if you had any other thoughts here!

ludwig/features/sequence_feature.py
```diff
@@ -212,6 +212,7 @@ def get_feature_meta(
             ngram_size=preprocessing_parameters["ngram_size"],
             processor=backend.df_engine,
         )
+        logger.info(f"Max length of dataset: {max_length} (without start and stop symbols)")
```
w4nderlust, line 215: "of dataset" -> "of feature {feature_name}"?

ludwig/features/text_feature.py
```diff
@@ -106,6 +106,7 @@ def get_feature_meta(
             padding_symbol,
             unknown_symbol,
         ) = tf_meta
+        logger.info(f"Max length of dataset: {max_len} (without start and stop symbols)")
```
w4nderlust, line 109: "of dataset" -> "of feature {feature_name}"?

w4nderlust:
@tgaddair @justinxzhao FYI

Added a couple minor comments. one thing I haven't checked though is backwards compatibility, so be mindful to check if this change makes models with the old `max_sequence_length` parameter obsolete. 

overall this looks good for now, but I still believe that this could be solved differently and potentially better in the future. Writing here so that there's a record of it (maybe we can create an issue somewhere to capture it too).

A couple options:
1. creating another instance of the same encoder, with all the same parameters but sequence length, and then setting manually the weights to point to the same weights of the tied encoder (as the weights shapes themselves shouldn't depend on the sequence length)
2. adding a global preprocessing parameter like `concatenate_text_features` that when multiple text / sequence features are specified creates a derived column that concatenates them and uses that (with a single encoder) instead of the original ones. It's arguable if the default would need to be true or false. Also In most cases it would be simpler and cleaner to do it before Ludwig, but still this would solve the current issue too.

Both things could complement the current solution instead of replacing them. Moreover, I like the idea of an explicit sequence length parameter as it's similar to what we do for images height and width, so even if in the future we were to implement 1 and 2, this work is still valuable and still holds.

geoffreyangus:
Sounds good, thanks @w4nderlust! Regarding backwards compatibility, we should be in the clear. The `max_sequence_length` parameter still defaults to `256`, and `sequence_length` here defaults to `None`. Given those two values, the derived `max_sequence_length` will be equal to the `min(preprocessing_parameters['max_sequence_length'], max_len)`, where `max_len` is the max length of the feature in the dataset, which is what we had before this change.

Will address your comments and merge after tests pass. Thanks!

PR #3222, opened 2023-03-07
Tagger decoder config override and auxiliary validation checks

arnavgarg1:
This PR ensures that tagger decoders for text and sequence features are used correctly within Ludwig. Here are the the main changes addressed by this PR:

1. If a config has no text/sequence/timeseries input features but uses a text/sequence tagger decoder output feature, it will raise a ConfigValidationError. 
2. If at least one text/sequence/timeseries feature does not use `reduce_output=None`, then it will raise a ConfigValidationError because the tagger decoder won't receive a b x s x h (3D) input from the combiner.
3. The tagger decoder requires `reduce_input` for text/sequence output features to be set to None to prevent the hidden tensors from being reshaped to 2D. The default value for `reduce_input` is set to `sum` since it is the more popular case, but will be overridden during ModelConfig object creation when a tagger decoder is being used, and logs a warning. This is because this is an implementation detail Ludwig users should not have to worry about since it relates to tensor shapes directly.

I've also added a few different tests for text and sequence output features, as well as a test for a config that does not have a text/sequence/timeseries input feature to make sure all of these different cases are correctly tested and accounted for.


ludwig/config_validation/checks.py
```diff
@@ -374,3 +374,36 @@ def is_divisible(hidden_size: int, num_heads: int) -> bool:
...
+    for output_feature in config.output_features:
+        if output_feature.type in {TEXT, SEQUENCE} and output_feature.decoder.type == "tagger":
+            output_feature_with_tagger_decoder = True
+
+    if not output_feature_with_tagger_decoder:
+        return
+
+    # Check that there is at least one sequence, text or timeseries input feature that doesn't reduce the
+    # output of the encoder.
+    one_feature_with_reduce_output_none = False
```
justinxzhao, line 394: nit: Can we make do without this additional boolean?

justinxzhao (approved):

abidwael (approved):

ludwig/config_validation/checks.py
arnavgarg1: Done!

w4nderlust (approved):

PR #3224, opened 2023-03-08
feat: Allow users to change validation strictness policy

ksbrar:
Resolves https://github.com/ludwig-ai/ludwig/issues/3218

Needs more testing atm.
--

Ludwig will now check if a user has an environment variable called `LUDWIG_SCHEMA_POLICY` set and, if so, use its value as the strictness policy for validation.

The only supported values are `include`, `exclude`, and `raise`. Is there a slick way to raise an error or otherwise warn the user about this in the CLI?

### DIFF
--- a/ludwig/schema/utils.py
+++ b/ludwig/schema/utils.py
@@ -1,4 +1,5 @@
 import copy
+import os
 import warnings
 from abc import ABC, abstractmethod
 from dataclasses import field, Field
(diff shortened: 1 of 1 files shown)


tgaddair:
Just saw this (resolves must not be a keyword GH recognizes). I put together #3226 which is similar, but only checks the env var once. So might make sense to go with that implementation.

PR #3225, opened 2023-03-08
follow up: unregister passthrough decoder for all features

abidwael:
Following up on
- https://github.com/ludwig-ai/ludwig/pull/3151
- https://github.com/ludwig-ai/ludwig/pull/3217

Disabling needs to happen both at the decoder and schema config levels.


arnavgarg1 (approved):
Good catch!

PR #3226, opened 2023-03-08
feat: Add env var `LUDWIG_SCHEMA_VALIDATION_POLICY` to change marshmallow validation strictness

### LINKED ISSUE #3218: [FR] Allow users to set the unmarshalling mode as RAISE for BaseConfig
Currently, when loading the user config, Ludwig allows unknown attributes to be set. There is a TODO item in the code at `ludwig/schema/utils.py:161`

```
unknown = INCLUDE  # TODO: Change to RAISE and update descriptions once we want to enforce strict schemas.
```

Since changing this to RAISE by default might impact lots of users I still believe there's value in having this set up as RAISE. That being said it'd be great if users can overwrite this default via an environment variable.

**Describe the use case**
Allow users to enforce the schema on their config.

The ability to overwrite the default behaviour of INCLDUE

I've tried monkeypatching ludwig. Although it's easily doable for a local environment, it becomes harder when sharing the monkeypatch with a team that runs everything in AWS Sagemaker.

**Additional context**
--

tgaddair:
Closes #3218.


ksbrar (commented):
My only question is how extensively have you tested this locally? (Last thing I was fiddling around with pytest env variables and was about to try Ludwig CLI).

In particular, I want to check if there are any breakages when `RAISE` is set for schemas that explicitly allow additional params via JSON. That means (based on a ctrl-f):
* input features
* output features
* preprocessing
* augmentation

Really, just one of these cases suffices as a sanity check for the rest. 

(Also, I believe we're covered but we have to be careful that any internal parameters are in fact declared on their respective schemas (e.g. `proc_column` for input features) - otherwise `RAISE` will reject them.)

tgaddair:
It's a good point @ksbrar, I haven't tested this extensively! If you have some cycles to test it out, please take a look and let me know what you find.

ksbrar:
I've tested out this PR. It doesn't work as is because the deprecation warning PR I merged a little while ago is actually performing the wrong task and is too aggressive in filtering parameters.

This code (https://github.com/ludwig-ai/ludwig/pull/3118) will filter out and log any and every unknown parameter as deprecated - which is the incorrect thing to do, as there is a clear distinction between a purely "unknown" or invalid parameter and a "deprecated" one - before marshmallow even has a chance to raise an error about the unknown parameters.

I have pushed a simple fix for this PR that gets us to what we want (a user sets this environment flag to `raise`, and they will see marshmallow validation errors for unknown parameters), but I want to sync with @justinxzhao (or @tgaddair ) about how deprecated parameters should be marked and parsed.

justinxzhao:
Synced offline -- upgrading the config, removing known deprecated parameters, and logging warnings for them is handled by `upgrade_config_dict_to_latest_version`.

We should be good to proceed with this PR (just blocked on writing a test).

Follow-up PRs to continue progress towards making `STRICT` the default policy for Ludwig configs:
1. Make `STRICT` the default
    - Clean up preload deprecation logging (https://github.com/ludwig-ai/ludwig/pull/3118) as it's no longer needed
    - Set policy to `STRICT` by default
    - set `additionalProperties` on all jsonschema to False
    - Remove os.environ var
2. Remove explicit call to JSON validation and rely purely on Marshmallow deserialziation
    - Remove call to `check_schema()` in `model_types::base.py` (assuming parity)
    - Keep `check_schema()` around for unit tests to verify pure-JSON-validation behavior.

ksbrar:
@justinxzhao I spent some time and couldn't get the environment variable switching to work properly. If someone else wants to give it a go, here is some rough code showing a couple different attempts. I can clean it up/give some clarification if need be:

```
def rreload(module):
    from importlib import reload

    """Recursively reload modules."""
    reload(module)
    for attribute_name in dir(module):
        attribute = getattr(module, attribute_name)
        if type(attribute) is ModuleType:
            rreload(attribute)


# @mock.patch.dict(os.environ, {LUDWIG_SCHEMA_UNKNOWN_POLICY: "raise"})
def test_user_config_validation_policy():
    from importlib import reload, import_module
    import sys
    from ludwig.constants import LUDWIG_SCHEMA_VALIDATION_POLICY

    # from ludwig.schema.model_types import base

    config = {
        "input_features": [{"name": "foo", "type": "category"}],
        "output_features": [category_feature(encoder={"vocab_size": 2}, reduce_input="sum")],
        "combiner": {"type": "concat", "output_size": 14},
        "trainer": {"unknown": "parameter"},
    }

    # Test default policy (unknown parameters are ignored):
    # ModelConfig.from_dict(config)

    # reload(lsb)

    print(os.environ.get(LUDWIG_SCHEMA_VALIDATION_POLICY))
    os.environ[LUDWIG_SCHEMA_VALIDATION_POLICY] = RAISE
    print(os.environ.get(LUDWIG_SCHEMA_VALIDATION_POLICY))
    # reload(sys.modules["ludwig.schema.model_types.base"])
    # reload(sys.modules["ludwig.schema.model_types.ecd"])
    # reload(sys.modules["ludwig.schema.utils"])
    # reload(sys.modules["ludwig.schema"])

    # mod = import_module("ludwig.schema.model_types.base")
    import ludwig.schema.model_types.base as base

    rreload(base)

    # for name, module in sys.modules.items():
    #     if "ludwig.schema" in name:
    #         reload(module)

    # Test RAISE policy:
    # monkeypatch.setenv(LUDWIG_SCHEMA_UNKNOWN_POLICY, "raise")
    with pytest.raises(ConfigValidationError):
        base.ModelConfig.from_dict(config)
```

But it's easy enough to verify locally in a terminal, so I think we can proceed with just landing this PR.

ksbrar (approved):
Good to go from my standpoint.

justinxzhao (approved):

PR #3227, opened 2023-03-08
Refactor loss implementation to use the schema config for all parameters

tgaddair:
Part of a broader refactor towards "full schemaficiation" whereby the config will be plumbed through as many layers of the stack as possible to avoid maintaining separate sources of truth for parameter defaults, etc.


connor-mccorm (approved):
LGTM!

PR #3228, opened 2023-03-09
Refactored combiner registry and broke circular dep with schema

tgaddair:


ksbrar (commented):
So can we get rid of this now? https://github.com/ludwig-ai/ludwig/blob/397e46b44a2e39fc8f6c81b94b398c491ff671c0/ludwig/config_validation/validation.py#L26

tgaddair:
@ksbrar good call!

connor-mccorm (approved):
LGTM!

tests/integration_tests/test_custom_components.py
```diff
@@ -61,7 +55,15 @@ class CustomLossConfig(BaseLossConfig):
     type: str = "custom_loss"
 
 
-@register_combiner(name="custom_test")
+@register_combiner_schema("custom_combiner")
```
ksbrar, line 58: Is this going to end up in the combiner registry? (No, right? - since it's under `tests`)

tgaddair: That's right.

ksbrar (approved):
LGTM. Looks like the tests are OOMing 😞

tgaddair:
Yep, gonna see what's up with that :(

PR #3231, opened 2023-03-10
Add columns and data types to ludwig datasets

connor-mccorm:


ludwig/datasets/configs/poker_hand.yaml
```diff
@@ -28,6 +16,29 @@ description: |
...
+  - name: S4
+    type: category
+  - name: C4
+    type: category
+  - name: S5
+    type: category
+  - name: C5
+    type: category
+  - name: hand
+    type: category
```
arnavgarg1, line 41: From internal benchmarks, I'd make all of these number types since the model seems to perform better, but leave the output as category since that is correct!

connor-mccorm: Sounds good, will change!

arnavgarg1 (approved):

PR #3234, opened 2023-03-11
Fixes torch DDP distributed metric computation for AUROC

geoffreyangus:
This PR ensures that AUROC can be computed for the torch DDP strategy. Before this change, `ludwig.modules.metric_modules.BinaryAUROCMetric` was not being instantiated correctly, meaning that the override for `BinaryAUROCMetric.update` was not called. This is in turn meant that the target was not being correctly cast into a boolean.

The fix was directly subclassing `torchmetrics.classification.BinaryAUROC`. Before this change, we were subclassing `torchmetrics.AUROC`, which overrides the `__new__` method and messes up class inheritance.

Once class inheritance was working correctly, we unveiled an issue with `horovod_utils.all_gather_tensors`. The issue was previously hidden because of the broken class inheritance issue. This issue meant that we were actually using torchmetrics' `all_gather_tensors` function for binary AUROC computation.

The proposed fix for `horovod_utils.all_gather_tensors` was to ensure that the returned list of tensors matched the shape of those returned by torchmetrics' `all_gather_tensors`.


ludwig/api.py
```diff
@@ -1425,7 +1425,9 @@ def preprocess(
 
             return PreprocessedDataset(proc_training_set, proc_validation_set, proc_test_set, training_set_metadata)
         except Exception as e:
-            raise RuntimeError(f"Caught exception during model preprocessing: {str(e)}") from e
+            raise RuntimeError(
+                f"Caught exception during model preprocessing: {str(e)}.\nFull traceback: {traceback.format_exc()}"
+            ) from e
```
tgaddair, line 1430: You should already get this with the `from e`, just need to scroll up a bit.

ludwig/features/base_feature.py
```diff
@@ -352,7 +353,10 @@ def get_metrics(self):
             try:
                 metric_vals[metric_name] = get_scalar_from_ludwig_metric(metric_fn)
             except Exception as e:
-                logger.error(f"Caught exception computing metric: {metric_name}. Exception: {e}")
+                logger.error(
```
tgaddair, line 356: This should just be:

```
logger.exception(f"Caught exception computing metric: {metric_name}")
```

Then you get the stack trace for free.

ludwig/utils/horovod_utils.py
```diff
@@ -54,33 +54,51 @@ def gather_all_tensors(result: torch.Tensor, group: Optional[Any] = None) -> Lis
...
     # sync and gather all
-    gathered = _HVD.allgather(result)
-    gathered_result = list(gathered.split(1, dim=0))
+    gathered_result = _HVD.allgather(result)
+
+    print("\tgathered_result:", gathered_result)
+
+    # This is to match the output of the torchmetrics gather_all_tensors function
+    # and ensures that the return value is usable by torchmetrics.compute downstream.
+    if len(result.shape) >= 2:
```
tgaddair, line 91: Interesting, so my understanding is that DDP allgather contract is like:

```
tensor[*shape] -> [tensor[*shape]] * world_size
```

While Horovod is:

```
tensor[shape] -> tensor[world_size, *shape]
```

In the previous implementation, we split the tensor along the `rank` dimension (dim=0) into `[tensor[shape]] * world_size` to match the DDP format. But with this change it seems if the input is a scalar, meaning the allgathered output is a 1D vector, then we just return it as-is (unless it's a bool, in which case we iterate over it and turn it into a list of bool tensors).

Seems like the output format is potentially inconsistent then, right? It could be a list of tensors or it could be a single tensor. Am I missing something.

The other aspect here is the difference between casting to a list vs calling split. Running this in a terminal, I get:

```
>>> t
tensor([[1., 1., 2.],
        [2., 3., 1.],
        [5., 2., 4.]])
>>> list(t)
[tensor([1., 1., 2.]), tensor([2., 3., 1.]), tensor([5., 2., 4.])]
>>> t.split(1, dim=0)
(tensor([[1., 1., 2.]]), tensor([[2., 3., 1.]]), tensor([[5., 2., 4.]]))
```

So it seems like the difference here is that `split` preserves the rank of the tensor, while `list` removes the rank that's being split upon entirely. So that does seem to better align with the DDP format.

As such, is the right thing to do here just to change lines 77 and 78 to:

```
gathered = _HVD.allgather(result)
gathered_result = list(gathered)
```

In other words, why only do this when `len(result.shape) >= 2`?

Otherwise, it makes sense, as I do believe DDP allgather does not add an extra dimension to the tensors being gathered, as the code seemed to be doing previously.

ludwig/api.py
geoffreyangus: Got it! Will revert this change.

ludwig/utils/horovod_utils.py
geoffreyangus: Oh interesting– so double-checking this, it seems like the guard I've added is pretty much always triggered. I believe this is because of these two lines:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/horovod_utils.py#L63-L66

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/horovod_utils.py#L73-L75

Tensors always have a rank of at least one because of the first code block, and the second code block ensures that the rank is at least two. I can remove the guard, but FYI I don't think it will make too much of a difference.

ludwig/features/base_feature.py
tgaddair: `logger.exception != logger.error`

geoffreyangus: Ah, didn't catch that. Okay will try that. Thanks!

ludwig/modules/metric_modules.py
```diff
@@ -142,7 +144,8 @@ def __init__(self, **kwargs):
 
 
 @register_metric(ROC_AUC, [BINARY], MAXIMIZE, PROBABILITIES)
-class BinaryAUROCMetric(AUROC, LudwigMetric):
+class BinaryAUROCMetric(BinaryAUROC, LudwigMetric):
+    # class BinaryAUROCMetric(AUROC, LudwigMetric):
```
tgaddair, line 148: Remove commented code

ludwig/utils/horovod_utils.py
```diff
@@ -54,33 +54,51 @@ def gather_all_tensors(result: torch.Tensor, group: Optional[Any] = None) -> Lis
     :return: list with size equal to the process group where gathered_result[i]
              corresponds to result tensor from process i
     """
+    print("ASDFASDF inside gather_all_tensors")
     if group is not None:
         raise ValueError("Horovod does not support allgather using a subcommunicator at this time. " "Unset `group`.")
 
+    print("\tgroup is None")
```
tgaddair, line 61: Remove debug prints

tgaddair (approved):
LGTM

arnavgarg1:
Should we add this for 0.7.3?

PR #3235, opened 2023-03-11
Adds support for TEXT features when using GBM with tf-idf encoder

tgaddair:
Introduces the following:

- `tf_idf` text encoder that computes [Tf-Idf](https://en.wikipedia.org/wiki/Tf%E2%80%93idf) on tokenized text features.
- Made `tf_idf` servable in TorchScript by using pure-PyTorch implementation of the embedding.
- Enabled text input features for GBMs and made `tf_idf` the default encoder.

Example usage:

```
input_features:
  - name: message
    type: text
    encoder:
      type: tf_idf
```

### DIFF
--- a/ludwig/features/bag_feature.py
+++ b/ludwig/features/bag_feature.py
@@ -43,7 +43,7 @@ class BagFeatureMixin(BaseFeatureMixin):
     def get_feature_meta(
         column, preprocessing_parameters: PreprocessingConfigDict, backend, is_input_feature: bool
     ) -> FeatureMetadataDict:
-        idx2str, str2idx, str2freq, max_size, _, _, _, _ = create_vocabulary(
+        vocabulary = create_vocabulary(
             column,
             preprocessing_parameters["tokenizer"],
             num_most_frequent=preprocessing_parameters["most_common"],
@@ -51,11 +51,11 @@ class BagFeatureMixin(BaseFeatureMixin):
             processor=backend.df_engine,
         )
         return {
-            "idx2str": idx2str,
-            "str2idx": str2idx,
-            "str2freq": str2freq,
-            "vocab_size": len(str2idx),
-            "max_set_size": max_size,
+            "idx2str": vocabulary.vocab,
+            "str2idx": vocabulary.str2idx,
+            "str2freq": vocabulary.str2freq,
+            "vocab_size": len(vocabulary.str2idx),
+            "max_set_size": vocabulary.line_length_max,
         }
 
     @staticmethod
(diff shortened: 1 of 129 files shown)


ludwig/features/bag_feature.py
```diff
@@ -43,19 +43,19 @@ def cast_column(column, backend):
     def get_feature_meta(
         column, preprocessing_parameters: PreprocessingConfigDict, backend, is_input_feature: bool
     ) -> FeatureMetadataDict:
-        idx2str, str2idx, str2freq, max_size, _, _, _, _ = create_vocabulary(
+        vocabulary = create_vocabulary(
```
justinxzhao, line 46: Nice cleanup.

ludwig/models/embedder.py
```diff
@@ -27,7 +28,14 @@ def __init__(self, feature_configs: List[FeatureConfigDict], metadata: TrainingS
...
         for feature in feature_configs:
             feature_cls = get_from_registry(feature[TYPE], get_input_type_registry())
-            feature_obj = feature_cls.get_schema_cls().from_dict(feature)
+
+            # TODO(travis): this assumes ECD is the selected model type, which is not problematic for now, as
+            # the only thing GBM vs ECD changes is the default encoder, but luckily at this point the encoder types
+            # have been fully materialized. However, this could change in the future as ECD and GBM feature configs
+            # diverge, so we should find a way to remove this. The best solution is to the change the input params from
+            # FeatureConfigDict types to BaseInputFeatureConfig types, which will require a refacto of preprocessing to
+            # us the schema, not the dict types.
```
justinxzhao, line 37: nit: "refactor", "use"

ludwig/schema/encoders/text_encoders.py
```diff
@@ -2960,6 +2963,12 @@ def module_name():
 class AutoTransformerConfig(HFEncoderConfig):
     """This dataclass configures the schema used for an AutoTransformer encoder."""
 
+    def __post_init__(self):
+        if self.pretrained_model_name_or_path is None:
+            raise ConfigValidationError(
+                "`pretained_model_name_or_path` must be specified for encoder: `auto_transformer`."
+            )
```
justinxzhao, line 2970: Interesting.

This does seems potentially cleaner than as an aux check as with the aux check you would need to first parse through the config to find text features that use the "auto_transformer" encoder.

Potentially this would be the preferred convention for validation checks for configs with incompatible interdependent parameter values within the same section of the config, like `check_training_runway`.

Does `base.py::ModelConfig`'s `__post_init__` get called after this one?

justinxzhao (approved):
Looks great! Just one nit and one question.

ludwig/schema/encoders/text_encoders.py
tgaddair: Yes, that's right. I like this approach for two reasons:

1. As you said, it means we don't need to parse through the config to hunt for these features, which makes things better encapsulated.
2. If in the future we want users to interact with the config directly by instantiating these schema components, then being able to perform validation at the component level (rather than just the config level) will make it more user-friendly.

PR #3242, opened 2023-03-14
fix: transform `onehot` encoder outputs to float32 tensor

abidwael:
When we specify a `onehot` encoder for category features and have dense layers down the line, we get the following error
```
RuntimeError: mat1 and mat2 must have the same dtype
```
This PR makes sure to convert the outputs of the `onehot` encoder to `float32`

### DIFF
--- a/tests/ludwig/config_validation/test_checks.py
+++ b/tests/ludwig/config_validation/test_checks.py
@@ -198,20 +198,3 @@ def test_dense_binary_encoder_0_layer():
     }
     with pytest.raises(ConfigValidationError):
         ModelConfig.from_dict(config)
-
-
-def test_onehot_category_encoder():
-    config = {
-        "defaults": {"category": {"encoder": {"type": "onehot"}}},
-        "input_features": [
-            {"name": "MSSubClass", "type": "category"},
-            {"name": "MSZoning", "type": "category"},
-            {"name": "Street", "type": "category"},
-            {"name": "Neighborhood", "type": "category"},
-        ],
-        "model_type": "ecd",
-        "output_features": [{"name": "SalePrice", "type": "number"}],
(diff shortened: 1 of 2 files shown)


tests/ludwig/config_validation/test_checks.py
```diff
@@ -198,3 +198,20 @@ def test_dense_binary_encoder_0_layer():
...
+            {"name": "MSZoning", "type": "category"},
+            {"name": "Street", "type": "category"},
+            {"name": "Neighborhood", "type": "category"},
+        ],
+        "model_type": "ecd",
+        "output_features": [{"name": "SalePrice", "type": "number"}],
+        "trainer": {"train_steps": 1},
+        "combiner": {"type": "concat", "num_fc_layers": 2},
+    }
+    ModelConfig.from_dict(config)
```
arnavgarg1, line 217: @abidwael does just doing `.from_dict()` repro the error? I would image this would require at least 1-2 steps of training for the error in your PR comment to be reproed

arnavgarg1 (changes requested):

tests/ludwig/config_validation/test_checks.py
abidwael: Agreed, it doesn't. Will remove this test as it will be captured in https://github.com/ludwig-ai/ludwig/pull/2991

arnavgarg1: I would probably add a test here as well since these will run on every commit, whereas the tests in #2991 will run on merge to master? or is that not the case?

abidwael: It will run on every commit

arnavgarg1 (approved):

PR #3249, opened 2023-03-15
Improve image/audio read throughput by 50% for image/audio features using Daft

arnavgarg1:
There are many advantages to do this from a code maintainability perspective:

1. We can get rid of a lot of custom custom and unnecessary Dask transformations
2. Still get to use Ray, but Daft is much more optimized for data querying and aggregations.
3. We only use Daft for image/audio reads

----

In the local single-partition Ray cluster setup: Daft is a 50% speedup over the existing Ludwig code for URL downloading, and an overall 40% speedup for all of preprocessing.

In the distributed multi-partition Ray cluster setup: Daft is a 50% speedup over the existing Ludwig code for URL downloading, and an overall 40% speedup for all of preprocessing.

See [benchmarking doc](https://docs.google.com/document/d/1gu2Zi75v0SVVIVm8lqU8rLH9GZ9Mx4kCeyTLgMZrXsk/edit?usp=sharing) for more details.

---

Co-authored-by: @jaychia


ludwig/backend/ray.py
```diff
@@ -889,59 +886,69 @@ def supports_multiprocessing(self):
...
-
-            # The resulting column is named "value", which is a dict with two keys: "idx" and "data".
-            ds = ray.data.read_datasource(BinaryIgnoreNoneTypeDatasource(), **read_datasource_fn_kwargs)
-            df = self.df_engine.from_ray_dataset(ds)
-            # Persist the dataframe to prevent re-reading binary files on each subsequent map_objects call
-            df = self.df_engine.persist(df)
-            df["idx"] = self.df_engine.map_objects(df["value"], lambda row: int(row["idx"]))
-            df["value"] = self.df_engine.map_objects(df["value"], lambda row: row["data"])
-            df = df.rename(columns={"value": column.name})
+            import daft
```
jppgks, line 898: might want to put this in a try/except block telling the user to install the `distributed` extra

jppgks (approved):

ludwig/backend/ray.py
arnavgarg1: Nice! @jppgks totally agree, do you want to push up a change with the try-except block?

ludwig/backend/ray.py
```diff
@@ -904,61 +919,85 @@ def supports_multiprocessing(self):
...
+
+                if map_fn is not None:
+                    df = df.with_column(column.name, df[column.name].apply(map_fn, return_dtype=daft.DataType.python()))
+
+                # Executes and convert Daft Dataframe to Dask DataFrame - note that this preserves partitioning
+                if is_dask_df:
+                    df = df.to_dask_dataframe()
+                    df = self.df_engine.persist(df)
+                else:
+                    df = df.to_pandas()
```
tgaddair, line 986: Is this only needed for the Modin df engine? I think Pandas shouldn't use this codepath at all, right?

requirements_distributed.txt
```diff
@@ -8,6 +8,10 @@ tensorboardX<2.3
 GPUtil
 tblib
 awscli
+modin[ray]
```
tgaddair, line 11: We had modin in extras since it's not a default option. Do we need it here? Can we keep it in `requirements_extra.txt`?

requirements_distributed.txt
```diff
@@ -8,6 +8,10 @@ tensorboardX<2.3
...
+getdaft[ray]
 
 # https://github.com/databrickslabs/dolly/issues/125
 deepspeed<0.9
+modin[ray]
+getdaft[ray]
```
tgaddair, line 17: Why are these listed twice?

ludwig/backend/ray.py
```diff
@@ -904,61 +919,85 @@ def supports_multiprocessing(self):
...
+            assert ray.is_initialized(), "Ray should be initialized by Ludwig already at application start"
+            daft.context.set_runner_ray(address="auto", noop_if_initialized=True)
+
+            # Convert Dask Series to Dask Dataframe
+            df = column.to_frame(name=column.name)
+            df["idx"] = column.index
+
+            is_dask_df = is_dask_series_or_df(df, self)
+
+            with self.storage.cache.use_credentials():
```
tgaddair, line 957: Why are we using the cache credentials here, are we reading / writing to the cache?

ludwig/backend/ray.py
```diff
@@ -904,61 +919,85 @@ def supports_multiprocessing(self):
...
-            # The resulting column is named "value", which is a dict with two keys: "idx" and "data".
-            ds = ray.data.read_datasource(BinaryIgnoreNoneTypeDatasource(), **read_datasource_fn_kwargs)
-            df = self.df_engine.from_ray_dataset(ds)
-            # Persist the dataframe to prevent re-reading binary files on each subsequent map_objects call
-            df = self.df_engine.persist(df)
-            df["idx"] = self.df_engine.map_objects(df["value"], lambda row: int(row["idx"]))
-            df["value"] = self.df_engine.map_objects(df["value"], lambda row: row["data"])
-            df = df.rename(columns={"value": column.name})
+            try:
+                import daft
```
jeffkinnison, line 937: Are we importing daft here to avoid an ImportError during local training?

ludwig/backend/ray.py
```diff
@@ -904,61 +919,85 @@ def supports_multiprocessing(self):
...
+
+            # Set the runner for executing Daft dataframes to a Ray cluster
+            # Prevent re-initialization errors if the runner is already set
+            # This can happen if there are 2 or more audio/image columns
+            assert ray.is_initialized(), "Ray should be initialized by Ludwig already at application start"
+            daft.context.set_runner_ray(address="auto", noop_if_initialized=True)
+
+            # Convert Dask Series to Dask Dataframe
+            df = column.to_frame(name=column.name)
+            df["idx"] = column.index
```
tgaddair, line 953: Would be good to get a brief comment on why this line is needed.

ludwig/backend/ray.py
arnavgarg1: I wasn't sure what else to use to get the right credentials for reading image data from S3 where the user has their own credentials. Would it instead be default credentials?

```
cred_inputs = {
            DEFAULTS: defaults,
            ARTIFACTS: load_creds(artifacts),
            DATASETS: load_creds(datasets),
            CACHE: load_creds(cache),
        }
```

ludwig/backend/ray.py
arnavgarg1: Yes

arnavgarg1: But I'm going to move this to general requirements since image processing should be possible even when we don't have distributed/ray.

ludwig/backend/ray.py
arnavgarg1: Actually it's needed to convert from Daft back to either Dask or Pandas, so even Pandas would hit this code path

ludwig/backend/ray.py
arnavgarg1: Added

arnavgarg1: See https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/dataframe_methods/daft.DataFrame.to_dask_dataframe.html

ludwig/backend/ray.py
arnavgarg1: Before, we created this FSSpec object and passed it into the RayDataSourceReader

```
fs, _ = get_fs_and_path(sample_fname)

read_datasource_fn_kwargs = {
       "path_and_idxs": list(zip(fnames, idxs)),
       "filesystem": PyFileSystem(FSSpecHandler(fs)),
}
```

arnavgarg1: Maybe we don't need this? But then I don't understand how it was working before..

justinxzhao (approved):
Excited for the speedups!!

jppgks (approved):

PR #3251, opened 2023-03-15
Filter entities from comparator combiner when not listed in input_features

tgaddair:


arnavgarg1 (approved):

jppgks (approved):

ksbrar:
Wondering why this behavior is not enforced by this validation check: https://github.com/ludwig-ai/ludwig/blob/d5f61eb5c4a8bde671b99414b8384bee91294cfb/ludwig/config_validation/checks.py#L194

tgaddair:
@ksbrar that would validate the config, but this PR makes the behavior more permissive by letting these configs go through.

ksbrar:
Ah gotcha, was thinking backwards - purpose IS to make the parameter more permissive 👍

tests/ludwig/config_validation/test_checks.py
```diff
@@ -198,3 +198,33 @@ def test_dense_binary_encoder_0_layer():
...
+            {"name": "out1", "type": "binary"},
+        ],
+        "combiner": {
+            "type": "comparator",
+            "entity_1": entity_1,
+            "entity_2": entity_2,
+        },
+    }
+
+    with pytest.raises(ConfigValidationError) if not expected else contextlib.nullcontext():
```
justinxzhao, line 227: Maximally parameterizing tests to minimize code duplication is tempting, but it can make tests more difficult to read and understand when it involves adding logic to the test itself.

See ["Avoid logic in tests"](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests) from Microsoft's testing best practices guide.

The behavior we are testing for is more readable/understandable as three separate tests, imo:

```python
def test_comparator_combiner_raises_missing_features():
def test_comparator_combiner_removes_extra_features():
def test_comparator_combiner_raises_duplicated_features():
```

tgaddair: Can't say I agree. I was thinking of adding a handful of additioanl tests to this. Parametrizing makes this super easy. Separate tests means in practice, no one's ever going to bother.

PR #3252, opened 2023-03-15
Removes invalid keys from GBM defaults in the schema

arnavgarg1:
Title :) 

Essentially `encoder`, `decoder` and `loss` are not valid keys for GBMs.

### DIFF
--- a/ludwig/utils/backward_compatibility.py
+++ b/ludwig/utils/backward_compatibility.py
@@ -811,8 +811,15 @@ def upgrade_defaults_config_for_gbm(config: ModelConfigDict) -> ModelConfigDict:
     defaults = copy.deepcopy(defaults_ref)
     gbm_feature_types = GBMDefaultsConfig.Schema().fields.keys()
     for feature_type in defaults_ref:
+        # GBM only supports binary, number, category and text features
         if feature_type not in gbm_feature_types:
             del defaults[feature_type]
+            continue
+
+        # Remove encoder, decoder and loss from defaults since they only apply to ECD
+        defaults[feature_type].pop("encoder", None)
+        defaults[feature_type].pop("decoder", None)
+        defaults[feature_type].pop("loss", None)
     config[DEFAULTS] = defaults
     return config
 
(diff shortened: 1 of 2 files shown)


ludwig/utils/backward_compatibility.py
```diff
@@ -811,8 +811,15 @@ def upgrade_defaults_config_for_gbm(config: ModelConfigDict) -> ModelConfigDict:
     defaults = copy.deepcopy(defaults_ref)
     gbm_feature_types = GBMDefaultsConfig.Schema().fields.keys()
     for feature_type in defaults_ref:
+        # GBM only supports binary, number, category and text features
         if feature_type not in gbm_feature_types:
             del defaults[feature_type]
+            continue
+
+        # Remove encoder, decoder and loss from defaults since they only apply to ECD
+        defaults[feature_type].pop("encoder", None)
```
tgaddair, line 820: Encoders do apply to GBMs

tgaddair: @ksbrar is this something that need to be fixed in the defaults?

arnavgarg1: That's true yeah, ok let me take a look to make sure we allow encoder as key

arnavgarg1: @tgaddair seems like it, will make it a part of this PR

ksbrar: @tgaddair I was under the impression they don't, but if not then yes they should be added back.

arnavgarg1: @tgaddair to be clear, we only support one hot and tf-idf for text feature encoders for GBMs right?

tgaddair: We support onehot and passthrough for category, ATM. TF-IDF is the only one supported for text for now, but we're going to add more soon.

Note that none of this is in v0.7, we I might now make this a cherry-pick into v0.7.

justinxzhao (approved):

PR #3260, opened 2023-03-16
Exclude `embed` from list of acceptable text encoders for the tagger decoder.

justinxzhao:


justinxzhao:
Closing as embed is fine to use with a tagger decoder.

PR #3266, opened 2023-03-17
feat: Add `kwargs` option to all file readers and feed `nrows` where possible

ksbrar:


ludwig/utils/data_utils.py
```diff
@@ -190,12 +198,22 @@ def read_xsv(data_fp, df_lib=PANDAS_DF, separator=",", header=0, nrows=None, ski
...
+        kwargs["chunksize"] = kwargs["get_preview"]
+        del kwargs["get_preview"]
+
     try:
         df = df_lib.read_csv(data_fp, **kwargs)
     except ParserError:
         logger.warning("Failed to parse the CSV with pandas default way," " trying \\ as escape character.")
         df = df_lib.read_csv(data_fp, escapechar="\\", **kwargs)
 
+    from collections.abc import Iterator
```
abidwael, line 212: Could you import this at the top

ludwig/utils/data_utils.py
```diff
@@ -190,12 +198,22 @@ def read_xsv(data_fp, df_lib=PANDAS_DF, separator=",", header=0, nrows=None, ski
...
+
+    if isinstance(df, Iterator):
```
abidwael, line 214: Will this condition be satisfied? Seems like the return type of `pd.read_csv` is `TextFileReader` only when `iterator=True`

ludwig/utils/data_utils.py
```diff
@@ -190,12 +198,22 @@ def read_xsv(data_fp, df_lib=PANDAS_DF, separator=",", header=0, nrows=None, ski
     if nrows is not None:
         kwargs["nrows"] = nrows
 
+    # If fetching a preview, create an iterator that will immediately be used to read the first chunk:
+    if "get_preview" in kwargs:
+        kwargs["chunksize"] = kwargs["get_preview"]
```
abidwael, line 203: [read_csv](https://docs.dask.org/en/stable/generated/dask.dataframe.read_csv.html) in Dask doesn't have `chunksize` but Pandas read_csv does.

abidwael (commented):
@ksbrar is this change only intended for `read_xsv`?

ludwig/utils/data_utils.py
ksbrar: Per docs and disc.: Dask's version of the function will still use pandas and forward `chunksize` to it

ludwig/utils/data_utils.py
```diff
@@ -159,6 +159,14 @@ def wrapped_fn(*args, **kwargs):
     return wrapped_fn
 
 
+def inner_read(fp, read_fn, **kwargs):
+    if "get_preview" in kwargs:
+        chunksize = kwargs["get_preview"]
```
jeffkinnison, line 164: This and the following line could be condensed to `chunksize = kwargs.pop("get_preview", None)`, which should return `None` rather than raise a `KeyError` if `"get_preview"` does not exist.

ludwig/utils/data_utils.py
```diff
@@ -159,6 +159,14 @@ def wrapped_fn(*args, **kwargs):
...
+        del kwargs["get_preview"]
+        return next(read_fn(fp, chunksize=chunksize, **kwargs))
```
jeffkinnison, line 166: This will only ever return the first chunk. Just double-checking, is that the intended behavior?

ludwig/utils/data_utils.py
```diff
@@ -190,12 +198,20 @@ def read_xsv(data_fp, df_lib=PANDAS_DF, separator=",", header=0, nrows=None, ski
     if nrows is not None:
         kwargs["nrows"] = nrows
 
+    # If fetching a preview, create an iterator that will immediately be used to read the first chunk:
+    if "get_preview" in kwargs:
```
jeffkinnison, line 202: Same as in `inner_method`, you could get rid of this if block with `kwargs["chunksize"] = kwargs.pop("get_preview", None)`.

ludwig/utils/data_utils.py
```diff
@@ -205,22 +221,28 @@ def read_xsv(data_fp, df_lib=PANDAS_DF, separator=",", header=0, nrows=None, ski
 
 @DeveloperAPI
 @spread
-def read_json(data_fp, df_lib, normalize=False):
+def read_json(data_fp, df_lib, normalize=False, **kwargs):
+    # Chunking is not supported for JSON files unless lines=True:
```
jeffkinnison, line 225: Do we want to add logic to support chunking with `lines=True` here?

jeffkinnison (commented):
Great work! Just a couple of minor changes and questions.

ludwig/utils/data_utils.py
ksbrar: My impression is that that is the purpose of `read_jsonl` , so I'd say no

justinxzhao (approved):

tests/ludwig/utils/test_data_utils.py
```diff
@@ -169,3 +171,13 @@ def test_dataset_synthesizer_output_feature_decoder():
     }
     build_synthetic_dataset_df(dataset_size=100, config=config)
     LudwigModel(config=config, logging_level=logging.INFO)
+
+
+def test_chunking():
```
ksbrar, line 176: @abidwael is it this simple for a public dataset?

tests/ludwig/utils/test_data_utils.py
```diff
@@ -169,3 +171,13 @@ def test_dataset_synthesizer_output_feature_decoder():
...
+    # Try basic reads:
+    read_csv("s3://ludwig-tests/datasets/synthetic_1k.csv")
```
jeffkinnison, line 178: Could you add `shape` tests here to ensure the whole file is being read?

tgaddair (approved):
LGTM!

tests/ludwig/utils/test_data_utils.py
abidwael: This should work. Could you parametrize this test? You can pass in the paths as arguments, then use `ludwig.utils.data_utils.figure_data_format_dataset` to get the file type and use the corresponding read function based on that.

abidwael (approved):

tests/ludwig/utils/test_data_utils.py
ksbrar: Assuming it works, will come back and improve this test, right now want to get it in ASAP.

ksbrar:
Remaining test failures:
* 3.8 / not distributed - segmentation fault at the end of the tests
* 3.9 / not distrubted - segmentation fault at the end of the tests
* 3.9 / distributed - timeout after running almost all tests

In the interest of getting this in quickly, I will merge this PR as all other tests passed (even in the scenarios above) and the specific test I added to validate `read_csv`/`read_parquet` completed in all scenarios.

PR #3269, opened 2023-03-18
Fixed non-scalar (text, vector, set) output feature explanations

tgaddair:


ludwig/explain/captum.py
```diff
@@ -104,7 +116,14 @@ def forward(self, *args):
         for of_name in self.model.output_features:
             predictions[of_name] = self.model.output_features.get(of_name).predictions(outputs, of_name)
 
-        return get_pred_col(predictions, self.target)
+        pred_t = get_pred_col(predictions, self.target)
+
+        # If the target feature is a non-scalar type (vector, set, etc.), sum it to get a scalar value.
+        # https://github.com/pytorch/captum/issues/377
```
arnavgarg1, line 122: QQ: Does this change the interpretation/explanation for these kinds of features?

justinxzhao (approved):
Seems reasonable to me! Thanks for finding and linking to a relevant discussion on the captum project.

ludwig/explain/captum.py
tgaddair: It would be different from looking at each element of the output individually, yes, but it's difficult to reason about what that alternative means for things like vectors.

jppgks (approved):

PR #3271, opened 2023-03-18
Check that concat combiner doesn't receive a mixture of non-reduced sequence and non-sequence features.

justinxzhao:


ludwig/config_validation/checks.py
```diff
@@ -320,3 +320,33 @@ def check_tagger_decoder_requirements(config: "ModelConfig") -> None:  # noqa: F
...
+            has_unreduced_sequence_feature = True
+        else:
+            has_non_sequence_feature = True
+
+    if has_unreduced_sequence_feature and has_non_sequence_feature:
+        raise ConfigValidationError(
+            "The concat combiner cannot receive a mix of unreduced sequence features and non-sequence features. "
+            "Options: 1) Set reduce_output in sequence encoder to a value other than None, 2) Choose a different "
+            "combiner like `sequence_concat` which can handle, 3) Remove all non-sequence features, or 4) Remove all "
+            "sequence features."
```
arnavgarg1, line 351: probably a typo/incomplete phrase in "combiner like `sequence_concat` which can handle,"?

arnavgarg1:
@justinxzhao can you add a PR description explaining exactly what the issue is / why it occurs?

justinxzhao:
> @justinxzhao can you add a PR description explaining exactly what the issue is / why it occurs?

Added -- let me know if anything's unclear!

tests/ludwig/config_validation/test_checks.py
```diff
@@ -233,3 +235,34 @@ def test_comparator_combiner_entities(entity_1: List[str], entity_2: List[str],
...
+  train_steps: 2
+model_type: ecd
+"""
+    )
+
+    with pytest.raises(ConfigValidationError):
+        ModelConfig.from_dict(config)
+
+    config[COMBINER][TYPE] = "sequence_concat"
+    ModelConfig.from_dict(config)
```
arnavgarg1, line 268: I feel like this adds a second dimension to this test which doesn't feel like it fits (although I get the purpose of it). Should we just break this up into two different tests that use the same config?

arnavgarg1 (approved):

tests/ludwig/config_validation/test_checks.py
justinxzhao: This extra block confirms that Line 264 is indeed caused by the choice of the concat combiner (and not due to any other ConfigValidationError). I've added a comment to make this clearer.

arnavgarg1 (approved):

PR #3274, opened 2023-03-20
Fixed handling of datetime types in input parquet files

tgaddair:
- Fixes issue with datetime splitting on timestamp type columns
- Fixes issue with automl type inference for datetime types
- Fixes issue with automl type inference for datetimes formatted as strings


jppgks (approved):
Thanks Travis!

arnavgarg1 (approved):

PR #3275, opened 2023-03-20
Ray nightly compatibility

arnavgarg1:


.github/workflows/pytest.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           - python-version: "3.10"
             pytorch-version: nightly
             torchscript-version: 1.10.2
-            ray-version: 2.3.1
+            ray-version: nightly
```
justinxzhao, line 39: Should we do?

```
- ray-version: 2.3.1
- ray-version: 2.4.0
- ray-version: nightly
```

justinxzhao (approved):

.github/workflows/pytest.yml
jeffkinnison: I was planning to do that in a followup PR so we could encapsulate 2.4 fixes. How does that sound?

PR #3279, opened 2023-03-21
Update all combiners to use `.get` to access `LudwigFeatureDict` contents

abidwael:
Fix for the following failure
`TypeError: 'LudwigFeatureDict' object is not subscriptable`
on the following config (just an example. Also fails with other combiners)
```
{'combiner': {'dropout': 0.23429196302654826,
              'norm': 'batch',
              'num_fc_layers': 2,
              'output_size': 128,
              'type': 'project_aggregate'},
 'input_features': [{'encoder': {'idx2str': ['shPZ',
                                             'tktQ',
                                             'nvZSdg',
                                             'jdddezacG',
                                             'OaP',
                                             'uPbBjZGPoj',
                                             'RzrsXi',
                                             'GLZWCEDiWu',
                                             'Xi',
                                             'dsrekI'],
                                 'vocab_size': 10},
                     'name': 'category_1',
                     'type': 'category'},
                    {'name': 'number_1', 'type': 'number'},
                    {'name': 'binary_1', 'type': 'binary'}],
 'model_type': 'ecd',
 'output_features': [{'decoder': {'idx2str': ['bFuiUG',
                                              'yANeMZgjRc',
                                              'ARDdfzasO',
                                              'Vluy',
                                              'IhE',
                                              'vv',
                                              'jYi',
                                              'YtOaymtGo',
                                              'wzTGEyr',
                                              'aJg'],
                                  'vocab_size': 10},
                      'name': 'category_output_1',
                      'type': 'category'}],
 'trainer': {'train_steps': 1}}
```

### DIFF
--- a/ludwig/combiners/combiners.py
+++ b/ludwig/combiners/combiners.py
@@ -478,7 +478,8 @@ class TransformerCombiner(Combiner):
             # after flattening the encoder output tensor
             [
                 Linear(
-                    torch.prod(torch.Tensor([*input_features[inp].output_shape])).type(torch.int32), config.hidden_size
+                    torch.prod(torch.Tensor([*input_features.get(inp).output_shape])).type(torch.int32),
+                    config.hidden_size,
                 )
                 for inp in input_features
             ]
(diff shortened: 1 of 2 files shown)


arnavgarg1 (approved):
Interesting. The real question is, why haven't our existing tests caught this error?
Can you add some quick unit tests/parametrized tests that would've caught this error?

abidwael:
The reason why they're not caught is because the data structure we pass in is a dict and not a `LudwigFeatureDict` which inherits from `torch.nn.Module` ([here](https://github.com/ludwig-ai/ludwig/blob/7276d0218ab7be67432b03697d4f0c233ef07983/tests/ludwig/combiners/test_combiners.py#L93))

PR #3282, opened 2023-03-21
Allow different splits for hyperopt metric reporting

abidwael:
Enables metric reporting for all splits: `train`, `validation`, and `test`.


tests/integration_tests/test_hyperopt.py
```diff
@@ -143,6 +146,7 @@ def test_hyperopt_search_alg(
     ray_cluster_7cpu,
     validate_output_feature=False,
     validation_metric=None,
+    split="validation",
```
tgaddair, line 149: Is this used? Looks like it's overridden by line 181 bwlow.

ludwig/hyperopt/execution.py
```diff
@@ -267,13 +267,10 @@ def _has_eval_metric(self, stats):
...
         return isinstance(stats, float)
 
-    def get_metric_score(self, train_stats) -> float:
-        if self._has_metric(train_stats, VALIDATION):
-            logger.info("Returning metric score from training (validation) statistics")
-            return self.get_metric_score_from_train_stats(train_stats, VALIDATION)
-        elif self._has_metric(train_stats, TRAINING):
-            logger.info("Returning metric score from training split statistics, " "as no validation was given")
-            return self.get_metric_score_from_train_stats(train_stats, TRAINING)
+    def get_metric_score(self, train_stats, split=VALIDATION) -> float:
```
tgaddair, line 270: How does this work if the dataset doesn't have the chosen split? I assume it should fail if the user manually selected a bad split, but the default behavior seems like it would fail if unspecified.

arnavgarg1: I was going to ask the same thing. Will be good to raise appropriate errors here when splits are missing, or even better, log a warning and default to the training split incase the split requested doesn't exist. WDYT?

abidwael: @tgaddair `split` is enforced to be one of `train`, `validation`, `test` at config validation time. If unspecified, it will default to `validation`. If any of `train`, `validation`, `test` is not there, it will go to the else statement and raise the error.

@arnavgarg1 we raise an error if the split specified doesn't have metrics. I'm on the fence about whether to raise an error if say the `validation` split is not there or to raise the warning and report the training metrics. The former will keep the integrity of the reports and what the user expects. The latter tries to prevent a failure. I'm leaning towards the first approach.

Do you think this is the right behavior?

tests/integration_tests/test_hyperopt.py
justinxzhao: +1 to @tgaddair's comment. It's not being used. Maybe remove line 181?

abidwael: fixed

justinxzhao (approved):

PR #3288, opened 2023-03-24
Adds category_prob output feature for providing labels as a probability distribution

tgaddair:


w4nderlust:
not having class names in the values themselves makes it so that a vocabulary has to be provided, otherwise an error is raised, right? Or class names would just be 0 1 2 ... N ?

An alternative would be to require names to be provided by having each row be something like
```
class1:0.3 class2:0.1 ... classN:0.2
```
or
```
{class1: 0.3,  class2: 0.1, ..., classN: 0.2}
```
but that would require some data processing from the user.

The other consideration is that the output doesn't really look like the original data. It is ok, and if someone really wants the output to look like the original data they can use the vector feature instead, but it is worth noting s it is different from all other output features.

ludwig/features/category_feature.py
```diff
@@ -471,3 +496,112 @@ def get_schema_cls():
...
+        self,
+        output_feature_config: Union[CategoryProbOutputFeatureConfig, Dict],
+        output_features: Dict[str, OutputFeature],
+        **kwargs,
+    ):
+        self.num_classes = output_feature_config.num_classes
+        self.top_k = output_feature_config.top_k
+        super().__init__(output_feature_config, output_features, **kwargs)
+        if hasattr(output_feature_config.decoder, "num_classes"):
+            output_feature_config.decoder.num_classes = output_feature_config.num_classes
```
w4nderlust, line 512: in which situation is this needed?

tgaddair:
@w4nderlust I think it would be fine to allow the labels to be 0, 1, 2, etc. if the user omits the vocab. It's a good question.

Agree we should callout that the predictions will differ from the inputs.

ludwig/features/category_feature.py
tgaddair: Copy-pasta. Cleaned it up.

PR #3293, opened 2023-03-28
Pass in `kwargs` to `read_parquet`

abidwael:


justinxzhao (approved):
Should we make this change to other read functions as well?

PR #3294, opened 2023-03-28
fix: Fix `read_parquet` for remote filesystems

ksbrar:


ludwig/utils/data_utils.py
```diff
@@ -252,10 +252,14 @@ def read_excel(data_fp, df_lib, **kwargs):
 def read_parquet(data_fp, df_lib, **kwargs):
     if "nrows" in kwargs:
         import pyarrow.parquet as pq
+        import s3fs
+
+        filesystem = s3fs.S3FileSystem() if 's3:' in data_fp else None
```
ksbrar, line 257: Checking if I need to handle other filesystems

ludwig/utils/data_utils.py
```diff
@@ -252,10 +252,14 @@ def read_excel(data_fp, df_lib, **kwargs):
 def read_parquet(data_fp, df_lib, **kwargs):
     if "nrows" in kwargs:
         import pyarrow.parquet as pq
+        import s3fs
+
+        filesystem = s3fs.S3FileSystem() if "s3:" in data_fp else None
```
tgaddair, line 257: use fsspec

abidwael: +1, namely
```
import fsspec
from ludwig.utils.fs_utils import get_fs_and_path

protocol, _ = fsspec.core.split_protocol(path)
fs, _ = get_fs_and_path(path)
```

ksbrar: thanks @abidwael !, significantly cleaner than what I was doing :)

tgaddair (approved):

PR #3295, opened 2023-03-28
fix: Add config transformation that removes extra `type` param if it exists in the defaults config

ksbrar:


tgaddair (approved):
LGTM!

ksbrar:
Closed in favor of #3296

PR #3301, opened 2023-03-29
Remove the flatten/unflatten ray backend tensor communication workaround

justinxzhao:
This was added in https://github.com/ludwig-ai/ludwig/pull/1260, and may no longer be needed for newer versions of ray.

Also see: https://github.com/ludwig-ai/ludwig/issues/1353


arnavgarg1:
This is pretty interesting @justinxzhao. If this is the case and tests pass, that would be good to know.

I'm currently running into a variety of issues while enabling Ray backend for text generation where it specifically fails during the unflatten operation that you're intending on removing here.

I would be curious to see if it fixes the issue I'm seeing once this lands :)

arnavgarg1:
Yup, just double checked, this seems to fix all of the issues I was running into! Hope we can get this merged

arnavgarg1 (approved):
LGTM!

PR #3305, opened 2023-03-29
Added `trainer.gradient_accumulation_steps` for increasing effective batch size

tgaddair:
Benefits include:

- Lower network bandwidth overhead by reducing frequency of allreduce / gradient synchronization
- Increase effective batch size to smooth out variance when training very large models

### DIFF
--- a/tests/integration_tests/test_trainer.py
+++ b/tests/integration_tests/test_trainer.py
@@ -328,3 +328,33 @@ def test_compile(tmpdir):
     # convergence like gradient magnitudes, etc. Should also add distributed tests.
     model = LudwigModel(config, backend=LocalTestBackend(), logging_level=logging.INFO)
     model.train(training_set=data_csv, validation_set=val_csv, test_set=test_csv, output_directory=tmpdir)
+
+
+@pytest.mark.parametrize("gradient_accumulation_steps", [1, 2, 3])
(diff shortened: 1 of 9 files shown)


tests/integration_tests/test_trainer.py
```diff
@@ -328,3 +328,33 @@ def test_compile(tmpdir):
...
+    config = {
+        "input_features": input_features,
+        "output_features": output_features,
+        "combiner": {"type": "concat", "output_size": 14},
+        TRAINER: trainer,
+    }
+
+    # Just test that training completes without error.
+    # TODO(travis): We may want to expand upon this in the future to include some checks on model
+    # convergence like gradient magnitudes, etc. Should also add distributed tests.
```
justinxzhao, line 358: Should we re-use the test utility for distributed tests?

```python
run_test_suite(config, dataset, "ray")
```

justinxzhao (approved):
C00L! 👍

tests/integration_tests/test_trainer.py
tgaddair: I think we should refactor to be able to do this, yeah. The reason I didn't do it here is because that function is a lot more expensive, since it runs a lot more additional tests. But we should in general rely on some standard test suite functions that can run with different levels of checks.

PR #3306, opened 2023-03-29
fix: Add separate output config registries for ECD and GBM

ksbrar:
Separate `output_config_registry` into `ecd_output_config_registry` and `gbm_output_config_registry`.


abidwael (approved):

tgaddair (approved):

ksbrar:
Test failures in both cases are due to tests reaching ~80% coverage (all passing) but then hitting the max timeout. Other workflows all passed and I have high confidence in this change, so merging it in.

PR #3307, opened 2023-03-30
Use `frac` arg in `df.sample` instead of `n`

abidwael:


justinxzhao (approved):
Did you observe a boost in speed?

abidwael:
Only frac is supported by dask: https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.sample.html

PR #3309, opened 2023-03-30
Allow providing encoders / decoder params as strings

tgaddair:
Note that this currently works due to backward compatibility logic, but we may remove this at some point in the future, so worth supporting it officially.

### DIFF
--- a/ludwig/schema/utils.py
+++ b/ludwig/schema/utils.py
@@ -1138,10 +1138,12 @@ class TypeSelection(fields.Field):
         key: str = "type",
         description: str = "",
         parameter_metadata: ParameterMetadata = None,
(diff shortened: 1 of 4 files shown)


ludwig/schema/utils.py
```diff
@@ -1163,6 +1165,11 @@ def __init__(
     def _deserialize(self, value, attr, data, **kwargs):
         if value is None:
             return None
+
+        if self.allow_str_value and isinstance(value, str):
```
ksbrar, line 1169: nit: below, update the final validation error message

ksbrar (approved):
LGTM. Worth leaving a TODO comment somewhere in the code?

connor-mccorm (approved):

PR #3312, opened 2023-03-31
Add `features_to_strcat` option for Ludwig Datasets transformations to synthesize stringified and concatenated columns of multiple features.

justinxzhao:
And add this as an optionally specified argument to `agnews`.


tgaddair (changes requested):
Hey @justinxzhao , I don't think this is a super useful place to put this, as the datasets api is not very commonly used. Why not make this part of global preprocessing in the config? I have some thoughts on what this could look like. Let's discuss.

PR #3313, opened 2023-03-31
int: Prevent encoder enums from being reshuffled

ksbrar:


ludwig/schema/combiners/sequence.py
```diff
@@ -15,7 +15,7 @@
 SEQUENCE encoders that always return 2D [batch_size, hidden_size] tensors, regardless of how they are parameterized.
 These should never be used with modules that expect 3D tensors, such as the SequenceCombiner.
 """
-_2D_SEQUENCE_ENCODERS = {"embed"}
```
ksbrar, line 18: On second thought, I can change this back to a set but would like to understand if and why one is preferable over another

abidwael (commented):
This looks good, could you merge master to install the right pandas version?

PR #3316, opened 2023-04-02
int: try force clearing the internal `marshmallow-dataclass` cache

ksbrar:
Curious to see if this will help with the memory issues.

As suggested in https://github.com/lovasoa/marshmallow_dataclass/pull/187#issuecomment-1138411268


ksbrar:
Looks like that did not work, although all distributed workflows simply halted at `tests/integration_tests/test_experiment.py::test_sequence_tagger_text_ray PASSED [ 17%]` until they hit their timeout.

PR #3321, opened 2023-04-04
Pass encoding dimensions to SequenceCombiner

RXminuS:
If this argument isn't passed along then the `attention` reducer doesn't work for me.

Note that although this fixes it for me I suspect there's lots of other places this would have to be applied. But maybe someone more experienced could chip in with if this is indeed an acceptable solution and other things to consider?


justinxzhao:
Hi @RXminuS,

The change looks reasonable to me!

We'd probably want to make the same change to all of the encoders in text_encoders that have reducers, but doing this for the Auto encoder first makes sense.

Not totally sure why all of the tests are failing in your PR though. I'll re-run the CI.

RXminuS:
Cool, happy to expand the PR to the others once tests pass 👍 thanks for reviewing

tgaddair:
Thanks for the contribution @RXminuS!

Most of these test failures look like issues with some recently added tests that use S3. So the fix on our side will simply be to exclude those tests when running from a fork. cc @abidwael 

The only one I'm not sure on is `tests/ludwig/profiling/test_metrics.py::test_get_column_profile_attributes`. Seems unrelated to either this PR or credentials, so @justinxzhao would you be able to take a quick look to see what's going on?

justinxzhao:
@tgaddair thanks for the pointers.

I've removed this test `tests/ludwig/profiling/test_metrics.py::test_get_column_profile_attributes`, so rebasing this PR (I've gone ahead and done this) should fix that test.

justinxzhao:
I've also gone ahead and applied the `private_param` and `private_test` annotations which should hopefully take care of the other failing tests that depend on specific credentials.

w4nderlust:
Thanks for the contribution Rik!

PR #3322, opened 2023-04-04
Pandas 2.0 update

jeffkinnison:
Updates to make Ludwig compatible with Pandas 2.0. The current changes include:

- Remove `experiment_impact_tracker` as a dependency
- Update `LudwigProfiler` to replicate `experiment_impact_tracker` CPU/GPU attribute collection
- `df.iteritems()` -> `df.items()`
- `df.append(df2)` -> `df_engine.concat([df, df2])`
- Support for writing `xls` files was [removed](https://github.com/pandas-dev/pandas/pull/49296). This only seems to impact one integration test.


ludwig/benchmarking/profiler.py
```diff
@@ -20,19 +22,38 @@
...
-sys.stdout = sys.__stdout__
-
 STOP_MESSAGE = "stop"
 logger = logging.getLogger()
 
 
+def get_gpu_info():
+    """Gathers general hardware information about an nvidia GPU.
+
+    :param args:
```
justinxzhao, line 32: nit: Remove empty docs.

If this is copied from the other repo, please reference it as a comment i.e.
`Adapted from <link>, with additional adjustments for compatibility with Pandas 2.0.`

ludwig/benchmarking/profiler.py
```diff
@@ -20,19 +22,38 @@
 from ludwig.globals import LUDWIG_VERSION
 from ludwig.utils.data_utils import save_json
 
-# disabling print because the following imports are verbose
```
justinxzhao, line 23: Nice cleanup.

justinxzhao (approved):
Sweet! Just one comment, otherwise LGTM.

jeffkinnison:
The changes listed above should cover the 2.0 update, and as far as I can tell we should still be able to support 1.x because `df.items` and `pd.concat` have been in the API since at least 1.0.

`requirements.txt` does allow for versions all the way back to 0.25, though, and `concat` doesn't exist that far back. Do we want to bump the minimum version to 1.0?

@justinxzhao

justinxzhao:
> The changes listed above should cover the 2.0 update, and as far as I can tell we should still be able to support 1.x because `df.items` and `pd.concat` have been in the API since at least 1.0.
> 
> `requirements.txt` does allow for versions all the way back to 0.25, though, and `concat` doesn't exist that far back. Do we want to bump the minimum version to 1.0?
> 
> @justinxzhao

Bumping the min version to 1.0 SGTM!

PR #3323, opened 2023-04-05
Prevent GBM memory leaks by forcing actor cleanup after each epoch

arnavgarg1:
Ray Actors get re-initialized after each epoch/boosting rounds, but the actors leave a lot of garbage hanging around in Ray's object store. This results in a few problems:

1. Increased in-memory data usage epoch after epoch since artifacts/objects aren't cleaned up, potentially leading to OOMs
2. Increased disk-spillage from object store to disk once the 20% threshold has passed, increasing time per epoch by an order of magnitude after each epoch

Together, these result in models either OOMing, or taking forever to train.

This PR forces cleanup after each epoch to prevent memory leaks by wrapping the LightGBM train step when using Ray with a `ray.remote(max_calls=1)` to kill the Ray Actors and GC at the end of every epoch. This involves two changes:

1. The creation of a modified `train_loop` function from the base class that enables making `ray.remote` calls.
2. Moving the `train_step` function outside of the `LightGBMRayTrainer` class and making it its own ray remote function that is called by the `train_loop` function inside the `LightGBMRayTrainer` class


ludwig/trainers/trainer_lightgbm.py
```diff
@@ -47,6 +47,11 @@
     ProgressTracker,
 )
 
+try:
+    import ray
+except Exception:
```
jeffkinnison, line 52: Can we change this to `ImportError`?

ludwig/trainers/trainer_lightgbm.py
```diff
@@ -945,64 +950,76 @@ def __init__(
...
         return GBMTrainerConfig
 
-    def train_step(
+    def _train_loop(
         self,
         params: Dict[str, Any],
-        lgb_train: "RayDMatrix",  # noqa: F821
-        eval_sets: List["RayDMatrix"],  # noqa: F821
+        lgb_train: lgb.Dataset,
+        eval_sets: List[lgb.Dataset],
```
jeffkinnison, line 957: Why are we cutting support for `RayDMatrix`? Is it just because of the index error?

ludwig/trainers/trainer_lightgbm.py
arnavgarg1: Good call!

ludwig/trainers/trainer_lightgbm.py
arnavgarg1: Good catch! this is a mistake because of some copy pasta from LightGBMTrainer which uses lgb.Dataset instead of RayDMatrix. Just updated!

ludwig/trainers/trainer_lightgbm.py
```diff
@@ -1057,3 +1074,107 @@ def _construct_lgb_datasets(
...
+    boost_rounds_per_train_step: int,
+    evals_result: Dict,
+    ray_params: "RayParams",  # noqa
+    evaluate_training_set: bool,
+    device: Optional[str] = None,
+) -> lgb.LGBMModel:
+
+    # We need to add max_calls here to ensure that the Ray actors that get created by the LightGBM class
+    # for each boosting round don't leave dangling resources in the object store memory.
+    @ray.remote(max_calls=1)
```
justinxzhao, line 1096: Once the memory leak from `RayLGBMClassifier`/`RayLGBMRegressor` has been fixed, does this mean we can simply remove `max_calls=1`? Would be ideal to add a link to an open issue on the Ray project if one exists.

justinxzhao (approved):
Nice fix!

ludwig/trainers/trainer_lightgbm.py
arnavgarg1: Yup! Ideally, we would remove the ray.remote decorator and move away from the need to use a remote function entirely!

PR #3324, opened 2023-04-06
Update type hinting to Python 3.9 conventions

abidwael:
See
- https://peps.python.org/pep-0585/#reference-implementation
- https://docs.python.org/3/library/typing.html


abidwael:
Filed an issue for the CI failure in https://github.com/pytorch/pytorch/issues/98521

PR #3329, opened 2023-04-07
Filter auto_transformer kwargs based on forward signature

### LINKED ISSUE #3328: TypeError: forward() got an unexpected keyword argument 'token_type_ids'
I am trying to fine tune bloom using your examples but  getting thi error TypeError: forward() got an unexpected keyword argument 'token_type_ids'

tgaddair:
Fixes #3328.

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -2170,6 +2170,8 @@ class AutoTransformerEncoder(HFTextEncoder):
             self.reduce_sequence = SequenceReducer(reduce_mode=reduce_output)
         self.max_sequence_length = max_sequence_length
 
+        # Precompute the set of params that are included in the forward signature of the AutoModel implementation so
+        # we can filter out unused params during the `forward` call.
         self.forward_kwargs = set(inspect.signature(self.transformer.module.forward).parameters.keys())
 
     def _maybe_resize_token_embeddings(self, transformer, vocab_size: Optional[int] = None):
@@ -2184,6 +2186,8 @@ class AutoTransformerEncoder(HFTextEncoder):
         if mask is not None:
             mask = mask.to(torch.int32)
 
+        # The forward signature of AutoModel is not consistent across implementations, so we need to make sure we're
+        # only passing in params included in the forward signature.
         kwargs = dict(
             input_ids=inputs,
             attention_mask=mask,
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import inspect
 import logging
 from typing import Any, Callable, Dict, List, Optional, Type, TYPE_CHECKING, TypeVar, Union
 
@@ -2169,6 +2170,8 @@ class AutoTransformerEncoder(HFTextEncoder):
             self.reduce_sequence = SequenceReducer(reduce_mode=reduce_output)
         self.max_sequence_length = max_sequence_length
 
+        self.forward_kwargs = set(inspect.signature(self.transformer.module.forward).parameters.keys())
+
     def _maybe_resize_token_embeddings(self, transformer, vocab_size: Optional[int] = None):
         """Overridden because AutoModel should use its own vocab size unless vocab size is explicitly specified."""
         if vocab_size is not None:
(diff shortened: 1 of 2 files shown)


ludwig/encoders/text_encoders.py
```diff
@@ -2169,6 +2170,8 @@ def __init__(
             self.reduce_sequence = SequenceReducer(reduce_mode=reduce_output)
         self.max_sequence_length = max_sequence_length
 
+        self.forward_kwargs = set(inspect.signature(self.transformer.module.forward).parameters.keys())
```
jeffkinnison, line 2173: nit: Could you add a comment here explaining why we're inspecting? Might be helpful down the road since this was motivated by a bloom-specific fix.

tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -158,6 +158,47 @@ def test_hf_ludwig_model_reduce_options(tmpdir, csv_filename, encoder_name, redu
...
+@pytest.mark.parametrize(
+    "pretrained_model_name_or_path",
+    [
+        "hf-internal-testing/tiny-random-bloom",
+        "hf-internal-testing/tiny-random-OPTModel",
+        "hf-internal-testing/tiny-random-GPTJModel",
+    ],
+)
+def test_hf_ludwig_model_auto_transformers(tmpdir, csv_filename, pretrained_model_name_or_path):
+    input_features = [
```
jeffkinnison, line 170: nit: Could you add a short explanation of the motivation behind this test?

ludwig/encoders/text_encoders.py
```diff
@@ -2180,11 +2183,15 @@ def _maybe_resize_token_embeddings(self, transformer, vocab_size: Optional[int]
     def forward(self, inputs: torch.Tensor, mask: Optional[torch.Tensor] = None):
         if mask is not None:
             mask = mask.to(torch.int32)
-        transformer_outputs = self.transformer.module(
+
+        kwargs = dict(
```
jeffkinnison, line 2187: Would it make sense to add any aux checks for incompatible config parameters passed to specific encoders?

jeffkinnison (commented):
LGTM! Just some nits and a quick question.

jeffkinnison (approved):

ludwig/encoders/text_encoders.py
tgaddair: Hmm, not sure what you had in mind here. The issue is predominately that there are many different HF model implementations that all take different params. So we need to make sure we don't accidentally pass in the wrong params.

ludwig/encoders/text_encoders.py
tgaddair: Added comments.

tests/ludwig/encoders/test_text_encoders.py
tgaddair: Added comment.

ludwig/encoders/text_encoders.py
jeffkinnison: Oh, apologies, on followup look this seems like something that would be difficult to catch in the schema. I was misremembering how we used `AutoTransformer`.

tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -158,6 +158,51 @@ def test_hf_ludwig_model_reduce_options(tmpdir, csv_filename, encoder_name, redu
...
 
+@pytest.mark.parametrize(
+    "pretrained_model_name_or_path",
+    [
+        "hf-internal-testing/tiny-random-bloom",
+        "hf-internal-testing/tiny-random-OPTModel",
+        "hf-internal-testing/tiny-random-GPTJModel",
+    ],
+)
+def test_hf_ludwig_model_auto_transformers(tmpdir, csv_filename, pretrained_model_name_or_path):
```
abidwael, line 169: Could you mark this with `@pytest.mark.slow`?

tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -158,6 +158,51 @@ def test_hf_ludwig_model_reduce_options(tmpdir, csv_filename, encoder_name, redu
...
+    model = LudwigModel(config=config, backend=LocalTestBackend())
+
+    # Validates that the defaults associated with the encoder are compatible with Ludwig training.
+    with mock.patch(
+        "ludwig.encoders.text_encoders.load_pretrained_hf_model_with_hub_fallback",
+        side_effect=_load_pretrained_hf_model_no_weights,
+    ):
+        model.train(dataset=rel_path, output_directory=tmpdir)
+
+
```
abidwael, line 205: Code here is a duplicate of that in `test_hf_ludwig_model_reduce_options` and `test_hf_ludwig_model_e2e`. Perhaps separate into a separate function and reuse.

tests/ludwig/encoders/test_text_encoders.py
tgaddair: I think this should only take 30s, so I would want to run it on every commit if we can get away with it.

arnavgarg1 (approved):

pushkarraj:
Hey guys, I am working in the domain of large language models. Thanks for fixing this issue. I would love to contribute in the open source. It would be a matter of pride to be the part of this.

pushkarraj:
Also this issue is resolved.after trying  the latest code I got  the following error:

**AttributeError: 'BloomModel' object has no attribute 'n_head'**

PR #3330, opened 2023-04-07
Use Python 3.8 in pytest

abidwael:


arnavgarg1 (commented):
Should we make this py3.8, py3.9 and py3.10?

abidwael:
Currently, we only have tests on Ray 2.2.0 and 2.3.1 and we'll add the nightly tests back. We can do 3.8 and 3.10 for now. WDYT?

.github/workflows/pytest.yml
```diff
@@ -22,18 +22,21 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.8", "3.8"]
```
arnavgarg1: I think this needs to be 3.8, 3.9 and 3.10?

arnavgarg1 (approved):

.github/workflows/pytest.yml
abidwael: yeah just pushed a change

arnavgarg1: perfect, approved

PR #3335, opened 2023-04-10
Support zero-shot learning and text generation through LLMs

arnavgarg1:
This PR introduces LLMs to Ludwig! In particular, Ludwig now has the ability to support text-to-text language generation tasks via LLMs, as well as the ability to perform zero-shot learning-styled classification tasks via text-to-category configurations.

## Text Generation
You can use your LLM to generate text using LLMs. This can be done via a simple config like this:

```yaml
input_features:
    - name: Question
      type: text
output_features:
    - name: Answer
      type: text
model_type: llm
generation_config:
    temperature: 0.1
    top_p: 0.75
    top_k: 40
    num_beams: 4
    max_new_tokens: 5
model_name: facebook/opt-350m
```

This will produce a maximum of 5 generated tokens per `Question` in the data, and then compute metrics by comparing the predicted `Answer` to the actual `Answer`. 

## Zero-Shot Learning (ZSL)

You can also use your LLM to generate text using LLMs, and then use that for downstream tasks such as classification. The idea behind ZSL is that you're training a model to recognize and classify new objects or tasks without having seen any examples of them during training. This can be done via a config like this:

```yaml
input_features:
    - name: review
      type: text
output_features:
    - name: label
      type: category
      preprocessing:
          vocab: [positive, neutral, negative]
          fallback_label: neutral
          prompt_template: |
              Context information is below.
              ###
              {review}
              ###
              Given the context information and not prior knowledge, classify the context as one of: {labels}
      decoder:
          match:
              positive:
                  type: contains
                  value: positive
              neutral:
                  type: regex
                  value: neutral
              negative:
                  type: contains
                  value: negative
model_type: llm
generation_config:
    temperature: 0.1
    top_p: 0.75
    top_k: 40
    num_beams: 4
    max_new_tokens: 5
model_name: facebook/opt-350m
```

There are a few new preprocessing and decoding options that have been introduced specifically for ZSL:

**Preprocessing**:
1. __prompt_template__: The template to use for the prompt. The underlying text feature will be augmented to include the prompt, and fill in values for feature names mentioned between { } braces. The labels will be inserted into the template if `{vocab}` is specified. This enriched version of the prompt will be passed to the LLM for generation.
2. __vocab__: This is the set of labels that you want to add to the prompt, and typically represents the list of labels that the model can predict.
3. __fallback_label__: This is one of the labels in `vocab` that the prediction should fallback to incase the LLM doesn't produce one of the items in `vocab`.

**Decoder**:
1. __match__: This is a way to define label classes and their corresponding match patterns. Once the LLM output is decoded, the match patterns are used to greedily check if the labels are present in the output. Currently, this supports `contains` and `regex`.  

## Implementation Details

1. Introduces a new `llm` model type: at the config level, users must provide a `model_name`, the hugging face hub path to a CausalLM. It also introduces a new top-level key for `llm` model types called `generation_config`, which can influence the text generated by your LLM.
4. Two new trainer types `LLMTrainer` and `ZeroShotTrainer`. The `ZeroShotTrainer` is responsible for both Text Generation and Zero-Shot Learning, `LLMTrainer` is just a base class that can be used for other LLM tasks.
5. Introduces two new decoders - `TextParserDecoder` and `CategoryParserDecoder`. The `TextParserDecoder` is used to extract generated token IDs during the forward pass from the LLM, while the `CategoryParserDecoder` is used to extract generated token IDs, decode them into plain text, and then extract the label from the prediction and match it to a label via a predefined list of possible output labels.
6. The ability to pass in prompts via `prompt_template` to category output features when using LLMs.


ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
         raise ValueError(f"Invalid missing value strategy {missing_value_strategy}")
 
 
+def handle_llm_prompt_injection(
+    dataset_cols, feature_configs, feature_name_to_preprocessing_parameters, backend
```
tgaddair, line 1673: Type hints.

ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
...
+    # dataset_cols with the updated injected prompt. Assumes just text input features for now.
+    for input_feature in input_features:
+        if input_feature[COLUMN] in feature_injection_feature_names and input_feature[TYPE] == TEXT:
+            match_str = "{" + input_feature[COLUMN] + "}"
+            dataset_cols[input_feature[COLUMN]] = dataset_cols[input_feature[COLUMN]].apply(
+                lambda x: prompt.replace(match_str, x)
+            )
+
+
+def _extract_prompt_injection_feature_names(prompt_template: str) -> List[str]:
```
tgaddair, line 1722: It's not immediately clear what we're doing or why with this function. Please add comments.

ludwig/features/category_feature.py
```diff
@@ -352,6 +368,11 @@ def update_config_with_metadata(feature_config, feature_metadata, *args, **kwarg
         feature_config.num_classes = feature_metadata["vocab_size"]
         feature_config.top_k = min(feature_config.num_classes, feature_config.top_k)
 
+        # If labels are provided, then this is a classification task for LLMs
+        if hasattr(feature_config.preprocessing, "labels"):
```
tgaddair, line 372: Let's actually call this `vocab` to be consistent with the cateogry distribution type. We can also make this generic such that ECD can also use it, and we just skip computing the str2idx stuff when that happens.

ludwig/features/category_feature.py
```diff
@@ -445,6 +466,22 @@ def update_config_with_metadata(feature_config, feature_metadata, *args, **kwarg
...
+    def update_metrics(self, targets: Tensor, predictions: Dict[str, Tensor]) -> None:
+        """Updates metrics with the given targets and predictions.
+
+        Args:
+            targets: Tensor with target values for this output feature.
+            predictions: Dict of tensors returned by predictions().
+        """
+        for metric_name, metric_fn in self._metric_functions.items():
+            # Skip loss metric for LLMs, since we don't compute it in the forward pass
+            if self.is_llm_task and metric_name == "loss":
```
tgaddair, line 478: Maybe we just have the loss be 0 for now to avoid this hack?

ludwig/features/text_feature.py
```diff
@@ -293,6 +301,24 @@ def update_config_with_metadata(feature_config, feature_metadata, *args, **kwarg
...
+
+        Args:
+            targets: Tensor with target values for this output feature.
+            predictions: Dict of tensors returned by predictions().
+        """
+        for metric_name, metric_fn in self._metric_functions.items():
+            # Skip loss metric for LLMs, since we don't compute it in the forward pass
+            # Perplexity needs to be computed from "probabilities" tensor, but right now
+            # we're only producing a "predictions" tensor as output from the forward pass.
+            if self.is_llm_task and metric_name in ("loss", "perplexity"):
```
tgaddair, line 315: Same as above, would be great if we could avoid all this special casing. For perplexity, can just similarly give it an arbitrary value for now.

ludwig/models/llm.py
```diff
@@ -0,0 +1,330 @@
...
+        **_kwargs,
+    ):
+        self.config_obj = config_obj
+        self._random_seed = random_seed
+
+        super().__init__(random_seed=self._random_seed)
+
+        self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.model_name)
+        self.generation_config = GenerationConfig(
+            temperature=1.0,
```
tgaddair, line 45: Should probably default to 0.1 to reduce randomness.

ludwig/models/llm.py
```diff
@@ -0,0 +1,330 @@
...
+            top_p=0.75,
```
tgaddair, line 46: In general this should be schemafied.

ludwig/models/llm.py
```diff
@@ -0,0 +1,330 @@
...
+        if of_feature_type == CATEGORY:
+            predictions = self.extract_category_predictions(inputs, outputs.sequences)
+            return {
+                self.config_obj.output_features[0].name: {
+                    "predictions": predictions["predictions"],
+                    "probabilities": predictions["probabilities"],
+                }
+            }
+
+    def extract_text_predictions(self, inputs, output_sequences):
```
tgaddair, line 171: This stuff should all live in the decoders.

ludwig/models/llm.py
```diff
@@ -0,0 +1,330 @@
...
+        sequence."""
+        input_ids = self.get_input_ids(inputs)
+
+        if input_ids.size()[0] == 1:
+            return output_sequences[:, input_ids.size()[1] :]
+
+        generated_predictions = []
+        input_ids_lens = [input_id.size()[0] for input_id in input_ids]
+        for idx, input_id_len in enumerate(input_ids_lens):
+            # Remove the input sequence from the generated sequence
```
tgaddair, line 182: Seems generic and duplicated with category, can pull it out into a common function.

ludwig/config_validation/checks.py
```diff
@@ -455,3 +456,15 @@ def check_hyperopt_nested_parameter_dicts(config: "ModelConfig") -> None:  # noq
         raise ConfigValidationError(
             f"Nested hyperparameter search spaces must be of type 'choice'. Requested space type: {space['space']}"
         )
+
+
+@register_config_check
+def check_llm_atleast_one_input_text_feature(config: "ModelConfig"):  # noqa: F821
```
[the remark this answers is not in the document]
geoffreyangus: I think the more pressing thing here is actually limiting the output features to 1 for now.

I'm worried that we'll have trouble parsing the response of multi-task LLM until we can figure out a principled way to do this.

ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
         raise ValueError(f"Invalid missing value strategy {missing_value_strategy}")
 
 
+def handle_llm_prompt_injection(
+    dataset_cols, feature_configs, feature_name_to_preprocessing_parameters, backend
+) -> None:
+    """If output feature is a category feature and it's preprocessing has prompt_template, then we need to inject
+    the prompt into the input text feature(s)."""
```
geoffreyangus, line 1676: Can you put an example of the input prompt and output prompt in the comment?

ludwig/config_validation/checks.py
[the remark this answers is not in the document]
arnavgarg1: @geoffreyangus Would it be fine to lift this constraint in a follow up PR?

examples/llm_text_to_text/simple_model_training.py
```diff
@@ -0,0 +1,99 @@
...
+
+import pandas as pd
+import yaml
+
+from ludwig.api import LudwigModel
+
+# clean out prior results
+shutil.rmtree("./results", ignore_errors=True)
+
+qa_pairs = {
```
w4nderlust, line 21: small nit. here and for the next examples I would change the datasets to be passed to pandas to lists ofr dictionaries instead of dictionaries of lists

ludwig/data/preprocessing.py
```diff
@@ -1195,6 +1199,10 @@ def build_dataset(
         preprocessing_parameters = feature_name_to_preprocessing_parameters[feature_config[NAME]]
         handle_missing_values(dataset_cols, feature_config, preprocessing_parameters, backend)
 
+    # If we're using LLMs for zero-shot/few-shot learning, update text input features with the prompt
+    # ahead of time so that we can compute metadata and build the preprocessed data correctly.
+    handle_llm_prompt_injection(dataset_cols, feature_configs, feature_name_to_preprocessing_parameters, backend)
```
w4nderlust, line 1204: prompt injection is the name of a technique for "hacking" llms. i would rename this

ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
...
+
+    # If output feature is a category feature and it's preprocessing has prompt_template, then we need to
+    # inject the prompt into the input text feature(s). Currently assumes only one output feature.
+    feature_injection_feature_names = None
+    labels = None
+    prompt = None
+    for output_feature in output_features:
+        if output_feature[TYPE] != CATEGORY:
+            continue
+        if "prompt_template" not in feature_name_to_preprocessing_parameters[output_feature[COLUMN]]:
```
w4nderlust, line 1695: may be wrong, but are you sure that the right key here and in the following lines is column instead of name or processed column?

ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
...
+        prompt = prompt.replace("{labels}", ", ".join(labels))
+        feature_injection_feature_names.pop(feature_injection_feature_names.index("labels"))
+
+    # Substitute values from the input features into the prompt template, and update the values in
+    # dataset_cols with the updated injected prompt. Assumes just text input features for now.
+    for input_feature in input_features:
+        if input_feature[COLUMN] in feature_injection_feature_names and input_feature[TYPE] == TEXT:
+            match_str = "{" + input_feature[COLUMN] + "}"
+            dataset_cols[input_feature[COLUMN]] = dataset_cols[input_feature[COLUMN]].apply(
+                lambda x: prompt.replace(match_str, x)
```
w4nderlust, line 1718: pretty sure this could be sped up by using re.compile and re.sub

ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
...
+            )
+
+
+def _extract_prompt_injection_feature_names(prompt_template: str) -> List[str]:
+    pattern = r"{([^}]*)}"
+    matches = re.findall(pattern, prompt_template)
```
w4nderlust, line 1724: compiling the re here too can help

ludwig/decoders/llm_decoders.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+
+            if is_match:
+                return label
+        return None
+
+
+@DeveloperAPI
+@register_decoder("parser", [CATEGORY])
+class ParserDecoder(Decoder):
```
w4nderlust, line 72: it's an open question to me how to call this thing. i understand why parser, but it's not really the decoder that is a parser, it's the extractor function after the decoding. We can keep it for now but let's discuss about it

ludwig/config_validation/checks.py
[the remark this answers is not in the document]
geoffreyangus: Oh yeah, it's fine to lift this constraint later. I was actually saying that we should add a constraint for output features to ensure len(output_features) == 1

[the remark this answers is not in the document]
arnavgarg1: Ah this is already taken care during config creation, see https://github.com/ludwig-ai/ludwig/pull/3335/files#diff-ebe7db1ebcf2017c6c29300f0ab2ad277675f102488c05695709c52747e4539dR317

ludwig/models/llm.py
arnavgarg1: @tgaddair would you recommend creating a custom decoder for Text Output Features (LLMs specifically)?

tgaddair: Yeah, I would.

arnavgarg1: Makes sense, I agree it both makes more sense and will be exponentially cleaner. About halfway into it at the moment

ludwig/schema/features/preprocessing/category.py
```diff
@@ -113,3 +114,34 @@ def __post_init__(self):
...
+
+    prompt_template: str = schema_utils.String(
+        default="",
+        allow_none=False,
+        description="The template to use for the prompt. The labels will be inserted into the template.",
+    )
+
+    is_llm_task: bool = schema_utils.Boolean(
+        default=True,
+        description="Whether the model is a language modeling task or not.",
```
w4nderlust: semantics here are a bit meh. a model is not a task. what is this trying to capture?

ludwig/schema/features/text_feature.py
```diff
@@ -66,6 +69,17 @@ class GBMTextInputFeatureConfig(TextInputFeatureConfig):
     )
 
 
+@DeveloperAPI
+@llm_input_config_registry.register(TEXT)
+@ludwig_dataclass
+class LLMTextInputFeatureConfig(TextInputFeatureConfig):
```
w4nderlust, line 75: I may be missing context here, but why parallel cnn?

ludwig/schema/features/preprocessing/category.py
arnavgarg1: Got rid of it, don't need it anymore

ludwig/data/preprocessing.py
arnavgarg1: Added!

examples/llm_text_to_text/simple_model_training.py
arnavgarg1: Updated

ludwig/data/preprocessing.py
arnavgarg1: Agreed! Just updated the names of the functions and variables to better reflect what the function is doing

ludwig/data/preprocessing.py
arnavgarg1: Column seems to work, but I think `name` is safer so I updated this

ludwig/data/preprocessing.py
arnavgarg1: Updated

ludwig/data/preprocessing.py
arnavgarg1: Updated

ludwig/decoders/llm_decoders.py
arnavgarg1: That's true.

I've also changed this now to be `category_parser` and `text_parser`, so we should probably think through it further. One potential name change could be "Extractor" instead of "Parser", since that is technically what the forward pass of both decoders are doing. Happy to change it to this if you think it is better.

ludwig/schema/features/text_feature.py
arnavgarg1: We can change this to anything we'd like. Beneath the surface, we're not actually making use of the encoder at all. Let me see if we can actually skip initializing any encoders. 

The only thing we actually care about here is to to use the tokenizer from the underlying model to preprocess the data since we want to make use of the pretrained vocab

ludwig/features/category_feature.py
arnavgarg1: @tgaddair are you suggesting renaming `labels` to `vocab`?

ludwig/data/preprocessing.py
```diff
@@ -1661,6 +1669,62 @@ def _handle_missing_values(
...
 
 
+def handle_llm_prompt_injection(
+    dataset_cols, feature_configs, feature_name_to_preprocessing_parameters, backend
+) -> None:
+    """If output feature is a category feature and it's preprocessing has prompt_template, then we need to inject
+    the prompt into the input text feature(s)."""
+
+    # Recover input and output features
+    input_features = []
```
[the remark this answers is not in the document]
arnavgarg1: Ideally yes, although propagating model type down to this point isn't very clear, I've been trying to see how it can get passed down. Might require quite a few changes to get that through, but I do think it's probably necessary.

ludwig/features/category_feature.py
tgaddair: We should probably consolidate on one. But it's okay to leave it as-is for now.

ludwig/schema/features/text_feature.py
tgaddair: We should use passthrough as GBMs do (with similar "ignores it" behavior).

tgaddair (commented):
Nice! There are some things we'll want to clean up, but I wouldn't block merging this for now, since it's pretty complete and doesn't conflict with the rest of the codebase. I would just get tests to pass and address the encoder comment (don't use parallel_cnn). Otherwise LGTM!

ludwig/schema/features/text_feature.py
arnavgarg1: Just updated, this is a good catch @w4nderlust @tgaddair

tgaddair (approved):
LGTM, let's land and fast follow.

examples/llm_text_generation/simple_model_training.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
+
```
skanjila, line 2: Should we have an equivalent of this file for llm models

ludwig/data/preprocessing.py
```diff
@@ -1195,6 +1199,12 @@ def build_dataset(
         preprocessing_parameters = feature_name_to_preprocessing_parameters[feature_config[NAME]]
         handle_missing_values(dataset_cols, feature_config, preprocessing_parameters, backend)
 
+    # If we're using LLMs for zero-shot/few-shot learning, update text input features with the prompt
+    # ahead of time so that we can compute metadata and build the preprocessed data correctly.
+    handle_data_augmentation_with_prompt(
```
skanjila, line 1204: how do we track errors inside this function

ludwig/features/category_feature.py
```diff
@@ -136,6 +136,15 @@ def get_feature_meta(
             num_most_frequent=preprocessing_parameters["most_common"],
             processor=backend.df_engine,
         )
+
+        if "vocab" in preprocessing_parameters:
```
skanjila, line 140: should we make this a string constant

arnavgarg1: We can do this in a follow-up PR

ludwig/data/preprocessing.py
arnavgarg1: @skanjila Preprocessing in Ludwig is wrapped around a try/except block, so any errors that happen will trigger the except block and we should be able to diagnose it from there. That being said, it would be good to add some implicit checks where needed

examples/llm_text_generation/simple_model_training.py
arnavgarg1: @skanjila I don't think I understand what you mean. Isn't this a file for LLM models? 
We don't support training an LLM from scratch yet, just the ability to use a pretrained LLM for text generation, zero shot learning and few shot learning (which should land in the next day). 

These example notebooks should be good enough for users to learn how to use Ludwig for these kinds of tasks

pushkarraj:
Can we also of similar examples for finetuning  LLMs on ray cluster.

PR #3336, opened 2023-04-10
Fix sequence decoder attribute error

jeffkinnison:
`SequenceOutputFeature.unflatten` would try to reshape its output using `self.decoder_obj.config.num_classes` and throw an `AttributeError`. This PR updates the sequence decoder integration tests and fixes the error by reshaping with `self.decoder_obj.config.vocab_size` instead.


justinxzhao (approved):
LGTM.

Perhaps in a follow up PR, it seems good to look into removing `num_classes` from the OutputFeatureConfig base class -- why is it there to begin with? @connor-mccorm, do you know?

connor-mccorm:
I think we can remove it, I don't remember exactly why I put it in, but the way the schema is structured now, we can just keep it on the relevant output feature configs.

connor-mccorm (approved):

PR #3339, opened 2023-04-12
Add input and output shapes for attention reducer (combiner)

abidwael:


ludwig/modules/attention_modules.py
```diff
@@ -39,6 +39,14 @@ def forward(self, inputs, mask=None):
         gated_inputs = torch.sum(attention * inputs, dim=1)
         return gated_inputs  # [b, h]
 
+    @property
```
tgaddair, line 42: Is this correct though? Seems like it should be `input_size` for the input?

abidwael: Fixed

connor-mccorm (approved):
LGTM, though you may want to wait for someone with more familiarity with this part of the stack to review since my familiarity is limited.

ludwig/modules/attention_modules.py
justinxzhao: For Ludwig modules thus far, input and output shape properties are expected to be available after the module is initialized, not dependent on the first `forward()` pass.

It's certainly a much larger change to have to pass in the input shape into the constructor from anywhere this module can be initialized. That would be the most "right" thing to do, but also not the most practical.

Based on the passing tests from before and after [your last commit](https://github.com/ludwig-ai/ludwig/pull/3339/commits/3898a53b0b21e32f06473a235c2bb4ce027aaa3b), it seems like the test is insensitive to what `input_shape` actually returns, and model exporting simply requires that all `@property` of modules (like `input_shape()` in our case) are defined. Have you confirmed that the test fails when you remove the `input_shape()` method entirely?

If deployment/model export succeeds regardless of what `input_shape()` returns, I say we should go back to the original dummy implementation you started with, but add a comment about it, which is similar to what we've [done for sequence_decoders](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/decoders/sequence_decoders.py#L332) where computing the true input shape is non-trivial.

abidwael: As discussed offline, there seems to be no harm with doing this dynamically.

justinxzhao: There's no evident downsides for setting input and output shape dynamically, but it does violate the original abstraction currently adopted by `LudwigModule`s that input and output shapes, if defined, are available at construction time.

Perhaps that policy was overbearing, so going with either the dummy shape or the dynamic shape to unblock is ok with me. For the future, it would be good to leave a comment about exceptional cases like this.

PR #3351, opened 2023-04-17
[LLM] Few-shot learning via Retrieval-augmented ICL

geoffreyangus:
This PR adds the ability to run few-shot learning style prompts for classification tasks using Large Language Models.


ludwig/models/llm.py
```diff
@@ -40,9 +40,30 @@ def __init__(
...
+        if hasattr(self.model.config, "max_sequence_length"):
+            self.context_len = self.model.config.max_sequence_length
+        elif hasattr(self.model.config, "max_position_embeddings"):
+            self.context_len = self.model.config.max_position_embeddings
+        else:
+            self.context_len = 2048
+
+        self.generation = GenerationConfig(**self.config_obj.generation.to_dict())
+        self.max_new_tokens = self.config_obj.generation.max_new_tokens
+        self.max_input_length = self.context_len - self.max_new_tokens - 8
```
arnavgarg1, line 66: Why do we subtract 8? Is it because of padding alignment?

ludwig/api.py
```diff
@@ -661,8 +661,8 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                                 calibrator.train_calibration(training_set, TRAINING)
                             else:
                                 calibrator.train_calibration(validation_set, VALIDATION)
-                        if not skip_save_model:
-                            self.model.save(model_dir)
+                        # if not skip_save_model:
+                        # self.model.save(model_dir)
```
geoffreyangus, line 665: @arnavgarg1 I think you mentioned that you had a fix in mind for this– is there something long-term we can do here?

ludwig/models/llm.py
geoffreyangus: Not sure, just copied this from @tgaddair's code. Maybe he can chime in?

examples/llm_zero_shot_learning/simple_model_training.py
```diff
@@ -50,40 +50,40 @@
...
+    preprocessing:
+        lowercase: false
+        prompt:
+            task: "Classify the sample input as either negative, neutral, or positive."
+output_features:
+-
+    name: label
+    type: category
+    preprocessing:
+        fallback_label: "neutral"
```
arnavgarg1, line 74: Did we move this into the decoder?

ludwig/api.py
arnavgarg1: Yeah, I'll have something in as a part of the prompt-tuning PR. 
Essentially, we will only save model checkpoints if there is an `adapter` being used for fine-tuning, same with loading. Otherwise, it should just do the right thing with just the `model_name` being provided afaik

ludwig/trainers/trainer_llm.py
```diff
@@ -370,7 +365,7 @@ def evaluation(
         progress_tracker: ProgressTracker,
     ):
         predictor = Predictor(
-            self.model, batch_size=batch_size, distributed=self.distributed, report_tqdm_to_ray=self.report_tqdm_to_ray
+            self.model, batch_size=1, distributed=self.distributed, report_tqdm_to_ray=self.report_tqdm_to_ray
         )
         metrics, _ = predictor.batch_evaluation(dataset, collect_predictions=False, dataset_name=dataset_name)
```
arnavgarg1, line 370: Why are we using a batch size of 1 here? Is there a way to grab it from the schema potentially?

ludwig/schema/prompt.py
```diff
@@ -0,0 +1,74 @@
...
+@DeveloperAPI
+@ludwig_dataclass
+class RetrievalConfig(schema_utils.BaseMarshmallowConfig):
+    """This Dataclass is a schema for the nested retrieval config under prompt."""
+
+    type: str = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description="The type of retrieval to use for the prompt.",
+    )
```
arnavgarg1, line 15: Does it auto default to `random`? Or should we set a default type by default?

examples/llm_zero_shot_learning/simple_model_training.py
geoffreyangus: Not yet– this felt out of scope for this PR (which is already big enough as-is). We can fast-follow with that change.

ludwig/schema/prompt.py
geoffreyangus: It defaults to None, which means no samples are retrieved and it becomes zero-shot learning. I can make this description more clear on that front

ludwig/trainers/trainer_llm.py
geoffreyangus: good catch– leftover from debugging. fixed.

ludwig/models/llm.py
geoffreyangus: Upshot: it was in the fast chat code. Leaving as-is for now.

ludwig/data/prompt.py
```diff
@@ -0,0 +1,178 @@
...
+    Returns:
+        Tuple[RetrievalModel, str]: A tuple containing the retrieval model and the name of the index.
+    """
+    retrieval_model = get_retrieval_model(
+        retrieval_config["type"],
+        model_name=retrieval_config["model_name"],
+    )
+
+    index_name = retrieval_config["index_name"]
+    index_cache_directory = os.path.join(get_default_cache_location(), "index")
```
tgaddair, line 85: Use `fs_utils` instead of native Python filesystem API, so we can support s3 and other remote filesystems for the cache.

ludwig/models/llm.py
```diff
@@ -66,6 +91,33 @@ def __init__(
...
+            model_kwargs.update(
+                dict(
+                    low_cpu_mem_usage=True,
+                    torch_dtype=torch.float16,
+                    device_map="auto",
+                    max_memory={i: "13GiB" for i in range(num_gpus)},
+                )
+            )
+
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 114: Worth adding a comment explaining why we need to do this.

ludwig/schema/prompt.py
```diff
@@ -0,0 +1,79 @@
...
+    index_name: str = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description="The name of the index to use for the prompt. Indices are stored in the ludwig cache by default.",
+    )
+
+    model_name: str = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description="The model to use for the prompt.",
```
tgaddair, line 31: Should this be "the model to use to generate embeddings from the prompt for retrieval"?

ludwig/utils/tokenizers.py
```diff
@@ -792,7 +792,14 @@ class HFTokenizer(BaseTokenizer):
     def __init__(self, pretrained_model_name_or_path, **kwargs):
         super().__init__()
         self.pretrained_model_name_or_path = pretrained_model_name_or_path
-        self.tokenizer = load_pretrained_hf_tokenizer(self.pretrained_model_name_or_path)
+
+        # HACK(geoffrey): `LlamaTokenizer` hangs on `use_fast=True`
+        # TODO(geoffrey): figure out if there's a systematic way to determine if fast tokenizer is available
+        pretrained_kwargs = {}
+        if self.pretrained_model_name_or_path in {"eachadea/vicuna-13b-1.1"}:
```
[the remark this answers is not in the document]
arnavgarg1: Does't `use_fast=True` basically not cause problems if there are no Rust tokenizers? My understanding is that it just falls back to `use_fast=False` so it should be safe to always set it to True?

ludwig/utils/tokenizers.py
```diff
@@ -804,7 +811,7 @@ def get_pad_token(self) -> str:
         # HACK(geoffrey): gpt2 has no pad token. Recommendation is to use eos token instead.
         # https://github.com/huggingface/transformers/issues/2630#issuecomment-1290809338
         # https://github.com/huggingface/transformers/issues/2648#issuecomment-616177044
-        if self.pretrained_model_name_or_path == "gpt2":
+        if self.pretrained_model_name_or_path in {"gpt2", "eachadea/vicuna-13b-1.1"}:
```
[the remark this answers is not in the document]
arnavgarg1: Agreed!

My suggestion is to add a few new functions and call them at the end of __init__():

1. set_pad_token()
2. set_pad_token_id()

Arguably setting pad_token should set pad_token_id automatically, but my understanding is that this is not always the case (it can get messed up sometimes), so it's best to explicitly set both

tests/integration_tests/test_llm.py
```diff
@@ -177,3 +180,80 @@ def test_llm_zero_shot_classification(tmpdir, backend, ray_cluster_4cpu):
...
+    df["label"] = np.random.choice([1, 2, 3, 4, 5], size=len(df)).astype(str)  # ensure labels match the feature config
+    df.to_csv(dataset_path, index=False)
+
+    model = LudwigModel(config, backend={**backend, "cache_dir": str(tmpdir)})
+    results = model.train(dataset=dataset_path, output_directory=str(tmpdir), skip_save_processed_input=True)
+
+    # TODO: fix LLM model loading
+    # model = LudwigModel.load(os.path.join(results.output_directory, "model"), backend=backend)
+    preds, _ = model.predict(dataset=dataset_path)
+    preds = convert_preds(backend, preds)
```
arnavgarg1, line 259: Is it worth adding asserts on `preds` to make sure we're getting back non-empty values?

ludwig/utils/tokenizers.py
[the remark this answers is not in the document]
geoffreyangus: I was actually seeing hanging behavior for this tokenizer with `use_fast=True`. So that, coupled with the fact that the tokenizer only has access to the `pretrained_model_name_or_path`, makes it difficult to check ahead of time. 

Another option is to load with `use_fast=False`, do some sort of type check, and then if it is not a `LlamaTokenizer`, then reload the tokenizer with `use_fast=True`. This is sub-optimal because we essentially load every tokenizer twice. Curious what people think of that approach, or any alternatives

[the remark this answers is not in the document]
arnavgarg1: That's super interesting. 

On some of my testing (will message you), I actually didn't run into this issue with vicuna 7b or 13b when I set `use_fast=True`. I did notice however, that it just takes a bit longer to get it loaded up initially, which is maybe why you were seeing hanging behavior? Would be great if you can try this again just to double check as well!

I do agree that re-initializing the tokenizer twice is not worth it, also because I think of preprocessing as a fixed cost and it's okay if it takes longer to tokenize

ludwig/models/llm.py
geoffreyangus: Good point. Added.

ludwig/schema/prompt.py
geoffreyangus: Ah yes, copy-paste. Fixed!

ludwig/data/preprocessing.py
```diff
@@ -1671,88 +1679,65 @@ def _handle_missing_values(
...
-        vocab: List = feature_name_to_preprocessing_parameters[output_feature[NAME]]["vocab"]
-        names_of_features_to_substitute: List = _extract_feature_names_from_prompt(prompt)
+    Args:
+        dataset_cols (Dict[str, Series]): Dataset columns.
+        feature_name_to_preprocessing_parameters (Dict[str, PreprocessingConfigDict]): Mapping from feature name to
+            preprocessing parameters.
+        features (List[FeatureConfigDict]): List of feature configurations.
+        df_engine (DataFrameEngine): Dataframe engine.
+        split_col (Optional[Series], optional): Split column. Defaults to None.
+    """
```
arnavgarg1, line 1700: Is it worth leaving an example of before and after enrichment in the doc-string like before?

ludwig/data/prompt.py
```diff
@@ -0,0 +1,176 @@
...
+
+ASSISTANT:
+"""
+
+
+DEFAULT_FEW_SHOT_PROMPT_TEMPLATE = """Below is relevant context:
+
+CONTEXT: {context}
+
+The context is comprised of labeled samples whose embeddings were similar to that of the sample input. The labels in
```
arnavgarg1, line 31: OOC: should context be CONTEXT to match? Or do we lowercase everything pre-tokenization?

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+            df["data"] = result.tolist()
+            return df
+
+    return _RetrievalModelFn
+
+
+def get_semantic_retrieval_model(model_name):
+    from sentence_transformers import SentenceTransformer
+
+    return SentenceTransformer(model_name, device=get_torch_device())
```
arnavgarg1, line 216: OOC, in multi-gpu setups, where does the semantic retrieval model get placed with the call to `get_torch_device()` since say it will return `cuda`. Does it place it on the CPU if no GPUs are available, or the first GPU device available?

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+        # https://numpy.org/doc/stable/reference/generated/numpy.save.html
+        with open(index_file_path, "wb") as f:
+            np.save(f, self.index)
+
+        index_data_file_path = os.path.join(cache_directory, name + "_data.csv")
+        self.index_data.to_csv(index_data_file_path, index=False)
+
+    def load_index(self, name: str, cache_directory: str):
+        index_file_path = os.path.join(cache_directory, name + ".index")
+        self.index = np.load(index_file_path)
```
arnavgarg1, line 95: It might be worth wrapping this in a try-except block incase there is no index at the cache path

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+    Gets k random indices from the dataset regardless of the query.
+    """
+
+    def __init__(self, **kwargs):
+        self.index = None
+        self.index_data = None
+
+    def create_dataset_index(self, df: pd.DataFrame, backend: "Backend", columns_to_index: Optional[List[str]] = None):
+        self.index = np.array(range(len(df)))
+        self.index_data = df
```
arnavgarg1, line 66: OOC, why don't we filter the `index_data` to just be `columns_to_index`?

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+
+class RandomRetrieval(RetrievalModel):
+    """Random retrieval model.
+
+    Gets k random indices from the dataset regardless of the query.
+    """
+
+    def __init__(self, **kwargs):
+        self.index = None
+        self.index_data = None
```
arnavgarg1, line 62: Should we also maintain a checksum here as well since we are writing the CSV form of the df + indices to disk? That way if the underlying data changes, we can recompute a new random index and also just return a pre-existing index if we've already saved it. I know the computation cost of Random retrieval is practically negligible, but it might be useful for consistency

ludwig/models/retrieval.py
arnavgarg1: Oh I see, maybe this is fine, no need to filter

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+        self.index_data = df
+
+    def _encode(self, row_strs: List[str], backend: "Backend") -> np.ndarray:
+        # only do this step once
+        if self.best_batch_size is None:
+            self.best_batch_size = backend.tune_batch_size(
+                create_semantic_retrieval_model_evaluator(self.model_name, row_strs), len(row_strs)
+            )
+
+        transform_fn = create_semantic_retrieval_model_fn(self.model_name, self.best_batch_size)
```
arnavgarg1, line 144: Not sure how expensive it is to initialize a model using SentenceTransformer, but since `create_semanatic_retrieval_model_evaluator()` and `create_semantic_retrieval_model_fn()` both call `get_semantic_retrieval_model()`, should we just call it once upfront, then pass the initialized model into both of these functions?

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def load_index(self, name: str, cache_directory: str):
+        index_file_path = os.path.join(cache_directory, name + ".index")
+        self.index = faiss.read_index(index_file_path)
+
+        index_data_file_path = os.path.join(cache_directory, name + "_data.csv")
+        self.index_data = pd.read_csv(index_data_file_path)
+
+
+def create_semantic_retrieval_model_evaluator(model_name, samples):
```
arnavgarg1, line 186: nit: type hints

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,225 @@
...
+
+        new_checksum = df_checksum(df_to_index)
+        if self.checksum == new_checksum:
+            # Reuse existing index
+            # TODO(travis): could use an LRU cache to support multiple datasets concurrently
+            return
+        self.checksum = new_checksum
+
+        rows = df_to_index.to_dict(orient="records")
+        row_strs = [json.dumps(r) for r in rows]
```
arnavgarg1, line 129: nit - could potentially turn this into a helper function since search() also does the same thing

ludwig/models/predictor.py
```diff
@@ -77,7 +77,9 @@ def __init__(
 
         # TODO (jeffkinnison): revert to using the requested device for GBMs when device usage is fixed
         self.device = get_torch_device() if not model.type() == MODEL_GBM else "cpu"
-        self.model = model.to(self.device)
+
+        self.model = model
+        self.model = model.to_device(self.device)
```
arnavgarg1, line 82: Can just keep one?

ludwig/models/llm.py
```diff
@@ -219,8 +269,10 @@ def outputs_to_predictions(self, outputs: Dict[str, torch.Tensor]) -> Dict[str,
 
     def save(self, save_path):
         """Saves the model to the given path."""
-        weights_save_path = os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME)
-        torch.save(self.state_dict(), weights_save_path)
+        # TODO: figure out what we want to do about weight saving for LLMs.
```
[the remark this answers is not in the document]
arnavgarg1: Agreed - I have this fixed in my PEFT work

ludwig/models/llm.py
```diff
@@ -66,6 +91,34 @@ def __init__(
...
+                    torch_dtype=torch.float16,
+                    device_map="auto",
+                    max_memory={i: "13GiB" for i in range(num_gpus)},
+                )
+            )
+
+        # we save and reload the weights to ensure that they can be sharded across the GPUs using `from_pretrained`
+        with tempfile.TemporaryDirectory() as tmpdir:
+            self.model.save_pretrained(tmpdir)
+            self.model = AutoModelForCausalLM.from_pretrained(tmpdir, **model_kwargs)
```
arnavgarg1, line 117: A couple of questions/thoughts here:

1. We should only do this if `torch.device` is CUDA, so I'd nest it in the if condition since we can also have device as `mps` in which case I don't think we need to reload like this? In the case of MPS, we might just be able to `self.model.to(device)`. 
2. Is it necessary to save the pretrained model and then load it back in? If I understand the motivation for saving the pretrained model to disk, it's to prevent downloading the files from hf hub again. If that's the case, then we can actually get away without doing this because huggingface maintains its own cache where the downloaded files are saved and it defaults to checking the cache first before downloading from hf hub, so it should be roughly the same performance wise to just do `AutoModelForCausalLM.from_pretrained(self.model_name, **kwargs).

ludwig/models/llm.py
```diff
@@ -33,16 +34,40 @@ def __init__(
...
+        if hasattr(self.model.config, "max_sequence_length"):
+            self.context_len = self.model.config.max_sequence_length
+        elif hasattr(self.model.config, "max_position_embeddings"):
+            self.context_len = self.model.config.max_position_embeddings
+        else:
+            self.context_len = 2048
+        self.max_input_length = self.context_len - self.max_new_tokens - 8
+
+        # Used only for its metadata about the vocabulary
+        self.tokenizer = AutoTokenizer.from_pretrained(self.config_obj.model_name, use_fast=False)
```
arnavgarg1, line 68: Just as we discussed, I think we can set `use_fast` to True if we set the protobuf dependency to be pinned to `3.20.3` with `tokenizers` set to 0.13.3. But also ok to leave it like this for now with a TODO to come back to making this `use_fast` by default.

arnavgarg1: Also damn, do we have to initialize the tokenizer just to grab the `bos_token_id`? That kinda sucks and is super unfortunate

ludwig/data/prompt.py
```diff
@@ -0,0 +1,176 @@
...
+    return result
+
+
+def _validate_prompt_template(template: str, is_few_shot: bool):
+    """Validates that the template contains the necessary fields for the prompt."""
+    if is_few_shot:
+        required_fields = {"context", "sample_input", "task"}
+    else:
+        required_fields = {"sample_input", "task"}
+    template_fields = {field for _, field, _, _ in string.Formatter().parse(template) if field is not None}
```
arnavgarg1, line 173: Slick!

ludwig/data/prompt.py
```diff
@@ -0,0 +1,176 @@
...
+    try:
+        _validate_prompt_template(template, is_few_shot=is_few_shot)
+    except ValueError as e:
+        raise ValueError(f"template invalid for {'few-shot' if is_few_shot else 'zero-shot'} prompt: {e}")
+
+    def generate_prompt(series: pd.Series):
+        df = series.to_frame(name=input_col_name)
+
+        if is_few_shot:
+            df["context"] = retrieval_model.search(df, backend, k=3, return_data=True)
```
arnavgarg1, line 145: Probably don't want to hard-code k right?

ludwig/data/prompt.py
```diff
@@ -0,0 +1,176 @@
...
+        index_name = f"embedding_index_{uuid.uuid4()}"
+        logger.info(f"Saving index to cache directory '{index_cache_directory}' with name '{index_name}'")
+        retrieval_model.save_index(index_name, cache_directory=index_cache_directory)
+    else:
+        logger.info(f"Loading index from cache directory '{index_cache_directory}' with name '{index_name}'")
+        retrieval_model.load_index(index_name, cache_directory=index_cache_directory)
+    return retrieval_model, index_name
+
+
+def format_input_with_prompt(
```
arnavgarg1, line 105: This function is so neat 😲

ludwig/schema/prompt.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    index_name: str = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description="The name of the index to use for the prompt. Indices are stored in the ludwig cache by default.",
+    )
+
+    model_name: str = schema_utils.String(
+        default=None,
```
arnavgarg1, line 29: If we're going to leave this as None, then we should make sure there's some sort of validation check somewhere to raise an error at the time of building the retrieval index with a useful error message.

In hindsight, I should probably add that for LLM creation as well to `model_name` make sure a value is passed in

arnavgarg1:
Would also be great if you can add an example for few shot classification!

ludwig/data/prompt.py
geoffreyangus: Made context CONTEXT for clarity.

We do lowercase everything by default, though I think with LLMs we may want to explore the performance deltas between lowercase=True and lowercase=False, since I think it might make a difference particularly in cases like this where it is more like a variable name/identifier than an actual word.

ludwig/data/prompt.py
geoffreyangus: Done.

ludwig/models/llm.py
geoffreyangus: Yeah that's right– because its just for the `bos_token_id`, maybe I'll just leave it as-is since it will load much faster. Here's the Linear ticket to improve this downstream: https://linear.app/predibase/issue/MLX-1082/add-tokenizer-cache-to-ensure-that-fast-llamatokenizer-is-only-loaded

ludwig/models/predictor.py
geoffreyangus: Done

ludwig/models/retrieval.py
geoffreyangus: Done

ludwig/models/retrieval.py
geoffreyangus: It should be placed on whatever device is visible to Ray Datasets, so in theory it should be placed on the next free GPU (if GPU resources were requested by the job)

ludwig/utils/tokenizers.py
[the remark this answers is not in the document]
geoffreyangus: Done

ludwig/utils/tokenizers.py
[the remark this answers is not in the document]
geoffreyangus: Done

tests/integration_tests/test_llm.py
geoffreyangus: Done

ludwig/models/llm.py
geoffreyangus: Ah, so I had left a comment that may have not been pushed. The reason we save it and load it back in is so that we can  place it onto the new device, since HF's `from_pretrained` implements device placement in a simple way that pytorch does not (see torch docs for their very low-level approach: https://pytorch.org/tutorials/intermediate/model_parallel_tutorial.html)

geoffreyangus: In any case, I've added the `else` code path to ensure we call `self.model.to(device)` for non-CUDA device placements.

ludwig/models/retrieval.py
geoffreyangus: I think we do want it to fail with the default `FileNotFoundError` if there is no index at the cache path, so I'm leaning towards leaving it as-is. Is there a reason we might want to do otherwise?

ludwig/models/retrieval.py
geoffreyangus: Ah yes– let's try that. Implemented it in the most recent PR revision, will test it out and see if it works.

ludwig/models/retrieval.py
arnavgarg1: That makes sense yeah, we can leave it as is!

ludwig/schema/prompt.py
geoffreyangus: Added a config validation check for this

ludwig/schema/prompt.py
```diff
@@ -0,0 +1,79 @@
...
+        allow_none=True,
+        description="The model to use for the prompt.",
+    )
+
+    k: int = schema_utils.NonNegativeInteger(
+        default=0,
```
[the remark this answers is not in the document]
geoffreyangus: Done.

ludwig/backend/ray.py
```diff
@@ -163,6 +164,7 @@ def train_fn(
     distributed_strategy: Union[str, Dict[str, Any]],
     executable_kwargs: Dict[str, Any] = None,
     model_ref: ObjectRef = None,  # noqa: F821
+    remote_trainer_cls: BaseTrainer = None,  # noqa: F821
```
tgaddair, line 167: Nit: `Type[BaseTrainer]`

ludwig/config_validation/checks.py
```diff
@@ -234,6 +235,33 @@ def check_validation_metric_exists(config: "ModelConfig") -> None:  # noqa: F821
...
+        if input_feature.type == TEXT:
+            if input_feature.preprocessing.prompt.task is not None:
+
+                _check_k_retrieval_config(input_feature.preprocessing.prompt.retrieval)
+                _check_model_name_retrieval_config(input_feature.preprocessing.prompt.retrieval)
+
+
+def _check_k_retrieval_config(retrieval_config: "RetrievalConfig") -> None:
+    """Checks that k is greater than zero if retrieval type is not None."""
+    if retrieval_config.type is None and retrieval_config.k != 0:
```
tgaddair, line 251: We can clean this up in the future by having a dynamically loaded schema based on the selection of the `type` param. That way we can auto-enforce that `k` isn't set if the type is empty, etc. This can be a follow-up PR, but I would leave a TODO for now.

ludwig/models/llm.py
```diff
@@ -33,16 +34,40 @@ def __init__(
...
+        self.max_new_tokens = self.config_obj.generation.max_new_tokens
+
+        # Determines the maximum length of the context (input + output tokens)
+        if hasattr(self.model.config, "max_sequence_length"):
+            self.context_len = self.model.config.max_sequence_length
+        elif hasattr(self.model.config, "max_position_embeddings"):
+            self.context_len = self.model.config.max_position_embeddings
+        else:
+            self.context_len = 2048
+        self.max_input_length = self.context_len - self.max_new_tokens - 8
```
tgaddair, line 65: Worth adding a comment about where this logic came from (FastChat), or explain why we subtract 8.

ludwig/models/llm.py
```diff
@@ -66,6 +91,35 @@ def __init__(
...
+            return self
+        else:
+            logger.info(f"Moving LLM from '{self.curr_device}' to '{device}'.")
+
+        model_kwargs = {}
+        if device == torch.device("cuda"):
+            num_gpus = torch.cuda.device_count()
+            model_kwargs.update(
+                dict(
+                    low_cpu_mem_usage=True,
```
tgaddair, line 107: TODO here about making this configurable in the future. I would also reference FastChat as the source of these params.

ludwig/models/retrieval.py
```diff
@@ -0,0 +1,224 @@
...
+    return hashlib.sha1(pd.util.hash_pandas_object(df).values).hexdigest()
+
+
+def df_to_row_strs(df: pd.DataFrame) -> List[str]:
+    rows = df.to_dict(orient="records")
+    row_strs = [json.dumps(r) for r in rows]
+    return row_strs
+
+
+class RetrievalModel:
```
tgaddair, line 32: Nit: use `ABC` as parent and decorate all methods as `@abstractmethod` with implementation `pass`.

ludwig/schema/utils.py
```diff
@@ -51,10 +51,15 @@ def load_trainer_with_kwargs(
...
     otherwise passes all other parameters through without change.
     """
-    from ludwig.constants import MODEL_ECD
-    from ludwig.schema.trainer import ECDTrainerConfig, GBMTrainerConfig
-
-    trainer_schema = ECDTrainerConfig if model_type == MODEL_ECD else GBMTrainerConfig
+    from ludwig.constants import MODEL_ECD, MODEL_GBM, MODEL_LLM
+    from ludwig.schema.trainer import ECDTrainerConfig, GBMTrainerConfig, LLMTrainerConfig
+
+    if model_type == MODEL_ECD:
```
tgaddair, line 57: Good candidate for a registry pattern. Maybe as a TODO.

requirements_test.txt
```diff
@@ -42,3 +42,6 @@ scikit-optimize
 zoopt
 
 s3fs>=2022.8.2
+
+# required for explain tests
+captum
```
tgaddair, line 47: Should you just install the `requirements_explain.txt` deps for the LLM test suite?

tgaddair (approved):
LGTM! Few nits.

ludwig/models/llm.py
geoffreyangus: Done.

requirements_test.txt
geoffreyangus: Done

ludwig/models/retrieval.py
geoffreyangus: Done

ludwig/models/llm.py
geoffreyangus: Done

ludwig/config_validation/checks.py
geoffreyangus: Done

ludwig/backend/ray.py
geoffreyangus: Done

ludwig/schema/utils.py
geoffreyangus: Done

arnavgarg1 (approved):
Great work!! 🚢

PR #3352, opened 2023-04-17
[LLM] Fix tokenizer padding, target/tensor alignment and enable Ray backend support for text generation tasks

arnavgarg1:
1. Fixes an issue when tokenizers for LLMs don't have pad_token_ids set
2. Improves the example LLM notebooks
3. Fixes an issue with tensor alignment where the probability/logit tensors weren't being re-aligned


arnavgarg1:
Only land once https://github.com/ludwig-ai/ludwig/pull/3301 lands since it fixes issues with Ray backend support for text generation tasks

PR #3356, opened 2023-04-19
fix: Fix dataset reading for parquet directories

ksbrar:


justinxzhao (approved):

abidwael (approved):

jeffkinnison (approved):

ksbrar:
Last test failures are due to flakiness. Seems like this PR fixes the underlying parquet read issue but can also fail without the right credentials context.

PR #3359, opened 2023-04-20
[LLM] Fine-Tune LLMs via Prompt tuning

arnavgarg1:
This PR adds the ability to fine-tune LLM model types via Prompt Tuning using PEFT. The only change required to the config is as follows:

```yaml
adapter:
    type: prompt_tuning
    num_virtual_tokens: 32
    prompt_tuning_init_text: "Rate this review from 1 through 5:"
```

This considerably reduces the number of parameters required to fine-tune by keeping the pretrained LLMs weights frozen while adding an adapter (in this case, a PromptEmbeddingLayer) to fine-tune the weights of the prompt instead of tuning all the weights of the LLM.

With most LLMs, this should result in only tuning < 1% of total model parameters during back-propagation.


ludwig/models/llm.py
```diff
@@ -49,7 +48,29 @@ def __init__(
...
 
-        self.max_new_tokens = self.config_obj.generation.max_new_tokens
+        # If an adapter config is provided, we want to wrap the model with a PEFT model
+        # for fine-tuning.
+        if self.config_obj.adapter:
+            from peft import get_peft_config, get_peft_model
+
+            # If the adapter config specifies a tokenizer name or path, it must match the model name
+            # Otherwise, we will set the tokenizer name or path to the model name
+            if (
```
tgaddair, line 58: I would just take this out of the user's control. Don't let them specify it in the config, and always set it to model name. We can do this in a follow-up, though.

ludwig/schema/trainer.py
```diff
@@ -753,12 +768,38 @@ class LLMTrainerConfig(BaseTrainerConfig):
...
 class ZeroShotTrainerConfig(LLMTrainerConfig):
     """Dataclass that configures most of the hyperparameters used for zero-shot LLM model training."""
 
-    pass
+    # Required for lookup during trainer initialization
+    type: str = schema_utils.ProtectedString("zeroshot")
+
+
+@DeveloperAPI
+@register_llm_trainer_schema("fewshot")
```
tgaddair, line 781: Another follow-up: we could just have a single trainer, call it "PassthroughTrainer" or something, and then have it used by default when no trainer is specified.

tgaddair (approved):
LGTM. Following the LoRA PR, we can consolidate some of the PEFT stuff. Few follow-up items as well that I don't think are blocking.

ludwig/models/llm.py
arnavgarg1: I agree, Done!

ludwig/schema/trainer.py
arnavgarg1: Agreed, will add this in a follow-up PR

PR #3361, opened 2023-04-20
Fixed distributed strategy registration to be explicit

tgaddair:
Previously were seeing issues were DDP was getting picked up even though FSDP was being selected.

### DIFF
--- a/ludwig/backend/ray.py
+++ b/ludwig/backend/ray.py
@@ -36,7 +36,7 @@ from ray.train.torch import TorchCheckpoint
 from ray.util.dask import ray_dask_get
 from ray.util.placement_group import placement_group, remove_placement_group
 
-from ludwig.distributed import get_current_dist_strategy, get_default_strategy_name, get_dist_strategy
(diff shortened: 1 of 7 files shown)


ludwig/backend/ray.py
```diff
@@ -244,7 +245,7 @@ def tune_batch_size_fn(
     # inside of a RayTrainer class, so we manually set the local_rank to 0 so that it picks up the right
     # device to tune batch size on.
     initialize_pytorch(local_rank=0, local_size=_local_size())
-    distributed = get_current_dist_strategy(allow_local=True)()
+    distributed = init_dist_strategy("local")
```
abidwael, line 248: Will this still be `local` after having merged https://github.com/ludwig-ai/ludwig/pull/2934?

tgaddair: Yes, good catch, we resolved this when merging the distirubted batch size tuning pr, so should be good:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L249

PR #3362, opened 2023-04-22
Added DeepSpeed distributed strategy and backend

tgaddair:
This PR introduces two ways to run DeepSpeed to train large models (like LLMs):

- DeepSpeed backend
- Ray backend + DeepSpeed strategy

**DeepSpeed backend:**

Config:

```
backend:
    type: deepspeed
    zero_optimization:
        stage: 3
```

CLI:

```
deepspeed --num_gpus 4 ludwig train --config config.yaml --dataset dataset.csv
```

**Ray backend + DeepSpeed strategy:**

Config:

```
backend:
    type: ray
    trainer:
        strategy:
            type: deepspeed
            zero_optimization:
                stage: 3
        num_workers: 4
        use_gpu: True
```

CLI:

```
ray exec ludwig train --config config.yaml --dataset dataset.csv
```

Note that using the DeepSpeed backend will use Pandas for data processing, and as such does not support larger-than-memory datasets. For large datasets, use the Ray backend with the DeepSpeed strategy.

TODO:

- [x] Fix saving and loading for sharded models
- [x] Fix Ray backend
- [ ] Fix batch size tuning for DeepSpeed backend
- [x] Fix batch size tuning with Ray backend (requires #2934).
- [x] Support DeepSpeed [optimizers](https://deepspeed.readthedocs.io/en/latest/optimizers.html).

Known issues / limitations:

- DeepSpeed cannot run the model in eval mode (see: https://github.com/microsoft/DeepSpeed/issues/3068). This means dropout, batch norm, etc. will still be enabled when computing validation / test metrics.


examples/fintuning/train_imdb.py
```diff
@@ -0,0 +1,96 @@
...
+    experiment_name="imdb_sentiment",
+    model_name="parallel_cnn",
+)
+
+# list contents of output directory
+print("contents of output directory:", output_directory)
+for item in os.listdir(output_directory):
+    print("\t", item)
+
+# batch prediction
```
geoffreyangus, line 78: Remove?

examples/fintuning/train_imdb.py
```diff
@@ -0,0 +1,96 @@
...
+from ludwig.api import LudwigModel
+
+# from ludwig.datasets import imdb
+
+# clean out prior results
+shutil.rmtree("./results", ignore_errors=True)
+# model_dir = "/src/results/simple_experiment_simple_model/model"
+
+# Download and prepare the dataset
+# training_set, test_set, _ = titanic.load(split=True)
```
geoffreyangus, line 16: Remove?

examples/titanic/simple_model_training.py
```diff
@@ -13,13 +13,15 @@
 import yaml
 
 from ludwig.api import LudwigModel
-from ludwig.datasets import titanic
+
+# from ludwig.datasets import titanic
```
geoffreyangus, line 17: Remove?

ludwig/distributed/base.py
```diff
@@ -20,11 +26,24 @@ class DistributedStrategy(ABC):
...
     @abstractmethod
-    def wrap_model(self, model: nn.Module) -> nn.Module:
+    def prepare(
+        self, model: nn.Module, optimizer: Optimizer, lr_scheduler: "LRScheduler", trainer_config: "ECDTrainerConfig"
+    ) -> Tuple[nn.Module, Optimizer, "LRScheduler"]:
         pass
 
-    @abstractmethod
-    def wrap_optimizer(self, optimizer: Optimizer, model: nn.Module, gradient_accumulation_steps: int) -> Optimizer:
+    def to_device(self, model: nn.Module) -> nn.Module:
```
geoffreyangus, line 34: should this take device as optional?

ludwig/features/base_feature.py
```diff
@@ -97,6 +98,16 @@ def add_feature_data(
         raise NotImplementedError
 
 
+@dataclass
+class ModuleWrapper:
+    """Used to prevent the PredictModule from showing up an attribute on the feature module.
+
+    This is necessary to avoid inflight errors from DeepSpeed.
```
geoffreyangus, line 105: What is an inflight error?

ludwig/features/base_feature.py
```diff
@@ -110,6 +121,9 @@ def __init__(self):
         self.probabilities_key = PROBABILITIES
         self.logits_key = LOGITS
 
+    def forward(self, *args, **kwargs):
```
geoffreyangus, line 124: How is this not infinitely recursive?

ludwig/models/predictor.py
```diff
@@ -65,23 +66,33 @@ class Predictor(BasePredictor):
...
-        self.device = get_torch_device() if not model.type() == MODEL_GBM else "cpu"
-        self.model = model.to(self.device)
+        device = get_torch_device()
+        if model.type() == MODEL_GBM:
+            # TODO (jeffkinnison): revert to using the requested device for GBMs when device usage is fixed
+            device = "cpu"
+            dist_model = dist_model.to(device)
+
+        self.device = device
+        self.dist_model = dist_model
```
geoffreyangus, line 90: Can you leave a comment on the difference between these two models? Maybe as part of an overall class docstring

ludwig/backend/deepspeed.py
```diff
@@ -0,0 +1,27 @@
...
+
+import deepspeed
+
+from ludwig.backend.base import DataParallelBackend
+from ludwig.constants import FALLBACK_BATCH_SIZE
+from ludwig.distributed import init_dist_strategy
+from ludwig.utils.batch_size_tuner import BatchSizeEvaluator
+
+
+class DeepSpeedBackend(DataParallelBackend):
```
geoffreyangus, line 11: How does this new backend interact with Ray/Ray Datasets? Noticed that DataParallelBackend uses the LocalPreprocessingMixin, so curious what you're thinking here.

ludwig/utils/state_dict_backward_compatibility.py
```diff
@@ -24,7 +24,19 @@ def _update_transformers_to_freeze_module(state_dict):
...
+def _update_combiner_no_device_tensor(state_dict):
+    """Removed device_tensor from state_dict following DeepSpeed integration."""
+    return {k: v for k, v in state_dict.items() if not k.endswith("device_tensor")}
+
+
 def update_state_dict(state_dict):
     """Checks state_dict on load, updates state dict if needed."""
     state_dict = _update_transformers_to_freeze_module(state_dict)
+    state_dict = _update_combiner_no_input_features(state_dict)
+    state_dict = _update_combiner_no_device_tensor(state_dict)
```
geoffreyangus, line 41: Nice change

geoffreyangus (commented):
Nice work on this! Left some questions, but overall looking really good.

ludwig/backend/deepspeed.py
tgaddair: It works the same as the existing HorovodBackend. You run with the command line tool `deepspeed`. It processes the data in-memory (like local training). It doesn't use Ray (that's the DeepSpeedStrategy).

ludwig/features/base_feature.py
tgaddair: Hmm, good question! Shouldn't be needed now that we made it an nn.Module again. Removed!

examples/llm_finetuning/README.md
arnavgarg1, line 1: Assuming you're going to land instructions to run the configs in the README?

examples/llm_finetuning/train_imdb_api.py
```diff
@@ -0,0 +1,98 @@
...
+# }
+# backend = {
+#     "type": "deepspeed",
+#     "zero_optimization": {"stage": 3},
+# }
+
+# print(model_dir)
+# model = LudwigModel.load(model_dir, backend=backend)
+# results, _ = model.predict("/home/ray/titanic.csv", skip_save_predictions=False)
+# print(results)
```
arnavgarg1, line 98: Cleanup?

tgaddair: Yeah, this is all messy as I'm still benchmarking and testing some things.

ludwig/data/dataset/ray.py
```diff
@@ -375,7 +375,7 @@ def _fetch_next_batch(self):
             self._next_batch = next(self.dataset_batch_iter)
             # If the batch has only one row and self.ignore_last, skip the batch
             # to prevent batchnorm / dropout related Torch errors
-            if self.ignore_last and len(self._next_batch[self._sample_feature_name]) == 1:
+            if self.batch_size > 1 and self.ignore_last and len(self._next_batch[self._sample_feature_name]) == 1:
```
arnavgarg1, line 378: nice catch! i noticed this with my prompt tuning PR as well that we go into an infinite loop with batch size 1

ludwig/distributed/base.py
```diff
@@ -20,11 +27,27 @@ class DistributedStrategy(ABC):
     """
 
     @abstractmethod
-    def wrap_model(self, model: nn.Module) -> nn.Module:
+    def prepare(
+        self,
+        model: nn.Module,
+        trainer_config: "ECDTrainerConfig",
+        base_learning_rate: float,
+    ) -> Tuple[nn.Module, Optimizer]:
```
arnavgarg1, line 35: It might be useful to add a docstring here indicating what the `prepare` function is supposed to do. From below, it seems like we're returning the model/wrapping the model with the relevant strategy initialization + creating the optimizer?

ludwig/backend/deepspeed.py
geoffreyangus: Nice, just read the PR description. Thanks!

geoffreyangus (approved):

ludwig/distributed/deepspeed.py
```diff
@@ -0,0 +1,183 @@
...
+)
+
+
+class DeepSpeedStrategy(DDPStrategy):
+    def __init__(self, zero_optimization: Optional[Dict[str, Any]] = None, **kwargs):
+        # If we're initializing from a `deepspeed` CLI command, deepspeed will have already been initialized, as
+        # indicated by the presence of the LOCAL_RANK var. Otherwise, we're initializing from Ray / torchrun, and will
+        # need to set this var ourselves, then init DeepSpeed here.
+        local_rank, local_size = os.environ.get("LOCAL_RANK"), os.environ.get("LOCAL_SIZE")
+        init_deepspeed = local_rank is None or local_size is None
```
arnavgarg1, line 43: Do we need both to be set for `deepspeed.init_distributed()` to work?

arnavgarg1, line 43: If both of these are set, is the reason we don't need to call `deepspeed.init_distributed()` because deepspeed is already initialized? When would that be the case?

ludwig/distributed/deepspeed.py
```diff
@@ -0,0 +1,183 @@
...
+        logging.info("Using DeepSpeed strategy")
+
+    def prepare(
+        self,
+        model: nn.Module,
+        trainer_config: "ECDTrainerConfig",
+        base_learning_rate: float,
+    ) -> Tuple[nn.Module, Optimizer]:
+        # If `batch_size=auto`, we set to 2 temporarily until auto-tuning adjusts it`
+        batch_size = trainer_config.batch_size if isinstance(trainer_config.batch_size, int) else 2
```
arnavgarg1, line 63: Should we log for visibility that we are setting the batch_size to 2 if it was `auto`?

ludwig/modules/optimization_modules.py
```diff
@@ -13,50 +13,57 @@
...
-from ludwig.schema.optimizers import BaseOptimizerConfig, GradientClippingConfig, optimizer_registry
 from ludwig.utils.misc_utils import get_from_registry
 from ludwig.utils.torch_utils import LudwigModule
 
+if TYPE_CHECKING:
+    from ludwig.schema.optimizers import BaseOptimizerConfig, GradientClippingConfig
+
+
+def create_clipper(gradient_clipping_config: Optional["GradientClippingConfig"]):
+    from ludwig.schema.optimizers import GradientClippingConfig
```
arnavgarg1, line 28: Do we need to import again if it's already imported under TYPE_CHECKING?

ludwig/trainers/trainer.py
```diff
@@ -386,15 +388,26 @@ def tune_batch_size(
...
+                return self.distributed.broadcast_object(best_batch_size)
+            finally:
+                # Restore original parameters to defaults
+                self.skip_save_model = skip_save_model
+                self.skip_save_progress = skip_save_progress
+                self.skip_save_log = skip_save_log
+                if snapshot_weights:
+                    if self.distributed.prepare_before_load():
+                        self.prepare()
+                    self.resume_weights_and_optimizer(str(tmpdir), checkpoint)
```
arnavgarg1, line 410: Can we add a comment for why we're doing this / when we have to do this?

ludwig/modules/optimization_modules.py
tgaddair: `TYPE_CHECKING` doesn't import it at all, it's only used for static analysis of the type hints.

arnavgarg1: Ah sorry, my bad

ludwig/distributed/deepspeed.py
tgaddair: We don't need to `init_distributed` when using Ray because we're already in a `torch.distributed` environment, but when using the `deepspeed` launch we need to call `init_distributed` to be in that context.

ludwig/data/dataset/ray.py
tgaddair: Yeah, this was casuing me problems today!

examples/llm_finetuning/README.md
tgaddair: Yep, done.

examples/llm_finetuning/train_imdb_api.py
tgaddair: Updated.

ludwig/distributed/deepspeed.py
tgaddair: It shouldn't be needed because we override it, it's just so DeepSpeed doesn't comnplain.

ludwig/trainers/trainer.py
```diff
@@ -540,23 +555,36 @@ def run_evaluation(
...
+                checkpoint_manager,
             )
         else:
             # There's no validation, so we save the model.
-            if self.is_coordinator() and not self.skip_save_model:
-                self.model.save(save_path)
+            if not self.skip_save_model:
+                logger.info("Saving model.\n")
+                checkpoint_manager.save_best(progress_tracker.steps)
+                self.callback(lambda c: c.on_save_best_checkpoint(self, progress_tracker, save_path))
```
arnavgarg1, line 565: Does this save on the coordinator automatically?

ludwig/trainers/trainer.py
```diff
@@ -765,19 +801,28 @@ def train(self, training_set, validation_set=None, test_set=None, save_path="mod
...
+        if not self.skip_save_model:
+            state_dict = checkpoint_manager.get_best_checkpoint_state_for_inference(self.return_device)
+            if not return_state_dict:
+                if self.distributed.is_model_parallel():
+                    # Assume the full weights cannot fit in memory on GPU
+                    self.model = self.model.cpu()
+                self.model.load_state_dict(state_dict)
+        elif return_state_dict:
+            state_dict = self.model.cpu().state_dict()
+        return_value = self.model if not return_state_dict else state_dict
```
arnavgarg1, line 818: Are you able to leave a comment on why/when we'd want to return the model state dict as opposed to the model with the loaded state dict?

arnavgarg1: I assume it is because each of the workers can return their individual state dicts (because of the model parallelism), and then it's all combined back on the coordinator, but I can't seem to see the conditional here for is_coordinator() to load the combined state dicts back into the model object

ludwig/trainers/trainer.py
arnavgarg1: Ohhhh I see, they're all CoordinatorCheckpoint classes that internally do the `is_coordinator()` checks, so this makes sense!

ludwig/utils/checkpoint_utils.py
```diff
@@ -65,28 +71,65 @@ def get_latest_checkpoint_path(directory: str) -> str:
...
+
+
+@DeveloperAPI
+class CoordinatorCheckpoint(Checkpoint):
+    def prepare(self, directory: str):
+        if self.distributed.is_coordinator():
+            super().prepare(directory)
+
+    def load(self, save_path: str, device: Optional[torch.device] = None) -> bool:
+        """Load state from a saved checkpoint.
```
arnavgarg1, line 124: With model parallelism, how does the model get loaded back? Is the idea that we load on the main coordinator device, then shard across devices again? If yes, is there a need to wrap the load call with `self.distributed.is_coordinator()`?

ludwig/trainers/trainer.py
tgaddair: Ray sends the state dict back to the caller, not the model, so loading the weights back into the model is wasteful.

ludwig/utils/checkpoint_utils.py
tgaddair: Each worker loads is part of the model independently, so each worker needs to call it.

ludwig/trainers/trainer.py
tgaddair: Good point, though, for hyperopt we only want to save the model on the coordinator. Updated.

ludwig/distributed/deepspeed.py
arnavgarg1: Ahhh got it, nice!

arnavgarg1 (approved):
Amazing work, super excited to try deepspeed 🚢  🚀 . Thanks for leaving comments - they definitely make it much more clear!

PR #3366, opened 2023-04-25
temporary `pip install expecttest` for torch nightly tests

abidwael:


tgaddair (approved):
Can you file an issue with pytorch and ask them to include this dependency as a requirement?

tgaddair:
Okay, I see there's already an issue. Maybe instead of having an issue tracked in ludwig, let's just reference the upstream issue, since that's what will tell us when the issue is fixed.

abidwael:
Filed an issue here: https://github.com/pytorch/pytorch/issues/100002

PR #3369, opened 2023-04-27
Ensure worker synchronization during resume

abidwael:
This PR ensures
- We're not accessing keys that don't exist in the progress tracker dictionary.
- All keys of the progress tracker dictionary are available for initialization.
- All workers are synced on whether to resume or create a new progress tracker (due to inconsistencies in the output of `resume_files_exist`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,6 @@ tags
 # examples
 examples/*/data/
 examples/*/visualizations/
+
+# benchmarking configs
+ludwig/benchmarking/configs/
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -605,12 +605,31 @@ class Trainer(BaseTrainer):
                 test_summary_writer = SummaryWriter(os.path.join(tensorboard_log_dir, TEST))
 
         # ================ Resume logic ================
-        if self.resume and self.resume_files_exist(training_progress_tracker_path, training_checkpoints_path):
-            logger.info("Resuming training from previous run.")
-            progress_tracker = self.resume_training_progress_tracker(training_progress_tracker_path)
-            self.resume_weights_and_optimizer(training_checkpoints_path, checkpoint)
+        self.callback(lambda c: c.on_resume_training(self.is_coordinator()))
+
+        should_resume = self.resume and self.resume_files_exist(
+            training_progress_tracker_path, training_checkpoints_path
+        )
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -136,3 +136,6 @@ tags
 # examples
 examples/*/data/
 examples/*/visualizations/
+
+# benchmarking configs
+ludwig/benchmarking/configs/
```
justinxzhao, line 141: Is this right?

ludwig/trainers/trainer.py
```diff
@@ -605,19 +605,39 @@ def train(self, training_set, validation_set=None, test_set=None, save_path="mod
...
+        )
+        # make sure all workers are on the same page about resuming.
+        should_resume = self.distributed.broadcast_object(should_resume, name="should_resume")
+
+        if should_resume:
+            try:
+                progress_tracker = self.resume_training_progress_tracker(training_progress_tracker_path)
+                self.resume_weights_and_optimizer(training_checkpoints_path, checkpoint)
+                logger.info("Resuming training from previous run.")
+            except Exception:
```
justinxzhao, line 621: Nice, I think this makes sense.

ludwig/trainers/trainer.py
```diff
@@ -605,19 +605,39 @@ def train(self, training_set, validation_set=None, test_set=None, save_path="mod
...
+                progress_tracker = get_new_progress_tracker(
+                    batch_size=self.batch_size,
+                    learning_rate=self.base_learning_rate,
+                    best_eval_metric_value=get_initial_validation_value(self.validation_metric),
+                    best_increase_batch_size_eval_metric=get_initial_validation_value(
+                        self.increase_batch_size_eval_metric
+                    ),
+                    output_features=output_features,
+                )
+                logger.info("Failed to resume training from previous run. Creating fresh model training run.")
```
justinxzhao, line 631: nit: Could you add a comment about when resuming would fail, even when `self.resume and self.resume_files_exist`, i.e. "This may happen if model training is interrupted after the progress tracker is initialized but before any real training progress is made"?

justinxzhao (approved):
Thanks!

.gitignore
abidwael: It's working as expected on my local so I believe so

ludwig/trainers/trainer.py
abidwael: added

PR #3373, opened 2023-04-28
Default to LocalStrategy for GBM metrics

abidwael:
There's no initialized distributed strategy for GBM. We default to `LocalStrategy` to get the right `gather_all_tensors_fn` for GBM model metrics.

Fixing `RuntimeError: Distributed strategy not initialized (type: RayTaskError(RuntimeError), retryable: true)`


justinxzhao (approved):

justinxzhao:
Is there a test that you could add?

jeffkinnison (commented):
Echoing others: this looks good, just needs a test :)

ludwig/modules/metric_modules.py
```diff
@@ -109,7 +110,11 @@ def sync_context(
         distributed_available: Optional[Callable] = jit_distributed_available,
     ) -> Generator:
         """Override the behavior of this in the base class to support custom distributed strategies."""
-        dist_strategy = get_current_dist_strategy()
+        try:
+            dist_strategy = get_current_dist_strategy()
+        except RuntimeError:
+            # We don't initialize distributed strategy for GBM models. Default to LocalStrategy.
+            dist_strategy = LocalStrategy()
```
tgaddair, line 117: I don't like this solution. The idea behind making the distributed strategy explicit is that we should always be setting it. The problem with this approach is that it allows us to accidentally forget to set it, then we get unexpected behavior. I would prefer we find out why we're not setting the distributed strategy for GBMs when we should be.

tgaddair (changes requested):

.github/workflows/pytest.yml
```diff
@@ -230,9 +230,14 @@ jobs:
         run: |
           pytest -rx --durations 100 -m "combinatorial" --junitxml pytest.xml tests/training_success
 
-  test-minimal-install:
-    name: Test Minimal Install
+  test-distributed-minimal-install:
```
tgaddair, line 233: Please revert this. This minimal install is needed to verify that what we put in the Getting Started works as-advertised. This is not the same thing. You can add a new test if you want to test the full dependencies.

examples/getting_started/benchmarking_config_titanic.yaml
```diff
@@ -0,0 +1,11 @@
+experiment_name: titanic_gbm_benchmarking_run
```
tgaddair, line 1: This is a bit overhwelming to put in getting started. I would move it to a `benchmarking` directory in the examples folder.

examples/getting_started/run.sh
```diff
@@ -18,3 +18,6 @@ ludwig train --config ${SCRIPT_DIR}/rotten_tomatoes.yaml --dataset rotten_tomato
 
 # Predict and Evaluate
 ludwig predict --model_path results/experiment_run/model --dataset rotten_tomatoes_test.csv
+
+# Train and evaluate distributed ECD and GBM models
+ludwig benchmark --benchmarking_config benchmarking_config_titanic.yaml
```
tgaddair, line 23: Same.

ludwig/trainers/trainer_lightgbm.py
```diff
@@ -565,6 +566,7 @@ def train(
         save_path="model",
         **kwargs,
     ):
+        init_dist_strategy("local")
```
tgaddair, line 569: Why is this not being executed by `LocalTrainingMixin.initialize()`? The distribution strategy initialization is intended to live in the backend, not the trainer.

abidwael: The pattern that I followed was how we call `init_dist_strategy` in `train_fn` [here](https://github.com/ludwig-ai/ludwig/blob/1529e02e0eff40182c7b607af530ab12aad42bca/ludwig/backend/ray.py#L172). We have an `initialize` function in `LocalTrainingMixin` and `HorovodBackend`. For the `RayBackend`, `initialize` just calls `ray.init`. I’ll move `init_dist_strategy` to be in backend.initialize functions.

tests/integration_tests/test_gbm.py
```diff
@@ -174,6 +174,13 @@ def test_local_gbm_number(tmpdir, local_backend):
     run_test_gbm_number(tmpdir, local_backend)
 
 
+@pytest.mark.distributed
+def test_ray_gbm_number_remote(tmpdir, ray_backend, ray_cluster_5cpu):
```
tgaddair, line 178: Would be good to add a comment here explaining why we want to test this specifically.

tgaddair (approved):
LGTM!

PR #3375, opened 2023-05-01
Implemented CORN loss for ordinal classification

tgaddair:
See: https://github.com/Raschka-research-group/coral-pytorch

When using the CORN loss, the vocab needs to be explicitly provided to ensure that the ordering of the categories is properly honored:

```
output_features:
  - name: rating
     type: category
     preprocessing:
       vocab: ["terrible", "bad", "okay", "good", "great"]
     loss:
       type: corn
```

We chose to replicate the code here with attribution rather than add an additional dependency due to the self-contained nature of the implementation to just a couple torch-only functions.


ludwig/modules/loss_implementations/corn.py
```diff
@@ -0,0 +1,150 @@
...
+    *Rank Consistent Ordinal Regression for Neural Networks
+       with Application to Age Estimation*
+    Pattern Recognition Letters, https://doi.org/10.1016/j.patrec.2020.11.008
+
+    Parameters
+    ----------
+    logits : torch.tensor, shape(num_examples, num_classes-1)
+        Outputs of the CORAL layer.
+
+    levels : torch.tensor, shape(num_examples, num_classes-1)
```
arnavgarg1, line 25: Is it also worth adding a CoralLoss Config?

tgaddair: Maybe, removed for now as not sure where it would make sense to use it.

arnavgarg1 (approved):

PR #3378, opened 2023-05-03
Temporarily switch to `eval` mode when using `batch_size: 1`

abidwael:
fixes the following error
```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss. You can enable unused parameter detection by passing the keyword argument `find_unused_parameters=True` to `torch.nn.parallel.DistributedDataParallel`, and by 
making sure all `forward` function outputs participate in calculating loss. 
If you already have done the above, then the distributed data parallel module wasn't able to locate the output tensors in the return value of your module's `forward` function. Please include the loss function and the structure of the return value of `forward` of your module when reporting this issue (e.g. list, dict, iterable).
Parameter indices which did not receive grad for rank 0: 48 49 51 52 54 55 57 58 60 61 62 63 64 65 67 68 70 71 72 73 74 75 77 78 80 81 82 83 84 85 87 88 90 91 93 94 96 97 99 100
 In addition, you can set the environment variable TORCH_DISTRIBUTED_DEBUG to either INFO or DETAIL to print out information about which particular parameters did not receive gradient on this rank as part of this error
```
The affected parameters are batch norm parameters.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -151,15 +151,15 @@ jobs:
 
       - name: Unit Tests
         run: |
-          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=5400 pytest -v --timeout 300 --durations 100 -m "$MARKERS and not slow and not combinatorial and not horovod" --junitxml pytest.xml tests/ludwig
+          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=5400 pytest -v --timeout 300 --durations 100 -m "$MARKERS and not slow and not combinatorial and not horovod and not llm" --junitxml pytest.xml tests/ludwig
 
       - name: Integration Tests
         run: |
-          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=6000 pytest -v --timeout 300 --durations 100 -m "$MARKERS and not slow and not combinatorial and not horovod" --junitxml pytest.xml tests/integration_tests
+          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=6000 pytest -v --timeout 300 --durations 100 -m "$MARKERS and not slow and not combinatorial and not horovod and not llm" --junitxml pytest.xml tests/integration_tests
 
       - name: Regression Tests
         run: |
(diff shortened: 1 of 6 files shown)


tgaddair (approved):
Clever!

tests/integration_tests/test_experiment.py
```diff
@@ -848,6 +848,31 @@ def on_resume_training(self, is_coordinator):
         )
 
 
+@pytest.mark.distributed
+def test_tabnet_with_batch_size_1(tmpdir, ray_cluster_5cpu):
```
tgaddair, line 852: Need to be consistnet with the ray cluster used in the module, which should be `ray_cluster_4cpu`.

.github/workflows/pytest.yml
```diff
@@ -151,15 +151,15 @@ jobs:
 
       - name: Unit Tests
         run: |
-          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=5400 pytest -v --timeout 300 --durations 100 -m "$MARKERS and not slow and not combinatorial and not horovod and not llm" --junitxml pytest.xml tests/ludwig
+          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=5400 pytest -v --timeout 300 --durations 100 -m "$MARKERS and not slow and not combinatorial and not horovod" --junitxml pytest.xml tests/ludwig
```
tgaddair, line 154: Why did you remove this? We don't want to run LLM tests here. This is what's causing the test failures.

abidwael: I was just trying to run the relevant tests but I must've not included it back when I reverted my change

.github/workflows/pytest.yml
```diff
@@ -258,8 +258,7 @@ jobs:
           python --version
           pip --version
           python -m pip install -U pip
-          pip install -r requirements.txt
-          pip install -r requirements_test.txt
+          pip install '.[test]'
```
tgaddair, line 261: This pulls in all the distributed things as well, if I rmemeber correctly. Is that intentional?

abidwael: Yes, so we can use DDP

tgaddair: But why? That's just going to baloon the test runtime to run combinatorial tests with DDP. We already run too many distributed tests (which is why we see so many timeouts).

abidwael: I would like to use DDP because the combinatorial tests is one of the few tests that run end to end model training with different parameters and architectures and that can help us catch issues similar to these.

tgaddair: True, but the cost is probably going to be prohibitve it be honest. We're talking orders of magnitude slower because of all the overhead Ray incurs. We can try it, but let's change it back if it ends up being too slow.

abidwael: This run took 25min and before introducing this it takes 17min.

PR #3379, opened 2023-05-03
Fix for `best.ckpt`  not found

abidwael:


ludwig/trainers/trainer.py
```diff
@@ -812,7 +812,9 @@ def train(
         # Load the best weights from saved checkpoint
         state_dict = None
         if not self.skip_save_model:
-            state_dict = checkpoint_manager.get_best_checkpoint_state_for_inference(self.return_device)
+            if self.is_coordinator():
```
tgaddair, line 815: Ah, okay, this makes sense as we only save the checkpoint on the coordinator for DDP. Can you do this here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/utils/checkpoint_utils.py#L158?

Reason being that DeepSpeed is different, the model gets saved on every worker independently.

The new logic should be:

```
def get_state_for_inference(self, save_path: str, device: Optional[torch.device] = None) -> Mapping[str, Any]:
        state = {}
        if self.distributed.is_coordinator():
            state = torch.load(save_path, map_location=device)
        state_dict = self.distributed.broadcast_object(state_dict, "state_dict")
        return state["model_weights"]
```

tgaddair:
Closed in favor of #3381.

PR #3381, opened 2023-05-04
Fixed DDP checkpointing to save to local disk on each node

tgaddair:
When loading the checkpoint, we assume it exists on each node, so we should make sure we save on each node.

This assumes we're not using a global filesystem like NFS. We could consider ways of checking this (like try writing files to the directory from each node and see if nodes can read each-other's files), but for now this should suffice for most use cases.

### DIFF
--- a/ludwig/utils/checkpoint_utils.py
+++ b/ludwig/utils/checkpoint_utils.py
@@ -212,6 +212,15 @@ class MultiNodeCheckpoint(Checkpoint):
         return self.distributed.local_rank() == 0
 
     def safe_move_file(self, src: str, dst: str):
+        """Move a file from one directory to another, possibly across filesystems.
+
+        This implementation specifically addresses the following issue with distributed training:
+
+        1. The `save_path` is a directory local to the node, in which case every node should write
+           checkpoints separately.
+        2. The `save_path` is a remote / global filesystem like NFS, in which case only the coordinator
+           should write checkpoints.
+        """
         try:
             os.replace(src, dst)
         except OSError as err:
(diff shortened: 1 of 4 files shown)


ludwig/utils/checkpoint_utils.py
```diff
@@ -169,42 +170,44 @@ def save(self, save_path: str, global_step: int):
...
+                # atomic save
+                with tempfile.TemporaryDirectory() as tmpdir:
+                    # Save to a temporary directory outside of the checkpoint dir so
+                    # async processes do not try and copy a partially-written checkpoint.
+                    # See Ray Tune and MLFlow for examples of background processes that
+                    # are affected by this.
+                    tmp_path = os.path.join(tmpdir, "temp.ckpt")
+                    torch.save(state, tmp_path)
+
+                    safe_move_file(tmp_path, save_path)
```
abidwael, line 201: My understanding is that `torch.save(state, tmp_path)` will save the checkpoint to a local path on each node. My concern is that when we call `safe_move_file(tmp_path, save_path)` from every node, we'll have multiple nodes trying to write to the same path. So do we need to add an `if self.distributed.is_coordinator:` before this call?

tgaddair: It's a good point, basically there are two possibilities:

1. The `save_path` is a directory local to the node (true most of the time) in which case the existing behavior works
2. The `save_path` is a remote and global filesystem like NFS, in which case this will not work

The right solution to handle (2) is to use a custom implementation of `safe_move_file` that only writes once for external filesystems. Let me give that a shot.

abidwael: I think my assumption was wrong, this happens during training so we can just save to the same path if it's local and that won't be a problem. But I think it might be a problem if it's a remote path?

abidwael: Perfect, you beat me to it :)

abidwael (approved):

PR #3382, opened 2023-05-04
Ensure captum's halved batch size retry fn wrapper accepts the minimum batch size of 1.

justinxzhao:
Add a constant for `MINIMUM_BATCH_SIZE=1`, and use that as the basis for captum's batch size retry wrapper.

### DIFF
--- a/ludwig/explain/captum.py
+++ b/ludwig/explain/captum.py
@@ -58,10 +58,12 @@ class ExplanationRunConfig:
 def retry_on_cuda_oom(run_config: ExplanationRunConfig):
     def retry_on_cuda_oom_fn(fn):
         def retry_on_cuda_oom_wrapper(*args, **kwargs):
-            while run_config.batch_size >= 2:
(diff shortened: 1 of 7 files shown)


ludwig/explain/captum.py
```diff
@@ -71,11 +83,14 @@ def retry_on_cuda_oom_wrapper(*args, **kwargs):
                         # Not a CUDA error
                         raise
 
-            raise RuntimeError("CUDA OOM raised during explanation, but batch size cannot be reduced any further")
+            raise RuntimeError(
+                f"Ran into latest error {latest_error} during explanation. "
```
jeffkinnison, line 87: Could we log the full traceback instead of the error message only (I think that's what this string formatting does). That could save us some time when debugging explain errors  by surfacing the relevant code path.

jeffkinnison (approved):
LGTM! Pretty straightforward changes, I just had one nit.

PR #3385, opened 2023-05-05
Added LoRA tuner for HuggingFace text encoders

tgaddair:
Using LoRA speeds up fine-tuning, allowing you to train with larger batch sizes and adjust fewer weights during training.

Usage:

```
encoder:
    type: bert
    trainable: true
    tuner: lora
```

You can also specify additional parameters:

```
encoder:
    type: bert
    trainable: true
    tuner:
        type: lora
        r: 16
        alpha: 32
        dropout: 0.1
        bias_type: all
```

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -135,7 +135,6 @@ class HFTextEncoder(Encoder):
 
             peft_config = tuner.to_config()
             transformer = get_peft_model(transformer, peft_config)
-            print("!!! WRAPPED WITH LORA !!!", peft_config)
         return FreezeModule(transformer, frozen=not trainable)
 
     def get_embedding_layer(self) -> nn.Module:
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -20,7 +20,7 @@ from typing import Any, Callable, Dict, List, Optional, Type, TYPE_CHECKING, Typ
 import numpy as np
 import torch
 from torch import nn
-from peft import LoraConfig, get_peft_model
+from peft import get_peft_model
 
 from ludwig.api_annotations import DeveloperAPI
 from ludwig.constants import TEXT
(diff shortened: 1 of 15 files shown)


ludwig/schema/encoders/text_encoders.py
```diff
@@ -60,6 +61,8 @@ class HFEncoderImplConfig(HFEncoderConfig):
         parameter_metadata=ENCODER_METADATA["HFEncoder"]["trainable"],
     )
 
+    tuner: Optional[BaseTunerConfig] = TunerDataclassField()
+
```
arnavgarg1, line 65: If we're adding the `tuner` attribute to `HFEncoderImplConfig` and all other encoders are subclasses of this, maybe we just need it to add it here instead of adding it to every single encoder schema config?

ludwig/encoders/text_encoders.py
```diff
@@ -128,6 +130,12 @@ def _maybe_resize_token_embeddings(self, transformer, vocab_size: int):
         if vocab_size != transformer.config.vocab_size:
             transformer.resize_token_embeddings(vocab_size)
 
+    def _wrap_transformer(self, transformer: nn.Module, tuner: Optional[BaseTunerConfig], trainable: bool):
```
arnavgarg1, line 133: Oh wow, I did not realize we could do this, this is awesome!

arnavgarg1 (approved):
Minor nits, but this looks awesome! Really nice change!

ludwig/encoders/text_encoders.py
```diff
@@ -19,6 +19,7 @@
 
 import numpy as np
 import torch
+from peft import get_peft_model
```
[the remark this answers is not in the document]
tgaddair: Good catch!

ludwig/schema/encoders/text_encoders.py
tgaddair: Sadly only `DebertaV2Config` inherits from `HFEncoderImplConfig` currently. For whatever reason these classes weren't made with an inheritance hierarchy. I added this stuff when I added Deberta so we could use this pattern going forward and refactor the existing stuff to use it, but unfortunately we haven't done that refactoring yet, so there's still a ton of duplication.

PR #3386, opened 2023-05-06
[LLM] Add Prefix Tuning, PTuning, LoRA, AdaLoRA and Adaption Prompt for LLM fine-tuning

tgaddair:
This PR adds support for parameter efficient fine-tuning strategies like Prefix Tuning, PTuning, LoRA, AdaLoRA and Adaption Prompt.

To use them in Ludwig with the LLM model types, you need to specify the strategy name in the adapter config:

- Prompt Tuning: `prompt_tuning`
- Prefix Tuning: `prefix_tuning`
- P-Tuning: `p_tuning`
- LoRA: `lora`
- AdaLoRA: `adalora`
- Adaption Prompt: `adaption_prompt`

Here's an example config:

```
model_type: llm
model_name: facebook/opt-66b
tuner: lora
```

Things to note:

1. Prompt Tuning, Prefix Tuning and PTuning require the num_virtual_tokens parameter to be set to a non-zero value
2. Lora is only supported for the following model types: t5, mt5, bart, opt, roberta and deberta-v2
3. Adaption Prompt is only supported for Llama model types and requires adapter_len and adapter_layers parameters to be set.

Co-authored by @arnavgarg1


tgaddair:
LGTM!

PR #3397, opened 2023-05-13
fix: sequence feature explain error

jeffkinnison:
Sequence feature explanations always fail with the error

```
RuntimeError: One of the differentiated Tensors appears to not have been used in the graph. Set allow_unused=True if this is the desired behavior.
```

This happened because sequence features were not treated as embedded features, which led to the wrong encoder layer being passed into captum's `LayerIntegratedGradients`. This fix updates explanations to treat sequences as embedded features and adds explanation integration tests for all sequence encoders.


justinxzhao (approved):
Sweet!

PR #3399, opened 2023-05-15
[llm] Move prompt into top-level of config

tgaddair:
This is the first step towards supporting multiple input features that resolve to a single text feature (under the hood) via the prompt as a coalescing function.


geoffreyangus:
Nice, so just to be clear– the scope of this PR is just to make the prompt a global parameter?

It seems like the prompt is still formatted the same way (i.e. unable to inject data from the different columns into the prompt of the primary text input feature).

tgaddair:
@geoffreyangus correct. This also retains the previous location in preprocessing for ECD. In the future, though, I wan to revisit the relationship between the prompt and input features as it pertains to LLMs, but we can chat more about that in a separate thread, as it's a bit of a bigger refactor.

justinxzhao (approved):

geoffreyangus (approved):
Sounds good. Ship it

PR #3400, opened 2023-05-15
[LLM] Ensure prompt is only supported for LLM text input features

arnavgarg1:
This ensures that

1. `prompt` is only an attribute for text input features for LLM models
2. `prompt` does not show up for text output features (LLM or otherwise)


ksbrar (approved):

tgaddair:
This is subsumed by #3399.

PR #3404, opened 2023-05-16
[llm] Parameter metadata

tgaddair:


martindavis (approved):

ludwig/schema/metadata/configs/llm.yaml
```diff
@@ -0,0 +1,98 @@
...
+      ui_display_name: Top K
+      expected_impact: 2
+  task:
+    ui_display_name: Task
+    expected_impact: 3
+    component_type: textarea
+  template:
+    ui_display_name: Template
+    expected_impact: 3
+    component_type: textarea
```
ksbrar, line 69: So based on a separate conversation with @martindavis and I had earlier today, I would have expected the new `component_type` property to appear at the _schema_ level, not in the parameter metadata.

By schema level I mean in the `json_schema_type_mapping` for `PromptConfigField`. So it would become a "sibling" property of "parameter_metadata" rather than one of the latter's child props.

We can plumb it through either way, but by putting it in parameter metadata the correct thing to do would be to define a default for this prop in the `ParameterMetadata` class as well, which I suppose could be `null`. But doing so would also imply we're establishing a general standard of tying parameters to their renderable HTML elements.

Up to you guys, I'm fine going either way. I kind of lean in this direction since we are already putting other "UI" focused metadata in their like the "ui_display_name". But maybe we don't to more strongly tie (quite light) business logic in like this.

tgaddair: Thanks! I went ahead and added `ui_component_type` to `ParameterMetadata`, which I feel is consistent with how we handle the display name, etc.

PR #3406, opened 2023-05-17
LLMFineTurningQuestions--NotToMerge

skanjila:
# Code Pull Requests

Please provide the following:

- Some questions from me to start the fine tuning llm workflow discussions, look for ###SXK statements to start our discussions

# Documentation Pull Requests

### DIFF
--- a/ludwig/decoders/llm_decoders.py
+++ b/ludwig/decoders/llm_decoders.py
@@ -125,6 +125,9 @@ class TextParserDecoder(Decoder):
         }
 
 
+
+###SXK where is this being used, for the decoder how does one build a custom one, one idea here might
+###be for the user to get a list of already available decoders before they start to build one
 @DeveloperAPI
 @register_decoder("category_parser", [CATEGORY])
 class CategoryParserDecoder(Decoder):
--- a/ludwig/decoders/llm_decoders.py
+++ b/ludwig/decoders/llm_decoders.py
@@ -125,7 +125,6 @@ class TextParserDecoder(Decoder):
         }
 
 
-
 ###SXK where is this being used, for the decoder how does one build a custom one, one idea here might
 ###be for the user to get a list of already available decoders before they start to build one
 @DeveloperAPI
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -46,12 +46,16 @@ class LLM(BaseModel):
 
         logger.info("Loading large language model...")
         self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.model_name)
+        ####SXK should we make this configurable ######
         self.curr_device = torch.device("cpu")  # model initially loaded onto cpu
         logger.info("Done.")
 
+        ####SXK why do we need this adapter and PEFT model, sorry about my ignorance here, how many types
+        ####of adapters can there be
         self.initialize_adapter()
 
         # Determines the maximum length of the context (input + output tokens)
+        ####SXK Is there any validation needed to be done here
         if hasattr(self.model.config, "max_sequence_length"):
             self.context_len = self.model.config.max_sequence_length
         elif hasattr(self.model.config, "max_position_embeddings"):
@@ -67,6 +71,7 @@ class LLM(BaseModel):
         # Used only for its metadata about the vocabulary
         self.tokenizer = AutoTokenizer.from_pretrained(self.config_obj.model_name, use_fast=False)
 
+        ####SXK what is the significance of the generation config, again just trying to understand high level context
         self.generation = GenerationConfig(**self.config_obj.generation.to_dict())
 
         # ================ Inputs ================
@@ -95,6 +100,7 @@ class LLM(BaseModel):
         _, self.output_feature_decoder = self.output_features.items()[0]
 
         # ================ Combined loss metric ================
+        ####SXK should we have a class to capture all the loss metrics and output to a dashboard during a run
         self.eval_loss_metric = torchmetrics.MeanMetric()
         self.eval_additional_losses_metrics = torchmetrics.MeanMetric()
 
@@ -102,6 +108,8 @@ class LLM(BaseModel):
 
     def initialize_adapter(self):
         """If an adapter config is provided, we want to wrap the model with a PEFT model for fine-tuning."""
+        ###SXK do we think we can extend the peft module but taylor it more towards ludwig
+        ###what would that look like or will this work out of the box
         if self.config_obj.adapter:
             from peft import get_peft_model
 
@@ -178,6 +186,9 @@ class LLM(BaseModel):
         """Returns the input ids for the text feature input."""
         return inputs[self.config_obj.input_features[0].name].type(torch.int32)
 
+    ###SXK just a thought here, based on the current use cases does it make sense to have a BaseLLM
+    ###class that we can build custom LLM classes, maybe the base LLM class contains a generic implementation
+    ###of the forward method, lets discuss this some more if it makes sense
     def forward(
         self,
         inputs: Union[
@@ -237,6 +248,7 @@ class LLM(BaseModel):
 
             return outputs
 
+        ####SXK why dont we need the calculation of gradients here
         with torch.no_grad():
             input_lengths = []
             sequences_list = []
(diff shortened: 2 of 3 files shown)


ludwig/decoders/llm_decoders.py
```diff
@@ -125,6 +125,8 @@ def forward(self, inputs: List[torch.Tensor], **kwargs):
         }
 
 
+###SXK where is this being used, for the decoder how does one build a custom one, one idea here might
+###be for the user to get a list of already available decoders before they start to build one
```
arnavgarg1, line 129: This is being automatically set depending on the output feature type for LLM in the schema definition for that particular output feature type. For e.g. https://github.com/ludwig-ai/ludwig/blob/master/ludwig/schema/features/text_feature.py#L175 for LLM text output features.

Yeah, we'll have this in the docs.

ludwig/models/llm.py
```diff
@@ -46,12 +46,16 @@ def __init__(
 
         logger.info("Loading large language model...")
         self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.model_name)
+        ####SXK should we make this configurable ######
```
arnavgarg1, line 49: This is actually a requirement when using the Ray backend. In theory one could directly load it onto GPUs, but the issue here is that Ray isn't aware of how many GPUs are available to it until later, so we load it onto CPU first and then move it onto GPUs in a later function (to_device). 

In the future, we plan to support meta tensor based initialization since this will not actually load the weights onto the CPU, but just the shapes. Then, if we're doing model parallel training, it'll move the tensors onto the right GPU/CPU devices when it needs to, along wit the optimizer and activations.

ludwig/models/llm.py
```diff
@@ -46,12 +46,16 @@ def __init__(
...
         self.curr_device = torch.device("cpu")  # model initially loaded onto cpu
         logger.info("Done.")
 
+        ####SXK why do we need this adapter and PEFT model, sorry about my ignorance here, how many types
+        ####of adapters can there be
```
arnavgarg1, line 54: PEFT is parameter efficient fine-tuning. It basically allows you to plug and play different "adapters" that are used for fine-tuning as opposed to adjusting the weights of the base LLM model. The forward pass goes through both, but the backward pass just goes through the PEFT adapter weights.

There's many - we already support 5, and more will be added in the future.

ludwig/models/llm.py
```diff
@@ -46,12 +46,16 @@ def __init__(
...
         self.initialize_adapter()
 
         # Determines the maximum length of the context (input + output tokens)
+        ####SXK Is there any validation needed to be done here
```
arnavgarg1, line 58: I think this validation should suffice - what do you have in mind?

ludwig/models/llm.py
```diff
@@ -67,6 +71,7 @@ def __init__(
         # Used only for its metadata about the vocabulary
         self.tokenizer = AutoTokenizer.from_pretrained(self.config_obj.model_name, use_fast=False)
 
+        ####SXK what is the significance of the generation config, again just trying to understand high level context
```
arnavgarg1, line 74: It tells the model how to generate tokens - it controls things such as the temperature, the number of new tokens, the minimum number of tokens, the top p / top k tokens to look it, whether to do greedy decoding or multinomial sampling, etc.

ludwig/models/llm.py
```diff
@@ -95,13 +100,16 @@ def __init__(
         _, self.output_feature_decoder = self.output_features.items()[0]
 
         # ================ Combined loss metric ================
+        ####SXK should we have a class to capture all the loss metrics and output to a dashboard during a run
```
arnavgarg1, line 103: We already do that using a set of predefined metric function classes. These are just maintained separately

ludwig/models/llm.py
```diff
@@ -95,13 +100,16 @@ def __init__(
...
         self.eval_loss_metric = torchmetrics.MeanMetric()
         self.eval_additional_losses_metrics = torchmetrics.MeanMetric()
 
         clear_data_cache()
 
     def initialize_adapter(self):
         """If an adapter config is provided, we want to wrap the model with a PEFT model for fine-tuning."""
+        ###SXK do we think we can extend the peft module but taylor it more towards ludwig
+        ###what would that look like or will this work out of the box
```
arnavgarg1, line 112: This works out of the box already since PEFT is a wrapper on top of transformers, and Ludwig is already compatible with transformers. We could adapt this in the future to a more Ludwig friendly design if we decide we should provide implementations for custom adapters that users might want to design, but I think that's a lot of work and 99.9% users won't want to do that at the moment

ludwig/models/llm.py
```diff
@@ -178,6 +186,9 @@ def get_input_ids(
         """Returns the input ids for the text feature input."""
         return inputs[self.config_obj.input_features[0].name].type(torch.int32)
 
+    ###SXK just a thought here, based on the current use cases does it make sense to have a BaseLLM
+    ###class that we can build custom LLM classes, maybe the base LLM class contains a generic implementation
+    ###of the forward method, lets discuss this some more if it makes sense
```
arnavgarg1, line 191: Hmm, we could, but curious where/how you think that might be useful. I don't see why not though

ludwig/models/llm.py
```diff
@@ -237,6 +248,7 @@ def forward(
 
             return outputs
 
+        ####SXK why dont we need the calculation of gradients here
```
arnavgarg1, line 251: We're computing a forward pass using the model, and we don't need to store the activations since in the generation case, we're not trying to update weights or compute gradients. The torch.no_grad() just freezes the pretrained models weights. Otherwise it's going to generate logits etc which we don't actually need. It also effectively treats the model like it's in "eval" mode

ludwig/models/llm.py
```diff
@@ -317,6 +329,7 @@ def train_loss(
         of_train_losses = {}
         for of_name, of_obj in self.output_features.items():
             _targets, _predictions = targets, predictions
+            ###SXK how can we extend this to other OutputFeature types
             if isinstance(of_obj, TextOutputFeature):
```
arnavgarg1, line 333: It already supports Text and category output features, with more support coming shortly. This if block is specially for text features since they need to be handled separately since they have variable length output. Category/Numbers etc will have fixed length output so this provision doesn't need to be made.

ludwig/models/llm.py
```diff
@@ -407,6 +420,8 @@ def load(self, save_path):
             weights_save_path = os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME)
             config = PeftConfig.from_pretrained(weights_save_path)
             config.inference_mode = False
+            ###SXK I noticed we are tightly coupling to hugging face in places, I wonder
+            ###if we want to make this part configurable as well
```
arnavgarg1, line 424: It is true that we are tighly coupling, but I'm not sure it's worth doing that today since. We could make this a registry pattern where we load LLMs different based on their "types", just like Ludwig does today for a lot of its tokenizers. Additionally, most open source large language models are either on huggingface or on github (and then eventually make their way to huggingface anyway), so this being tightly coupled seems fine for now.

In the future, I do think your suggestion of having a base LLM class and this being a subclass of that called HFLLM or something could make sense though!

ludwig/trainers/trainer_llm.py
```diff
@@ -28,6 +28,7 @@
 logger = logging.getLogger(__name__)
 
 
+###SXK do we want to have a spark trainer as well
```
arnavgarg1, line 31: This would be a pretty big Ludwig contribution, since today we only have support for local and Ray backends for all parts of Ludwig. Would be a great contribution if you're interested in adding it!

ludwig/trainers/trainer_llm.py
```diff
@@ -373,6 +374,8 @@ def run_evaluation(
         return False
 
 
+###SXK I'd like to help out in adding more use cases around this fine tuning workflow, lets talk more about this
+###as well,one thing I was wondering about is how tightly coupled are we to the trainer and the corresponding backend
```
arnavgarg1, line 378: As things go in Ludwig, we are flexible with the following:

Backend: local, ray, horovod, deepspeed
Trainer: local and ray (ray, horovod and deepspeed all run through Ray)

PR #3407, opened 2023-05-17
[LLM] Fine-Tuning fixes: Correct inference mode, enable weight adjustment for adapters when fine-tuning for > 1 epoch, etc.

arnavgarg1:
This PR makes a few updates:

1. Fixes/Pins missing dependency for `tokenizers`, and sets it to a minimum version of `0.13.3` for LlamaTokenizer support
2. During adapter initialization, we explicitly set the PEFT inference mode to False to make the PEFT adapter weights trainable.
3. On LLM model load, we were setting the config's inference mode to False to make the weights trainable after each epoch, but the config wasn't being used. This meant that the weights were trainable for the first epoch, but not for any subsequent epoch. The new fix always sets this to False for the loaded model so that the weights are trainable beyond the first epoch.
4. During prediction, we dynamically set inference_mode for LLM model types when using adapters. It gets set to True since we're doing inference. Peft makes optimizations, which are useful since they speed up batch prediction.
5. Fixes an issue during `model.predict()` where the trained LLM goes through the forward pass of the "fine-tuning" route as opposed to the forward pass for "generation", which is what we should be doing. When it goes through the code path for fine-tuning, it ends up producing non-sensical outputs since the call to the model returns logits as opposed to the output token IDs. To make this fix, we hotswap the adapter in the config object to None, and then set it back in the config object post `model.predict()` incase a user decides to now go back to `model.train()`. This fix is a bit hacky, but it is the fastest solution I could think of to fix this.

### DIFF
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -108,6 +108,8 @@ class LLM(BaseModel):
             peft_config = self.config_obj.adapter.to_config(
                 task_type="CAUSAL_LM", tokenizer_name_or_path=self.model_name
             )
+            # Incase inference_mode is True, set it to False
+            peft_config.inference_mode = False
             self.model = get_peft_model(self.model, peft_config)
 
             logger.info("==================================================")
@@ -412,6 +414,7 @@ class LLM(BaseModel):
 
             # Do this since we want to train the PEFT adapter and model. During inference,
             # we will explicitly override this to True.
+            # Note: default is the name of the adapter (always the case)
             self.model.peft_config["default"].inference_mode = False
 
     def get_args(self):
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -406,9 +406,13 @@ class LLM(BaseModel):
 
             weights_save_path = os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME)
             config = PeftConfig.from_pretrained(weights_save_path)
-            config.inference_mode = False
+
             self.model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path)
             self.model = PeftModel.from_pretrained(self.model, weights_save_path)
+
+            # Do this since we want to train the PEFT adapter and model. During inference,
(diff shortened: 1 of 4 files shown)


tgaddair (changes requested):
Too hacky for my tastes. Let's chat about a better solution.

arnavgarg1:
> Too hacky for my tastes. Let's chat about a better solution.

Makes sense. What if we split this PR into two:

Part 1 (which I assume is fine since it's not hacky): Fix requirements.txt, and keep the changes in llm.py (always set inference_mode to False).

Part 2: Change inference mode and remove adapter at prediction time (we can come up with a better solution for this)?

requirements_llm.txt
```diff
@@ -1,6 +1,7 @@
 sentence-transformers
 faiss-cpu
 
+tokenizers>=0.13.3
```
tgaddair, line 4: We can put this in main requirements. Also, shouldn't this be handled by transformers itself?

arnavgarg1: I thought so too, but that doesn't seem to be the case

ludwig/models/llm.py
```diff
@@ -406,10 +408,15 @@ def load(self, save_path):
...
             config = PeftConfig.from_pretrained(weights_save_path)
-            config.inference_mode = False
+
             self.model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path)
             self.model = PeftModel.from_pretrained(self.model, weights_save_path)
 
+            # Do this since we want to train the PEFT adapter and model. During inference,
+            # we will explicitly override this to True.
+            # Note: default is the name of the adapter (always the case)
+            self.model.peft_config["default"].inference_mode = False
```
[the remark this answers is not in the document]
arnavgarg1: Agreed, I can remove this

arnavgarg1:
Closing in favor of https://github.com/ludwig-ai/ludwig/commit/f5081d87dbcc3ff2a2e16c24a23b62c11d9eb78f

PR #3408, opened 2023-05-18
[LLM] Fix Loss Computation for LLM Fine-tuning using shifted tensors/new loss function

arnavgarg1:
This PR makes the following changes:

1. Updates the forward pass of the LLM during fine-tuning to use a tensor that merges input and target tokens as opposed to just passing in the input tensor with token IDs. 
2. It also creates attention masks to focus on non pad/bos token IDs during the forward pass. 
3. Adds a new loss function inspired by the loss computed in the BertLMHead module here: https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/bert/modeling_bert.py#LL1253C1-L1260C1, which computes the loss between the target tokens and the predicted token logits using an offset of 1 for the targets.
4. Fixes an issue during preprocessing where we use the wrong pad token ID for text features during preprocessing when using a hf tokenizer
5. Sets pad token ID for the LLM tokenizer if its missing
6. Optimizes tokenizer loading to use the fast tokenizer when possible, skipping llama tokenizer initialization since the rust implementation takes 3-4 minutes to load.

This is the recommended approach for PEFT based fine-tuning.


ludwig/decoders/llm_decoders.py
```diff
@@ -87,7 +87,12 @@ def __init__(
         # Load tokenizer required for decoding the output from the generate
         # function of the text input feature for LLMs.
         self.tokenizer = get_tokenizer(self.tokenizer_type, self.vocab_file, self.pretrained_model_name_or_path)
-        self.tokenizer_vocab_size = self.tokenizer.tokenizer.vocab_size
+        try:
```
tgaddair, line 90: Let's use `hasattr` instead of try-except.

ludwig/models/llm.py
```diff
@@ -142,6 +143,11 @@ def to_device(self, device):
                 self.model.save_pretrained(tmpdir)
                 self.model = AutoModelForCausalLM.from_pretrained(tmpdir, **model_kwargs)
 
+            # If loading from checkpoint did not place it on the desired device, forcefully place it on
+            # the desired device. On Cuda, this will place it on the device with the lowest index.
+            if self.model.device != device:
```
tgaddair, line 148: What would cause this to happen? Line above should handle this. right? My concern would be what would happen if the model was too big for one gpu and we tried to do this.

ludwig/models/llm.py
```diff
@@ -213,30 +210,48 @@ def forward(
...
+            target_ids = self.get_target_ids(targets)
 
-            if self.output_feature_type != TEXT:
-                # Pass generated tokens through decoder after averaging the token probabilities
-                # This is required for the classification head for the classifier decoder
-                model_outputs = torch.mean(model_outputs, dim=1)
+        # TODO: Figure out how to get rid of using the additional self.config_obj.adapter check
+        # The issue is that when we run evaluation on the validation/test sets, we set the mode to eval
+        # and use the forward_generate path instead of the forward_train path.
+        if self.model.training or self.config_obj.adapter:
```
tgaddair, line 220: We'll address this in #3409.

tgaddair (approved):
LGTM!

ludwig/models/llm.py
arnavgarg1: I'm not 100% sure, but it happens specifically when using the local backend on a pod with GPUs

arnavgarg1: Actually I agree, this can probably go. It's only needed for models where device_map isn't implemented, in which case we'd want to get rid of it from `model_kwargs` to begin wtih

ludwig/models/llm.py
arnavgarg1: Agreed

ludwig/modules/loss_modules.py
```diff
@@ -200,6 +201,22 @@ def forward(self, preds: Tensor, target: Tensor) -> Tensor:
...
+
+    def forward(self, preds: Tensor, target: Tensor) -> Tensor:
+        """
+        Params:
+            preds: Tensor of shape [batch x sequence_length x vocab_size]
+            target: Tensor of shape [batch x sequence_length], where each element is integral between 0 and vocab_size.
+        """
+        target = target.long()
+        _, _, vocab_size = preds.shape
+        return self.loss_fn(preds.reshape(-1, vocab_size), target.reshape(-1))
```
geoffreyangus, line 217: Don't we need to shift preds and targets?

geoffreyangus (changes requested):

ludwig/modules/loss_modules.py
arnavgarg1: @geoffreyangus  Hmm, that's true. What would this change to?

arnavgarg1: We want to shift preds by 1, and keep targets as is right?

ludwig/models/llm.py
```diff
@@ -142,6 +158,13 @@ def to_device(self, device):
                 self.model.save_pretrained(tmpdir)
                 self.model = AutoModelForCausalLM.from_pretrained(tmpdir, **model_kwargs)
 
+            # TODO(Arnav): Looking into this later, but it specifically happens when using local backend on a setup
```
geoffreyangus, line 161: This happened because BERT didn't implement device_map, and therefore didn't load itself onto GPU when we asked it to.

Probably not an issue for real LLMs, but important to keep in mind when debugging in the future

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
 
+    def _set_pad_token(self):
+        """Sets the pad token for the tokenizer if it is not already set."""
+        # HACK(Arnav): gpt, gpt2 and llama tokenizers had no pad tokens.
+        # These recommend using eos tokens instead
+        # https://github.com/huggingface/transformers/issues/2648#issuecomment-616177044
+        # https://github.com/huggingface/transformers/issues/2630#issuecomment-1290809338
+        if any(
+            isinstance(self.tokenizer, t)
+            for t in [GPT2Tokenizer, GPT2TokenizerFast, LlamaTokenizer, LlamaTokenizerFast]
```
geoffreyangus, line 500: Oh I see– this is because we use the raw tokenizer at this point instead of the `HFTokenizer` wrapper.

Can we move this logic into a tokenizer util function that both the HFTokenizer and this function call to prevent duplication?

ludwig/utils/strings_utils.py
```diff
@@ -504,8 +505,21 @@ def build_sequence_matrix(
         logger.debug(f"max length of {format}: {max_length} < limit: {length_limit}")
     max_length = length_limit
 
+    # Set padding token id based on tokenizer_type. Huggingface tokenizers typically have a pad_token_id attribute.
+    if tokenizer_type == "hf_tokenizer":
+        if hasattr(tokenizer.tokenizer, "pad_token_id"):
```
geoffreyangus, line 510: I think this is actually only an issue for `hf_tokenizer` objects that end up being hot-swapped for TorchText-based tokenizers, like the BERTTokenizer.

For `HFTokenizer` objects, we do set this value already: https://github.com/ludwig-ai/ludwig/blob/f829345c7df645835d50fd78d7c483ed2d12cbc7/ludwig/utils/tokenizers.py#L823

geoffreyangus: Probably okay for now, but I would leave a TODO here for ensuring that this works with the torchtext tokenizers.

ludwig/models/llm.py
```diff
@@ -452,19 +621,39 @@ def realign_target_and_prediction_tensors(
...
     if target_length > prediction_length:
         # Pad the predictions.
         zeros_to_add = target_length - prediction_length
-        predictions[of_name][PREDICTIONS] = F.pad(predictions[of_name][PREDICTIONS], (0, zeros_to_add))
+
+        if pad_direction == "right":
+            predictions[of_name][PREDICTIONS] = F.pad(
+                predictions[of_name][PREDICTIONS], (0, zeros_to_add), value=pad_value
+            )
+            predictions[of_name][PROBABILITIES] = F.pad(predictions[of_name][PROBABILITIES], (0, 0, 0, zeros_to_add))
```
geoffreyangus, line 639: Can you add a test for this function? I can't easily tell if this actually works

ludwig/models/llm.py
arnavgarg1: Totally agree

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
+            for t in [GPT2Tokenizer, GPT2TokenizerFast, LlamaTokenizer, LlamaTokenizerFast]
+        ):
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+            self.tokenizer.pad_token_id = self.tokenizer.eos_token_id
+
+    def _generate_merged_ids(self, input_ids, target_ids):
+        """This function merges the input_ids and target_ids together to create a unified tensor to pass into the
+        model.
+
+        This is required for PEFT based fine-tuning. It also returns attention masks for the merged tensors.
```
geoffreyangus, line 509: This is required for all kinds of fine-tuning right? Not just PEFT?

Also can you put the shapes of these tensors in the docstring, and whether or not they are coming in with padding, etc. as part of a `Args:` subsection, as well as the return value in a `Returns:` subsection

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
+        if not pad_value:
+            padding = torch.zeros((max_length - input_ids.shape[0]), dtype=torch.int32, device=input_ids.device)
+        else:
+            padding = torch.tensor(
+                [pad_value] * (max_length - input_ids.shape[0]), dtype=torch.int32, device=input_ids.device
+            )
+        return torch.cat((padding, input_ids), dim=-1)
+
+    def _create_attention_mask(self, input_ids):
+        """Creates attention mask for the input_ids tensor."""
```
geoffreyangus, line 580: Can you add the shapes here and in `_add_left_padding` and any other tensor op methods?

Also it would probably be good to break all of these out into an llm_utils file so that the functions can be unit tested in isolation

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
 def realign_target_and_prediction_tensors(
-    targets: Dict[str, torch.Tensor], predictions: Dict[str, torch.Tensor], of_name: str
+    targets: Dict[str, torch.Tensor],
+    predictions: Dict[str, torch.Tensor],
+    of_name: str,
+    tokenizer: AutoTokenizer,
+    pad_direction: str = "right",
+    pad_value: int = None,
 ) -> Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]:
     """Realigns the target tensor with the predictions.
```
geoffreyangus, line 604: Can you describe what this means?

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
         """Returns the augmentation pipeline for this model."""
         return AugmentationPipelines({})
 
 
 def realign_target_and_prediction_tensors(
-    targets: Dict[str, torch.Tensor], predictions: Dict[str, torch.Tensor], of_name: str
+    targets: Dict[str, torch.Tensor],
+    predictions: Dict[str, torch.Tensor],
+    of_name: str,
+    tokenizer: AutoTokenizer,
```
geoffreyangus, line 600: nit: tokenizer is probably actually of type BaseTokenizer or something along those lines

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
+            merged_input_and_targets[i] = self._add_left_padding(merged_sample_ids, max_length)
+            attention_masks.append(self._create_attention_mask(merged_input_and_targets[i]))
+
+        return torch.stack(merged_input_and_targets), torch.stack(attention_masks)
+
     def _remove_left_padding(self, input_ids_sample: torch.Tensor):
-        bos_idxs = torch.where(input_ids_sample == self.tokenizer.bos_token_id)[0]  # all BOS token locations
+        """Removes left padding from the input_ids tensor."""
+        # Remove all PAD tokens
+        pad_idxs = torch.where(input_ids_sample == self.tokenizer.pad_token_id)[0]  # all PAD token locations
```
geoffreyangus, line 552: What if there is no left padding because all of the sequences are the same length (or worse, they are right padded somehow)

Let's make sure we add some checks here (would also really like some unit tests to increase confidence in this since its quite complex)

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
             bos_idx = 0
 
-        input_ids_sample_no_padding = input_ids_sample[bos_idx:].unsqueeze(0)
-        return input_ids_sample_no_padding
+        input_ids_sample_no_bos = input_ids_sample_no_padding[bos_idx:].unsqueeze(0)
+        return input_ids_sample_no_bos
+
+    def _add_left_padding(self, input_ids, max_length, pad_value=0):
+        """Adds left padding to the input_ids tensor."""
+        if not pad_value:
```
geoffreyangus, line 571: couldn't we always do the `else` block?

```
padding = torch.tensor(
  [pad_value] * (max_length - input_ids.shape[0]), dtype=torch.int32, device=input_ids.device
)
```

since pad_value being 0 would be equivalent?

ludwig/models/llm.py
```diff
@@ -418,23 +489,117 @@ def get_args(self):
...
+
+    def _create_attention_mask(self, input_ids):
+        """Creates attention mask for the input_ids tensor."""
+        attention_mask = torch.ones_like(input_ids)
+        bos_index = (input_ids == self.tokenizer.bos_token_id).nonzero()
+
+        if len(bos_index) > 0:
+            first_bos_index = bos_index[0][0]
+            # Set attention mask to 0 for the part of the input that is padding
+            attention_mask[:first_bos_index] = 0
```
geoffreyangus, line 587: instead of relying on bos_token_id (which tokenizer may not have) lets mask all PAD tokens out.

geoffreyangus (approved):

PR #3409, opened 2023-05-18
WIP: Create separate predictors for LLMs

tgaddair:


ludwig/models/llm.py
```diff
@@ -268,10 +308,37 @@ def forward_train(self, input_ids, target_ids) -> Dict[str, torch.Tensor]:
...
+            # Convert targets to tensors.
+            for target_feature_name, target_value in targets.items():
+                if not isinstance(target_value, torch.Tensor):
+                    targets[target_feature_name] = torch.from_numpy(target_value)
+                else:
+                    targets[target_feature_name] = target_value
+        else:
+            targets = None
+
+        assert list(inputs.keys()) == self.input_features.keys()
```
arnavgarg1, line 338: nit: this can probably move into it's own utility function that gets shared between `generate` and `forward`, but we can do that in another PR

ludwig/models/llm.py
```diff
@@ -167,9 +217,9 @@ def to_device(self, device):
             # if self.model.device != device:
             #     self.model = self.model.to(device)
 
-            self.eval_loss_metric = self.eval_loss_metric.to(device)
-            self.eval_additional_losses_metrics = self.eval_additional_losses_metrics.to(device)
-            self.output_features.update({k: v.to(device) for k, v in self.output_features.items()})
+            # self.eval_loss_metric = self.eval_loss_metric.to(device)
+            # self.eval_additional_losses_metrics = self.eval_additional_losses_metrics.to(device)
+            # self.output_features.update({k: v.to(device) for k, v in self.output_features.items()})
```
arnavgarg1, line 222: Delete instead of uncomment? Also, is the reason we no longer need to do

```self.output_features.update({k: v.to(device) for k, v in self.output_features.items()})```

this is because we're not initializing these in the LLM model anymore?

File: ludwig/models/llm.py [hunk 3]
```diff
@@ -36,6 +37,44 @@
 logger = logging.getLogger(__name__)
 
 
+class DictWrapper:
+    """Wrapper for a LudwigFeatureDict module that allows for iteration over keys.
+
+    The purpose of this class is to avoid exposing input and output features as modules of the LLM. This is because we
+    only wish to train the underlying model, and having these additional modules can confuse systems like DeepSpeed.
+    """
```
arnavgarg1, line 45: Very clever way of getting around initializing all of the other modules

ludwig/models/llm.py
```diff
@@ -106,14 +145,25 @@ def __init__(
...
+    def create_feature_dict(self) -> LudwigFeatureDict:
+        return DictWrapper(LudwigFeatureDict())
+
+    @property
+    def eval_loss_metric(self) -> torchmetrics.MeanMetric:
+        return self._eval_loss_metric.module
+
+    @property
+    def eval_additional_losses_metrics(self) -> torchmetrics.MeanMetric:
+        return self._eval_additional_losses_metrics.module
```
arnavgarg1, line 161: Do we need these if they're already defined in the base model class?

ludwig/distributed/base.py
```diff
@@ -171,9 +171,6 @@ def prepare_before_load(self) -> bool:
     def is_model_parallel(cls) -> bool:
         return False
 
-    def eval(self, model: nn.Module):
-        model.eval()
```
arnavgarg1, line 175: Is the reason we're getting rid of this over here because we're just going to be running this through the Predictor classes now, which will change model to eval mode at that point?

In that case, would it not make sense to still do something like self.model.eval()? Although I guess yeah, you don't need a function for this

tgaddair: We removed this because DeepSpeed fixed the issue that required it. Now the eval flag is set directly instead of going through the distributed backend.

ludwig/models/llm.py
tgaddair: Good catch! Removed.

ludwig/models/llm.py
tgaddair: Yeah, they exist separately.

arnavgarg1 (approved):
LGTM 🚀

arnavgarg1:
Failed test looks like a transient issue, it should be safe to merge

PR #3414, opened 2023-05-21
fix: Count full dataset in DatasetShard.__len__

jeffkinnison:
Running the new test with the previous code, `DatasetShard.__len__` returned `256` rather than the full dataset size defined in the test. Summing the size of the batches returned by `DatasetIterator.iter_batches` returns the full dataset size.

### DIFF
--- a/ludwig/backend/datasource.py
+++ b/ludwig/backend/datasource.py
@@ -163,10 +163,14 @@ class _BinaryIgnoreNoneTypeDatasourceReader:
                 file_size = None  # unknown file size is None
             else:
                 resolved_path, filesystem = _resolve_paths_and_filesystem([raw_path], filesystem)
-                read_path, file_size = meta_provider.expand_paths(resolved_path, filesystem)
-                # expand_paths returns two lists, so get the first element of each
-                read_path = read_path[0]
-                file_size = file_size[0]
+                try:
+                    # Ray >= 2.4: expand_paths returns `Iterator[Tuple[str, int]]`, so unpack the first tuple
+                    read_path, file_size = next(meta_provider.expand_paths(resolved_path, filesystem))
+                except (TypeError, ValueError):
+                    # Ray < 2.4: expand_paths returns two lists, so get the first element of each
+                    read_path, file_size = meta_provider.expand_paths(resolved_path, filesystem)
+                    read_path = read_path[0]
+                    file_size = file_size[0]
 
             self._paths.append((read_path, idx))
             self._file_sizes.append(file_size)
--- a/ludwig/backend/datasource.py
+++ b/ludwig/backend/datasource.py
@@ -163,14 +163,10 @@ class _BinaryIgnoreNoneTypeDatasourceReader:
                 file_size = None  # unknown file size is None
             else:
                 resolved_path, filesystem = _resolve_paths_and_filesystem([raw_path], filesystem)
-                try:
-                    # Ray >= 2.4: expand_paths returns `Iterator[Tuple[str, int]]`, so unpack the first tuple
-                    read_path, file_size = next(meta_provider.expand_paths(resolved_path, filesystem))
-                except (TypeError, ValueError):
-                    # Ray < 2.4: expand_paths returns two lists, so get the first element of each
-                    read_path, file_size = meta_provider.expand_paths(resolved_path, filesystem)
-                    read_path = read_path[0]
-                    file_size = file_size[0]
+                read_path, file_size = meta_provider.expand_paths(resolved_path, filesystem)
+                # expand_paths returns two lists, so get the first element of each
+                read_path = read_path[0]
+                file_size = file_size[0]
 
             self._paths.append((read_path, idx))
             self._file_sizes.append(file_size)
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -2,6 +2,9 @@ import logging
 import os
 import queue
 import threading
+from typing import Any, Callable, Dict
+
+from packaging import version
 
 from ludwig.api_annotations import DeveloperAPI, PublicAPI
 from ludwig.callbacks import Callback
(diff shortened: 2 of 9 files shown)


ludwig/data/dataset/ray.py
```diff
@@ -280,7 +280,8 @@ def __len__(self):
         try:
             count = next(self.epoch_iter).count()
         except TypeError:
-            count = next(self.epoch_iter.iter_batches()).count()
+            # Sum over all batches when using a DatasetIterator
+            count = sum(map(lambda b: b.count(), self.epoch_iter.iter_batches()))
```
arnavgarg1, line 284: Couple of quick questions:

1. When would we hit the TypeError? Is it in the case that self.epoch_iter is not a DatasetIterator object?
2. Does this end up doing a pass over the entire data and call count on each batch, then sum that total?

jeffkinnison: 1. Yes, that's for backwards compatibility. I'm leaving it in for now just in case, but I think 2.3+ should follow the DatasetIterator path.
2. Yes. `iter_batch` returns an iterator over the batches of one epoch. It batches by block if batch_size isn't provided, but we pass batch_size to the actual batcher in a separate iter_batches call.

jeffkinnison: 2. The cost is relative to dataset size and object store memory, but running time for a number of different dataset sizes seems to be roughly the same as the other methods of getting dataset size from Ray Data.

ludwig/data/dataset/ray.py
```diff
@@ -275,7 +278,22 @@ def initialize_batcher(
...
+                _ray_230
+                and isinstance(self.epoch_iter, ray.data.DatasetIterator)
+                and hasattr(self.epoch_iter, "_base_dataset_pipeline")
+            ):
+                count = next(self.epoch_iter._base_dataset_pipeline).count()
+            else:
+                count = next(self.epoch_iter).count()
+        except TypeError:
+            # Sum over all batches when using a DatasetIterator
+            count = sum(map(lambda b: b.count(), self.epoch_iter.iter_batches()))
```
arnavgarg1, line 293: At this point, what is the type of epoch_iter?

ludwig/contribs/mlflow/__init__.py
```diff
@@ -171,16 +174,50 @@ def on_visualize_figure(self, fig):
...
-            "mlflow": {
-                "experiment_id": self.experiment_id,
-                "experiment_name": self.experiment_name,
-                "tracking_uri": mlflow.get_tracking_uri(),
-            },
-        }
+        import ray
+
+        if version.parse(ray.__version__) >= version.parse("2.3.0"):
+
```
tgaddair, line 180: Is `2.3.0` the right min version? It was working before with 2.3, right? So maybe we just want to check against 2.4?

ludwig/hyperopt/execution.py
```diff
@@ -896,10 +896,10 @@ def _register(name, trainable):
             result_grid = tuner.fit()
             # Get ExperimentAnalysis object from ResultGrid
             analysis = result_grid._experiment_analysis
-        except Exception as e:
+        except Exception:
             # Explicitly raise a RuntimeError if an error is encountered during a Ray trial.
             # NOTE: Cascading the exception with "raise _ from e" still results in hanging.
-            raise RuntimeError(f"Encountered Ray Tune error: {e}")
+            raise RuntimeError(f"Encountered Ray Tune error: {traceback.format_exc()}")
```
tgaddair, line 902: You can change this to:

```
raise RuntimeError(...) from e
```

That way you get the whole traceback without needing to turn it into a stirng.

ludwig/backend/datasource.py
```diff
@@ -163,10 +163,14 @@ def __init__(
                 file_size = None  # unknown file size is None
             else:
                 resolved_path, filesystem = _resolve_paths_and_filesystem([raw_path], filesystem)
-                read_path, file_size = meta_provider.expand_paths(resolved_path, filesystem)
-                # expand_paths returns two lists, so get the first element of each
-                read_path = read_path[0]
-                file_size = file_size[0]
+                try:
```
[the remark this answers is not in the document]
jeffkinnison: Sure, I'll convert it.

ludwig/contribs/mlflow/__init__.py
jeffkinnison: Will do. The new mlflow integration also works with 2.3, but we can introduce it with the 2.4 bump.

ludwig/data/dataset/ray.py
```diff
@@ -275,7 +278,22 @@ def initialize_batcher(
...
+            # Try to count using a pipeline object if possible.
+            if (
+                _ray_230
+                and isinstance(self.epoch_iter, ray.data.DatasetIterator)
+                and hasattr(self.epoch_iter, "_base_dataset_pipeline")
+            ):
+                count = next(self.epoch_iter._base_dataset_pipeline).count()
+            else:
+                count = next(self.epoch_iter).count()
+        except TypeError:
```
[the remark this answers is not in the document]
jeffkinnison: I can break this out into more fine-grained conditions.

ludwig/data/dataset/ray.py
```diff
@@ -434,10 +455,12 @@ def producer():
 
             try:
                 # if augmentation is specified, setup prefetching batch of data
-                if self.augmentation_pipeline:
+                if not _ray_230 and self.augmentation_pipeline:
                     pipeline = pipeline.map_batches(augment_batch, batch_size=batch_size, batch_format="pandas")
 
                 for batch in pipeline.iter_batches(prefetch_blocks=0, batch_size=batch_size, batch_format="pandas"):
+                    if _ray_230:
+                        batch = augment_batch(batch)
```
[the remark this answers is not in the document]
jeffkinnison: So `DatasetIterator` objects don't have a `map_batches` method, and calling `map_batches` on the underlying `_base_dataset_pipeline` in the iterator leads to downstream problems reusing the pipeline that aren't resolved by calling `repeat`. I'll play around with this and see if we can get it into a pipeline.

PR #3415, opened 2023-05-22
Fix Missing Param Metadata

connor-mccorm:
What was going wrong here is that the default for param metadata was `None` which would directly get put into the schema, so unless you explicitly defined param metadata, it would come without param metadata and cause issues downstream. This basically sets the default param metadata value to an empty parameter metadata entry with all the default values for that data class.


arnavgarg1 (approved):
Good catch! So now it will have the dataclass with the default values for all the fields in parameter metadata right?

connor-mccorm:
Yes exactly!

PR #3417, opened 2023-05-23
Set default model for semantic retrieval

arnavgarg1:


arnavgarg1:
@geoffreyangus do we really need a validation error that is model name shouldn't be set if retrieval type is none? shouldn't it not matter and just be ignored?

PR #3419, opened 2023-05-23
Order prompt schema and add title

connor-mccorm:
Reasoning for adding a title param to `unload_jsonschema_from_marshmallow_class`: Basically, in UI, when there is a nested JSON object - so think the Retrieval being a field under prompt - the schema processor looks to create a specific section for the nested object so that it can display all it's properties. Without a title however, it just displays a random UI divider bar which looks weird / doesn't make sense. particular line adds that title so it's clear that the user is adding in parameters for the retrieval section. Broadly speaking, the addition of a title param to this function allows us to add titles in for this reason in other parts of the schema when necessary.

### DIFF
--- a/ludwig/schema/prompt.py
+++ b/ludwig/schema/prompt.py
@@ -67,15 +67,13 @@ class RetrievalConfigField(schema_utils.DictMarshmallowField):
         super().__init__(RetrievalConfig)
 
     def _jsonschema_type_mapping(self):
-        return schema_utils.unload_jsonschema_from_marshmallow_class(RetrievalConfig)
+        return schema_utils.unload_jsonschema_from_marshmallow_class(RetrievalConfig, title="Retrieval")
 
 
 @DeveloperAPI
 @ludwig_dataclass
 class PromptConfig(schema_utils.BaseMarshmallowConfig):
     """This Dataclass is a schema for the nested prompt config under preprocessing."""
-
-    retrieval: RetrievalConfig = RetrievalConfigField().get_default_field()
 
     task: str = schema_utils.String(
         default=None,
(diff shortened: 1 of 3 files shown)


ludwig/schema/prompt.py
```diff
@@ -67,16 +67,14 @@ def __init__(self):
         super().__init__(RetrievalConfig)
 
     def _jsonschema_type_mapping(self):
-        return schema_utils.unload_jsonschema_from_marshmallow_class(RetrievalConfig)
+        return schema_utils.unload_jsonschema_from_marshmallow_class(RetrievalConfig, title="Retrieval")
```
arnavgarg1, line 70: @connor-mccorm what's the motivation to add this?

connor-mccorm: Basically, in UI, when there is a nested JSON object - so think the Retrieval being a field under prompt - the schema processor looks to create a specific section for the nested object so that it can display all it's properties. Without a title however, it just displays a random UI divider bar which looks weird / doesn't make sense. particular line adds that title so it's clear that the user is adding in parameters for the retrieval section. Broadly speaking, the addition of a title param to this function allows us to add titles in for this reason in other parts of the schema when necessary.

arnavgarg1: Got it! How does it work for other parts of the schema that don't explicitly set title like this?

connor-mccorm: So if you don't explicitly set the title, then we just leave the json object without a title like we were doing before. Some of the objects get titles added in the FE which we want to avoid since it just adds extra processing time which makes the UI less snappy, so we should expect to see some more of those moving over and using this mechanism in the near future.

arnavgarg1: Got it!

arnavgarg1 (approved):

PR #3420, opened 2023-05-24
Fix: Device placement for metrics during evaluation

arnavgarg1:
This fixes an issue where the metric functions are on the wrong device during evaluation, and ensures that metrics, metric functions and the model are all on the same device.

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -116,6 +116,7 @@ class Predictor(BasePredictor):
         self.device = device
         self.dist_model = dist_model
         self.model = model
+        self.model.metrics_to_device(device)
 
         if remote:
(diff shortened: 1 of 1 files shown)


tgaddair (approved):

jeffkinnison (approved):

ludwig/models/predictor.py
```diff
@@ -116,6 +116,7 @@ def __init__(
         self.device = device
         self.dist_model = dist_model
         self.model = model
+        self.model.metrics_to_device(device)
```
justinxzhao, line 119: As much as I love 1-line bug fixes, it behooves us to leave a comment so that future readers (which includes ourselves) have more context for why this line of code is important.

PR #3421, opened 2023-05-24
[llm] Fixed loading when performing full fine-tuning

tgaddair:


arnavgarg1 (approved):
LGTM! Yeah I don't know another way except than to check the trainer type

PR #3422, opened 2023-05-24
[llm] Use lower default learning rate and rename llm decoders to extractors

arnavgarg1:


ludwig/schema/trainer.py
```diff
@@ -711,6 +711,21 @@ class GBMTrainerConfig(BaseTrainerConfig):
 class LLMTrainerConfig(BaseTrainerConfig):
     """Base class for all LLM trainer configs."""
 
+    learning_rate: Union[float, str] = schema_utils.OneOfOptionsField(
+        default=0.0001,
```
justinxzhao, line 715: Should this be even lower? Some of our BERT baseline configs use `0.00002`.

justinxzhao (approved):

tgaddair (approved):

ludwig/schema/trainer.py
arnavgarg1: I think this is fine for now from what I've tried, we could go to maybe 0.00005 if we'd like

PR #3423, opened 2023-05-25
[llm] Replace `model_name` with *required* `base_model`, add preset LLM registry, update internal adapter modules

tgaddair:
#### Summary

The new `base_model` parameter is also just a string, backed by a `anyOf` schema that either validates whether the provided string is a Ludwig-supported preset through JSON or an otherwise valid model in the Huggingface repo at runtime (i.e. via validation checks).

It is the first *required* parameter in Ludwig, outside of input and output features, and it uses a bespoke schema tweak to enforce that. It is also the first parameter to use an `anyOf` implementation; this is necessary because `oneOf` validation requires a value to match *exactly one* option schema (a preset LLM name would partially match both schemas and cause an error).

If the user does choose a preset LLM then after JSON validation but *before config object initialization* this value will be swapped for the full slash-delimited path defined in the `MODEL_PRESETS` dictionary.

#### Usage

```
base_model: llama-7b
```

is valid and can be verified by JSON or by the validation checks. By the end of config object initialization it has been replaced by `huggyllama/llama-7b` (which is defined in `MODEL_PRESETS`).

```
base_model: bigscience/bloom-3b
```

is also valid but can only be verified during config-object initialization by the extra validation checks.

and

```
base_model:  <anything other than a string>
```

will cause an immediate validation failure.

#### Necessary follow-up PRs, prompted by changes here
- [x] Looks like the `required` field is missing for model-level parameters (e.g. `input_features`). Probably an accidental regression. - Fix: https://github.com/ludwig-ai/ludwig/pull/3442
- [ ] Add official support for `required` parameters using `marshmallow`'s in-built options
- [ ] Add support for arbitrary selector or "controller" fields (other than `type`).
- [ ] Investigate whether the extra `metadata` dicts inserted into marshmallow fields are necessary anymore


ludwig/schema/llms/base_model.py
```diff
@@ -0,0 +1,166 @@
...
+
+    name: str = schema_utils.String(
+        default=None,
+        allow_none=True,
+        description=(
+            "The name of the model to use. This can be a local path or a "
+            "remote path. If it is a remote path, it must be a valid HuggingFace "
+            "model name. If it is a local path, it must be a valid HuggingFace "
+            "model name or a path to a local directory containing a valid "
+            "HuggingFace model."
```
martindavis, line 76: Should we include a longer description with a link to the HF models?

tgaddair: We can do this, yeah. I would do it in the parameter metadata, which is consistent with how we do it for text encoders, etc. We can do it in a follow-up.

ludwig/schema/model_types/llm.py
```diff
@@ -28,25 +29,22 @@ class LLMModelConfig(ModelConfig):
...
+        default=None,
         description=(
-            "The name of the model to use. This can be a local path or a "
-            "remote path. If it is a remote path, it must be a valid HuggingFace "
-            "model name. If it is a local path, it must be a valid HuggingFace "
-            "model name or a path to a local directory containing a valid "
-            "HuggingFace model."
+            "Base pretrained model to use. This can be one of the presets defined by Ludwig, a fully qualified "
+            "name of a pretrained model from the HuggingFace Hub, or a path to a directory containing a "
+            "pretrained model."
```
ohho, line 47: Since this description shows up in the UI, I feel like it's a little confusing as written. Am curious why we need to say this is a "defined by Ludwig". Also should we tell the user they can choose "None" if they'd like to provide the path/name of a model that we don't have.

For example:
image

arnavgarg1 (approved):
LGTM

ludwig/schema/llms/base_model.py
```diff
@@ -0,0 +1,159 @@
...
+            model_name, parameter_metadata=LLM_METADATA["base_model"]["default"]["name"]
+        )
+
+
+@DeveloperAPI
+def get_base_model_conds():
+    conds = []
+    for base_model_type, base_model_cls in base_model_registry.items():
+        other_props = schema_utils.unload_jsonschema_from_marshmallow_class(base_model_cls)["properties"]
+        schema_utils.remove_duplicate_fields(other_props, fields=["preset"])
```
martindavis, line 98: Just to call this out: we should only include the `name` attribute when the `preset` is `None`. Otherwise, we're telling down stream consumers of the Schema that they should be filling in `name` even though it will be ignored.

ludwig/config_validation/validation.py
```diff
@@ -28,11 +28,16 @@ def get_schema(model_type: str = MODEL_ECD):
 
     cls = model_type_schema_registry[model_type]
     props = unload_jsonschema_from_marshmallow_class(cls)["properties"]
+    required = []
+    if model_type == MODEL_LLM:
+        required = ["base_model"]
+    # TODO: Add required back for ECD/GBM too?
```
ksbrar, line 34: will do in a follow up, surprised that this hasn't caused an error yet

ludwig/schema/model_types/llm.py
ksbrar: @arnavgarg1 @tgaddair thoughts here?

ludwig/utils/backward_compatibility.py
```diff
@@ -848,6 +849,21 @@ def remove_extra_type_param_in_defaults_config(defaults: FeatureTypeDefaultsDict
     return defaults_copy
 
 
+@register_config_transformation("0.8")
```
ksbrar, line 852: @arnavgarg1 @tgaddair not actually sure if this config upgrade is necessary, can I delete it?

also note to self: need to update this as well.

tests/ludwig/schema/test_model_config.py
```diff
@@ -784,3 +785,45 @@ def test_encoder_decoder_values_as_str():
...
+)
+def test_llm_base_model_config_error(base_model_config):
+    config = {
+        "model_type": "llm",
+        "base_model": base_model_config,
+        "input_features": [{"name": "text_input", "type": "text"}],
+        "output_features": [{"name": "text_output", "type": "text"}],
+    }
+
+    # if base_model_config is None:
```
ksbrar, line 825: delete this

tests/ludwig/schema/test_model_config.py
```diff
@@ -784,3 +785,45 @@ def test_encoder_decoder_values_as_str():
 
     assert isinstance(config_obj.input_features[0].encoder, BERTConfig)
     assert isinstance(config_obj.output_features[0].decoder, ClassifierConfig)
+
+
+@pytest.mark.parametrize(
+    "base_model_config,model_name",
+    [
+        ("bigscience/bloom-3b", "bigscience/bloom-3b"),
```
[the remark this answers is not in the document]
ksbrar: good catch

ludwig/config_validation/validation.py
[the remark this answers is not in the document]
arnavgarg1: What's required for ECD/GBM? Nothing except for input/ouput feature names right?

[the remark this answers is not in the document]
ksbrar: right

tests/integration_tests/test_preprocessing.py
```diff
@@ -890,7 +890,7 @@ def test_prompt_template(input_features, expected, model_type, backend, tmpdir,
     model_name = "hf-internal-testing/tiny-random-OPTModel"
     if model_type == MODEL_LLM:
         # For LLMs, specify the prompt at the top level
-        config["model_name"] = model_name
+        config["base_model"] = model_name
```
arnavgarg1, line 893: One thought - given the number of places you've had to make the change from `model_name` to `base_model` across all of the files, I'm wondering if it would be a good idea to just make it a constant and use it everywhere so that if we ever make a change to the name in the future we just have to make it in one place

ksbrar: it actually already exists as a constant and I forgot to use it, good catch

ludwig/schema/llms/base_model.py
```diff
@@ -0,0 +1,102 @@
...
+from marshmallow import fields, ValidationError
+from transformers import AutoConfig
+
+from ludwig.api_annotations import DeveloperAPI
+from ludwig.constants import BASE_MODEL
+from ludwig.error import ConfigValidationError
+from ludwig.schema.metadata import LLM_METADATA
+from ludwig.schema.metadata.parameter_metadata import convert_metadata_to_json
+
+MODEL_PRESETS = {
```
justinxzhao, line 12: nit: Can we document this global variable?

> Maps a preset LLM name to the full slash-delimited HF path. If the user chooses a preset LLM, the preset LLM name is replaced with the full slash-delimited HF path using this map, after JSON validation but before config object initialization.

justinxzhao (approved):

ludwig/schema/llms/base_model.py
ksbrar: done

ludwig/schema/llms/base_model.py
```diff
@@ -0,0 +1,105 @@
...
+                    {
+                        "type": "string",
+                        "enum": list(MODEL_PRESETS.keys()),
+                        "description": "Pick an LLM with first-class Ludwig support.",
+                        "title": "preset",
+                        "parameter_metadata": convert_metadata_to_json(LLM_METADATA[BASE_MODEL]),
+                    },
+                    {
+                        "type": "string",
+                        "description": "Enter the full (slash-delimited) path to a huggingface LLM.",
```
arnavgarg1, line 80: In theory it doesn't have to be slash-delimited, but they do tend to be slash-delimited (for e.g., `bert-base-uncased` isn't slash delimited). I'd say just the full LLM model path.

ludwig/schema/llms/base_model.py
```diff
@@ -0,0 +1,81 @@
...
+                return value
+            raise ValidationError(f"Value to deserialize is not a string: {value}")
+
+        def _jsonschema_type_mapping(self):
+            return {
+                "anyOf": [
+                    {
+                        "type": "string",
+                        "enum": list(MODEL_PRESETS.keys()),
+                        "description": "Pick an LLM with first-class Ludwig support.",
```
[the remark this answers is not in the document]
arnavgarg1: Hmm, maybe more than first-class Ludwig support, I'd probably lean more towards something along the lines of `popular LLMs of different sizes across a variety of architecture types`

ludwig/schema/model_types/llm.py
arnavgarg1: I think this description makes sense for Ludwig users, but I guess it's a bit confusing for Predibase users. 
On the Predibase side, what we care about is `This can be one of the presets or a fully qualified name of a pretrained model from the HuggingFace Hub`.

What do we think?

ludwig/schema/model_types/utils.py
```diff
@@ -304,8 +304,8 @@ def set_llm_tokenizers(config: "ModelConfig") -> None:
     if config.model_type != "llm":
         return
 
-    pretrained_model_name_or_path = config.model_name
-    if pretrained_model_name_or_path is None:
+    pretrained_model_name_or_path = config.base_model
+    if not isinstance(pretrained_model_name_or_path, str) or pretrained_model_name_or_path is None:
         raise ValueError("Must set model_name when using the LLM model.")
```
arnavgarg1, line 309: Update this to be "Must set `base_model` when using the LLM model type"?

ludwig/utils/backward_compatibility.py
arnavgarg1: I think we can get rid of it since we haven't pushed out Ludwig 0.8 yet!

tests/ludwig/schema/test_model_config.py
```diff
@@ -784,3 +788,42 @@ def test_encoder_decoder_values_as_str():
...
+    config = {
+        MODEL_TYPE: MODEL_LLM,
+        BASE_MODEL: base_model_config,
+        INPUT_FEATURES: [{NAME: "text_input", TYPE: "text"}],
+        OUTPUT_FEATURES: [{NAME: "text_output", TYPE: "text"}],
+    }
+
+    config_obj = ModelConfig.from_dict(config)
+
+    assert config_obj.base_model == model_name
```
arnavgarg1, line 810: Can we add a similar test for a fully-qualified name and make sure the custom `AutoConfig` check works?

ludwig/schema/llms/base_model.py
ksbrar: fixed

ludwig/schema/model_types/utils.py
ksbrar: fixed

tests/ludwig/schema/test_model_config.py
ksbrar: Added one more test config with a fully-qualified name that exists. The test below this one checks that the logic raises an error if the path is invalid.

ludwig/utils/backward_compatibility.py
ksbrar: removed

PR #3426, opened 2023-05-30
Support combining multiple columns into a single text feature

tgaddair:
For LLMs and other text models, it's common to treat multiple inputs as a single string of text to feed into the model as input, rather than construct separate "towers" for each feature. This PR provides a standard way to do this via the `prompt` config.

For ECD:

```
input_features:
    - name: context
      type: text
      preprocessing:
          prompt:
            template: The {color} {animal} jumped over the {size} {object}
```

For LLM:

```
model_type: llm
prompt:
    template: The {color} {animal} jumped over the {size} {object}
input_features:
    - name: context
      type: text
```

Here, it is not required that `context` be a valid column of the input dataset, as it will be generated from the `prompt`. However, each of `color`, `animal`, `size`, and `object` are assumed to be columns of the input dataset.

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1231,7 +1231,7 @@ def build_dataset(
     # update input features with prompt configs during preprocessing (as opposed to during the model forward pass)
     # so that we can compute metadata and build the dataset correctly.
     logger.debug("handle text features with prompt parameters")
-    dataset_cols = handle_features_with_prompt_config(
+    synthesized_dataset_cols = handle_features_with_prompt_config(
         config, dataset_df, features, split_col=split_col, backend=backend
     )
 
@@ -1243,8 +1243,12 @@ def build_dataset(
             feature_configs.append(feature)
             feature_hashes.add(feature[PROC_COLUMN])
 
+    dataset_cols = {}
     for feature_config in feature_configs:
-        dataset_cols[feature_config[COLUMN]] = dataset_df[feature_config[COLUMN]]
+        col_name = feature_config[COLUMN]
+        dataset_cols[col_name] = (
+            synthesized_dataset_cols[col_name] if col_name in synthesized_dataset_cols else dataset_df[col_name]
+        )
 
     logger.debug("build preprocessing parameters")
     feature_name_to_preprocessing_parameters = build_preprocessing_parameters(
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -49,6 +49,7 @@ from ludwig.constants import (
     SPLIT,
     SRC,
     TEST,
+    TEXT,
     TRAINING,
     TYPE,
     VALIDATION,
@@ -1217,18 +1218,6 @@ def build_dataset(
         raise ValueError(f"Invalid mode {mode}")
     global_preprocessing_parameters = merge_dict(default_preprocessing_parameters, global_preprocessing_parameters)
 
-    # Get all the unique preprocessing features to compute
-    feature_configs = []
-    feature_hashes = set()
-    for feature in features:
-        if feature[PROC_COLUMN] not in feature_hashes:
-            feature_configs.append(feature)
-            feature_hashes.add(feature[PROC_COLUMN])
-
-    dataset_cols = {}
-    for feature_config in feature_configs:
-        dataset_cols[feature_config[COLUMN]] = dataset_df[feature_config[COLUMN]]
-
     split_col = None
     if global_preprocessing_parameters["split"]["type"] == "fixed":
         if global_preprocessing_parameters["split"]["column"] in dataset_df.columns:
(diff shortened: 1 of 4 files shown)


tests/integration_tests/test_preprocessing.py
```diff
@@ -910,17 +913,84 @@ def test_handle_features_with_few_shot_prompt_config(backend, retrieval_kwargs,
...
+        ]
+    )
+
+    config = {
+        "model_type": MODEL_LLM,
+        "model_name": "gpt2",
+        "input_features": [text_feature(name="question", encoder={"type": "passthrough"})],
+        "output_features": [text_feature(name="answer")],
+        "prompt": {
+            "template": "You are a helpful chatbot. USER: {instruction}: {country}, {year} ASSISTANT:",
```
justinxzhao, line 968: Could we also add a test for prompt templates within an input feature definition?

ludwig/data/prompt.py
```diff
@@ -169,44 +174,53 @@ def format_input_with_prompt(
...
+            df[CONTEXT] = retrieval_model.search(df, backend, k=k, return_data=True)
+        if SAMPLE in template_fields:
+            df[SAMPLE] = df[input_col_name].map(lambda entry: json.dumps(entry, indent=2))
+        if TASK in template_fields:
+            df[TASK] = task_str
 
-        if is_few_shot:
-            df["context"] = retrieval_model.search(df, backend, k=k, return_data=True)
+        def generate_prompt_for_row(row):
+            kwargs = {col: row[col] for col in template_fields}
```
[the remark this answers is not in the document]
tgaddair: It would be like: `{size:.2f}`. I can add some tests for this.

tests/integration_tests/test_preprocessing.py
tgaddair: This is supported, see my PR description above:

```
input_features:
    - name: context
      type: text
      preprocessing:
          prompt:
            template: The {color} {animal} jumped over the {size} {object}
```

tgaddair: Yes, going to add that today.

ludwig/data/prompt.py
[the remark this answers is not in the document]
tgaddair: Added to tests.

tests/integration_tests/test_preprocessing.py
justinxzhao: Nice, thanks!

ludwig/data/prompt.py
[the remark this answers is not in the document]
justinxzhao: Thanks!

ludwig/data/prompt.py
```diff
@@ -169,44 +174,64 @@ def format_input_with_prompt(
             template = DEFAULT_ZERO_SHOT_PROMPT_TEMPLATE
```
justinxzhao, line 174: OOC: Would a user be able to use the `DEFAULT_ZERO_SHOT_PROMPT_TEMPLATE` for the prompt template within an input feature definition like so?

```
input_features:
    - name: context
      type: text
      preprocessing:
          prompt:
            template: None  # Use default.
```

tgaddair: Yes, though I believe in this case they need to provide a `task`.

ludwig/data/preprocessing.py
```diff
@@ -1783,26 +1795,36 @@ def handle_features_with_prompt_config(
...
 def _get_prompt_config(config: ModelConfigDict, input_feature_config: Dict) -> Dict:
-    if "prompt" in config and config["prompt"]["task"] is not None:
-        return config["prompt"]
+    if input_feature_config[TYPE] != TEXT:
+        # Prompt config is only applied to text features
+        return None
 
     preprocessing = input_feature_config["preprocessing"]
-    if "prompt" in preprocessing and preprocessing["prompt"]["task"] is not None:
+    if _has_prompt_section(preprocessing):
```
geoffreyangus, line 1815: How does this work? Are we accepting `prompt` keys at both the local and global level?

ludwig/data/prompt.py
```diff
@@ -169,44 +174,64 @@ def format_input_with_prompt(
...
+            df[CONTEXT] = retrieval_model.search(df, backend, k=k, return_data=True)
+        if SAMPLE in template_fields:
+            df[SAMPLE] = df[input_col_name].map(lambda entry: json.dumps(entry, indent=2))
+        if TASK in template_fields:
+            df[TASK] = task_str
 
-        if is_few_shot:
-            df["context"] = retrieval_model.search(df, backend, k=k, return_data=True)
+        def generate_prompt_for_row(row):
+            kwargs = {col: field_to_dtype[col](row[col]) for col in template_fields}
```
geoffreyangus, line 192: why do we need to cast the values?

ludwig/data/preprocessing.py
tgaddair: The LLM schema requires the prompt at the top-level, and does not allow feature-level prompts.

The ECD schema allows prompts at the feature-level, but not the top level.

ludwig/data/prompt.py
tgaddair: If you have a number like `0.1234` in the DF and you want to format it like `{number:.2f}` it will fail if the number is represented as a string in the DF. So we need to cast it to a float. I can add a comment to this effect.

geoffreyangus (approved):

PR #3431, opened 2023-06-05
Make Reward Model Preprocessing Modifications

asdataminer:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1205,50 +1205,6 @@ def build_dataset(
             logger.debug(f"sample {sample_ratio} of data")
             dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
 
-        # If training a reward model, perform grouping and joining on dataset
-        if "reward" in global_preprocessing_parameters:
-            reward_parameter_names = [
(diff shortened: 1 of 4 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1205,6 +1205,50 @@ def build_dataset(
             logger.debug(f"sample {sample_ratio} of data")
             dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
 
+        # If training a reward model, perform grouping and joining on dataset
```
tgaddair, line 1208: Nice! This looks like the right set of transformations, but I think we'll likely want to do this at the end of preprocessing, using the processed text input feature, rather than here at the beginning. Specifically, I would consider doing this here:

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/preprocessing.py#L1364

I would also suggest adding a test in `test_preprocessing.py` to verify it works end-to-end on a synthetic dataset.

asdataminer: Sounds good! I have made the edits and refactored the code a bit to be simpler.

PR #3438, opened 2023-06-22
Replace all non-word characters in feature names to ensure no downstream issues with external libraries.

justinxzhao:
## Context

Ludwig supports a diverse set of data sources. However, we've seen that sometimes the feature/column names of these data sources can have special characters that cause issues for downstream 3rd party services. For example:
- LightGBM dataset columns do not allow JSON characters, i.e. `:{}`: https://github.com/ludwig-ai/ludwig/pull/3326
- MLFlow logging does not support logging feature-specific metrics if the feature name has parentheses `()`: https://github.com/mlflow/mlflow/issues/1164
- torch `ModuleDict`, which Ludwig uses to represent encoders and decoders, are keyed by the feature name, which may cause issues: https://github.com/pytorch/pytorch/issues/71203

## Two-pronged approach

1. At `ModelConfig.from_dict()` construction time, modify the config to use sanitized feature names (input, output, tied, dependencies, entity lists).
2. At `build_dataset()` time, rename dataframe column names to use sanitized column names.

## Other changes

- Add a script to train a model with synthetically generated data.
- Some minor flake8 fixes

### DIFF
--- a/tests/integration_tests/test_explain.py
+++ b/tests/integration_tests/test_explain.py
@@ -4,10 +4,9 @@ import os
 import numpy as np
 import pandas as pd
 import pytest
-import torch
 
 from ludwig.api import LudwigModel
-from ludwig.constants import BATCH_SIZE, BINARY, CATEGORY, MINIMUM_BATCH_SIZE, MODEL_ECD, MODEL_GBM, NAME
+from ludwig.constants import BATCH_SIZE, BINARY, CATEGORY, MINIMUM_BATCH_SIZE, MODEL_ECD, MODEL_GBM, TYPE
 from ludwig.explain.captum import IntegratedGradientsExplainer
 from ludwig.explain.explainer import Explainer
 from ludwig.explain.explanation import Explanation
@@ -172,15 +171,13 @@ def run_test_explainer_api(
 
     if input_features is None:
         input_features = [
-            number_feature(),
-            category_feature(encoder={"type": "onehot", "reduce_output": "sum"}),
-            category_feature(encoder={"type": "passthrough", "reduce_output": "sum"}),
-        ]
-        input_features.update(
             # Include a non-canonical name that's not a valid key for a vanilla pytorch ModuleDict:
             # https://github.com/pytorch/pytorch/issues/71203
-            {NAME: attribute_name for attribute_name in dir(torch.nn.ModuleDict())}
-        )
+            {"name": "type", "type": "binary"},
+            number_feature(),
+            category_feature(encoder={TYPE: "onehot", "reduce_output": "sum"}),
+            category_feature(encoder={TYPE: "passthrough", "reduce_output": "sum"}),
+        ]
         if model_type == MODEL_ECD:
             # TODO(travis): need unit tests to test the get_embedding_layer() of every encoder to ensure it is
             #  compatible with the explainer


ludwig/utils/data_utils.py
```diff
@@ -1089,3 +1089,17 @@ def use_credentials(creds):
                     json.dump(old_conf, f)
                 conf.clear()
                 set_conf_files(tmpdir, conf)
+
+
+def get_sanitized_feature_name(feature_name: str) -> str:
```
jeffkinnison, line 1094: nit: We should stick with either `feature` or `column` in these function names

ludwig/utils/data_utils.py
```diff
@@ -1089,3 +1089,17 @@ def use_credentials(creds):
...
+    """Replaces non-word characters (anything other than alphanumeric or _) with _.
+
+    Used in model config initialization and sanitize_column_names(), which is called during dataset building.
+    """
+    return re.sub(r"[(){}.:]", "_", feature_name)
```
jeffkinnison, line 1099: I think we need to add `[` and `]` as well, both of those would likely break LightGBM. Do we want to sub whitespace too?

tests/integration_tests/test_explain.py
```diff
@@ -173,7 +173,7 @@ def run_test_explainer_api(
         input_features = [
             # Include a non-canonical name that's not a valid key for a vanilla pytorch ModuleDict:
             # https://github.com/pytorch/pytorch/issues/71203
-            {"name": "binary.1", "type": "binary"},
+            {"name": "__contains__", "type": "binary"},
```
jeffkinnison, line 176: Do we want to add in handling for magic method names? We should be able to enumerate them in either a set or a regex.

ludwig/utils/data_utils.py
jeffkinnison: We should probably include quotation marks as well.

tests/integration_tests/test_explain.py
justinxzhao: The special handling for magic method names is handled via a `LudwigFeatureDict` mechanism, which appends a `FEATURE_NAME_SUFFIX` to ensure there's no issues with torch's `ModuleDict`.

ludwig/utils/data_utils.py
justinxzhao: Done. I didn't include whitespace because this would break a backwards compatibility test, and I also haven't seen whitespace cause issues.

jeffkinnison: That makes sense. Thanks for adding those in!

tests/integration_tests/test_explain.py
justinxzhao: Interesting tidbit: I discovered that there are still some feature names that raise errors due to our own schema objects' restrictions with what `<schema object>.setattr()` accepts.

Code where feature names are called with `setattr()` on schema objects:

```python
class FeatureCollection(Generic[T], schema_utils.ListSerializable):
    def __init__(self, features: List[T]):
        self._features = features
        self._name_to_feature = {f.name: f for f in features}
        for k, v in self._name_to_feature.items():
            setattr(self, k, v)
```

The `ECDBinaryInputFeatureConfig` (and likely other input feature config objects as well) doesn't permit calling `setattr` with:
- `__class__`: `TypeError: __class__ must be set to a class, not 'ECDBinaryInputFeatureConfig' object`
- `__dict__`: `TypeError: __dict__ must be set to a dictionary, not a 'ECDBinaryInputFeatureConfig'`
- `__weakref__`: `AttributeError: attribute '__weakref__' of 'FeatureCollection' objects is not writable`
- `__reduce_ex__`: `TypeError: 'ECDBinaryInputFeatureConfig' object is not callable`

Supporting these feature names would be a rather simple change to move `LudwigFeatureDict` key management logic up to the schema level. I doubt anyone is looking to use these feature names with Ludwig (and be unwilling to rename them haha) CC: @arnavgarg1  @tgaddair @w4nderlust

ludwig/utils/data_utils.py
justinxzhao: Based on the discussion at standup today, I would say that current names seem ok. `get_sanitized_feature_name()` is called on a feature-wise basis while `sanitize_column_names()` is used for the underlying dataframe. They share the same sanitization step.

jeffkinnison: Agreed! It was good to get some clarification on the nomenclature

jeffkinnison (approved):

PR #3442, opened 2023-06-25
int: Add back `required` for input and output features to the Ludwig JSON schema

ksbrar:
At some point this was deleted 😓

### DIFF
--- a/ludwig/config_validation/validation.py
+++ b/ludwig/config_validation/validation.py
@@ -40,6 +40,7 @@ def get_schema(model_type: str = MODEL_ECD):
         "title": "model_options",
         "description": "Settings for Ludwig configuration",
         "required": required,
+        "additionalProperties": False,  # TODO: Set to true after 0.8 releases.
     }
 
 
--- a/ludwig/config_validation/validation.py
+++ b/ludwig/config_validation/validation.py
@@ -6,7 +6,7 @@ from jsonschema import Draft7Validator, validate
 from jsonschema.validators import extend
 
 from ludwig.api_annotations import DeveloperAPI
-from ludwig.constants import MODEL_ECD, MODEL_GBM, MODEL_TYPE
+from ludwig.constants import MODEL_ECD, MODEL_TYPE
 from ludwig.error import ConfigValidationError
 
 # TODO(travis): figure out why we need these imports to avoid circular import error
@@ -30,9 +30,7 @@ def get_schema(model_type: str = MODEL_ECD):
     props = unload_jsonschema_from_marshmallow_class(cls)["properties"]
 
     # TODO: Replace with more robust required logic later.
-    required = []
-    if model_type == MODEL_ECD or model_type == MODEL_GBM:
-        required = ["input_features", "output_features"]
+    required = ["input_features", "output_features"]
 
     return {
         "type": "object",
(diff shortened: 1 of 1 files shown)


ludwig/config_validation/validation.py
```diff
@@ -28,11 +28,19 @@ def get_schema(model_type: str = MODEL_ECD):
...
+    if model_type == MODEL_ECD or model_type == MODEL_GBM:
+        required = ["input_features", "output_features"]
+
     return {
         "type": "object",
         "properties": props,
         "title": "model_options",
         "description": "Settings for Ludwig configuration",
+        "required": required,
+        "additionalProperties": False,  # TODO: Set to true after 0.8 releases.
```
ksbrar, line 43: @tgaddair Actually, do we make this change for 0.8 or do it after?

ludwig/config_validation/validation.py
```diff
@@ -28,11 +28,19 @@ def get_schema(model_type: str = MODEL_ECD):
...
+    if model_type == MODEL_ECD or model_type == MODEL_GBM:
+        required = ["input_features", "output_features"]
+
     return {
         "type": "object",
         "properties": props,
         "title": "model_options",
         "description": "Settings for Ludwig configuration",
+        "required": required,
+        "additionalProperties": True,  # TODO: Set to false after 0.8 releases.
```
ksbrar, line 43: @tgaddair Actually, do we want to make this change for 0.8 or do it after?

justinxzhao: Making input and output features required fields? I'd say make this change now.

justinxzhao (approved):

ludwig/config_validation/validation.py
ksbrar: @justinxzhao Ah sorry, I should have been clearer. Do we want to set `additionalProperties` to false now or later?

connor-mccorm (approved):

jeffkinnison (approved):

ludwig/config_validation/validation.py
```diff
@@ -28,11 +28,19 @@ def get_schema(model_type: str = MODEL_ECD):
 
     cls = model_type_schema_registry[model_type]
     props = unload_jsonschema_from_marshmallow_class(cls)["properties"]
+
+    # TODO: Replace with more robust required logic later.
+    required = []
+    if model_type == MODEL_ECD or model_type == MODEL_GBM:
+        required = ["input_features", "output_features"]
```
arnavgarg1, line 35: This is actually true for all model types including LLMs

ksbrar: good catch

ludwig/config_validation/validation.py
justinxzhao: Ah I see.

Curious to know what other people think. I'd be in favor of making the change now since 1) there are robustness benefits, 2) we've wanted to make this change for a while, 3) there's an easy way to add a new parameter with schema_utils.Dict().

Is the error message for legacy users who might have inadvertently been including extra parameters in their configs clear?

ksbrar: @justinxzhao This is the deprecation warning that shows up right now:
https://github.com/ludwig-ai/ludwig/blob/dc0d685ac70729c667acc83f85c897cab7b38917/ludwig/schema/utils.py#L186

It actually mentions that 0.8 will be strict, so I would like to do this now too.

In any case, I'll raise a separate PR for it and we can have the discussion there if need be.

ksbrar:
Last failures were due to `daft` dependency upgrade. Second-to-last commit was green, confident that the last commit did not affect anything.

PR #3444, opened 2023-06-27
feat: Disallow additional properties everywhere, enforce strict validation by default

ksbrar:
#### Interesting tidbits
* We use `draft-7` version of the JSON spec. In this version, `additionalProperties` will collide with any nested `allOf` properties because it only semantically applies to elements specified at the same level of the schema that is it set. So we can't set it for any of our `allOf` fields - loading the model config object will still capture the errors, but this is unexpected. The next version of the spec, `draft-2019`, adds a new keyword that covers this scenario so at some point we should upgrade to it.


ksbrar:
This PR has major problems and ramifications, will have to rethink this later.

PR #3447, opened 2023-06-28
[llm] fix device placement issues when using CPUs and GPUs during LLM fine tuning

arnavgarg1:


justinxzhao (approved):

justinxzhao:
This is at least the second CI run in recent history I've seen `test_experiment_model_resume_distributed[horovod]` timeout now (it seems flaky, maybe we should disable it soon). The logs show that the run actually completed -- any ideas?

arnavgarg1:
> This is at least the second CI run in recent history I've seen `test_experiment_model_resume_distributed[horovod]` timeout now (it seems flaky, maybe we should disable it soon). The logs show that the run actually completed -- any ideas?

Right. I'm not 100% sure, but what it means is that this test is highly non-deterministic and it's something we should absolutely either fix or figure out if it's safe disable.

Issue #1924, opened 2022-04-13
Type hints and static analysis


dantreiman:
The task here is to add type hints for all functions.

This will enable static analysis of the codebase, to catch errors faster and make refactoring easier.

PR #3449, opened 2023-06-29
refactor: add type hints for encoders

Dennis-Rall:
- add missing type hints in the encoders package (see #1924)
- introduce a [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) for the encoder output, so that we can ensure that the `encoder_output` key is present in the returned dict


ludwig/encoders/types.py
```diff
@@ -0,0 +1,9 @@
+from typing import TypedDict
+
+import torch
+
+
+class EncoderOutputDict(TypedDict, total=False):
```
justinxzhao, line 6: It would be good to add a comment about which encoders return which tensors, or that only sequence encoders return `encoder_output_state` and `attentions`.

justinxzhao (approved):
Thanks for the change! This type of change is great for code readability.

ludwig/encoders/types.py
Dennis-Rall: Yeah, your're right! I've added the comments. `attentions` are only used by the ViTLegacy Encoder, which is planned to be removed in 0.8. So this return tensor can then be removed too

Dennis-Rall:
During implementation I noticed that the bag, set and categorical encoders return simple tensors instead of dicts. Is this intentional, or is there some work to do to make them return dicts as well? (I could take this on)

justinxzhao:
> During implementation I noticed that the bag, set and categorical encoders return simple tensors instead of dicts. Is this intentional, or is there some work to do to make them return dicts as well? (I could take this on)

Good observation! I'd say it's not intentional and probably came about once there were multiple tensors that needed to be returned from the encoders. CC: @w4nderlust @tgaddair @jimthompson5802 

The reason why the model still works with some encoders returning tensor dicts with an `encoder_output` key and other encoders returning raw tensors is because the encoders are all wrapped by their respective `<Type>InputFeature` module.

You'll notice that the `BagInputFeature`, `SetInputFeature`, and `CategoryInputFeature` modules package the encoder's output in a tensor dict (with `"encoder_output"` tensor key) while the other `<TYPE>InputFeature` modules like `NumberInputFeature` return the output from their encoders (which are already in tensor dict form) as is.

This is definitely a bit confusing -- if making encoder outputs consistent something you would be able to pick up as part of this PR, that would be awesome.

Dennis-Rall:
Thanks for the explanation. I'll give it a try

Dennis-Rall:
Some integration tests with the explainer functionality are failing. I am not very familiar with the explainer feature. Can you give me some hints on how to fix them?

justinxzhao:
> Some integration tests with the explainer functionality are failing. I am not very familiar with the explainer feature. Can you give me some hints on how to fix them?

Hi @Dennis-Rall, I pushed up a commit that should fix the explanation tests.

In Ludwig's implementation of [explainability with captum](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/explain/captum.py#L372-L428), we provide a list of torch modules that should be used to compute attribution scores for each feature.

For [embedded feature types](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/explain/captum.py#L46), the layers provided to captum are determined by `encoder.get_embedding_layer()` ([CategoricalOneHotEncoder.get_embedding_layer()](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/encoders/category_encoders.py#L219), for example). Internally, captum assumes these modules return tensors and captum tries to call `.clone()` on them.

Your PR modified category encoders to return a dictionary of tensors (to make this consistent with the rest of ludwig). So the additional change to fix explainability functionality is to add an additional `nn.Identity()` module shim to capture the encoder's non-dict output and, for captum, change `encoder.get_embedding_layer()` to return a reference to that module.

Hope that was clarifying!

Dennis-Rall:
Yeah, thanks for the explanation and the fix :smile:

PR #3457, opened 2023-07-06
Initial implementation of DaftDataFrameEngine

jaychia:
# Code Pull Requests

This PR introduces the DaftDataFrameEngine, which is a DataFrameEngine implementation that is backed by [Daft](https://github.com/Eventual-Inc/Daft/tree/main/daft)

This has several advantages:

1. Daft is easy to use locally as well as remotely
2. Daft is fast and provides native kernels such as those for URLs, images and tensors
3. Daft runs natively on Ray
4. Daft has a simpler API than the current distributed dataframe abstraction (Dask) because it has no concept of indexing, which complicates many operations in Ludwig such as joins etc

### DIFF
--- a/ludwig/data/dataframe/daft.py
+++ b/ludwig/data/dataframe/daft.py
@@ -58,44 +58,34 @@ class DaftEngine(DataFrameEngine):
         )
 
     def map_objects(self, series, map_fn, meta=None):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def map_partitions(self, series, map_fn, meta=None):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def map_batches(self, series, map_fn, enable_tensor_extension_casting=True):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def apply_objects(self, df, apply_fn, meta=None):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def reduce_objects(self, series, reduce_fn):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def split(self, df, probabilities):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def remove_empty_partitions(self, df):
-        # TODO(jay): Implement
-        pass
+        raise NotImplementedError("TODO: Needs implementation!")
 
     def to_parquet(self, df, path, index=False):
-        # TODO(jay): Implement
-        pass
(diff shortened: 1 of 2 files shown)


arnavgarg1:
Nice! This looks like a great starting point - let me do the work we discussed on my end and (1) cleanup unused functions and (2) move pandas/python specific logic behind DF engine functions where possible

jaychia:
> Nice! This looks like a great starting point - let me do the work we discussed on my end and (1) cleanup unused functions and (2) move pandas/python specific logic behind DF engine functions where possible

I added a shim class on top of the Daft Dataframe here to ease the transition! The shim classes:

1. `DaftShimDataframe`: primarily enables the `__setitem__` method to support operations in Ludwig that make strong assumptions about the underlying Dataframe abstraction supporting the `df["foo"] = series` syntax
2. `DaftShimSeries`: the Daft "series" is actually just the `daft.Expression`, which is "anonymous" and only takes on an identity ("this is the Series I refer to") when it is applied on a Dataframe. Some places in the ludwig code run functions directly on Series objects and have strong assumptions about the Series object supporting a Pandas API (e.g. "series.astype(np.int32)"), so this could be a good layer to help us with that abstraction, or at least to help us eventually migrate off places where that happens.

jaychia:
Quick question @arnavgarg1  - do you know how `.map_partitions` and `.map_batches` ([on the DataFrameEngines](https://github.com/ludwig-ai/ludwig/blob/master/ludwig/data/dataframe/base.py#L47-L53)) are different?

Based on the function signature/naming it seems like the intention is for `.map_partitions` to run on `Series`, and `.map_batches` to run on Dataframes. However, it seems in practice they’ve been used fairly interchangeably, and surprisingly without any bugs?

1. [Dask implementation](https://github.com/ludwig-ai/ludwig/blob/231fd62fb46182ea90171af670ed7ec1e52d26a9/ludwig/data/dataframe/dask.py#L156-L171) of `map_batches` says it runs on a Series, but the Pandas and Modin ones say it runs on Dataframes
2. Naming of `map_partitions` on the Dask, Pandas and Modin implementations seem to agree that it takes in a Series, but in practice it is actually used on Dataframes sometimes.
    - In [prompt.py](https://github.com/ludwig-ai/ludwig/blob/231fd62fb46182ea90171af670ed7ec1e52d26a9/ludwig/data/prompt.py#L201)
    - In [split.py](https://github.com/ludwig-ai/ludwig/blob/231fd62fb46182ea90171af670ed7ec1e52d26a9/ludwig/data/split.py#L220)
    - In [ray.py](https://github.com/ludwig-ai/ludwig/blob/231fd62fb46182ea90171af670ed7ec1e52d26a9/ludwig/backend/ray.py#L999)

It appears they are both used fairly interchangeably. I wonder if y'all wanted to maybe consolidate them?

Ideally also if they don't run on arbitrary Pandas dictionaries, and instead users of the API must specify the names of the columns that they need to map over, it would make query optimization more effective! A black-box "map this function over the entire dataframe" leaves Daft with no choice but to think that your query requires every single column, even if in practice your Python function only uses one specific column.

PR #3467, opened 2023-07-13
Updated Ludwig CI for torch nightly

Infernaught:
Addresses the failing unit tests [here](https://github.com/ludwig-ai/ludwig/actions/runs/5522490134/jobs/10072155042?pr=3459). 

Horovod installation and unit tests disabled for torch nightly.
https://github.com/ludwig-ai/ludwig/issues/3468


ludwig/utils/audio_utils.py
```diff
@@ -64,8 +63,11 @@ def read_audio_from_path(path: str) -> Optional[TorchAudioTuple]:
 
     Useful for reading from a small number of paths. For more intensive reads, use backend.read_binary_files instead.
     """
-    bytes_obj = get_bytes_obj_from_path(path)
-    return read_audio_from_bytes_obj(bytes_obj)
+    try:
+        return torchaudio.backend.sox_io_backend.load(path)
+    except Exception as e:
```
jeffkinnison, line 68: Is there a more specific exception we could catch here?

requirements.txt
```diff
@@ -13,10 +13,7 @@ torchvision
 transformers>=4.31.0
 tokenizers>=0.13.3
 spacy>=2.3
-
-# https://github.com/yaml/pyyaml/issues/601
-PyYAML>=3.12,<6.0.1
-
+PyYAML>=3.12,<6.0.1,!=5.4.*
```
jeffkinnison, line 16: nit: Since we're excluding a very small range of PyYAML versions, could you add a short comment explaining why 5.4.x is excluded?

tests/integration_tests/test_ray.py
```diff
@@ -574,19 +574,20 @@ def test_ray_audio(tmpdir, dataset_type, ray_cluster_2cpu):
...
-    ]
-    run_preprocessing(
-        tmpdir,
-        "dask",
-        input_features,
-        output_features,
-        dataset_type=dataset_type,
-        nan_percent=0.1,
-    )
+    with tempfile.TemporaryDirectory() as tmpdir:
```
jeffkinnison, line 577: Out of curiosity, can we not use [pytest's tmp_path fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html) here?

ludwig/features/audio_feature.py
```diff
@@ -145,7 +147,11 @@ def _process_in_memory(
         backend,
     ):
         df_engine = backend.df_engine
-        raw_audio = backend.read_binary_files(column, map_fn=read_audio_from_bytes_obj)
+        if version.parse(torch.__version__) > version.parse("2.0.0"):
```
jeffkinnison, line 150: Do we want to check for the`torchaudio` version rather than the `torch` version? Is there a case where `torch` is 2.0+ but we would still want to use the else block?

jeffkinnison (commented):
Great work! Just had a few comments.

ludwig/features/audio_feature.py
Infernaught: I don't recall making this change, so I'm not sure what the specific reasoning of this one is.

tests/integration_tests/test_ray.py
Infernaught: Let me give this a shot. I recall that there were issues using tmpdir, but I'm seeing some posts online about switching over to tmp_path

jeffkinnison: I wouldn't worry about updating it for now if this is working. The original question was just to satisfy my own curiosity :)

ludwig/features/audio_feature.py
jeffkinnison: No worries! Looks like @justinxzhao added this bit, we can revisit it later if necessary.

ludwig/utils/audio_utils.py
Infernaught: I’m noticing that the same kind of Exception throwing is occurring in the block on line 76. I’m not sure what specific exceptions are going to be caught here, but it appears that we want to be general about it based on the other test? Do you think it would be better to try to catch a more specific one?

jeffkinnison: I'm typically a fan of catching specific exception types so that we don't suppress errors we need to fix. For now, this shouldn't block the merge.

jeffkinnison (approved):

PR #3469, opened 2023-07-14
Zero copy initialization of models onto training workers for LLMs

arnavgarg1:
This PR enables zero copy loading of model weights (particularly for LLMs when using Deepspeed) across training workers. It works by first loading the model into CPU memory, then extracting the weights as a numpy array and serializing it into the Ray object store. When training workers are initialized, we deserialize these weights using the object store reference, so all of the training workers load the same copy of the weights in object store memory, making this a fast, zero-copy operation.

This is useful to prevent ballooning up of CPU memory because otherwise, each worker would load its own instance of the model weights which would in effect require compute resources of model size * num_workers which can be massive for larger LLMs.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -665,6 +665,7 @@ class LudwigModel:
                             else:
                                 calibrator.train_calibration(validation_set, VALIDATION)
                         if not skip_save_model:
+                            self.model = trainer.model
                             self.model.save(model_dir)
 
                     # Evaluation Frequency
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -638,12 +638,13 @@ class LudwigModel:
                         test_set=test_set,
                         save_path=model_dir,
                     )
+                    (self.model, train_trainset_stats, train_valiset_stats, train_testset_stats) = train_stats
 
                     # Calibrates output feature probabilities on validation set if calibration is enabled.
                     # Must be done after training, and before final model parameters are saved.
                     if self.backend.is_coordinator():
                         calibrator = Calibrator(
-                            trainer.model,
+                            self.model,
                             self.backend,
                             batch_size=trainer.eval_batch_size,
                         )
@@ -667,7 +668,6 @@ class LudwigModel:
                         if not skip_save_model:
                             # ensure that any changes to the model object held by the
                             # trainer class are reflected in the model in this class.
-                            self.model = trainer.model
                             self.model.save(model_dir)
 
                     # Evaluation Frequency
@@ -687,7 +687,6 @@ class LudwigModel:
                     # List[TrainerMetric], with one entry per training checkpoint, according to steps_per_checkpoint.
                     # We reduce the dictionary of TrainerMetrics to a simple list of floats for interfacing with Ray
                     # Tune.
-                    (self.model, train_trainset_stats, train_valiset_stats, train_testset_stats) = train_stats
                     train_stats = TrainingStats(
                         metric_utils.reduce_trainer_metrics_dict(train_trainset_stats),
                         metric_utils.reduce_trainer_metrics_dict(train_valiset_stats),
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -666,8 +666,6 @@ class LudwigModel:
                             else:
                                 calibrator.train_calibration(validation_set, VALIDATION)
                         if not skip_save_model:
-                            # ensure that any changes to the model object held by the
-                            # trainer class are reflected in the model in this class.
                             self.model.save(model_dir)
 
                     # Evaluation Frequency
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -665,7 +665,7 @@ class LudwigModel:
                             else:
                                 calibrator.train_calibration(validation_set, VALIDATION)
                         if not skip_save_model:
-                            # ensure that any changes to the model object held by the 
+                            # ensure that any changes to the model object held by the
                             # trainer class are reflected in the model in this class.
                             self.model = trainer.model
                             self.model.save(model_dir)
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -665,6 +665,8 @@ class LudwigModel:
                             else:
                                 calibrator.train_calibration(validation_set, VALIDATION)
                         if not skip_save_model:
+                            # ensure that any changes to the model object held by the 
+                            # trainer class are reflected in the model in this class.
                             self.model = trainer.model
                             self.model.save(model_dir)
 
--- a/ludwig/backend/__init__.py
+++ b/ludwig/backend/__init__.py
@@ -59,6 +59,15 @@ def _has_ray():
         return False
 
 
+def is_ray_backend(backend) -> bool:
+    if isinstance(backend, str):
+        return backend == "ray"
+    elif isinstance(backend, dict):
+        return backend.get("type", "local") == "ray"
+    else:
+        return False
+
+
 def get_local_backend(**kwargs):
     return LocalBackend(**kwargs)
 
--- a/ludwig/backend/__init__.py
+++ b/ludwig/backend/__init__.py
@@ -59,15 +59,6 @@ def _has_ray():
         return False
 
 
-def is_ray_backend(backend) -> bool:
-    if isinstance(backend, str):
-        return backend == "ray"
-    elif isinstance(backend, dict):
-        return backend.get("type", "local") == "ray"
-    else:
-        return False
-
-
 def get_local_backend(**kwargs):
     return LocalBackend(**kwargs)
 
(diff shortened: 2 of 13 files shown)


ludwig/models/llm.py
```diff
@@ -196,6 +212,15 @@ def initialize_adapter(self):
     def to_device(self, device):
```
geoffreyangus, line 212: pretty sure `to_device` isn't called when using the deepspeed trainer

ludwig/models/llm.py
```diff
@@ -89,17 +90,32 @@ def __init__(
         self._random_seed = random_seed
 
         self.model_name = self.config_obj.base_model
+        self.model_config = AutoConfig.from_pretrained(self.config_obj.base_model)
+
+        backend = self.config_obj.backend
+        self.backend_type = backend.type if hasattr(backend, "type") else "local"
```
geoffreyangus, line 96: backend is a dictionary coming in, it looks like– so this doesn't work

arnavgarg1: Intriguing, I feel like that's weird because it should just be the config object and not a dictionary

ludwig/models/llm.py
arnavgarg1: Agreed

ludwig/models/llm.py
```diff
@@ -89,17 +99,35 @@ def __init__(
...
-        self.curr_device = torch.device("cpu")  # model initially loaded onto cpu
+        if self.is_ray_backend:
+            # Extract weights as numpy tensors and place them in the Ray object store.
+            # If we store the weights of a model as NumPy arrays on Plasma, we can access those
+            # weights directly out of Plasma’s shared memory segments, without making any copies.
+            # This enables zero copy model loading on each training worker using shared
+            # memory from the Ray object store for model initialization.
+            import ray
+
+            self.model_ref = ray.put(
```
geoffreyangus, line 115: we should do something to make downstream calls to `self.model` not fail

ludwig/models/llm.py
```diff
@@ -89,17 +91,35 @@ def __init__(
...
```
tgaddair, line 107: Why did we get rid of `self.model`? After the extract function is called, it has no data. So there's no reason not to assign it to `self.model`, as far as I can tell.

ludwig/models/llm.py
```diff
@@ -152,15 +172,16 @@ def __init__(
...
         )
 
         # Extract the decoder object for the forward pass
         self._output_feature_decoder = ModuleWrapper(self.output_features.items()[0][1])
 
-        # Initialize the PEFT adapter is one is provided
-        self.initialize_adapter()
+        # Defer adapter initialization until after serialization for Ray backend
+        if not self.is_ray_backend:
+            self.initialize_adapter()
```
tgaddair, line 184: Why do we defer initialization to post serialization when using an adapter?

tgaddair (commented):
Main thing I would change is where this behavior is being applied. Currently we're doing this all within the LLM object, but I don't think this is necessary or desirable. Instead, what we can do is move this into the Ray backend here (for serialization / extract tensors):

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L482

And here (for deserialization / replace tensors):

https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L197

I think this is the right place to do it as the sequence of events will look something like this:

1. Model is initialized on the drive on CPU
2. Model is serialized, during which we call `extract_tensors`, put the weights into the object store, and strip them out of the model on the head node.
3. Model is deserialized, we call `replace_tensors` to load weights (zero-copy) back into the model on the remote workers.
4. Model is loaded from checkpoint here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/backend/ray.py#L497

Step (4) should be fine, as the model on the head node is already stripped of its weights, so there should be no additional memory overhead from loading the weights from the checkpoint. The only thing to watch out for is that the weights in the object store have been deleted, which should be the case as all references to the weights have been deleted.

ludwig/utils/model_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    tensors.
+
+    Reference implementation: https://medium.com/ibm-data-ai/how-to-load-pytorch-models-340-times-faster-with-
+    ray-8be751a6944c  # noqa
+    """
+
+    tensors = []
+    for _, module in model.named_modules():
+        # Store the tensors as numpy arrays in Python dictionaries
+        # Move the same tensors to a meta device since we no longer need them and we want to reduce memory pressure.
```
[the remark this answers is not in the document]
arnavgarg1: Ah yeah need to delete the comment. I was trying moving to meta device earlier but I was running into issues where it wasn't actually being placed, so decided to just delete the parameters.

ludwig/models/llm.py
arnavgarg1: I don't think we need to, will revert. I initially thought we'd end up with a model structure that is PEFT(LLM()) where the LLM parameters are None but PEFT parameters actually have the weight matrixes, and during replace_tensor we'd run into an issue where we're loading the wrong weights for the wrong layers, but I don't think that will be the case.

ludwig/models/llm.py
arnavgarg1: I just felt like we didn't need to keep a copy in CPU memory at all at this point - are you suggesting the following change?

```
self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.base_model)
self.model_ref = ray.put(extract_tensors(self.model))
```

And the idea is that self.model still has the skeleton, just that the underlying weights have been set to None in place? I think that would probably be ok!

ludwig/models/llm.py
arnavgarg1: Actually this may still be true

ludwig/backend/__init__.py
```diff
@@ -59,6 +59,15 @@ def _has_ray():
         return False
 
 
+def is_ray_backend(backend) -> bool:
+    if isinstance(backend, str):
+        return backend == "ray"
+    elif isinstance(backend, dict):
+        return backend.get("type", "local") == "ray"
+    else:
+        return False
```
arnavgarg1, line 68: Not using this, but going to leave it in here

ludwig/backend/ray.py
```diff
@@ -501,6 +512,12 @@ def train(
         # use `strict=False` to account for PEFT training, where the saved state in the checkpoint
         # might only contain the PEFT layers that were modified during training
         state_dict, *args = results
+
+        # Re-add the model weights to the model
+        model, model_weights = ray.get(model_ref)
+        replace_tensors(model, model_weights, torch.device("cpu"))
+        self.model = model
+        self.model.initialize_adapter()
```
geoffreyangus, line 520: this needs to be a gated step (i.e. check for if the model is an LLM or otherwise)

ludwig/api.py
```diff
@@ -665,6 +665,7 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                             else:
                                 calibrator.train_calibration(validation_set, VALIDATION)
                         if not skip_save_model:
+                            self.model = trainer.model
```
geoffreyangus, line 668: Ensure that any changes to the model object held by the trainer class are reflected in the model here.

tgaddair: Should this just be a return type of the trainer.train method? Otherwise it seems we're relying on some inherent statefulnes that shouldn't be guaranteed.

ludwig/backend/__init__.py
tgaddair: Probably better to remove if unused.

ludwig/models/llm.py
```diff
@@ -193,6 +193,9 @@ def initialize_adapter(self):
             self.model.print_trainable_parameters()
             logger.info("==================================================")
 
+    def prepare_for_training(self):
+        self.initialize_adapter()
```
tgaddair, line 197: Would add a TODO comment that this will not work if resuming from previous checkpoint.

tgaddair (commented):
Looks great! Just a couple remaining items.

ludwig/api.py
geoffreyangus: That's reasonable– I was basing it off of the Calibrator logic, which uses `trainer.model` directly for some reason: https://github.com/ludwig-ai/ludwig/blob/0fdcb98bcd926c7ad83974743cc21f0953261986/ludwig/api.py#L646

Happy to change this line of code as well if we think it is also a code smell.

ludwig/backend/__init__.py
geoffreyangus: removed.

ludwig/models/llm.py
geoffreyangus: Done.

ludwig/api.py
geoffreyangus: Alright, updated both this line and the Calibrator line to use `self.model`.

Issue #3440, opened 2023-06-23
pyarrow.lib.ArrowInvalid: GetFileInfo() yielded path 'data.training.parquet/part.00000000.parquet', which is outside base dir 'data.training.parquet'


xxivk:
Hello , 
am trying to run Ray as backend 

my yaml file : 
`input_features:
- type: number
  name: feature_1
  encoder:
    type: dense
    output_size: 64
- type: number
  name: feature_2
  encoder:
    type: dense
    output_size: 64
- type: number
  name: feature_3
  encoder:
    type: dense
    output_size: 64
combiner:
  type: transformer
  bias_initializer: zeros
  dropout: 0.2
  fc_activation: relu
  fc_dropout: 0
  fc_residual: false
  hidden_size: 64
  num_fc_layers: 0
  num_heads: 8
  num_layers: 5
  output_size: 64
  reduce_output: attention
  transformer_output_size: 256
  use_bias: true
  weights_initializer: xavier_uniform
defaults:
  number:
    preprocessing:
      normalization: zscore
trainer:
  epochs: 500
  early_stop: 200
  validation_field: assigned_label
  evaluate_training_set: true
  learning_rate: 0.0001
  batch_size: 128
  learning_rate_scheduler:
    warmup_evaluations: 5
  optimizer:
    type: adam
output_features:
- name: assigned_label
  type: number
  reduce_dependencies: sum
  reduce_input: sum
  preprocessing:
    computed_fill_value: 0
    fill_value: 0
    missing_value_strategy: drop_row
  loss:
    type: mean_squared_error
    weight: 1
  decoder:
    bias_initializer: zeros
    fc_activation: relu
    fc_bias_initializer: zeros
    fc_dropout: 0
    fc_output_size: 256
    fc_use_bias: true
    fc_weights_initializer: xavier_uniform
    num_fc_layers: 2
    type: regressor
    use_bias: true
    weights_initializer: xavier_uniform
backend:
  type: ray
  processor:
    type: dask
  trainer:
    strategy: horovod
`
[my csv data 
](https://github.com/dragosmc/ludwig-ray/blob/main/data.csv)

i run command :
**ludwig train --config config.yaml --dataset data.csv**

![image](https://github.com/ludwig-ai/ludwig/assets/42117836/9eb82836-b517-467f-bdaf-b6c7e2ee0e60)

i get error : 

`Users\crypto\Desktop\ludwig-ray-main>ludwig train --config config.yaml --dataset data.csv
[...]

PR #3470, opened 2023-07-19
fix: Read partitioned parquet files from relative paths

jeffkinnison:
Fix for #3440. Parquet reads fail under the following conditions:

1. The parquet data is partitioned into multiple files
2. A relative path is passed to `pyarrow.parquet.read_table`
3. A filesystem object is passed to `pyarrow.parquet.read_table`

This error can be addressed in two ways:

a. Convert the relative path to an absolute path
b. Do not pass a filesystem object

While determining the root cause, I [this StackOverflow question](https://stackoverflow.com/questions/72048415/arrowinvalid-getfileinfo-yielded-path-which-is-outside-base-dir-parquet) and [this pyarrow Jira issue](https://issues.apache.org/jira/browse/ARROW-10937), both of which indicate that there are issues with passing the filesystem object. Both indicate that there are potential issues with passing a filesystem object to `pyarrow`, so I decided to go for fix b.

This fix modifies `ludwig.data.dataset.ray.read_remote_parquet` to catch `pyarrow.lib.ArrowInvalid` errors and try the read again with no `filesystem` kwarg. It also adds a test for this issue.


ksbrar (approved):
awesome 🚀

PR #3476, opened 2023-07-23
Add QLoRA for 4-bit fine-tuning

tgaddair:


justinxzhao (approved):

arnavgarg1:
@tgaddair one though - can we change the top-level parameter to something like `model_parameters` with `quantization` as a nested key inside?

```
model_parameters:
    quantization:
        bits: 4
```

I ask because I want to add support for RoPE scaling, which can also be added under this top-level parameter, as well as any future ways to modify the model's representation. Wdyt?

PR #3477, opened 2023-07-24
Add RoPE scaling to increase context length up to 8K for training or inference.

arnavgarg1:
This PR adds support for dynamic RoPE-scaling (rotary position embeddings) to extend the context length of LLM like LLaMA, GPT-NeoX, or Falcon. 

To optionally configure these parameters, you can set them under the `model_parameters` section of the Ludwig config.

```yaml
model_type: llm
base_model: meta-llama/Llama-2-7b
model_parameters:
    rope_scaling:
        type: dynamic
        factor: 2.0

input_features:
  - name: instruction
    type: text
output_features:
  - name: output
    type: text
```

For more context on RoPE scaling, check out the following:
- https://kaiokendev.github.io/context
- https://www.reddit.com/r/LocalLLaMA/comments/14mrgpr/dynamically_scaled_rope_further_increases/

### DIFF
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -102,7 +102,18 @@ class LLM(BaseModel):
                 self.load_kwargs["rope_scaling"] = self.config_obj.model_parameters.rope_scaling.to_dict()
 
         logger.info("Loading large language model...")
-        self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.base_model, **self.load_kwargs)
+        try:
+            self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.base_model, **self.load_kwargs)
+        except TypeError as e:
+            error_message = str(e)
+            # Rope Scaling isn't supported for all model types
+            if "unexpected keyword argument 'rope_scaling'" in error_message:
+                logger.warning(
+                    f"rope_scaling is not supported for {self.config_obj.base_model}."
+                    "Loading the model without rope_scaling."
+                )
+                self.load_kwargs.pop("rope_scaling")
+            self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.base_model, **self.load_kwargs)
 
         # Model initially loaded onto cpu
         self.curr_device = torch.device("cpu")
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -96,6 +96,10 @@ class LLM(BaseModel):
             # Apply quanitzation configuration at model load time
             self.load_kwargs["torch_dtype"] = getattr(torch, self.config_obj.quantization.bnb_4bit_compute_dtype)
             self.load_kwargs["quantization_config"] = self.config_obj.quantization.to_bitsandbytes()
+        if self.config_obj.model_parameters:
+            # Apply RoPE scaling at model load time
+            if hasattr(self.config_obj.model_parameters, "rope_scaling"):
+                self.load_kwargs["rope_scaling"] = self.config_obj.model_parameters.rope_scaling.to_dict()
 
         logger.info("Loading large language model...")
         self.model = AutoModelForCausalLM.from_pretrained(self.config_obj.base_model, **self.load_kwargs)
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -85,7 +85,7 @@ def load_pretrained_from_config(
 
     if config_obj.model_parameters:
         # Add any model specific parameters to the load kwargs
-        for param_name, param_value in config_obj.model_parameters.to_config().items():
+        for param_name, param_value in config_obj.model_parameters.to_dict().items():
             # Not all parameters are supported by all models, so we only add the parameter to the load kwargs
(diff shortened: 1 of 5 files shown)


ludwig/schema/llms/model_parameters.py
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+
+@DeveloperAPI
+class ModelParametersConfigField(schema_utils.DictMarshmallowField):
+    def __init__(self):
+        super().__init__(ModelParametersConfig, default_missing=True)
+
+    def _jsonschema_type_mapping(self):
+        return schema_utils.unload_jsonschema_from_marshmallow_class(ModelParametersConfig)
```
tgaddair, line 73: Need to make the same changes to support `None` that were added for the quantization config:

https://github.com/ludwig-ai/ludwig/commit/279a9ba1cb516948bf71754f973bfc2d7868fee2

ludwig/models/llm.py
```diff
@@ -71,13 +71,27 @@ def update(self, modules: Dict[str, torch.nn.Module]) -> None:
...
 
+    if config_obj.model_parameters:
+        # Add any model specific parameters to the load kwargs
+        for param_name, param_value in config_obj.model_parameters.to_config().items():
+            # Not all parameters are supported by all models, so we only add the parameter to the load kwargs
+            # if it is supported by the model.
+            if hasattr(model_config, param_name):
+                load_kwargs[param_name] = param_value
+            else:
+                logger.warning(f"Parameter {param_name} is not supported by {config_obj.base_model}. Skipping.")
```
tgaddair, line 93: It seems like we're going to log this warning even if the user didn't specify a value for this param (because we're not checking that the `param_value is None`). I would suggest changing the logic to something like:

```
if param_value is None:
  continue

if hasattr(model_config, param_name):
  load_kwargs[param_name] = param_value
else:
  logger.warning(f"Parameter {param_name} is not supported by {config_obj.base_model}. Skipping.")
```

ludwig/schema/llms/model_parameters.py
```diff
@@ -0,0 +1,73 @@
...
+
+    This parameter is a dictionary containing the scaling configuration
+    for the RoPE embeddings. Currently supports three scaling strategies: linear and dynamic. Their
+    scaling factor must be an float greater than 1. The expected format is {'type': strategy name,
+    'factor': scaling factor}
+    """
+
+    def __post_init__(self):
+        if self.type and not self.factor:
+            raise ConfigValidationError(f"`rope_scaling`'s `factor` field must be an float > 1, got {self.factor}")
```
tgaddair, line 21: The error message seems incongruous with the condition. Shouldn't this be something like "if a type is provided, you must specify a factor" or something?

ludwig/schema/llms/model_parameters.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if self.factor and not self.type:
+            raise ConfigValidationError(
+                f"`rope_scaling`'s `type` field must be one of ['linear', 'dynamic'], got {self.type}"
+            )
+
+        if not isinstance(self.factor, float) or self.factor <= 1.0:
+            raise ConfigValidationError(f"`rope_scaling`'s `factor` field must be an float > 1, got {self.factor}")
```
tgaddair, line 29: Is this not handled by `FloatRange`?

ludwig/schema/llms/model_parameters.py
```diff
@@ -0,0 +1,73 @@
...
+
+    type: str = schema_utils.StringOptions(
```
tgaddair, line 31: Type should be `Optional[str]`

ludwig/schema/llms/model_parameters.py
```diff
@@ -0,0 +1,73 @@
...
+        options=["linear", "dynamic"],
+        default=None,
+        allow_none=True,
+        description="Currently supports two strategies: linear and dynamic scaling.",
+    )
+
+    factor: float = schema_utils.FloatRange(
```
tgaddair, line 38: Type should be `Optional[float]`

ludwig/schema/llms/model_parameters.py
```diff
@@ -0,0 +1,86 @@
...
+        super().__init__(RoPEScalingConfig, default_missing=True)
+
+    def _jsonschema_type_mapping(self):
+        return schema_utils.unload_jsonschema_from_marshmallow_class(RoPEScalingConfig, title="rope_scaling")
+
+
+@DeveloperAPI
+@ludwig_dataclass
+class ModelParametersConfig(schema_utils.BaseMarshmallowConfig):
+    rope_scaling: RoPEScalingConfig = RoPEScalingConfigField().get_default_field()
```
tgaddair, line 61: One thing I noticed about this: if we were to add another param to `model_parameters` and the user didn't specify `rope_scaling` it would raise an exception, which seems wrong. I think `rope_scaling` itself should default to `None` rather than a default. We should also add tests to verify these situations.

tgaddair (approved):

PR #3478, opened 2023-07-24
Enable evaluation with LLMs <7B

geoffreyangus:


ludwig/backend/ray.py
```diff
@@ -683,24 +691,40 @@ def get_resources_per_worker(self) -> Tuple[int, int]:
...
         output_columns = get_output_columns(self.model.output_features, include_logits=collect_logits)
+
+        num_cpus, num_gpus = self.get_resources_per_worker()
+
+        distributed_strategy = self.trainer_kwargs.get("strategy", get_default_strategy_name())
+        if distributed_strategy == "deepspeed" or (
+            isinstance(distributed_strategy, dict) and distributed_strategy.get("type", None) == "deepspeed"
+        ):
+            # make sure all gpus are used by a single Ray Datasets worker during batch predict
+            num_gpus = self.trainer_kwargs.get("num_workers", 1) * num_gpus
```
tgaddair, line 709: This doesn't quite work, as the Ray cluster could have multiple nodes, but the model can't be loaded across multiple nodes. Therefore, you want to compute the max number of gpus on a single node:

```
num_gpus = max(n["Resources"].get("GPU", 0) for n in ray.nodes())
```

ludwig/backend/ray.py
```diff
@@ -683,24 +691,40 @@ def get_resources_per_worker(self) -> Tuple[int, int]:
...
+    ):
         self._check_dataset(dataset)
 
         predictor_kwargs = self.predictor_kwargs
         output_columns = get_output_columns(self.model.output_features, include_logits=collect_logits)
+
+        num_cpus, num_gpus = self.get_resources_per_worker()
+
+        distributed_strategy = self.trainer_kwargs.get("strategy", get_default_strategy_name())
+        if distributed_strategy == "deepspeed" or (
```
tgaddair, line 705: Rather than do this for DeepSpeed specifically, it might make sense to do this for any LLM model type (we should also do it for LLMs as encoders, but at the moment they don't do model parallelism for inference, so it wouldn't work anyway).

ludwig/distributed/deepspeed.py
```diff
@@ -182,6 +187,17 @@ def create_checkpoint_handle(
     ) -> Checkpoint:
         return DeepSpeedCheckpoint(self, dist_model, optimizer, scheduler)
 
+    @classmethod
+    def extract_model_for_serialization(cls, model: nn.Module) -> Union[nn.Module, Tuple[nn.Module, List[Dict]]]:
```
tgaddair, line 191: Oops! That was definitely my mistake (in my defense, someone merged the PR before I had a chance to test it).

ludwig/api.py
```diff
@@ -1567,7 +1568,8 @@ def load(
         # Upgrades deprecated fields and adds new required fields in case the config loaded from disk is old.
         config_obj = ModelConfig.from_dict(config)
 
-        if backend_param is None and "backend" in config:
+        # Ensure that the original backend is used if it was specified in the config and user requests it
+        if use_backend_from_config or (backend_param is None and "backend" in config):
```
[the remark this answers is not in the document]
geoffreyangus: Yeah, so the issue comes from the evaluate CLI: https://github.com/ludwig-ai/ludwig/blob/adc82cc33c041f29480f734bf146a5d02b5e8641/ludwig/evaluate.py#L267-L274

The backend is initialized fresh when running `ludwig evaluate`. In doing this, the backend config is entirely ignored. This made it difficult to iterate quickly on this PR (my strategy was to run `ludwig train` once, then run `ludwig evaluate` to debug batch evaluation and prediction).

This is not a problem in Predibase engine code though, because backend is passed in as None and therefore the config is respected every time.

ludwig/backend/ray.py
geoffreyangus: got it!

ludwig/backend/ray.py
geoffreyangus: done

ludwig/api.py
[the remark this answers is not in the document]
tgaddair: Okay, going to push a quick change to fix the issue here, by not plumbing through the backend used in the CLI evaluate code path.

tgaddair (approved):
LGTM!

PR #3480, opened 2023-07-25
Add `ludwig upload` to push artifacts to HuggingFace Hub

arnavgarg1:
This PR adds a new command called `ludwig upload` to push trained model artifacts to Huggingface Hub. 

The API is very simple:

```
ludwig upload hf_hub --repo_id <repo_id> --model_path </path/to/saved/model>
```

When you run this command, it will check if you are already signed into Huggingface Hub. This can be done through `huggingface-cli login` in advance, otherwise it will make you follow the steps to sign in which looks something like this:

```
$ ludwig upload --service hf_hub -r arnavgrg/test -m /workspaces/trained_models/api_experiment_run

    _|    _|  _|    _|    _|_|_|    _|_|_|  _|_|_|  _|      _|    _|_|_|      _|_|_|_|    _|_|      _|_|_|  _|_|_|_|
    _|    _|  _|    _|  _|        _|          _|    _|_|    _|  _|            _|        _|    _|  _|        _|
    _|_|_|_|  _|    _|  _|  _|_|  _|  _|_|    _|    _|  _|  _|  _|  _|_|      _|_|_|    _|_|_|_|  _|        _|_|_|
    _|    _|  _|    _|  _|    _|  _|    _|    _|    _|    _|_|  _|    _|      _|        _|    _|  _|        _|
    _|    _|    _|_|      _|_|_|    _|_|_|  _|_|_|  _|      _|    _|_|_|      _|        _|    _|    _|_|_|  _|_|_|_|
    
    To login, `huggingface_hub` requires a token generated from https://huggingface.co/settings/tokens .
Token: 
Add token as git credential? (Y/n) Y
Token is valid (permission: write).
Your token has been saved in your configured git credential helpers (!f()).
Your token has been saved to /home/vscode/.cache/huggingface/token
Login successful
```

You can find your token at `https://huggingface.co/settings/tokens`.

Once you're signed in, this will create a new public model repository on Huggingface Hub called `repo_id` if it doesn't exist, and push the trained model weights and associated configs in `model_path` to `repo_id`. 

There are some other optional parameters you can pass to `ludwig upload`:

1. `private`: Whether to create a private repository or a public repository.
2.  `commit_message`: The summary / title / first line of the generated commit.
3. `commit_description`: The description of the generated commit.

### DIFF
--- a/ludwig/cli.py
+++ b/ludwig/cli.py
@@ -56,6 +56,7 @@ Available sub-commands:
    init_config           Initialize a user config from a dataset and targets
    render_config         Renders the fully populated config with all defaults set
    check_install         Runs a quick training run on synthetic data to verify installation status
+   commit                Push trained model artifacts to Huggingface Hub
 """,
         )
         parser.add_argument("command", help="Subcommand to run")
@@ -178,6 +179,11 @@ Available sub-commands:
         from ludwig import datasets
 
         datasets.cli(sys.argv[2:])
+
+    def commit(self):
+        from ludwig import commit
+
+        commit.cli(sys.argv[2:])
 
 
 def main():
--- a/ludwig/cli.py
+++ b/ludwig/cli.py
@@ -56,7 +56,7 @@ Available sub-commands:
    init_config           Initialize a user config from a dataset and targets
    render_config         Renders the fully populated config with all defaults set
    check_install         Runs a quick training run on synthetic data to verify installation status
-   commit                Push trained model artifacts to Huggingface Hub
+   upload                Push trained model artifacts to Huggingface Hub
 """,
         )
         parser.add_argument("command", help="Subcommand to run")
@@ -180,10 +180,10 @@ Available sub-commands:
 
         datasets.cli(sys.argv[2:])
 
-    def commit(self):
-        from ludwig import commit
+    def upload(self):
+        from ludwig import upload
 
-        commit.cli(sys.argv[2:])
+        upload.cli(sys.argv[2:])
 
 
 def main():
--- a/ludwig/cli.py
+++ b/ludwig/cli.py
@@ -56,7 +56,7 @@ Available sub-commands:
    init_config           Initialize a user config from a dataset and targets
    render_config         Renders the fully populated config with all defaults set
    check_install         Runs a quick training run on synthetic data to verify installation status
-   upload                Push trained model artifacts to Huggingface Hub
+   upload                Push trained model artifacts to a registry (e.g., HuggingFace Hub)
 """,
         )
         parser.add_argument("command", help="Subcommand to run")
--- a/ludwig/commit.py
+++ b/ludwig/commit.py
@@ -63,14 +63,14 @@ def cli(sys_argv):
     # ---------------
     # Required parameters
     # ---------------
-    parser.add_argument("-m", "--model_path", help="model to push to huggingface", required=True)
-
     parser.add_argument(
         "-r",
         "--repo_id",
         help="ID of the repo on HF. This will be created if it doesn't exist. " "Format: username/repo_name",
         required=True,
     )
+
+    parser.add_argument("-m", "--model_path", help="model to push to huggingface", required=True)
 
     # ---------------
     # Optional parameters
--- a/ludwig/commit.py
+++ b/ludwig/commit.py
@@ -1,0 +1,112 @@
+import argparse
+import logging
+import sys
+from typing import Optional
+
+from ludwig.utils.commit_utils import HuggingFaceHub
+from ludwig.utils.print_utils import get_logging_level_registry
+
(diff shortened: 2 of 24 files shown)


ludwig/commit.py
```diff
@@ -0,0 +1,112 @@
...
+            Set to `"dataset"` or `"space"` if uploading to a dataset or
+            space, `None` or `"model"` if uploading to a model. Default is
+            `None`.
+        commit_message (`str`, *optional*):
+            The summary / title / first line of the generated commit. Defaults to:
+            `f"Upload {path_in_repo} with huggingface_hub"`
+        commit_description (`str` *optional*):
+            The description of the generated commit
+    """
+
```
skanjila, line 43: I am wondering if we should allow to publish to our internal ludwig model-hub as well

ludwig/utils/commit_utils.py
```diff
@@ -0,0 +1,109 @@
...
+
+        with open(cached_token_path) as f:
+            hf_token = f.read()
+
+        hf_api = HfApi(token=hf_token)
+        assert hf_api.token == hf_token
+
+        self.api = hf_api
+
+    def upload_to_hfhub(
```
skanjila, line 30: one idea here might be to have a flag that additionally allows to publish to our model-hub or maybe we should discuss how the two should work together

ludwig/utils/commit_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        # Create empty model repo using repo_id, but it is okay if it already exists.
+        self.api.create_repo(
+            repo_id=repo_id,
+            private=private,
+            repo_type=repo_type,
+            exist_ok=True,
+        )
+
+        # Upload all artifacts in model weights folder
+        upload_path = self.api.upload_folder(
```
skanjila, line 101: should we return a status from this method indicating the different levels of success

ludwig/commit.py
arnavgarg1: We can!

ludwig/utils/commit_utils.py
arnavgarg1: Agreed. We can define a base class and expand the arguments to be:

1. `ludwig commit hf_hub ...`
2. `ludwig commit ludwig_hub ...`

ludwig/utils/commit_utils.py
arnavgarg1: Yeah we can do that, let me follow up

ludwig/cli.py
```diff
@@ -56,6 +56,7 @@ def __init__(self):
    init_config           Initialize a user config from a dataset and targets
    render_config         Renders the fully populated config with all defaults set
    check_install         Runs a quick training run on synthetic data to verify installation status
+   commit                Push trained model artifacts to Huggingface Hub
```
tgaddair, line 59: Maybe to avoid a backward compatibility issue in the future, we can go ahead and make this either:

`ludwig commit_hf`

OR

`ludwig commit hf_hub`

Something like that. What do you think?

ludwig/utils/commit_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        trained_model_artifacts_path = os.path.join(model_path, "model", "model_weights")
+        if not os.path.exists(trained_model_artifacts_path):
+            raise Exception(
+                f"Model artifacts not found at {trained_model_artifacts_path}. "
+                f"It is possible that model at '{model_path}' hasn't been trained yet, or something went"
+                "wrong during training where the model's weights were not saved."
+            )
+
+        # Make sure the model's saved artifacts either contain:
+        # 1. pytorch_model.bin -> regular model training, such as ECD or for LLMs
```
tgaddair, line 82: Does HF support non-HF models being uploaded to the hub?

ludwig/commit.py
```diff
@@ -0,0 +1,112 @@
...
+    hf = HuggingFaceHub()
```
tgaddair, line 44: Can you make an API version of this as well that does the same thing?

```
model = LudwigModel(...)
model.upload_to_hf(...)
```

ludwig/cli.py
tgaddair: See top-level comment, might make sense to change `commit` to `upload`.

tgaddair (commented):
This is awesome!

I'm kind of torn on the naming. I think `commit` is somewhat counterintuitive to me. The HF Hub API itself uses the terminology "upload" which feels more clear. Maybe we can do that instead? I also think making it clear we're uploading to HF (as opposed to Predibase or MLFlow or whatever) is important.

So to that end I would recommend the following:

```
ludwig upload hf_hub ...
```

What do you think?

tgaddair (approved):

PR #3486, opened 2023-07-29
Fixed QLoRA with mutli-gpu and reduce CUDA memory pressure during eval

tgaddair:


arnavgarg1 (approved):
LGTM since we're only doing device_map=auto with quantisation

PR #3487, opened 2023-07-30
Added quantization parameter metadata

tgaddair:


ludwig/schema/metadata/configs/llm.yaml
```diff
@@ -119,3 +119,17 @@ adapter:
     adapter_layers:
       ui_display_name: Adapter Layers
       expected_impact: 3
+quantization:
+  _oneOf:
+    allOf:
```
ksbrar, line 124: ultimately these are just meta properties that you can name whatever you want, but strictly speaking there is no `allOf` schema for `quantization` - it's just that the entire thing is disabled or you specify the full object.

ludwig/schema/llms/quantization.py
```diff
@@ -20,6 +21,7 @@ class QuantizationConfig(schema_utils.BaseMarshmallowConfig):
         options=[4, 8],
         default=4,
         description="The quantization level to apply to weights on load.",
+        parameter_metadata=LLM_METADATA["quantization"]["bits"],
```
ksbrar, line 24: will you add more for the other parameters?

ludwig/schema/metadata/configs/llm.yaml
```diff
@@ -119,3 +119,17 @@ adapter:
...
+      ui_display_name: Quantization
+      expected_impact: 3
+    none:
+      ui_display_name: No Quantization
+      expected_impact: 3
+  _meta:
```
ksbrar, line 130: nice, so this will be the "top-level" metadata for the `jsonschema_type_mapping`

ludwig/schema/metadata/configs/llm.yaml
```diff
@@ -119,3 +119,17 @@ adapter:
...
+    expected_impact: 3
+    ui_component_type: radio_string_combined
```
ksbrar, line 132: this is the same component type as `base_model`. Just confirming that you want that?

ludwig/schema/llms/quantization.py
tgaddair: The others are low impact, so wanted to leave them as-is for now.

ludwig/schema/metadata/configs/llm.yaml
tgaddair: Right, that's my understanding.

ludwig/schema/metadata/configs/llm.yaml
tgaddair: I think so? Was going off the `adapter` meta, which should work more or less the same as this.

ludwig/schema/metadata/configs/llm.yaml
tgaddair: Interesting, was just going off the pattern for `adapter`. What should it be changed to?

ludwig/schema/llms/quantization.py
ksbrar: ah gotcha

ludwig/schema/metadata/configs/llm.yaml
ksbrar: It's just a question of how you would like it to render. `adapter` and `base_model` render as a set of radio buttons where each button's label is the title of each option and when you select one the nested properties of that option are rendered below it.

Without `radio_string_combined`, you would see a dropdown in place of the radio buttons (because the dropdown is the default way a oneof choice is rendered)

tgaddair: Gotcha, yeah my preference would be a radio button for Quantization / No Quantization, then a dropdown for the bits field if Quantization is selected. Is that what the current behavior would be?

ludwig/schema/metadata/configs/llm.yaml
ksbrar: So then make sure to add it to the `jsonschema_type_mapping` here:
```
{
  "oneOf": {
    ...
  },
  "title": "...",
  "description": "...",
  "parameter_metadata": ${PUT HERE}
}
```

ludwig/schema/metadata/configs/llm.yaml
ksbrar: I would call it "enabled" or "object" or "non-null" or something like that. These middle-management words don't semantically *mean* anything - they're just a reference for the coder to be able to tell a schema where to pull the parameter metadata from.

So you haven't done this yet, but if I looked at the code for `jsonschema_type_mapping` and saw this:
```
{
  "oneOf": {
    {
      "title": "enabled"
      ...
      "parameter_metadata": LLM_METADATA["quantization"]["_oneOf"]["allOf"]
    },
    ...
  }
  "title": "quantization",
  ...
  "parameter_metadata":  LLM_METADATA["quantization"]["_meta"]
}
```

I would get slightly confused - I would think to myself "allOf? I don't see an allOf in this code".

ludwig/schema/metadata/configs/llm.yaml
ksbrar: yup, that would be the behavior with `radio_string_combined`

PR #3489, opened 2023-07-31
[fix] Multiple test fixes

jeffkinnison:
- ~~Recently, `ffmpeg` and `sox` integrations were removed from torchaudio nightly, which led to errors with audio file IO. This update installs `ffmpeg` and `sox` into the CI env to account for that change.~~ `ffmpeg` causes a segfault in the ubuntu container (could not repro locally), and `sox` isn't recognized by torchaudio. Pinning to a working torchaudio nightly version for the time being to unblock.
- `llama-7b` was removed from the base model list but remained a parameter of `test_llm_base_model_config`, leading to an error. `vicuna-7b` is now used in its place.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -79,11 +79,13 @@ jobs:
         if: runner.os == 'linux'
         run: |
           sudo apt-get install -y cmake libsndfile1
+          sudo apt-get install -y ffmpeg
 
       - name: Setup macOS
         if: runner.os == 'macOS'
         run: |
           brew install libuv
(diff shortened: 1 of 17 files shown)


tgaddair (approved):
Can you add this to the Docker images as well?

jeffkinnison:
Installing ffmpeg seemed to cause a segfault, so I've removed it for now to test with sox instead.

jeffkinnison:
@tgaddair I added a quick fix for the "llama-7b" bug that started showing up this evening.

PR #3493, opened 2023-08-02
Require text output feature for LLM finetuning

arnavgarg1:


ludwig/config_validation/checks.py
```diff
@@ -472,6 +472,22 @@ def check_llm_atleast_one_input_text_feature(config: "ModelConfig"):  # noqa: F8
...
+@register_config_check
+def check_llm_finetuning_output_feature_config(config: "ModelConfig"):  # noqa: F821
+    """Checks that the output feature config for LLM finetuning is valid."""
+    if config.model_type != MODEL_LLM:
+        return
+
+    if config.trainer.type != "finetune":
+        return
+
+    if config.output_features[0].type != TEXT:
```
connor-mccorm, line 484: Couple sanity checks here:
-  We only allow one output feature for LLMs currently right?
- I assume there's another validation check for this, but should we check to make sure `config.output_features` is not an empty list?

connor-mccorm (approved):
Just a thought, in the future, I feel like it may make sense to do some sort of casting behind the scenes for the user when they select category for a fine-tuning job since it seems a little clunky to be restricted to text, even though the task may be binary or multi-class classification.

Aside from that thought, LGTM!

arnavgarg1:
> Just a thought, in the future, I feel like it may make sense to do some sort of casting behind the scenes for the user when they select category for a fine-tuning job since it seems a little clunky to be restricted to text, even though the task may be binary or multi-class classification.

Hmm, so the way fine-tuning works is that a lot of these LLMs are encoder-decoder models, so we're just updating weights within the model architecture itself (either through full fine-tuning, or by inserting new layers through a PEFT adapter), which is why the input and output pairs need to be text-based because that's how it produces logits. However, if the LLM was an encoder-only LLM, then yeah in theory we could totally support category/binary output feature types because we can stack on a few layers (like a classification head or regression head) that could be fine-tuned. As it stands today, I think just doing text-to-text even for binary of multi-class classification by using a text-to-text model should be totally fine, and it is the recommended approach in the LLM community from what I know.

ludwig/config_validation/checks.py
arnavgarg1: > We only allow one output feature for LLMs currently right?

Technically yes, but I actually don't think we have any explicit checks for this today. Let me add another one for this particular scenario in another PR.

> I assume there's another validation check for this, but should we check to make sure config.output_features is not an empty list?

I would hope that we have a check for this already, or that normal config validation through marshmallow would fail if we got an empty list for output features. I'm quite sure that is already the case so wouldn't worry about this one too much.

arnavgarg1: Okay I just tried an LLM model type with 1 input feature and 2 output features and this is the error I got:

```
E           ludwig.error.ConfigValidationError: Failed to validate JSON schema for config. Error: [{'name': 'text_output', 'type': 'text'}, {'name': 'category_output', 'type': 'category'}] is too long
```

I attribute this to this check in the LLm model type schema: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/schema/features/base.py#L320

So seems like we don't have to take care of this, but I will say that the error message is non-obvious/quite cryptic.

tgaddair (approved):

arnavgarg1:
> Basically, we would still treat it as a text-to-text problem under the hood, but just where the loss is computed on exact match to the generated text. Does that make sense?

It makes sense to me, but atm it's not supported with the current implementation so we would need to build that in. Should be fairly straightforward to do if this is something we want to support

PR #3495, opened 2023-08-02
Remove falcon from presets as it requires running untrusted code

tgaddair:


connor-mccorm (approved):

arnavgarg1:
I know the current recommendation is to add a flag to make this work - is that not something we can do?

arnavgarg1:
Ycombinator news: https://news.ycombinator.com/item?id=36612627

I do believe this flag will be removed in the next transformers version

arnavgarg1 (approved):

arnavgarg1:
Another idea:

Check if the model architecture is RWForCausalLM and then conditionally add the trust-remote_code=True flag to the load kwargs. That way it'll only be supported for Falcon models. 

https://huggingface.co/tiiuae/falcon-40b/blob/main/config.json

PR #3498, opened 2023-08-04
Add use_pretrained attribute for AutoTransformers

arnavgarg1:
Fixes the following error:

```
'AutoTransformerConfig' object has no attribute 'use_pretrained'
```

when trying to train an custom transformer model from HF using a config that looks like this:

```
encoder:
      type: auto_transformer
      trainable: false
      pretrained_model_name_or_path: huggyllama/llama-7b
    preprocessing:
      tokenizer: space_punct
      max_sequence_length: null
```

### DIFF
--- a/ludwig/schema/encoders/text_encoders.py
+++ b/ludwig/schema/encoders/text_encoders.py
@@ -3092,6 +3092,13 @@ class AutoTransformerConfig(HFEncoderConfig):
         description=ENCODER_METADATA["AutoTransformer"]["type"].long_description,
     )
 
+    use_pretrained: bool = schema_utils.Boolean(
+        default=True,
+        description="Whether to use the pretrained weights for the model. If false, the model will train from "
+        "scratch which is very computationally expensive.",
+        parameter_metadata=ENCODER_METADATA["AutoTransformer"]["use_pretrained"],
+    )
+
     pretrained_model_name_or_path: str = schema_utils.String(
         default=None,
         allow_none=True,
--- a/ludwig/schema/encoders/text_encoders.py
+++ b/ludwig/schema/encoders/text_encoders.py
@@ -3092,12 +3092,9 @@ class AutoTransformerConfig(HFEncoderConfig):
         description=ENCODER_METADATA["AutoTransformer"]["type"].long_description,
     )
 
-    use_pretrained: bool = schema_utils.Boolean(
-        default=True,
-        description="Whether to use the pretrained weights for the model. If false, the model will train from "
-        "scratch which is very computationally expensive.",
-        parameter_metadata=ENCODER_METADATA["AutoTransformer"]["use_pretrained"],
-    )
+    # Always set this to True since we always want to use the pretrained weights
+    # We don't currently support training from scratch for AutoTransformers
+    use_pretrained: bool = True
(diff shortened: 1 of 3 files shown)


arnavgarg1:
Will add a test

arnavgarg1:
Okay this is probably wrong, going to close and discuss this first.

ludwig/schema/encoders/text_encoders.py
```diff
@@ -2574,7 +2574,7 @@ def module_name():
     )
 
     use_pretrained: bool = schema_utils.Boolean(
-        default=False,
+        default=True,
         description="Whether to use the pretrained weights for the model. If false, the model will train from "
         "scratch which is very computationally expensive.",
         parameter_metadata=ENCODER_METADATA["FlauBERT"]["use_pretrained"],
```
arnavgarg1, line 2580: We should always set use_pretrained to True by default right?

tgaddair: Good catch.

ludwig/schema/encoders/text_encoders.py
```diff
@@ -3092,6 +3097,10 @@ def module_name():
         description=ENCODER_METADATA["AutoTransformer"]["type"].long_description,
     )
 
+    # Always set this to True since we always want to use the pretrained weights
+    # We don't currently support training from scratch for AutoTransformers
+    use_pretrained: bool = True
```
tgaddair, line 3102: Let's make this a property so the user could never modify it.

tgaddair:
```
@property
def use_pretrained(self) -> bool:
    return True
```

tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -292,3 +289,42 @@ def test_tfidf_encoder(vocab_size: int):
     inputs = torch.randint(2, (batch_size, sequence_length)).to(DEVICE)
     outputs = text_encoder(inputs)
     assert outputs[ENCODER_OUTPUT].shape[1:] == text_encoder.output_shape
+
+
+def test_hf_auto_transformer_use_pretrained(tmpdir, csv_filename):
```
jeffkinnison, line 294: nit: The `True` case would be tested implicitly elsewhere, correct? If not, maybe we could parametrize the test with both cases.

jeffkinnison (approved):

tests/ludwig/encoders/test_text_encoders.py
```diff
@@ -292,3 +289,23 @@ def test_tfidf_encoder(vocab_size: int):
...
+
+def test_hf_auto_transformer_use_pretrained():
+    """This test ensures that use_pretrained is always True when using the auto_transformer text encoder even if a
+    user explicitly sets it to False."""
+    config = {
+        "input_features": [
+            text_feature(
+                encoder={
+                    "type": "auto_transformer",
+                    "use_pretrained": False,
```
tgaddair, line 302: Ideally this should be an error if we were more strict with our config validation rules. We should instead just leave this out of the config.

tgaddair (approved):

PR #3504, opened 2023-08-07
Return sentences instead of individual tokens for text features during inference

arnavgarg1:
Some questions:

1. Should we skip probabilities during LLM inference?
2. Should we only do this for LLMs?


tgaddair (approved):

ludwig/features/text_feature.py
```diff
@@ -337,11 +337,13 @@ def postprocess_predictions(
 
             def idx2str(pred):
                 if tokenizer is None:
-                    return [
-                        metadata["idx2str"][token] if token < len(metadata["idx2str"]) else UNKNOWN_SYMBOL
-                        for token in pred
-                    ]
-                return tokenizer.tokenizer.batch_decode(pred, skip_special_tokens=True)
+                    return " ".join(
```
justinxzhao, line 340: What do you think about having a separate key? The result dictionary would contain both the token list and the concatenated result.

Ideas for key names:
- `f"{feature_name}_concat"`
- `f"{feature_name}_concatenated"`
- `f"{feature_name}_expressed"`
- `f"{feature_name}_merged"`
- `f"{feature_name}"`

arnavgarg1: Nice, this is what I've decided to try and implement! Good suggestion

PR #3517, opened 2023-08-10
[feat] Support for numeric date feature inputs

jeffkinnison:
Currently, the date feature supports ISO date/time strings and custom date/time strings with a user-supplied format string. This adds support for preprocessing int and float second and millisecond POSIX timestamps.


justinxzhao (approved):
Thanks!

PR #3518, opened 2023-08-10
Set default local backend when doing quantization

Infernaught:
Previous: set default backend to ray if ray was installed

Now: set default backend to local if quantization is being used (quantization is set in the config)

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -316,7 +316,10 @@ class LudwigModel:
         self.set_logging_level(logging_level)
 
         # setup Backend
-        self.backend = initialize_backend(backend or self._user_config.get("backend"))
+        if self._user_config.get("quantization"):
+            self.backend = initialize_backend("local")
+        else:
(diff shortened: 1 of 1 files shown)


arnavgarg1:
As discussed offline, we want to move this down into a config validation check since this issue only happens when Ray is installed on the system, in which case we default to Ray backend instead of Local backend. 

The config validation should ensure that when quantisation is set, backend is not empty + that it must be set explicitly to local (double check that the second part of this doesn't already exist).

arnavgarg1 (changes requested):
As discussed offline, we want to move this down into a config validation check since this issue only happens when Ray is installed on the system, in which case we default to Ray backend instead of Local backend.

The config validation should ensure that when quantisation is set, backend is not empty + that it must be set explicitly to local (double check that the second part of this doesn't already exist).

arnavgarg1:
Closing in favor of https://github.com/ludwig-ai/ludwig/pull/3524

PR #3519, opened 2023-08-10
Move loss metric to same device as inputs

Infernaught:
Previously: evaluation loss metric could be on a different device than the inputs in a multi-GPU setup

Now: evaluation loss metric is moved to the same device as the inputs


arnavgarg1 (changes requested):
As discussed, the right thing to do here is to add a config validation check to ensure that the user explicitly sets the backend to local when using quantisation

PR #3520, opened 2023-08-11
Add mechanic to override default values for generation during model.predict()

justinxzhao:
Sample code:

```python
model = LudwigModel(config=config, logging_level=logging.INFO)
results = model.train(training_set=dataset)
model.predict(dataset=dataset[:1], generation_config={"max_new_tokens": 513})
```

I've added additional logging statements to make it easier for testers/users to verify behavior.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -831,6 +831,7 @@ class LudwigModel:
         data_format: str = None,
         split: str = FULL,
         batch_size: int = 128,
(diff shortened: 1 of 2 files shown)


arnavgarg1:
LGTM, thanks for adding this!

As discussed offline, if the logging gets noisy, we can just change the logging.info to logging.debug

arnavgarg1 (approved):

PR #3522, opened 2023-08-11
Move loss metric to same device as inputs

Infernaught:
Previously: evaluation loss metric could be on a different device than inputs

Now: evaluation loss metric moved to same device as inputs

### DIFF
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -467,7 +467,11 @@ class LLM(BaseModel):
                 continue
             of_obj.update_metrics(targets[of_name], predictions[of_name])
 
+        # HACK (Tim): get the device of the targets to transfer self.eval_loss_metric to the same device
+        target_device = list(targets.values())[0].device
+
         eval_loss, additional_losses = self.eval_loss(targets, predictions)
+        self.eval_loss_metric = self.eval_loss_metric.to(target_device)
         self.eval_loss_metric.update(eval_loss)
         self.eval_additional_losses_metrics.update(additional_losses)
 


ludwig/models/llm.py
```diff
@@ -467,7 +467,11 @@ def update_metrics(self, targets, predictions):
                 continue
             of_obj.update_metrics(targets[of_name], predictions[of_name])
 
+        # HACK (Tim): get the device of the targets to transfer self.eval_loss_metric to the same device
+        target_device = list(targets.values())[0].device
+
         eval_loss, additional_losses = self.eval_loss(targets, predictions)
+        self.eval_loss_metric = self.eval_loss_metric.to(target_device)
```
arnavgarg1, line 474: Nice catch! Could you also do this for `self.eval_additional_losses_metrics`? I assume that may also require being moved to the right device (once we support multi-feature outputs for LLM model types, but might as well do it now)

arnavgarg1 (approved):

ludwig/models/llm.py
```diff
@@ -467,7 +467,11 @@ def update_metrics(self, targets, predictions):
                 continue
             of_obj.update_metrics(targets[of_name], predictions[of_name])
 
+        # HACK (Tim): get the device of the targets to transfer self.eval_loss_metric to the same device
+        target_device = list(targets.values())[0].device
```
arnavgarg1, line 471: OOC, does `targets[of_name].device` not work?

arnavgarg1 (approved):

PR #3523, opened 2023-08-11
Add test that verifies that the generation config passed in at model.predict() is used correctly.

justinxzhao:


tests/integration_tests/test_llm.py
```diff
@@ -46,6 +47,11 @@
 }
 
 
+def get_num_non_empty_tokens(iterable):
+    """Returns the number of non-empty tokens."""
+    return len(list(filter(bool, iterable)))
```
arnavgarg1, line 52: slick

tests/integration_tests/test_llm.py
```diff
@@ -134,6 +140,20 @@ def test_llm_text_to_text(tmpdir, backend, ray_cluster_4cpu):
     assert preds["Answer_probabilities"]
     assert preds["Answer_probability"]
 
+    # Check that in-line generation parameters are used. Original prediction uses max_new_tokens = 5.
+    assert get_num_non_empty_tokens(preds["Answer_predictions"][0]) > 3
```
arnavgarg1, line 144: One more way to guarantee this actually, is to use `min_new_tokens`: https://huggingface.co/docs/transformers/main/main_classes/text_generation#transformers.GenerationConfig.min_new_tokens in the base generation config, that way we are not left to determinism. Wdyt?

arnavgarg1 (approved):

tests/integration_tests/test_llm.py
justinxzhao: Good idea! Done.

PR #3524, opened 2023-08-11
Ensure user sets backend to local w/ quantization

Infernaught:
Previously: users could omit backend from their config when running LLM finetuning with quantization

Now: users are prompted to set backend to local explicitly when running LLM finetuning with quantization


justinxzhao:
The change LGTM. It looks like there's some failing test. @Infernaught can you take a look?

arnavgarg1 (approved):
LGTM

tgaddair:
This is very aggressive check that is already causing user-facing issues (#3529). 

I understand the motivation for raising an error if the user selects quantization and the Ray backend, but that's not what this is doing. This is raising an error if the user doesn't EXPLCITLY select the local backend, which is quite extreme.

We should NEVER require user to specify the local backend explicitly if the local backend is going to be used anyway. Going to need to revert this and yank the v0.8.1 release.

PR #3533, opened 2023-08-15
Add `effective_batch_size` to auto-adjust gradient accumulation

tgaddair:
For LLM fine-tuning, it's often the case that the batch size per GPU (`batch_size` in Ludwig) is very small (1 or 2), but the ideal batch size for model convergence is 32 or so. In these cases, we want to use gradient accumulation to compensate for the low batch size. Additionally, adding training workers with DeepSpeed further increases the effective batch size, meaning the user needs to do a lot of quick math to figure out how to set the `gradient_accumulation_steps`.

This PR adds a new trainer param called `effective_batch_size` that auto-adjusts gradient accumulation based on this value.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -103,7 +103,7 @@ from ludwig.utils.misc_utils import (
 )
 from ludwig.utils.print_utils import print_boxed
 from ludwig.utils.torch_utils import DEVICE
-from ludwig.utils.trainer_utils import get_training_report
+from ludwig.utils.trainer_utils import get_rendered_batch_size_grad_accum, get_training_report
 from ludwig.utils.types import DataFrame, TorchDevice
 
 logger = logging.getLogger(__name__)
@@ -796,6 +796,12 @@ class LudwigModel:
         self.model = self._online_trainer.train_online(training_dataset)
 
     def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_seed):
+        # Render the batch size and gradient accumulation steps prior to batch size tuning. This is needed in the event
+        # the effective_batch_size and gradient_accumulation_steps are set explicitly, but batch_size is AUTO. In this
+        # case, we can infer the batch_size directly without tuning.
+        self.config_obj.trainer.batch_size, self.config_obj.trainer.gradient_accumulation_steps = \
+            get_rendered_batch_size_grad_accum(self.config_obj.trainer, self.backend)
+
         # TODO (ASN): add support for substitute_with_max parameter
         # TODO(travis): detect train and eval batch sizes separately (enable / disable gradients)
         if (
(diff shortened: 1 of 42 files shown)


ludwig/schema/trainer.py
```diff
@@ -159,8 +186,27 @@ def __post_init__(self):
...
+        allow_none=False,
+        description=(
+            "The effective batch size is the total number of samples used to compute a single gradient update "
+            "to the model weights. This differs from `batch_size` by taking `gradient_accumulation_steps` and number "
+            "of training worker processes into account. In practice, "
+            "`effective_batch_size = batch_size * gradient_accumulation_steps * num_workers`. "
+            "If 'auto', the effective batch size is derivied implicitly from `batch_size`, but if set explicitly, then "
+            "one of `batch_size` or `gradient_accumulation_steps` must be set to something other than 'auto', and "
+            "consequently will be set following the formula given above."
+        ),
```
arnavgarg1, line 200: Nice, this is a very clear description of how this works

ludwig/api.py
```diff
@@ -795,35 +791,60 @@ def train_online(
...
-            ):
-                self._tune_batch_size(self._online_trainer, dataset, random_seed=random_seed)
+            self._tune_batch_size(self._online_trainer, dataset, random_seed=random_seed)
 
         self.model = self._online_trainer.train_online(training_dataset)
 
     def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_seed):
+        if not self.config_obj.trainer.can_tune_batch_size():
+            # Models like GBMs don't have batch sizes to be tuned
+            return
```
arnavgarg1, line 801: Is it worth logging a message here to indicate this very clearly?

arnavgarg1 (commented):
LGTM! This is an awesome change.

One thing I wanted to confirm is, what is the exact behavior when batch_size, effective_batch_size and gradient_accumulation_steps are set to `auto`? Is it that we don't update the batch_size and grad_accumulation_steps when we make the first call to `self.config_obj.trainer.update_batch_size_grad_accum(num_workers)`, then actually perform batch size tuning if batch size tuning is auto to set it to an actual number, and now that we have workers + batch size, we recompute batch size (which doesn't change) but update gradient accumulation steps to 1 while leaving effective batch size to auto? but in the case that effective batch size is not auto, we calculate it because we have batch_size, num workers and the desired effective batch size?

ludwig/api.py
tgaddair: This code path always gets executed regardless of user config, so I wouldn't add a message. Would likely confuse the user.

tgaddair:
@arnavgarg1 if everything is set to `auto` (the default), then it's equivalent to the current behavior where we just set the batch size to maximize GPU utilization and then set gradient accumulation to 1.

PR #3539, opened 2023-08-17
Refactor evaluation metrics to support decoded generated text metrics like BLEU and ROUGE.

justinxzhao:
## Adding new decoded text metrics

New evaluation metrics for text output features can be added to the metrics registry under the `RESPONSE` keyword, like so:

```python
@register_metric("bleu", [TEXT], MAXIMIZE, RESPONSE)
class BLEUScoreMetric(BLEUScore, LudwigMetric):
    def __init__(self, **kwargs):
        super().__init__()
```

New metrics added:

- BLEU
- ROUGE
- WordErrorRate

## Changes to TextOutputFeature.update_metrics()

Since we need to pass in additional decoded inputs to text features, we explicitly define `update_metrics()` for `TextOutputFeature` instead of relying on the base `OutputFeature` implementation. The decoded texts are optional arguments and if not provided, decoded text metrics are skipped.

```python
def update_metrics(
        self,
        targets: Tensor,
        predictions: Dict[str, Tensor],
        decoded_targets: Optional[List[str]] = None,
        decoded_predictions: Optional[List[str]] = None,
    ) -> None:
        """Updates metrics with the given targets and predictions.

        If decoded_targets and decoded_predictions are provided, as through LLM model types, then additional
        response-based metrics like BLEU and ROUGE are also computed.

        Args:
            targets: Tensor with target values for this output feature.
            predictions: Dict of tensors returned by predictions().
        """
        for metric_name, metric_fn in self._metric_functions.items():
            prediction_key = get_metric_tensor_input(metric_name)
            if prediction_key != RESPONSE:
                # Non-RESPONSE metrics don't use decoded texts.
                metric_fn = metric_fn.to(predictions[prediction_key].device)
                metric_fn.update(predictions[prediction_key].detach(), targets)
                continue

            if decoded_targets is not None and decoded_predictions is not None:
                # RESPONSE metrics cannot be computed if decoded texts aren't provided.
                # We assume that RESPONSE metrics can only be calculated one example at a time.
                for i in range(len(decoded_predictions)):
                    metric_fn.update(decoded_predictions[i], decoded_targets[i])
```

Some torchmetrics text metrics return a dictionary of sub-metrics. These are individually unpacked in `model. get_metrics()`.

## Changes to the LLM class

The LLM class sets has two separate methods for evaluation metrics: 1) `update_metrics` for ZS/FS and 2) `update_metrics_finetune_llm()` for fine-tuning LLMs. These appear to have slightly different treatments for aligning tensors and handling pad tokens.

Once predictions and targets are computed, they are decoded and passed to `TextOutputFeature.update_metrics()`.

## Refactoring the `MetricsPrintedTable`

This PR also replaces the MetricsPrintedTable with a simpler, stateless method. Instead of:

```python
printed_table = MetricsPrintedTable(output_features)
printed_table.add_metrics_to_printed_table(train_metrics_log, TRAIN)
printed_table.add_metrics_to_printed_table(validation_metrics_log, VALIDATION)
printed_table.add_metrics_to_printed_table(test_metrics_log, TEST)
printed_table.log_info()
```

We consolidate printed metrics into one call:

```python
print_metrics_table(
    output_features,
    progress_tracker.train_metrics,
    progress_tracker.validation_metrics,
    progress_tracker.test_metrics,
)
```

I've also changed the metrics table to print out metrics in a transposed way with splits as columns. This looks substantially better with many additional metrics.

```
╒═══════════════════════╤════════════╤══════════════╤════════╕
│                       │      train │ validation   │ test   │
╞═══════════════════════╪════════════╪══════════════╪════════╡
│ bleu                  │     0.0000 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ char_error_rate       │     0.5089 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ loss                  │     2.3335 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ next_token_perplexity │ 30315.8379 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ perplexity            │ 37736.5156 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rouge1_fmeasure       │     0.5648 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rouge1_precision      │     0.4919 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rouge1_recall         │     0.6630 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rouge2_fmeasure       │     0.2991 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rouge2_precision      │     0.2602 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rouge2_recall         │     0.3516 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rougeL_fmeasure       │     0.4167 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rougeL_precision      │     0.3629 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rougeL_recall         │     0.4891 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rougeLsum_fmeasure    │     0.5093 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rougeLsum_precision   │     0.4435 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ rougeLsum_recall      │     0.5978 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ sequence_accuracy     │     0.0000 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ token_accuracy        │     0.2716 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ word_error_rate       │     1.1882 │              │        │
├───────────────────────┼────────────┼──────────────┼────────┤
│ combined_loss         │     2.3335 │              │        │
╘═══════════════════════╧════════════╧══════════════╧════════╛
```

Combined loss is also merged into this table as the last column instead of a separate table, which has frequently confused users.

```
╒═══════════════╤═════════╤══════════════╤════════╕
│               │   train │   validation │   test │
╞═══════════════╪═════════╪══════════════╪════════╡
│ accuracy      │  0.8157 │       0.6966 │ 0.8090 │
├───────────────┼─────────┼──────────────┼────────┤
│ loss          │  0.4619 │       0.5039 │ 0.4488 │
├───────────────┼─────────┼──────────────┼────────┤
│ precision     │  0.8274 │       0.6250 │ 0.7818 │
├───────────────┼─────────┼──────────────┼────────┤
│ recall        │  0.6680 │       0.4545 │ 0.6615 │
├───────────────┼─────────┼──────────────┼────────┤
│ roc_auc       │  0.8471 │       0.7706 │ 0.8592 │
├───────────────┼─────────┼──────────────┼────────┤
│ specificity   │  0.9105 │       0.8393 │ 0.8938 │
├───────────────┼─────────┼──────────────┼────────┤
│ combined_loss │  0.4619 │       0.5039 │ 0.4488 │
╘═══════════════╧═════════╧══════════════╧════════╛
```


ludwig/features/base_feature.py
```diff
@@ -380,9 +379,18 @@ def get_metrics(self):
...
+                logger.exception(f"Caught exception computing metric: {metric_name} with error: {e}.")
+                continue
+
+            # Metrics from torchmetrics can be a tensor.
+            if isinstance(computed_metric, Tensor):
+                metric_vals[metric_name] = computed_metric.detach().cpu().numpy().item()
+            else:
+                # Or a dict of tensors. Unpack if so.
+                for sub_metric_name, metric in computed_metric.items():
+                    metric_vals[sub_metric_name] = metric.detach().cpu().numpy().item()
```
arnavgarg1, line 393: Can we leave a comment here saying what kinds of metrics may be a dictionary (rouge)?

ludwig/features/text_feature.py
```diff
@@ -261,6 +264,36 @@ def get_output_dtype(cls):
...
+        response-based metrics like BLEU and ROUGE are also computed.
+
+        Args:
+            targets: Tensor with target values for this output feature.
+            predictions: Dict of tensors returned by predictions().
+        """
+        for metric_name, metric_fn in self._metric_functions.items():
+            prediction_key = get_metric_tensor_input(metric_name)
+            if prediction_key != RESPONSE:
+                # Non-RESPONSE metrics don't use decoded texts.
```
arnavgarg1, line 286: Again, would be good to expand on this comment a bit more

ludwig/features/text_feature.py
```diff
@@ -261,6 +264,36 @@ def get_output_dtype(cls):
...
+                metric_fn = metric_fn.to(predictions[prediction_key].device)
+                metric_fn.update(predictions[prediction_key].detach(), targets)
+                continue
+
+            if decoded_targets is not None and decoded_predictions is not None:
+                # RESPONSE metrics cannot be computed if decoded texts aren't provided.
+                # We assume that RESPONSE metrics can only be calculated one example at a time.
```
arnavgarg1, line 293: OOC, why do we make this assumption?

requirements.txt
```diff
@@ -24,6 +24,7 @@ marshmallow
 marshmallow-jsonschema
 marshmallow-dataclass==8.5.4
 tensorboard
+nltk
```
arnavgarg1, line 27: Is this required for bleu/rouge?

arnavgarg1 (approved):
LGTM, nice work @justinxzhao ! Just left some comments to add more comments to clarify some pieces of the code, but this is a good change overall.

ludwig/features/text_feature.py
justinxzhao: Ah, this appears to be an incorrect assumption. torchmetrics text docs all use single examples. However, batching `(preds, targets)` works great, consistent with other torchmetrics. Removing the extra for loop.

ludwig/features/text_feature.py
justinxzhao: Done. On a side note, `DECODED_TEXT` seems like a clearer naming (for both this code and the registry), but then we'd lose the consistency with the output of `model.predict()`.

ludwig/features/base_feature.py
justinxzhao: Done.

ludwig/features/text_feature.py
```diff
@@ -261,6 +264,34 @@ def get_output_dtype(cls):
...
+        If decoded_targets and decoded_predictions are provided, as through LLM model types, then additional
+        response-based metrics like BLEU and ROUGE are also computed.
+
+        Args:
+            targets: Tensor with target values for this output feature.
+            predictions: Dict of tensors returned by predictions().
+        """
+        for metric_name, metric_fn in self._metric_functions.items():
+            prediction_key = get_metric_tensor_input(metric_name)
+            if prediction_key != RESPONSE:
```
w4nderlust, line 285: nit, i'd rather have more explitly the response case and the else case than the non response case

ludwig/models/llm.py
```diff
@@ -480,13 +480,17 @@ def update_metrics(self, targets, predictions):
...
-            of_obj.update_metrics(targets[of_name], predictions[of_name])
+                # 1. Align the target length with the predictions length to enable text metric evaluation.
+                # 2. Decode both the target and predictions tokens to strings to calculate additional metrics.
+                (
+                    _targets,
+                    _decoded_targets,
+                    _predictions,
+                    _decoded_predictions,
+                ) = realign_target_and_prediction_tensors_for_inference(targets, predictions, of_name, self.tokenizer)
+                of_obj.update_metrics(_targets[of_name], _predictions[of_name], _decoded_targets, _decoded_predictions)
```
w4nderlust, line 491: do we streictly need to pass all this stuff? plus isn't there a more elegant way that having an if textfeature and having update_metrics to have a different signature only in that case? It's not amazing abstraction, and requires us to check if textoutput everywhere we call update metrics (it may be happening only here, but still, it's a red flag that comething is wrong with the abstraction). One option could be to have kwargs in the superclass update metrics and put the edditional things as parameters passed through kwargs. it would still need an if for determining the inputs, but at least we would not have textoutputs need a different signature for update_metrics

ludwig/models/llm.py
```diff
@@ -505,7 +509,22 @@ def update_metrics_finetune_llm(self, targets, predictions):
...
+                    # LLM.update_metrics() for decoding consistency.
+                    (
+                        _targets,
+                        _decoded_targets,
+                        _predictions,
+                        _decoded_predictions,
+                    ) = realign_target_and_prediction_tensors_for_inference(
+                        _targets, _predictions, of_name, self.tokenizer
+                    )
+                    of_obj.update_metrics(
```
w4nderlust, line 523: like above, this happens in at least two palces then

ludwig/features/text_feature.py
justinxzhao: Done.

ludwig/models/llm.py
justinxzhao: Having `TextOutputFeature` have access to its tokenizer (for decoding predictions) would enable the original `update_metrics()` signature to work, though `TextOutputFeature` would still need to override the base class's implementation of `update_metrics()` to decode texts with the tokenizer and use the decoded texts to compute `RESPONSE` metrics.

However, constructing `TextOutputFeature` with the tokenizer would require an even more invasive change.
- Output feature construction code paths are still fully unified under `BaseModel.build_outputs()`. 
- We would need to change the `build_outputs()` signature to take in an additional `training_set_metadata`  argument (which contains tokenizer information), so that it can be passed through the output feature constructors, even though TextOutputFeature is the only output feature who would need training_set_metadata in this way.
- Output feature construction call sites would need to be updated as well -- there are 3 (`ecd.py`, `gbm.py`, and `llm.py`).

justinxzhao: This refactoring seems worthwhile to do if we anticipate `TextOutputFeature` being used for multiple model types. For now, keeping the ownership of the tokenizer at the model level, relying on the model to decode predictions, and provide them for `TextOutputFeature`'s metrics calculations seems reasonable to proceed with -- WDYT?

ludwig/models/llm.py
justinxzhao: Acknowledged.

ludwig/models/llm.py
justinxzhao: I made one simplification, which is to do decoding within `update_metrics()` instead of outside at the model level. This should reduce some duplication.

w4nderlust: sounds good to me!

PR #3541, opened 2023-08-18
Fix pad + bos token issues for all models

arnavgarg1:
This fixes some very minor (but potentially serious) bugs in LLM fine-tuning, specifically for Llama-based models. This can cause less accurate text generation outputs during fine-tuning, and may potentially lead to NaN loss during fine-tuning (although there are fairly low chances of this happening since we have already seen good post-finetuning results).

##  Issue 1:
When removing padding tokens from the input and target tensors, we left the BOS token in for both the input and target tensors. This led to the curios case of the merged_tensor looking like `[bos_token, input_tokens, bos_token, target_tokens, pad_token]`. It turns out that this in fact incorrect, and we should remove it from both the input and target tensors before merging them together to instead be represented as `[input_tokens, target_tokens, pad_token]`. This PR fixes this.

##  Issue 2:
2. Llama-based tokenizers don't have padding tokens set by default. Traditionally, the approach has been to set the pad token to the EOS token, and this is what 90% of notebooks do online. However, what they never mention is that this only makes sense when the default padding side is right padding. This makes sense, because you end up with a sequence that is your input tokens + target tokens + EOS token. However, when you do left padding, this is actually causes an issue. Your sequence begins with EOS tokens (since it is what is used as padding), followed by a BOS token, followed by your input tokens + target tokens, and then an EOS token at the very end (which substitutes a final padding token). Effectively, this is going to teach your model not to generate anything, or rather, generate incorrectly because it just sees EOS tokens at the beginning of the sequence. This is not what we want. 

To fix this, there are two recommended approaches:

### Approach 1
Add a special token called "<pad>" to the tokenizer's vocabulary and preprocess the data. Then, when loading in the Llama based model, update the `embed_tokens` layer to use the right padding token/resize the embedding layer. See https://huggingface.co/docs/transformers/model_doc/llama2

> The original model uses pad_id = -1 which means that there is not padding token. We can’t have the same logic, make sure to add a padding token using tokenizer.add_special_tokens({"pad_token":"<pad>"}) and resize the token embedding accordingly. You should also set the model.config.pad_token_id. The embed_tokens layer of the model is initialized withself.embed_tokens = nn.Embedding(config.vocab_size, config.hidden_size, self.config.padding_idx), which makes sure that encoding the padding token will output zeros, so passing it when initializing is recommended.

This is actually non-trivial to do unless you're training the model from scratch. The reason is that the new token gets assigned a value of len(vocab) + 1, but the embedding layer has a shape of (vocab, hidden_size) with padding_idx set in advance to the last token in the vocab. Here are some issues I ran into:

- The default embedding layer uses (embed_tokens): Embedding(32000, 4096, padding_idx=31999) 31999 as the padding_idx. By doing the tokenizer.add_special_tokens({"pad_token": "<pad>"}) as HF suggests, it actually increases the vocab_size by 1 and sets the padding token ID to 32000, which is 1 more than the size of the embedding layer (32000).
- If you try and get the config using AutoConfig, then update the config (padding_token_id or vocab_size or both) and initialize the model using that, you run into the following error: `AssertionError: Padding_idx must be within num_embeddings`. That makes sense. Also, if we do it this way, we lose all of the other model_load_kwargs way of initializing the model which is not at all good.
- If you try and do it after initialization through something like `model.get_input_embeddings().padding_idx = tokenizer.pad_token_id`, it causes the same error but at `RunTime: AssertionError: Padding_idx must be within num_embeddings`.
- The final thing one could do is actually create a fresh `torch.nn.embeddings` layer `torch.nn.Embedding(32001, 4096, padding_idx=32000)` overwriting the existing layer, but then you lose the pre-trained weights which I'm absolutely not in favor of doing.

### Approach 2
Set the pad token to the UNK token since it's hardly used during pretraining, which means it will be treated in a similar fashion to what a real pad token would do. Here's an example of a notebook that does this exact thing for Llama-2-finetuning when the padding side is left: https://github.com/run-llama/modal_finetune_sql/blob/main/src/finetune_sql.py. This is the approach I've implemented.

Finally, I added a test to check a majority of the LLM util functions in sequence to imitate their calling pattern in the LLM model's forward pass. This is good for us to have both from an understanding perspective, but also to confirm that things are happening exactly as expected.

## Comparing Quality Before And After

I fine-tuned Llama-2-7b using the Code Alpaca dataset on 100 rows of data for 5 epochs.

Here are the results before this change:

```
Instruction: Create an array of length 5 which contains all even numbers between 1 and 10.
Input: 
Generated Output: sierp = [2, 4, 6, 8, 10]



Instruction: Create an array of length 15 containing numbers divisible by 3 up to 45.
Input: 
Generated Output: sierp = [3, 6, 9, 12, 15, 18, 21, 24, 27, 30, 33, 36, 39, 42, 45]



Instruction: Create a nested loop to print every combination of numbers between 0-9
Input: 
Generated Output: sierpana = [0,1,2,3,4,5,6,7,8,9]
for i in range(10):
    for j in range(10):
        print(augustana[i], augustana[j])



Instruction: Generate a function that computes the sum of the numbers in a given list
Input: 
Generated Output: sierp = [1, 2, 3, 4, 5]
sum = 0
for i in sierp:
    sum += i
return sum



Instruction: Create a class to store student names, ages and grades.
Input: 
Generated Output: sierpana.py

class student:
    def __init__(self, name, age, grade):
        self.name = name
        self.age = age
        self.grade = grade

    def __str__(self):
        return f"{self.name} is {self.age} years old and has a grade of {self.grade}."



Instruction: Print out the values in the following dictionary.
Input: my_dict = {
  'name': 'John Doe',
  'age': 32,
  'city': 'New York'
}
Generated Output: sierp_dict = {
  'name': 'john doe',
  'age': 32,
  'city': 'new york'
}
for key, value in my_dict.items():
    print(key, ":", value)
```

Here are the results using this branch:
```
Instruction: Create an array of length 5 which contains all even numbers between 1 and 10.
Input: 
Generated Output: arr = [2, 4, 6, 8, 10]

def create_array_of_length_5_which_contains_all_even_numbers_between_1_and_10():
    arr = [2, 4, 6, 8, 10]
    return arr

arr = create_array_of_length_5_which_contains_all_even_numbers_between_1_and_10()
print(arr)


Instruction: Create an array of length 15 containing numbers divisible by 3 up to 45.
Input: 
Generated Output: arr = [3,6,9,12,15,18,21,24,27,30,33,36,40,45]ITableView(arr)

def array_of_length_15_containing_numbers_divisible_by_3_up_to_45():
    arr = [3,6,9,12,15,18,21,24,27,30,33,36,40,45]
    print(arr)

array_of_length_15_containing_numbers_divisible_by_3_up_to_45()


Instruction: Create a nested loop to print every combination of numbers between 0-9
Input: 
Generated Output: for (int i = 0; i < 10; i++) {
    for (int j = 0; j < 10; j++) {
        std::cout << i << " " << j << " ";
    }
    std::cout << std::endl;
}

Instruction: Generate a function that computes the sum of the numbers in a given list
Input: 
Generated Output: def sum_of_numbers_in_list(numbers):
    total = 0
    for number in numbers:
        total += number
    return total

sum_of_numbers_in_list([1, 2, 3, 4, 5])
# output: 15


Instruction: Create a class to store student names, ages and grades.
Input: 
Generated Output: 
class student:
    def __init__(self, name, age, grade):
        self.name = name
        self.age = age
        self.grade = grade

    def __str__(self):
        return f"{self.name} is {self.age} years old and has a grade of {self.grade}."

Instruction: Print out the values in the following dictionary.
Input: my_dict = {
  'name': 'John Doe',
  'age': 32,
  'city': 'New York'
}
Generated Output: print(my_dict)
name: john doe
age: 32
city: new york
```

In my opinion, the latter results are objectively much better when training for the same duration on the exact same dataset. One clear observation is that because the last token is now a PAD token instead of a EOS token, the model tends to overgenerate a little bit. I wonder if we should just set the last token to a EOS token instead of a PAD token (PAD token is what is recommended by PEFT for CausalLM fine-tuning).

### DIFF
--- a/tests/ludwig/utils/test_llm_utils.py
+++ b/tests/ludwig/utils/test_llm_utils.py
@@ -312,7 +312,7 @@ def test_llm_utils_e2e():
 
     input_ids = tokenizer([input_str])
     input_ids["input_ids"] = torch.tensor(input_ids["input_ids"])
-    input_token_ids = input_ids["input_ids"]
+    input_token_ids_without_padding = input_ids["input_ids"]
     # Prepend some padding tokens
     input_ids["input_ids"] = torch.nn.functional.pad(
         input_ids["input_ids"], (10 - len(input_ids["input_ids"]), 0), value=tokenizer.pad_token_id
@@ -321,7 +321,7 @@ def test_llm_utils_e2e():
 
     target_ids = tokenizer([target_str])
     target_ids["input_ids"] = torch.tensor(target_ids["input_ids"])
-    target_token_ids = target_ids["input_ids"]
+    target_token_ids_without_padding = target_ids["input_ids"]
     # Prepend some padding tokens
     target_ids["input_ids"] = torch.nn.functional.pad(
         target_ids["input_ids"], (10 - len(target_ids["input_ids"]), 0), value=tokenizer.pad_token_id
@@ -337,12 +337,18 @@ def test_llm_utils_e2e():
 
     # Ensure the merged_ids is the correct length
     # - 2 for the two BOS tokens in the input and target, + 1 for the PAD token added to the end
-    assert len(merged_ids[0]) == len(input_token_ids[0]) + len(target_token_ids[0]) - 2 + 1
+    assert (
+        len(merged_ids[0]) == len(input_token_ids_without_padding[0]) + len(target_token_ids_without_padding[0]) - 2 + 1
+    )
 
     # Ensure the merged_ids is the correct value
     # This also ensures that there is no PAD or BOS token at the beginning of the merged_ids
     expected_merged_ids = torch.cat(
-        (input_token_ids[0][1:], target_token_ids[0][1:], torch.tensor([tokenizer.pad_token_id])),
+        (
+            input_token_ids_without_padding[0][1:],
+            target_token_ids_without_padding[0][1:],
+            torch.tensor([tokenizer.pad_token_id]),
+        ),
         dim=0,
     )
     assert torch.equal(merged_ids[0], expected_merged_ids)
--- a/tests/ludwig/utils/test_llm_utils.py
+++ b/tests/ludwig/utils/test_llm_utils.py
@@ -11,7 +11,7 @@ from ludwig.utils.llm_utils import (
     has_padding_token,
     pad_target_tensor_for_fine_tuning,
     realign_target_and_prediction_tensors_for_inference,
-    remove_left_padding,
+    remove_left_padding_and_bos_token,
     set_pad_token,
 )
 
@@ -84,13 +84,13 @@ def test_has_padding_token_without_padding_tokens(tokenizer):
         # Padding
         (torch.tensor([1, 5, 5, 3]), torch.tensor([5, 5, 3])),
         # EOS token
-        (torch.tensor([2, 5, 5, 3]), torch.tensor([2, 5, 5, 3])),
+        (torch.tensor([2, 5, 5, 3]), torch.tensor([5, 5, 3])),
         # Padding + EOS token
-        (torch.tensor([1, 2, 5, 5, 3]), torch.tensor([2, 5, 5, 3])),
+        (torch.tensor([1, 2, 5, 5, 3]), torch.tensor([5, 5, 3])),
     ],
 )
-def test_remove_left_padding(input_ids, expected, tokenizer):
-    assert torch.equal(remove_left_padding(input_ids, tokenizer).squeeze(0), expected)
+def test_remove_left_padding_and_bos_token(input_ids, expected, tokenizer):
+    assert torch.equal(remove_left_padding_and_bos_token(input_ids, tokenizer).squeeze(0), expected)
 
 
 @pytest.mark.parametrize(
--- a/tests/ludwig/utils/test_llm_utils.py
+++ b/tests/ludwig/utils/test_llm_utils.py
@@ -42,12 +42,22 @@ def target_ids():
     return torch.tensor([[9, 10, 11], [12, 13, 14]])
 
 
-def test_set_pad_token_doesnt_exist():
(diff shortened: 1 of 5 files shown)


tests/ludwig/utils/test_llm_utils.py
```diff
@@ -288,3 +298,54 @@ def test_realign_target_and_prediction_tensors_for_inference(tokenizer):
...
+    target_str = "over the lazy dog."
+
+    tokenizer = AutoTokenizer.from_pretrained("huggyllama/llama-7b", use_fast=False)
+    set_pad_token(tokenizer)
+
+    # # Simulate Output From Ludwig preprocessing
+
+    input_ids = tokenizer([input_str])
+    input_ids["input_ids"] = torch.tensor(input_ids["input_ids"])
+    input_token_ids = input_ids["input_ids"]
```
justinxzhao, line 315: nit: `input_token_ids_without_padding`

justinxzhao (approved):
Just one nit. Thank you for the extensive PR description.

w4nderlust:
A couple points to stress:
- is the BOS thing a llama tokenizer specific issue. Other models and relative tokenizers could be different. I saw that you added an if for those exception. It is fine for now, but Probably the best solution in the future would be to have a dictionary that maps from the tokenizer to the actual tokens they use for eos bos and pad, so that yo ucan have a single code path and multiple behaviors depending on the tokenizer (with a default case too)
- I don't fully appreciate why if llama-2 does right padding instead of left padding, we can't just do the same instead of doing left padding and then having to manipulate the padded sentence once more. It's not a big issue as long as it works, but I wonder if, similar to what I was saying above, this could be one additional parameter associated with specific llms/tokenizers so that the dictionary will tell us if to do right padding or left padding i nthe same way it can tell us what tokens to use for eos bos and pad

PR #3543, opened 2023-08-23
Add skip_all_evaluation as a mechanic to skip all evaluation.

justinxzhao:
Evaluating a model, especially on large validation or test sets, can be time-consuming.

This parameter can also be useful if you are training a model with a well-known configuration on a well-known dataset and are confident about the expected results, or if you are planning to evaluate the model separately, outside of ludwig.

### DIFF
--- a/ludwig/trainers/trainer_llm.py
+++ b/ludwig/trainers/trainer_llm.py
@@ -109,6 +109,23 @@ class NoneTrainer(BaseTrainer):
         self.checkpoints_per_epoch = self.config.checkpoints_per_epoch
         self.early_stop = self.config.early_stop
         self.evaluate_training_set = self.config.evaluate_training_set
+        self.skip_all_evaluation = self.config.skip_all_evaluation
+
+    def finish_training(
+        self, progress_tracker, save_path, train_summary_writer, validation_summary_writer, test_summary_writer
+    ):
+        # ================ Finished Training ================
+        self.callback(
+            lambda c: c.on_trainer_train_teardown(self, progress_tracker, save_path, self.is_coordinator()),
+            coordinator_only=False,
+        )
+
+        if train_summary_writer is not None:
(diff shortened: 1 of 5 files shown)


ludwig/trainers/trainer_llm.py
```diff
@@ -164,24 +197,10 @@ def train(
                 save_path,
             )
         finally:
-            # ================ Finished Training ================
-            self.callback(
-                lambda c: c.on_trainer_train_teardown(self, progress_tracker, save_path, self.is_coordinator()),
-                coordinator_only=False,
+            self.finish_training(
```
w4nderlust, line 200: we can find a better name for this function. Maybe saome thing more explicit of what actually happens inside it, like "close_writers" or something long those lines

justinxzhao: Done.

ludwig/trainers/trainer_llm.py
```diff
@@ -151,6 +168,22 @@ def train(
             output_features=output_features,
         )
 
+        # When running with Ray, we only need to return the state dict, as it's faster and cheaper to send the
+        # state dict over the network than to load the model state here, serialize it back to a state dict, then
+        # load it back on the head node.
+        return_value = self.model if not return_state_dict else self.model.cpu().state_dict()
```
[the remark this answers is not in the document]
justinxzhao: FYI, this is not a new comment (I'm only moving up from below). It looks like this was originally added by @geoffreyangus [here](https://github.com/ludwig-ai/ludwig/pull/3351/files#diff-1863d6aa28debac9851df20b967ea4f44b7ecc52a2c3a8c57f03b5311304ab5eR175-R178).

My reading of this code is that this is an optimization to prevent an extra round of serialization/deserialization when using Ray for multi-node training.

w4nderlust (approved):

PR #3552, opened 2023-08-28
Add create_pr=True to `ludwig upload`.

justinxzhao:
The command seems to fail for newly created model repositories.


arnavgarg1 (commented):
Is this with a newer version of huggingface hub? Should we also pin Ludwig to a minimum of that version in our requirements?

justinxzhao:
> Is this with a newer version of huggingface hub? Should we also pin Ludwig to a minimum of that version in our requirements?

It looks like the latest release was from July 7, so I'm not sure what changed. Are you able to run `ludwig upload` to a new repo without this change?

PR #3553, opened 2023-08-28
Add backwards compatibility check for effective batch size

arnavgarg1:


tests/ludwig/utils/test_backward_compatibility.py
```diff
@@ -929,3 +929,21 @@ def test_type_removed_from_defaults_config():
 
     for feature_type in config_obj_2.get("defaults"):
         assert "type" not in config_obj_2["defaults"][feature_type]
+
+
+def test_trainer_effective_batch_size():
```
tgaddair, line 934: Test doesn't fail without your change. Please make sure any time you add a test to fix a bug, it is increasing the surface area of the test suite, and that the test properly fails without the fix applied.

tests/ludwig/utils/test_backward_compatibility.py
```diff
@@ -929,3 +929,21 @@ def test_type_removed_from_defaults_config():
...
+        "output_features": [
+            {"name": "Survived", "type": "category"},
+        ],
+        "trainer": {
+            "batch_size": 64,
+        },
+    }
+
+    config_obj = ModelConfig.from_dict(config)
+    assert config_obj.trainer.effective_batch_size
```
jeffkinnison, line 948: Is this intended to check whether or not the attribute exists? If so, it might be a good idea to refactor to

```
assert hasattr(config_obj.trainer, "effective_batch_size")
```

This would be a bit clearer and avoids any collisions with `effective_batch_size` evaluating to `False` (admittedly also an issue, but we should be able to catch those with the second assert).

PR #3558, opened 2023-08-30
do not compare dicts in visualize compare_performance

thelinuxkid:
# Description

`ludwig visualize   --visualization compare_performance` collects `roc_curve` and `precision_recall_curve` (possibly others) as dicts instead of a list of floats. These two keys contain duplicate data-points such as `precision`, `recall`. Skip compare these two keys so that the call to `min` or `max` won't fail

# Change

Skips comparing dicts for `visualize`'s `compare_performance`

# POC

## Before Change
image

## After Change

image


thelinuxkid:
@arnavgarg1 @jeffkinnison

jeffkinnison:
@thelinuxkid +1 to @arnavgarg1's comment. If this unblocks I'm happy to adopt it as a temporary fix, but we should figure out why there are dupes.

thelinuxkid:
I can't reproduce this anymore so I will close but the issue was basically [`metric_val`](https://github.com/ludwig-ai/ludwig/pull/3558/files#diff-a6e3d9afdd95decfc7a71b4140a9d7eb6f38dc3de3c0eb8491bf581dfa6ca234R1513) would sometimes contain a `list[float]` and sometimes a `dict[str,list[float]]`. This might have been because I ran the trainer multiple times and it create multiple instances of the model.

PR #3559, opened 2023-08-30
Report loss in tqdm to avoid log spam

tgaddair:


justinxzhao (approved):
Thanks for the change!

arnavgarg1 (approved):

PR #3562, opened 2023-08-30
Wrap each metric update in try/except.

justinxzhao:


ludwig/features/text_feature.py
```diff
@@ -302,15 +302,22 @@ def update_metrics(
...
+                            if metric_name == "bleu":
+                                # BLEU takes in targets as a list.
+                                metric_fn.update(decoded_predictions, [decoded_targets])
+                            else:
+                                metric_fn.update(decoded_predictions, decoded_targets)
+                else:
+                    metric_fn = metric_fn.to(predictions[prediction_key].device)
+                    metric_fn.update(predictions[prediction_key].detach(), targets)
+            except Exception as e:
+                logger.info(f"Ran into error when calculating metric {metric_name}. Skipping. The error is: {e}")
```
arnavgarg1, line 320: Hmm, so will the metric function have no value then that we can collect later? Will that cause an error?

justinxzhao: I think we should be ok. `get_metrics()` is already wrapped in a try/except:

```python
    def get_metrics(self):
        metric_vals = {}
        for metric_name, metric_fn in self._metric_functions.items():
            try:
                computed_metric = metric_fn.compute()
            except Exception as e:
                logger.exception(f"Caught exception computing metric: {metric_name} with error: {e}.")
                continue
```

And it also looks like there's default values for torchmetrics, even if no update() was called.
```python
from torchmetrics.text import BLEUScore
bleu = BLEUScore()
bleu.compute()
/Users/justinzhao/ludwig/env/lib/python3.8/site-packages/torchmetrics/utilities/prints.py:36: UserWarning: The ``compute`` method of metric BLEUScore was called before the ``update`` method which may lead to errors, as metric states have not yet been updated.
  warnings.warn(*args, **kwargs)
tensor(0.)
```

PR #3564, opened 2023-08-30
schema: Add `prompt` validation check

ksbrar:


ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,20 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+
+    column_names = [feature.column for feature in config.input_features]
+    template_refs = set(Formatter().parse(config.prompt.template))
+    intersection = column_names & template_refs
+
+    if len(list(intersection)) == 0:
+        raise ConfigValidationError(
+            "Prompt template must include a reference to a column. This can be as simple as a string with just the "
+            'column and nothing else, e.g.: "{column_name}".'
+        )
```
arnavgarg1, line 626: One thing to consider is that we actually allow no `template` to be specified if the `task` parameter of `prompt` is specified. In this case, we actually have a default zero shot and few shot prompt template that we use (zero shot also gets used for fine-tuning), and this automatically inserts `{__task__}` and `{__sample__}` into the data (I think).

1. Would be good to confirm if this is indeed what happens
2. Does this check allow for that to happen? I might add a conditional block to account for the task + template, or just task case.

ksbrar: will add

tests/ludwig/config_validation/test_validate_config_misc.py
```diff
@@ -466,3 +466,31 @@ class TestClass:
...
+            "unknown_parameter_combiner": False,
+        },
+        TRAINER: {
+            "epochs": 1000,
+        },
+    }
+    with pytest.raises(ConfigValidationError):
+        ModelConfig.from_dict(config)
+
+    config.prompt.template = "invalid string"
```
jeffkinnison, line 490: nit: The values of `config.prompt.template` could be provided as pytest parameters, i.e.

```
@pytest.mark.parametrize("template", [None, "invalid template", "{{test}}"])
def test_template_string_validation(template):
```

tests/ludwig/config_validation/test_validate_config_misc.py
```diff
@@ -466,3 +466,31 @@ class TestClass:
     test_class = unload_jsonschema_from_marshmallow_class(TestClass)
 
     assert test_class["properties"]["test_schema_entry"]["parameter_metadata"] is not None
+
+
+def test_template_string_validation():
```
jeffkinnison, line 471: Could you add a test for a config that passes as well?

ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,38 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
         raise ConfigValidationError("Fine-tuning and LLM with quantization requires using the 'lora' adapter")
+
+
+@register_config_check
+def check_llm_template_references(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks that prompt.template includes a valid reference to a feature column."""
+    if config.model_type != MODEL_LLM:
+        return
+
+    template = config.prompt.template
```
arnavgarg1, line 618: I'd add a check to make sure `config.prompt` is not None first, and then proceed with this check

ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,38 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+    if config.prompt.task is not None and template is not None:
+        raise ConfigValidationError("Prompt cannot have both a task and a template!")
```
arnavgarg1, line 620: This isn't true, you are allowed to have both a task and a template. The idea is that task gets used in the template

ksbrar: that was the old check

ludwig/config_validation/checks.py
ksbrar: out of date

ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,38 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+
+    # If no prompt is provided, no validation necessary:
+    if not config.prompt:
+        return
+
+    template = config.prompt.template
+    task = config.prompt.task
+    # If no template is provided, task is required:
+    if template and not task:
+        raise ConfigValidationError("A prompt task is required if no template is provided!")
```
jeffkinnison, line 626: I don't think the error message here matches the condition. It should be `if not template and not task`, right?

ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,38 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+        template_refs = set(Formatter().parse(template))
+        intersection = column_names & template_refs
+
+        if task and "__task__" not in template_refs:
+            raise ConfigValidationError(
+                "When providing a template and a task, you must make sure that the task keyword `{{__task__}} is "
+                "present somewhere in the template string!"
+            )
+
+        if len(list(intersection)) == 0 and "__sample__" not in template_refs:
```
jeffkinnison, line 640: nit: `len(intersection)` should return the same value here (if I'm interpreting this right and `intersection` is a set)

tests/ludwig/config_validation/test_checks.py
```diff
@@ -459,3 +459,28 @@ def test_check_qlora():
         "type": "lora",
     }
     ModelConfig.from_dict(config)
+
+
+@pytest.mark.parametrize("template", [None, "invalid template", "{{test}}"])
```
jeffkinnison, line 464: Some other (admittedly paranoid) test cases we may want to consider:

- Other non-string data types (int, float, bool, list of string) as a sanity check
- A valid case with `"{{test}}"` embedded in a a larger string
- Adding a second input feature `"test2"` and making sure `"{{test2}}"` passes
- With the above, making sure `"{{test}} {{test2}}"` passes
- Also making sure `"{{test}} {{test2}}"` passes
- What happens if we reference a feature that doesn't exist in the template?

ludwig/config_validation/checks.py
ksbrar: yup already fixed

tests/ludwig/config_validation/test_checks.py
```diff
@@ -459,3 +459,28 @@ def test_check_qlora():
...
+        },
+    }
+
+    ModelConfig.from_dict(config)
+
+    config["prompt"] = {"task": "Some task", "template": ""}
+    with pytest.raises(ConfigValidationError):
+        ModelConfig.from_dict(config)
+
+    config["prompt"] = {"task": "Some task", "template": "{__task__}"}
```
arnavgarg1, line 486: Are there some more lines of code for this test to show success?

ksbrar: still working on it

tests/ludwig/config_validation/test_checks.py
jeffkinnison: @arnavgarg1 Just wanted to double-check with you: would tests for multiple input features or nonexistent input features be feasible using only the config?

jeffkinnison: nvm, we're skipping those for now

ksbrar:
@jeffkinnison @arnavgarg1 @justinxzhao PR is complete and affected tests have succeeded. Remaining test failures seem unrelated but they've happened on several runs now.

ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,56 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+    if config.model_type != MODEL_LLM:
+        return
+
+    # TODO: `prompt` by default should be set to null, not a default dict:
+    # # If no prompt is provided, no validation necessary:
+    # if not config.prompt:
+    #     return
+    from ludwig.schema.llms.prompt import PromptConfig, RetrievalConfig
+
+    if config.prompt == PromptConfig():
```
justinxzhao, line 624: Is the idea here that if it's the default prompt config, we don't need to check prompt requirements?

ludwig/config_validation/checks.py
```diff
@@ -606,3 +607,56 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+
+        # Otherwise, the template should at least contain the sample keyword or some input column:
+        # TODO: len(template_refs) is a hacky attempt to check that there are references to *something* in the
+        # string. The proper validation is to check the references against the features in the user's dataset - but we
+        # do not have access to the dataset in this code path right now.
+        if not task:
+            if len(template_refs) == 0 and "__sample__" not in template_refs:
+                raise ConfigValidationError(
+                    "A template must contain at least one reference to a column or, in the case of zero/few-shot "
+                    "learning, at least the sample keyword `{__sample__}`!"
```
justinxzhao, line 661: How about: 

"A template must contain at least one reference to a column or the sample keyword `{__sample__}` for a JSON-serialized representation of non-output feature columns."

Since `{__sample__}` can be used for regular fine-tuning as well.

CC: @arnavgarg1

justinxzhao (approved):
Nice, clear aux config validation check!

ludwig/config_validation/checks.py
ksbrar: Exactly. It's a confusing check, which is why ideally I think `prompt` should be null by default.

ludwig/config_validation/checks.py
ksbrar: added

PR #3565, opened 2023-08-31
fix: Add predictor-specific device placement handling

jeffkinnison:


ludwig/distributed/deepspeed.py
```diff
@@ -125,6 +126,9 @@ def prepare_for_inference(self, model: nn.Module) -> nn.Module:
     def to_device(self, model: nn.Module, device: Optional[torch.device] = None) -> nn.Module:
         return model
 
+    def to_device_for_prediction(self, model: "BaseModel", device: Optional[torch.device] = None) -> nn.Module:
+        return model
```
arnavgarg1, line 130: Is it worth adding a comment here on why we're doing this custom behaviour for DDP?

jeffkinnison:
Closing in favor of #3566

Issue #3568, opened 2023-08-31
local variable 'tokens' referenced before assignment


randy-tsukemen:
**Describe the bug**
LLM finetuning tutorial notebook failed.

**To Reproduce**

1. Open [ludwig_llama2_7b_finetuning_4bit.ipynb](https://colab.research.google.com/drive/1c3AO8l_H6V_x37RwQ8V7M6A-RmcBf2tG?usp=sharing)
2. Replace the huggingface tokens
3. Run all notebook cell
4. See error `RuntimeError: Caught exception during model preprocessing: local variable 'tokens' referenced before assignment`

**Expected behavior**
Notebook shouldn't fail.

**Screenshots**
![圖片](https://github.com/ludwig-ai/ludwig/assets/25134035/50b5eb33-4e42-4b66-923e-fa1a5e4446ce)
**Environment (please complete the following information):**

- OS: google colab

arnavgarg1:
Thanks for flagging this issue @randy-tsukemen. Could you give this branch a try? https://github.com/ludwig-ai/ludwig/pull/3569
It downgrades the transformers package version to `4.31.0`.

PR #3569, opened 2023-08-31
Pin Transformers to 4.31.0

arnavgarg1:
It seems like the latest `transformers` package has a bug. This can be fixed by either using `transformers` master, or downgrading to `4.31.0` which is stable and works correctly. 

This is the error we run into with 4.32.1:
```
site-packages/ludwig/api.py:1510), in LudwigModel.preprocess(self, dataset, training_set, validation_set, test_set, training_set_metadata, data_format, skip_save_processed_input, random_seed, **kwargs)
   1508     return PreprocessedDataset(proc_training_set, proc_validation_set, proc_test_set, training_set_metadata)
   1509 except Exception as e:
-> 1510     raise RuntimeError(f"Caught exception during model preprocessing: {str(e)}") from e
   1511 finally:
   1512     for callback in self.callbacks:
RuntimeError: Caught exception during model preprocessing: local variable 'tokens' referenced before assignment
```

Closes: #3568

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,8 @@ torchaudio
 torchtext
 torchvision
 pydantic<2.0
+# TODO(Arnav): Lift transformers package restriction once 4.32.2 is released.
+# Issue: https://github.com/ludwig-ai/ludwig/issues/3571
 transformers>=4.31.0,<4.32.0
 tokenizers>=0.13.3
 spacy>=2.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ torchaudio
 torchtext
 torchvision
 pydantic<2.0
-transformers<4.32.0
+transformers>=4.31.0,<4.32.0
 tokenizers>=0.13.3
 spacy>=2.3
 PyYAML>=3.12,<6.0.1,!=5.4.* #Exlude PyYAML 5.4.* due to incompatibility with awscli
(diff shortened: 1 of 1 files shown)


tgaddair (approved):
Can you file an issue with HF transformers? It might be an API issue on our end, but either way we probably need them to weigh in.

Also, please add a TODO to the requirements file to revert this once the issue is resolved.

requirements.txt
```diff
@@ -11,7 +11,7 @@ torchaudio
 torchtext
 torchvision
 pydantic<2.0
-transformers>=4.31.0
+transformers<4.32.0
```
tgaddair, line 14: Should be:

```
transformers>=4.31.0;<4.32.0
```

tgaddair (changes requested):
Please fix lower bound.

arnavgarg1:
@tgaddair seems like the issue was flagged and fixed on `transformers` master: https://github.com/huggingface/transformers/issues/25805

arnavgarg1:
Associated Ludwig issue: https://github.com/ludwig-ai/ludwig/issues/3571

arnavgarg1:
Associated Ludwig issue: https://github.com/ludwig-ai/ludwig/issues/3571

justinxzhao (approved):

tgaddair (approved):

PR #3572, opened 2023-08-31
Allow user to specify huggingface link or local path to pretrained lora weights

Infernaught:
Allows user to specify huggingface link or local path to adapter weights. Tested using [Arnav's Code Alpaca V3 model](https://huggingface.co/arnavgrg/codealpaca_v3) (loaded the V3 adapter weights using this change and ran the results through human-eval -- scores matched those of the original V3 model)


ludwig/models/llm.py
```diff
@@ -221,12 +221,36 @@ def initialize_adapter(self):
                     "`finetune` or remove the adapter config."
                 )
 
-            from peft import get_peft_model, TaskType
+            from peft import get_peft_model
+
+            pretrained = False
+            if self.config_obj.adapter.pretrained_weights:
+                print(f"PRETRAINED_WEIGHTS: {self.config_obj.adapter.pretrained_weights}")
```
jeffkinnison, line 228: nit: Are we leaving this `print` in? If so, could you change it to `logger.info`?

ludwig/schema/llms/peft.py
```diff
@@ -69,6 +69,18 @@ class LoraConfig(BaseAdapterConfig):
         description="Bias type for Lora.",
     )
 
+    pretrained_weights: Optional[str] = schema_utils.String(
+        default="none",
```
jeffkinnison, line 73: Should this be `None`? If so, you can pass `allow_none=True` to `schema_utils.String` so that it doesn't raise an exception.

ludwig/models/llm.py
```diff
@@ -221,12 +221,36 @@ def initialize_adapter(self):
...
+                # If pretrained adapter weights are provided, we want to load them into the model
+                from peft import MODEL_TYPE_TO_PEFT_MODEL_MAPPING, PeftConfig
+
+                pretrained = True
+                peft_config = PeftConfig.from_pretrained(self.config_obj.adapter.pretrained_weights)
+                peft_dict = peft_config.to_dict()
+                for param_name, param_value in self.config_obj.adapter.to_config().to_dict().items():
```
jeffkinnison, line 235: nit: It would be nice to have a short comment explaining what this loop is doing and why, just in case this needs to be modified down the road.

ludwig/schema/llms/peft.py
Infernaught: I was going by the convention that Travis used on line 68, but I can change this.

jeffkinnison: Ah, no worries! If it makes sense there then I wouldn't change it.

tests/integration_tests/test_llm.py
```diff
@@ -492,12 +493,46 @@ def test_default_max_sequence_length():
...
         },
+        ADAPTER: {TYPE: "lora", PRETRAINED_WEIGHTS: "Infernaught/test_adapter_weights"},
+        BACKEND: {TYPE: "local"},
     }
     config_obj = ModelConfig.from_dict(config)
     assert config_obj.input_features[0].preprocessing.max_sequence_length is None
     assert config_obj.output_features[0].preprocessing.max_sequence_length is None
 
 
+def test_load_pretrained_adapter_weights():
```
jeffkinnison, line 504: A couple of tests we should add (possibly in a followup PR):

- Checking a null input
- Checking an invalid weights path

jeffkinnison (approved):
Just one additional comment, it shouldn't be blocking if we want to land this.

ludwig/schema/llms/peft.py
```diff
@@ -69,6 +69,17 @@ class LoraConfig(BaseAdapterConfig):
...
+    pretrained_weights: Optional[str] = schema_utils.String(
+        default=None, description="Path to pretrained weights for Lora.", allow_none=True
+    )
+
+    target_modules: Optional[list] = schema_utils.List(
+        str,
+        default=None,
+        allow_none=True,
+        description="List of modules to apply Lora to. If None, apply to all modules.",
+    )
```
arnavgarg1, line 81: Is this needed?

ludwig/trainers/trainer.py
```diff
@@ -272,6 +272,10 @@ def closure():
                     targets, model_outputs, self.regularization_type, self.regularization_lambda
                 )
                 loss = loss / self.gradient_accumulation_steps
+                try:
+                    loss.requires_grad = True
+                except RuntimeError:
+                    pass
```
arnavgarg1, line 278: Why do we have to add this?

ludwig/schema/llms/peft.py
Infernaught: I recall this causing an error if this wasn't set.

ludwig/trainers/trainer.py
Infernaught: This was a workaround I had because, when the lora weights were loaded in, some of the loss functions did not have requires_grad set to True. However, without the try-except block, this would try to set requires_grad to True for some intermediate loss functions, which isn't valid.

ludwig/constants.py
```diff
@@ -282,6 +282,7 @@
 GENERATION = "generation"
 PROMPT = "prompt"
 ADAPTER = "adapter"
+PRETRAINED_WEIGHTS = "pretrained_weights"
```
arnavgarg1, line 285: nit, might be more clear to call it `pretrained_adapter_weights` since pretrained weights also come from the model! So just to avoid confusion

ludwig/trainers/trainer.py
justinxzhao: Questions:
1. Why would loading lora weights have loss functions with `requires_grad != True`?
2. What does the training error look like when some loss functions have `requires_grad != True`?
3. Do you have an example of an intermediate loss function that raises an error when you try to set `requires_grad=True`?

At least, we should be adding a comment explaining why this is here, e.g.

> "When loading adpater weights from huggingface or a local path, some of the loss functions do not have `requires_grad=True`. `requires_grad=True` is necessary for back-propogation, but we wrap this in a try/except because some intermediate losses like __ raise an error if `requires_grad` is explicitly set to `True` in this way."

ludwig/models/llm.py
```diff
@@ -222,12 +222,35 @@ def initialize_adapter(self):
...
+                # If pretrained adapter weights are provided, we want to load them into the model
+                from peft import MODEL_TYPE_TO_PEFT_MODEL_MAPPING, PeftConfig
+
+                peft_config = PeftConfig.from_pretrained(self.config_obj.adapter.pretrained_weights)
+                peft_dict = peft_config.to_dict()
+
+                # Need to update the peft config with some of the values from config_obj because not all of them are set
+                for param_name, param_value in self.config_obj.adapter.to_config().to_dict().items():
+                    if param_name is None:
+                        continue
```
arnavgarg1, line 238: when would param_name be None? This dictionary is the parameters for the pretrained adapter right?

ludwig/trainers/trainer.py
arnavgarg1: Can you elaborate a bit more? What are the intermediate loss functions? I'm also not totally sure how wrapping the model with the adapter is causing these issues?

arnavgarg1: I agree with @justinxzhao and have the same questions as well

ludwig/models/llm.py
Infernaught: You're correct. This should be param_value.

ludwig/constants.py
Infernaught: On it

ludwig/schema/llms/peft.py
Infernaught: If I recall correctly, there was an error involving target_modules not being a parameter of a LoraConfig.

ludwig/trainers/trainer.py
Infernaught: 1. I'm not quite sure on why this was the case. My hypothesis is that, when lora weights are loaded, they overwrite some of the parameters of certain layers.
2. When some loss functions have requires_grad != True, training stops and errors out.
3. I don't have an example of this, but I think my terminology was incorrect here. Specifically, requires_grad can only be changed on leaf variables, so if there is a loss function that is not a leaf node, setting requires_grad=True would cause an error.

arnavgarg1: By default, when a PEFT pretrained adapter is loaded in, is it set to training mode or inference mode? Does that maybe have something to do with requires_grad not being set?

ludwig/models/llm.py
```diff
@@ -222,12 +222,37 @@ def initialize_adapter(self):
...
+                from peft import MODEL_TYPE_TO_PEFT_MODEL_MAPPING, PeftConfig
+
+                peft_config = PeftConfig.from_pretrained(self.config_obj.adapter.pretrained_adapter_weights)
+                peft_dict = peft_config.to_dict()
+
+                # Need to update the peft config with some of the values from config_obj because not all of them are set
+                for param_name, param_value in self.config_obj.adapter.to_config().to_dict().items():
+                    # Not all parameters are supported by all models, so we only add the parameter to the load kwargs
+                    # if it is supported by the model.
+                    if param_value is None:
```
justinxzhao, line 239: Could you add a comment with examples for when this happens?

When are `param_value` from `self.config_obj.adapter.to_config().to_dict().items()` `None` and why should we skip adding them to the peft_dict?

When is the `param_name` not in peft_dict that comes from HF? (Is it just the `type` parameter?

ludwig/trainers/trainer.py
justinxzhao: Good call. Some useful references:
- https://discuss.pytorch.org/t/check-if-model-is-eval-or-train/9395
- https://stackoverflow.com/questions/60018578/what-does-model-eval-do-in-pytorch

ludwig/config_validation/checks.py
```diff
@@ -477,7 +477,7 @@ def check_llm_finetuning_output_feature_config(config: "ModelConfig"):  # noqa:
     if config.model_type != MODEL_LLM:
         return
 
-    if config.trainer.type != "finetune":
+    if config.trainer.type != "finetune" and config.adapter.pretrained_adapter_weights is not None:
```
justinxzhao, line 480: Should this be an OR? Why does specifying `pretrained_adapter_weights` no longer require that the first output feature be TEXT?

Or is it that we want to make it so that using the none trainer type doesn't require an output feature?

```python
if config.trainer.type == "none":
    return
```

CC: @arnavgarg1

ludwig/config_validation/checks.py
```diff
@@ -493,6 +493,9 @@ def check_llm_finetuning_trainer_config(config: "ModelConfig"):  # noqa: F821
     if config.model_type != MODEL_LLM:
         return
 
+    if config.trainer.type == "none" and config.adapter.pretrained_adapter_weights is not None:
```
justinxzhao, line 496: Should this be more simply

```python
if config.trainer.type == "none":
    # The NoneTrainer for ZS is valid.
    return
```

ludwig/models/llm.py
justinxzhao: The newly added comment LGTM. @Infernaught could you mark this comment as resolved?

tests/integration_tests/test_llm.py
```diff
@@ -492,12 +493,62 @@ def test_default_max_sequence_length():
...
     assert config_obj.input_features[0].preprocessing.max_sequence_length is None
     assert config_obj.output_features[0].preprocessing.max_sequence_length is None
 
 
+@pytest.mark.parametrize("adapter", ["lora", "adalora", "adaption_prompt"])
+def test_load_pretrained_adapter_weights(adapter):
+    from peft import PeftModel
+    from transformers import PreTrainedModel
+
+    print(f"ADAPTER: {adapter}")
```
justinxzhao, line 509: nit: Remove print.

ludwig/config_validation/checks.py
Infernaught: But in this case, we would load in untrained LoRA weights if pretrained adapter weights weren't specified in the config, right? Would that be a problem?

justinxzhao (approved):
Thanks!

PR #3585, opened 2023-09-05
Set default global_max_sequence_length to 512 for LLMs

arnavgarg1:
To prevent OOMs, and to ensure parity with huggingface's [SFTTrainer defaults](https://huggingface.co/docs/trl/v0.7.1/en/trainer#trl.SFTTrainer.max_seq_length), we set `global_max_sequence_length` to 512. This only affects LLMs.

For example:

```yaml
input_features:
  - name: instruction
    type: text
output_features:
  - name: output
    type: text
preprocessing:
   global_max_sequence_length: 512
```

will concat the all the input + output tokens (after stripping off the padding), then truncate to the first 512 tokens post-merge.

### DIFF
--- a/ludwig/schema/preprocessing.py
+++ b/ludwig/schema/preprocessing.py
@@ -39,11 +39,12 @@ class PreprocessingConfig(schema_utils.BaseMarshmallowConfig):
     )
 
     global_max_sequence_length: int = schema_utils.PositiveInteger(
-        default=None,
+        default=512,
         allow_none=True,
-        description="Specifically for LLMs. This is the maximum length of the input sequence going into the model's "
-        "forward pass during training. Sequences will be truncated to this length after merging inputs and targets. "
-        "If not set, the total length of the merged input and target token sequences will be used.",
+        description="Specifically for LLMs. Defaults to 512. This is the maximum length of the input sequence "
+        "going into the model's forward pass during training. Sequences will be truncated to this length after "
+        "merging inputs and targets. If set to None, the total length of the merged input and target token "
+        "sequences will be used.",
         parameter_metadata=PREPROCESSING_METADATA["global_max_sequence_length"],
     )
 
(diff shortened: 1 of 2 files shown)


tgaddair (approved):
Nice! We should definitley decouple LLM and ECD preprocessing params to avoid the odd coupling here (not this PR, but we should do it soon).

arnavgarg1:
@tgaddair Totally agree!

justinxzhao (approved):
SGTM.

BTW these tests are all known failures or timeouts (which I'm looking into this week), so good to merge.

piegu:
Hi,

The context window of Llama2 is 4096 tokens (input+output), right?

Why do we would like to truncate it to default `global_max_sequence_length` of 512? 
Because of GPU RAM?

If yes, which GPU has to be used for running the [finetuning Ludwig notebook](https://colab.research.google.com/drive/1Ly01S--kUwkKQalE-75skalp-ftwl0fE?usp=sharing) with 

512 < `global_max_sequence_length` <= 4096?

Thank you.

tgaddair:
@piegu yes the issue is GPU memory. If you're using a bigger card (A100, etc.) then you can go up to 4096.

Would it make more sense to you to set this to null by default and require the user to set it manually to avoid OOMs? That was the other approach we were considering.

piegu:
Thanks for your answer @tgaddair.

In your notebook (in the introduction and in the config file), I think you should be clear about this point: 

In order to get the notebook running in a Colab T4 GPU, we setup the `global_max_sequence_length` to 512. If you need to increase this variable (up to 4096 for Llama 2), you need to move to a GPU with more RAM. Here our recommendations of GPU:
- `global_max_sequence_length` of 1024: ...
- `global_max_sequence_length` of 2048: ...
- `global_max_sequence_length` of 4096: ...

What do you think?

arnavgarg1:
@piegu Definitely agree, that's a good point! We're going to work on updating our notebooks to be very clear about why we're setting this as the default and let the user know they need more memory for longer sequence lengths. Thanks for your suggestion!

On that note, I'm also going to close this PR

PR #3588, opened 2023-09-06
Add support for Paged Optimizers (Adam, Adamw), 8-bit optimizers, and new optimizers: LARS, LAMB and LION

arnavgarg1:
This PR adds support for a variety of new optimizers, including paged and 8-bit variants. All of these are useful for fine-tuning. I will add better descriptions for parameters in a follow-up PR when I better understand some of the underlying papers for LAMB, LARS and LION. I will also add tests once https://github.com/ludwig-ai/ludwig/pull/3578 PR lands.

# What are paged optimizers?

Paged Optimizers use NVIDIA unified memory 3 feature which does automatic page-to-page transfers between the CPU and GPU for error-free GPU processing in the scenario where the GPU occasionally runs out-of-memory. The feature works like regular memory paging between CPU RAM and the disk. This allocates paged memory for the optimizer states which are then automatically evicted to CPU RAM when the GPU runs out-of-memory and paged back into GPU memory when the memory is needed in the optimizer update step.

Screenshot 2023-09-05 at 8 23 14 PM

# What new optimizers are being added to Ludwig?

Here's a summary of the new optimizers that will be added with this PR:
- SGD: 8-bit support
- Adam: 8-bit, Paged Adam, Paged Adam 8-bit
- AdamW: 8-bit, Paged AdamW, Paged AdamW 8-bit
- Adagrad: 8-bit
- RMSProp: 8-bit
- LAMB, Lamb 8-bit
- LARS, LARS 8-bit
- LION, LION 8-bit, Paged LION, Paged LION 8-bit

Here's an example of how you can set these different variants:

Regular AdamW
```yaml
trainer:
  optimizer:
    type: adamw
```

8-Bit AdamW
```yaml
trainer:
  optimizer:
    type: adamw_8bit
```

Paged AdamW
```yaml
trainer:
  optimizer:
    type: paged_adamw
```

Paged AdamW 8-bit
```yaml
trainer:
  optimizer:
    type: paged_adamw_8bit
```

All of this has been made possible through a deeper integration with the bitsandbytes library directly. Note that the 8-bit and paged optimizers only work on GPU machines, and they are not compatible with Deepspeed.

### DIFF
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -4,6 +4,7 @@ from abc import ABC, abstractmethod
 from re import findall
 from typing import Callable, TYPE_CHECKING
 
+import torch
 from transformers import AutoConfig
 
 from ludwig.api_annotations import DeveloperAPI
@@ -109,6 +110,16 @@ def check_training_runway(config: "ModelConfig") -> None:  # noqa: F821
                 "trainer.steps_per_checkpoint. Please specify one or the other, or specify neither to "
                 "checkpoint/eval the model every epoch."
             )
+
+
+@register_config_check
+def check_optimizer_requirements(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks if GPU is available when using Paged Optimizer."""
+    # TODO: Figure out why this check won't work
+    if config.trainer.optimizer.is_paged and not torch.cuda.is_available():
+        raise ConfigValidationError(
+            "Paged Optimizers require GPU but no GPU was detected. Please use a different optimizer."
+        )
 
 
 @register_config_check
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -4,7 +4,6 @@ from abc import ABC, abstractmethod
 from re import findall
 from typing import Callable, TYPE_CHECKING
 
-import torch
 from transformers import AutoConfig
 
 from ludwig.api_annotations import DeveloperAPI
@@ -110,28 +109,6 @@ def check_training_runway(config: "ModelConfig") -> None:  # noqa: F821
                 "trainer.steps_per_checkpoint. Please specify one or the other, or specify neither to "
                 "checkpoint/eval the model every epoch."
             )
-
-
-@register_config_check
-def check_optimizer_requirements(config: "ModelConfig") -> None:  # noqa: F821
-    """Checks if GPU is available when using Paged Optimizer."""
-    # GBMs don't specify optimizers, so skip this check for them and potentially other
-    # config types that don't specify optimizers.
-    if not hasattr(config.trainer, "optimizer"):
-        return
-
-    # TODO: Figure out why this check won't work
-    # Also it seems like this check is needed for 8-bit versions of the optimizers
-    # as well, so add a new property to the base class and then add that check here.
-    if config.trainer.optimizer.is_8bit and not torch.cuda.is_available():
-        raise ConfigValidationError(
-            "8-bit Optimizers require GPU but no GPU was detected. Please use a different optimizer."
-        )
-
-    if config.trainer.optimizer.is_paged and not torch.cuda.is_available():
-        raise ConfigValidationError(
-            "Paged Optimizers require GPU but no GPU was detected. Please use a different optimizer."
-        )
 
 
 @register_config_check
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -118,6 +118,9 @@ def check_optimizer_requirements(config: "ModelConfig") -> None:  # noqa: F821
     # TODO: Figure out why this check won't work
     # Also it seems like this check is needed for 8-bit versions of the optimizers
     # as well, so add a new property to the base class and then add that check here.
+    if not hasattr(config.trainer, "optimizer"):
+        return
+
     if config.trainer.optimizer.is_paged and not torch.cuda.is_available():
         raise ConfigValidationError(
             "Paged Optimizers require GPU but no GPU was detected. Please use a different optimizer."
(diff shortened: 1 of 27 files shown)


ludwig/config_validation/checks.py
```diff
@@ -111,6 +112,21 @@ def check_training_runway(config: "ModelConfig") -> None:  # noqa: F821
...
+@register_config_check
+def check_optimizer_requirements(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks if GPU is available when using Paged Optimizer."""
+    # TODO: Figure out why this check won't work
+    # Also it seems like this check is needed for 8-bit versions of the optimizers
+    # as well, so add a new property to the base class and then add that check here.
+    if not hasattr(config.trainer, "optimizer"):
+        return
+
+    if config.trainer.optimizer.is_paged and not torch.cuda.is_available():
```
tgaddair, line 131: This isn't correct. GPU might be allocated later on (think Ray backend).

ludwig/schema/optimizers.py
```diff
@@ -50,6 +51,11 @@ class BaseOptimizerConfig(schema_utils.BaseMarshmallowConfig, ABC):
     a `ValidationError`.
     """
 
+    @property
+    def is_paged(self) -> bool:
```
tgaddair, line 55: Is this needed?

arnavgarg1: I only added it for the schema validation check, but it may not be necessary

ludwig/config_validation/checks.py
arnavgarg1: Hmm, I'm trying to think of how we can catch this early enough and return a more straightforward error (the current one on CPU is quite bad and doesn't indicate the need to run this on GPUs)

arnavgarg1: One alternative is to wrap optimizer initialization in a try-except, catch that specific error that's being thrown, and then raise a RuntimeError. 

But, from the code, it seems like we're not inside the Ray process yet at the time we make this call since we're still initializing the trainer, but we only have access to the right resources post training initialization once we make the call to trainer.train().

arnavgarg1: This is the error that gets thrown btw:

```
NameError: name 'str2optimizer32bit' is not defined
```

Stack Trace:
```
def optimizer_update_32bit(
        optimizer_name: str,
        g: Tensor,
        p: Tensor,
        state1: Tensor,
        beta1: float,
        eps: float,
        step: int,
        lr: float,
        state2: Tensor = None,
        beta2: float = 0.0,
        weight_decay: float = 0.0,
        gnorm_scale: float = 1.0,
        unorm_vec: Tensor = None,
        max_unorm: float = 0.0,
        skip_zeros=False,
    ) -> None:
        """
        Performs an inplace optimizer update with one or two optimizer states.
    
        Universal optimizer update for 32-bit state and 32/16-bit gradients/weights.
    
        Parameters
        ----------
        optimizer_name : str
            The name of the optimizer: {adam}.
        g : torch.Tensor
            Gradient tensor.
        p : torch.Tensor
            Parameter tensor.
        state1 : torch.Tensor
            Optimizer state 1.
        beta1 : float
            Optimizer beta1.
        eps : float
            Optimizer epsilon.
        weight_decay : float
            Weight decay.
        step : int
            Current optimizer step.
        lr : float
            The learning rate.
        state2 : torch.Tensor
            Optimizer state 2.
        beta2 : float
            Optimizer beta2.
        gnorm_scale : float
            The factor to rescale the gradient to the max clip value.
        unorm_vec : torch.Tensor
            The tensor for the update norm.
        max_unorm : float
            The maximum update norm relative to the weight norm.
        skip_zeros : bool
            Whether to skip zero-valued gradients or not (default: False).
        """
    
        param_norm = 0.0
        if max_unorm > 0.0:
            param_norm = torch.norm(p.data.float())
    
    
        optim_func = None
        if g.dtype == torch.float32:
>           optim_func = str2optimizer32bit[optimizer_name][0]
E           NameError: name 'str2optimizer32bit' is not defined
```

arnavgarg1: Okay I have an idea for a fix, let me push it up

tgaddair: Optimizer is initialized here: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/distributed/base.py#L31

That's where you would make such a check.

Also, you should add a check for DeepSpeed here as well as it doesn't support these optimizers.

arnavgarg1: Updated!

tgaddair (approved):

ludwig/utils/llm_utils.py
```diff
@@ -376,3 +378,46 @@ def realign_target_and_prediction_tensors_for_inference(
...
+            device=model.device,
+        )
+
+        # Update the model's original embedding layer to use the BNB embedding layer using the module_path
+        # returned by find_embedding_layer_with_path.
+        module_path = module_path.split(".")
+        module = model
+        for module_name in module_path[:-1]:
+            module = getattr(module, module_name)
+        setattr(module, module_path[-1], bnb_embedding)
```
arnavgarg1, line 418: I don't know a better way to do it than this because different transformer architectures have their embedding layers at different module_path locations/names. 

While this works, the issue is that we're going to end up holding almost 2x the memory until we're actually able to get to the embedding layer module that we want to replace. Not sure if there's a nicer way to do it than this.

ludwig/models/llm.py
```diff
@@ -100,6 +101,10 @@ def load_pretrained_from_config(
     logger.info("Loading large language model...")
     pretrained_model_name_or_path = weights_save_path or config_obj.base_model
     model: PreTrainedModel = AutoModelForCausalLM.from_pretrained(pretrained_model_name_or_path, **load_kwargs)
+
+    # We may need to replace the embedding layer when using 8-bit optimizers from bitsandbytes.
+    update_embedding_layer(model, config_obj)
+
```
arnavgarg1, line 107: I might move this downstream to the trainer.prepare() call

justinxzhao (approved):

Issue #3591, opened 2023-09-07
test_tabtransformer_combiner_number_or_binary_without_category parameter update unit tests fail with torch 2.1.0 nightly 9/1 and onwards


justinxzhao:
[Sample run](https://github.com/ludwig-ai/ludwig/actions/runs/6103233216/job/16563284341?pr=3588)

cc: @jimthompson5802

jimthompson5802:
@justinxzhao  @w4nderlust   Initial observation...all the failed tests are calling out a particular norm layer weights/bias parameter not being updated with pytorch nightly.  The same norm layers appear to be updated when using a GA version of PyTorch.

I still need to dig in more to determine root cause.

jimthompson5802:
Looks like `torch.nn.LayerNorm` in nightly behaves differently.  In the GA version, the output of this layer has "non-zero" values, albeit very small, on the order of 1e-6.  

In the nightly version the output is zero.  A result, I believe this results in a zero gradient and no parameter updates.

Anyway, going further down the rabbit hole.

jimthompson5802:
@justinxzhao @w4nderlust 

Here is a MWE of the zero output from `torch.nn.LayerNorm`

```
import torch
from torch import nn, functional as F

if __name__ == "__main__":
    print(f"Torch version: {torch.__version__}")
    torch.manual_seed(42)
    inputs = torch.randn(16, 1, dtype=torch.float32)

    layer_norm = nn.LayerNorm(1)

    outputs = layer_norm(inputs)

    print(f"layer norm output:\n{outputs}")
```

When run with **`torch-nightly`**, this is what I see
```
Torch version: 2.2.0.dev20230908+cpu
layer norm output:
tensor([[0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.],
        [0.]], grad_fn=<NativeLayerNormBackward0>)
```

When I run the same program with **`torch 2.0.0`**, this is what I see
```
Torch version: 2.0.0+cpu
layer norm output:
tensor([[-2.7311e-05],
        [-1.3996e-05],
        [-1.4542e-05],
        [-8.0787e-06],
        [ 5.3804e-06],
        [ 1.1374e-05],
        [ 1.2255e-08],
        [ 5.2535e-06],
        [-1.9136e-06],
        [ 1.2105e-05],
[...]

PR #3596, opened 2023-09-09
FIX: Failure in TabTransformer Combiner Unit test

jimthompson5802:
# Code Pull Requests

This is a temporary work-around for Issue #3591.

Comment out corner cases that cause failure in tabtransformer combiner unit test.

Root cause appears to be a recent change in PyTorch `LayerNorm` implementation.  See cited issue for details.


jimthompson5802:
Based on discussion here https://github.com/pytorch/pytorch/issues/108956, removing all test cases that generate single column batch for the parameter update check test for TabTransformerCombiner.

justinxzhao (approved):
Thanks for the PR fix @jimthompson5802!

PR #3600, opened 2023-09-12
Eliminate short-circuiting for loading from local

Infernaught:
Previously, we would raise an error if a local path was passed base_model in our config. This PR adds a check that verifies if the local path is valid before checking if it is a valid HF path.

### DIFF
--- a/ludwig/schema/llms/base_model.py
+++ b/ludwig/schema/llms/base_model.py
@@ -58,8 +58,8 @@ def BaseModelDataclassField():
                 return model_name
             except OSError:
                 raise ConfigValidationError(
-                    f"Specified base model `{model_name}` is not a valid pretrained CausalLM listed on huggingface. "
-                    "or a valid directory containing a pretrained CausalLM."
+                    f"Specified base model `{model_name}` is not a valid pretrained CausalLM listed on huggingface "
+                    "or a valid local directory containing the weights for a pretrained CausalLM from huggingface."
                     "Please see: https://huggingface.co/models?pipeline_tag=text-generation&sort=downloads"
                 )
         raise ValidationError(
--- a/ludwig/schema/llms/base_model.py
+++ b/ludwig/schema/llms/base_model.py
@@ -1,3 +1,4 @@
+import os
 from dataclasses import field
 
 from marshmallow import fields, ValidationError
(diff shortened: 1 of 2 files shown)


ludwig/schema/llms/base_model.py
```diff
@@ -45,17 +46,20 @@ def validate(model_name: str):
...
                 return MODEL_PRESETS[model_name]
+            if os.path.isdir(model_name):
+                return model_name
             try:
                 AutoConfig.from_pretrained(model_name)
                 return model_name
             except OSError:
                 raise ConfigValidationError(
                     f"Specified base model `{model_name}` is not a valid pretrained CausalLM listed on huggingface. "
+                    "or a valid directory containing a pretrained CausalLM."
```
arnavgarg1, line 62: nit: Probably want to remove the full stop from the previous line. 

Also, just to people are absolutely clear, should we say `or a valid local directory containing the weights for a pretrained CausalLM from huggingface`? wdyt?

arnavgarg1 (commented):
Would actually be good for us to add a test for this in `test_llm.py`, which downloads weights to local, then uses that path as the base_model name, and then just makes sure the there are no errors when we do `ModelConfig.from_dict`. Equally, I wouldn't be opposed to the test instead just training for 3-4 steps. I'd be okay with both, but a test will be good

ludwig/schema/llms/base_model.py
Infernaught: That makes sense to me

justinxzhao (approved):
If you rebase, you should be able to get a clean CI

ludwig/schema/llms/base_model.py
```diff
@@ -45,17 +46,20 @@ def validate(model_name: str):
...
 
         If the name exists in `MODEL_PRESETS`, returns the corresponding value from the dict; otherwise checks if the
-        given name (which should be a full path) exists in the transformers library.
+        given name (which should be a full path) exists locally or in the transformers library.
         """
         if isinstance(model_name, str):
             if model_name in MODEL_PRESETS:
                 return MODEL_PRESETS[model_name]
+            if os.path.isdir(model_name):
+                return model_name
```
arnavgarg1, line 55: Hmm, wondering if we should add some more checks here, the most basic one being that the directory should not be empty. In an ideal world, we also add validation to ensure that the model config exists in this directly and it can be loaded back in correctly from this directory using the same code block from line 57, what do you think?

arnavgarg1: I do think we can do this in a fast-follow, but I do think for completeness these additional checks are important. What do you think? @Infernaught @justinxzhao

Infernaught: Cool. Justin and I expect HF to give us a failure message if the model objects are bad, so it's probably not that bad if we don't have these checks for the time being. I'm going to merge this for now, but I definitely think we should keep thinking about how to properly verify this.

arnavgarg1: Okay with the merge for now as well, but I do think we should follow-up with custom validation here and raise clear resolution methods for what went wrong and how to fix it. 

I do think we should add these checks now because they're honestly so trivial to add and we want to fail fast, always. There's nothing wrong with leaving it up to HF, but I guarantee that we are going to get Ludwig users messaging us asking us why they're seeing cryptic import error or not found errors from HF without them when there is a user-side error. The reason I personally really like these kinds of validation checks is that it lets us get ahead and provide really clear and crisp resolution methods to unblock/self-serve yourself, and that is what I would lean towards.

PR #3602, opened 2023-09-13
Updated characters, underscore and comma preprocessors to be TorchScriptable.

martindavis:
The `comma`, `underscore` and `characters` preprocessors are not currently TorchScriptable because they do not implement the TorchScript Module methods. As part of this PR, we now have a general `StringSplitTokenizer` that can be re-used across the various defaults we support. Unfortunately, since TorchScript only supports basic data types, `CharactersToListTokenizer` had to be written as a separate class instead of using Lamda Function.

Will help users run into this error less often:
```
ValueError: comma is not supported by torchscript. Please use one of {'sentencepiece', 'space_punct', 'clip', 
'gpt2bpe', 'bert', 'space'}.
```

This covered by the `tests/integration_tests/test_torchscript.py::test_torchscript_e2e_text` test because this PR updates `TORCHSCRIPT_COMPATIBLE_TOKENIZERS` to include the new TorchScriptable tokenizers.

### DIFF
--- a/ludwig/utils/tokenizers.py
+++ b/ludwig/utils/tokenizers.py
@@ -29,11 +29,7 @@ from ludwig.utils.nlp_utils import load_nlp_pipeline, process_text
 logger = logging.getLogger(__name__)
 torchtext_version = torch.torch_version.TorchVersion(torchtext.__version__)
 
-SPACE_PUNCTUATION_REGEX = re.compile(r"\w+|[^\w\s]")
-COMMA_REGEX = re.compile(r"\s*,\s*")
-UNDERSCORE_REGEX = re.compile(r"\s*_\s*")
-
-TORCHSCRIPT_COMPATIBLE_TOKENIZERS = {"space", "space_punct"}
+TORCHSCRIPT_COMPATIBLE_TOKENIZERS = {"space", "space_punct", "comma", "underscore", "characters"}
 TORCHTEXT_0_12_0_TOKENIZERS = {"sentencepiece", "clip", "gpt2bpe"}
 TORCHTEXT_0_13_0_TOKENIZERS = {"bert"}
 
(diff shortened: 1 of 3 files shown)


ludwig/utils/tokenizers.py
```diff
@@ -50,14 +46,61 @@ def __call__(self, text: str):
         pass
 
 
-class CharactersToListTokenizer(BaseTokenizer):
-    def __call__(self, text):
-        return [char for char in text]
+class StringSplitTokenizer(torch.nn.Module):
```
justinxzhao, line 49: Let's add a test for this tokenizer in `tokenizer_test.py`.

In case you haven't done this on Ludwig yet -- tests can be run with `pytest <python_file>::name_of_test` e.g. 

```
pytest tests/ludwig/utils/test_tokenizers.py::test_english_lemmatize_filter_tokenizer
```

ludwig/utils/tokenizers.py
```diff
@@ -50,14 +46,61 @@ def __call__(self, text: str):
...
+    def __init__(self, split_string, **kwargs):
+        super().__init__()
+        self.split_string = split_string
+
+    def forward(self, v: Union[str, List[str], torch.Tensor]) -> Any:
+        if isinstance(v, torch.Tensor):
+            raise ValueError(f"Unsupported input: {v}")
+
+        inputs: List[str] = []
```
justinxzhao, line 58: I see that you are adapting an existing implementation, though this seems more complicated than I would expect (for example, why do we have a `get_tokens()` function that returns its own input?). 

@geoffreyangus, ooc does this also look strange to you, or is this imposed on us by torchscript?

ludwig/utils/tokenizers.py
martindavis: Good call! Done!

justinxzhao (approved):

ludwig/utils/tokenizers.py
geoffreyangus: It looks like `NgramTokenizer`, which subclasses `SpaceStringToListTokenizer` (which in turn subclasses the new `StringSplitTokenizer`, seems to override `get_tokens`: https://github.com/ludwig-ai/ludwig/pull/3602/files#diff-5cbace55f4f4fd07725c061b9f981b83fe43cb53b0045cf1257c9fb5d4931f0dR132-R142

PR #3606, opened 2023-09-13
fix: Load 8-bit quantized models for eval after fine-tuning

jeffkinnison:
# Errors

After training or fine-tuning, the best model checkpoint is loaded for evaluation. When loading an 8-bit quantized model that was fine-tuned on GPU, the following errors occur:

1. With no handling, the call to load_state_dict [here](https://github.com/ludwig-ai/ludwig/blob/03a0da09f9d01141717f7caed462aa83cf891c26/ludwig/trainers/trainer.py#L887) raises `RuntimeError: Loading a quantized checkpoint into non-quantized Linear8bitLt is not supported. Please call module.cuda() before module.load_state_dict()"`
2. If 1. is handled, a number of unexpected keys are returned from load_state_dict and an AssertionError is raised

# Causes

These issues can be reproduced by running `tests/integration_tests/test_llm.py::test_llm_finetuning_strategies` with 8-bit quantization. Both issues are the result of custom handling in `bitsandbytes`. They are caused by

1. Moving an 8-bit parameter object to GPU [creates a number of metadata matrices](https://github.com/TimDettmers/bitsandbytes/blob/18e827d666fa2b70a12d539ccedc17aa51b2c97c/bitsandbytes/nn/modules.py#L290-L302) behind the scenes. These are added to model state on the fly during the move to GPU, and thus do not exist in a version of the model that has not been put on GPU. A check during `load_state_dict` in `bitsandbytes` raises the RuntimeError.
2. When saving a state dict, `bitsandbytes` adds a number of `weight_format` entries to the state dict behind the scenes. These are metadata entries that are used in `load_state_dict` to reconstruct the quantized parameters. Since these `weight_format` entries are never registered in model state, on load they are returned in the `unexpected_keys` list. On load for eval, we assert that no unexpected keys were returned.

# Workaround

This update puts in a workaround that addresses both issues. For 8-bit quantized models only, at the call to load_state_dict we first move the model to GPU and back to solve 1., then we ensure that the only unexpected keys are `weight_format` keys to handle 2. This should unblock 8-bit quantization for the time being, though we should double-check model quality.

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -884,8 +884,18 @@ class Trainer(BaseTrainer):
                     if self.distributed.is_model_parallel():
                         # Assume the full weights cannot fit in memory on GPU
                         self.model = self.model.cpu()
-                    _, unexpected_keys = self.model.load_state_dict(state_dict, strict=False)
-                    assert unexpected_keys == [], f"Unexpected keys found in state dict: {unexpected_keys}"
+                    if hasattr(self.model.config_obj, "quantization") and self.model.config_obj.quantization.bits == 8:
+                        if torch.cuda.is_available():
+                            self.model.model.cuda()
+                            self.model.model.cpu()
+                        _, unexpected_keys = self.model.load_state_dict(state_dict, strict=False)
+                        only_weights_format_keys = ["weights_format" in k for k in unexpected_keys]
+                        assert (
+                            unexpected_keys == [] or only_weights_format_keys
+                        ), f"Unexpected keys found in state dict: {unexpected_keys}"
+                    else:
+                        _, unexpected_keys = self.model.load_state_dict(state_dict, strict=False)
+                        assert unexpected_keys == [], f"Unexpected keys found in state dict: {unexpected_keys}"
             elif return_state_dict:
                 state_dict = self.model.cpu().state_dict()
 
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -884,6 +884,9 @@ class Trainer(BaseTrainer):
                     if self.distributed.is_model_parallel():
                         # Assume the full weights cannot fit in memory on GPU
                         self.model = self.model.cpu()
+
+                    # For a full explanation of this 8-bit workaround, see https://github.com/ludwig-ai/ludwig/pull/3606
+                    # TODO (jeffkinnison): Investigate why `self.model` is seemingly not placed on GPU before this point
                     if hasattr(self.model.config_obj, "quantization") and self.model.config_obj.quantization.bits == 8:
                         if torch.cuda.is_available():
                             self.model.model.cuda()
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -886,14 +886,15 @@ class Trainer(BaseTrainer):
                         self.model = self.model.cpu()
 
                     # For a full explanation of this 8-bit workaround, see https://github.com/ludwig-ai/ludwig/pull/3606
-                    # TODO (jeffkinnison): Investigate why `self.model` is seemingly not placed on GPU before this point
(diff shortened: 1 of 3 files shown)


ludwig/trainers/trainer.py
```diff
@@ -884,8 +884,21 @@ def train(
...
                         self.model = self.model.cpu()
-                    _, unexpected_keys = self.model.load_state_dict(state_dict, strict=False)
-                    assert unexpected_keys == [], f"Unexpected keys found in state dict: {unexpected_keys}"
+
+                    # For a full explanation of this 8-bit workaround, see https://github.com/ludwig-ai/ludwig/pull/3606
+                    # TODO (jeffkinnison): Investigate why `self.model` is seemingly not placed on GPU before this point
+                    if hasattr(self.model.config_obj, "quantization") and self.model.config_obj.quantization.bits == 8:
+                        if torch.cuda.is_available():
+                            self.model.model.cuda()
+                            self.model.model.cpu()
```
arnavgarg1, line 893: Ooof, maybe one callout here might be that the .cuda() call is unique and overriden for the Linear8Bit layers which internally does some stuff for 8BitParameters?

tgaddair: Yeah, it's not clear to me why we need to move to GPU then back to CPU like this. Comment would be great so I don't need to read the full PR description.

ludwig/trainers/trainer.py
```diff
@@ -884,8 +884,21 @@ def train(
...
+                        _, unexpected_keys = self.model.load_state_dict(state_dict, strict=False)
+                        only_weights_format_keys = ["weights_format" in k for k in unexpected_keys]
+                        assert (
+                            unexpected_keys == [] or only_weights_format_keys
+                        ), f"Unexpected keys found in state dict: {unexpected_keys}"
```
tgaddair, line 898: Add something about the `only_weights_format_keys` to the error message.

justinxzhao (approved):

ludwig/trainers/trainer.py
jeffkinnison: Added a comment and removed the move to CPU. It turns out the model was on GPU all along: `self.model.device` reports that it is on CPU, but `self.model.model.device` and deeper modules in the model all report that that they are on GPU.

ludwig/trainers/trainer.py
jeffkinnison: Added in [b8b487d](https://github.com/ludwig-ai/ludwig/pull/3606/commits/b8b487d10d15a81fa3f68a5631e56880ad4f09f5)

ludwig/trainers/trainer.py
```diff
@@ -884,8 +884,31 @@ def train(
...
+                    # For a full explanation of this 8-bit workaround, see https://github.com/ludwig-ai/ludwig/pull/3606
+                    # TODO (jeffkinnison): Determine why `SCB` and `CB` are deleted from parameter state
+                    if hasattr(self.model.config_obj, "quantization") and self.model.config_obj.quantization.bits == 8:
+                        # If the model was previously placed on GPU, 8-bit parameter state will be updated with several
+                        # matrices containing quantization information. These are recorded matrices are recorded in the
+                        # training checkpoint state dicts, but do not necessarily exist in the parameter object, leading
+                        # to a RuntimeError in `load_state_dict`. Explicitly call `model.cuda()` to make sure the
+                        # matrices are part of model state. This workaround is necessary because the matrices are
+                        # deleted during the model's forward pass.
+                        if self.device == torch.device("cuda"):
```
tgaddair, line 897: Check `self.device.type == "cuda"` as the device might be `cuda:0`, etc.

ludwig/trainers/trainer.py
tgaddair: Hmm, am I missing something, I still don't see anything in the assert message about it?

tgaddair (approved):

PR #3607, opened 2023-09-14
Add pytorch profiler and additional tensorboard logs for GPU memory usage.

justinxzhao:
Add two memory profiling mechanics to Ludwig.

## 1. Official pytorch profiler

image

### Usage

Add `enable_profiling: True` to the trainer config.

```yaml
trainer:
  enable_profiling: True
  profiler:
    repeat: 5
    active: 3
```

### Notes
- Code is modeled off of [this tutorial](https://pytorch.org/tutorials/intermediate/tensorboard_profiler_tutorial.html).
- The full pytorch profiler can be rather heavyweight, so we disable it by default.
- There was some trial and error to determine default values for the [schedule](https://pytorch.org/docs/stable/profiler.html), which seem to work ok with basic models.

## Memory profiling using torch cuda memory stats

image

### Notes
- Getting information from torch.cuda memory data is lightweight and additional logging does not seem to negatively affect training time.
- torch.cuda memory information is logged every training step under its own section in the tensorboard called `cuda`.
- I'm unsure whether these logs will always be consistent with the official pytorch profiler. This may also hard to compare since the official pytorch profiler shows memory usage over a time window rather than on a per-step basis.

## Inspecting using tensorboard

Launch tensorboard:

```sh
tensorboard --logdir=results/
```

In colab:

```sh
!pip install tensorflow  # tensorflow must be installed for the tensorboard extension to load correctly.
%load_ext tensorboard
%tensorboard --logdir=results
```

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -425,7 +425,64 @@ jobs:
 
       - name: Integration Tests (D)
         run: |
-          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=7200 pytest -v --timeout 300 --durations 100 -m "not slow and not combinatorial and not horovod and not llm and not integration_tests_a and not integration_tests_b and not integration_tests_c" --junitxml pytest.xml tests/integration_tests
+          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=7200 pytest -v --timeout 300 --durations 100 -m "not slow and not combinatorial and not horovod and not llm and integration_tests_d" --junitxml pytest.xml tests/integration_tests
+
+  integration-tests-e:
+    name: Integration Tests (E)
+    runs-on: ubuntu-latest
+
+    env:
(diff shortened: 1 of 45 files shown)


tgaddair (commented):
One clarifying question: is enabling the pytorch profiler the only way to get GPU utilization metrics? Feels like this should be pretty cheap to obtain and ideally something we always collect. I can understand why the more granular memory metrics would be expensive, but just want to try and find a way to decouple those things.

justinxzhao:
> One clarifying question: is enabling the pytorch profiler the only way to get GPU utilization metrics? Feels like this should be pretty cheap to obtain and ideally something we always collect. I can understand why the more granular memory metrics would be expensive, but just want to try and find a way to decouple those things.

I tested logging torch.cuda.* data as tfevents on a colab T4 to produce additional GPU memory utilization graphs in tensorboard. The additional logging doesn’t slow down training.

I've attached an image that's fine-tuning lambda-2-7b on codealpaca for 1K steps.

I've updated the PR to keep the option to use the full torch profiler (off by default) and log torch.cuda.* stuff always.

arnavgarg1 (approved):
LGTM! I think the only thing to add is adding a logger.warning call at profiler initialization time to indicate that profiling is expensive and will slow down training.

PR #3608, opened 2023-09-14
[CI] Run sudo apt-get update in GHAs.

justinxzhao:


justinxzhao:
Thanks @martindavis for the fix!

PR #3610, opened 2023-09-14
[Need to test] Remove max_memory usage and adapter float16 initialization by default

arnavgarg1:


ludwig/models/llm.py
```diff
@@ -321,7 +321,6 @@ def to_device(self, device):
                     self.model = PeftModel.from_pretrained(
                         self.model,
                         tmpdir,
-                        torch_dtype=torch.float16,
```
tgaddair, line 324: We probably need a way to plumb this in, otherwise it's going to load in fp32, right?

arnavgarg1: That is true yeah, I think we actually need a way to plumb in 

```
quantization:
    bits: 16
    type: bfloat16
```

and

```
quantization:
    bits: 16
    type: float16
```

and then trickle this down appropriately to both the base model and the adapter.

This will also allow users to experiment with non-deepspeed-based training if they'd like (say they have a H100 or A100)

arnavgarg1: What do you think?

PR #3613, opened 2023-09-15
Add support for gradient checkpointing for LLM fine-tuning

arnavgarg1:
This PR adds support in the finetune trainer to optionally enabled `gradient_checkpointing`.

# What is gradient checkpointing?
Gradient checkpointing works by recomputing the activations of the model during the backward pass, rather than storing them in memory during the forward pass. This is a tradeoff between compute and memory, as the activations need to be recomputed during the backward pass, but the memory footprint is reduced. This is set to false by default because it is not always beneficial to use gradient checkpointing, and it can sometimes slow down training.
image

# How can you use gradient checkpointing in the config?
Gradient checkpointing is disabled by default. To enable it, you can simply set `enable_gradient_checkpointing` to True.

```yaml
trainer:
    enable_gradient_checkpointing: true
```

# When should I enable gradient checkpointing?
This is useful when training very large models that run into out of memory errors very quickly during training. It is particularly helpful when doing non-quantization based training (adapter based or full fine-tuning).

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -225,7 +225,7 @@ class Trainer(BaseTrainer):
             elif not hasattr(self.compiled_model, "model") and not hasattr(
                 self.compiled_model.model, "gradient_checkpointing_enable"
             ):
-                logger.warning("Gradient checkpointing is not supported for this model. Skipping...")
+                logger.warning("Gradient checkpointing is not supported by this model. Skipping...")
             elif hasattr(self.compiled_model.model, "gradient_checkpointing_enable"):
                 self.compiled_model.model.gradient_checkpointing_enable()
                 self.compiled_model.model.enable_input_require_grads()
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -216,10 +216,10 @@ class Trainer(BaseTrainer):
         update_embedding_layer(self.compiled_model, self.config)
 
         # Enable gradient checkpointing
-        logger.info("Enabling gradient checkpointing")
-        self.compiled_model.model.gradient_checkpointing_enable()
+        logger.info("Enable input requires grad to true; no gradient checkpointing")
+        # self.compiled_model.model.gradient_checkpointing_enable()
         self.compiled_model.model.enable_input_require_grads()
-        logger.info("Enabled gradient checkpointing")
+        logger.info("Enabled input requires grad to true")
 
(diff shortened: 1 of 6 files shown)


ludwig/trainers/trainer.py
```diff
@@ -214,6 +215,24 @@ def prepare(self):
...
+            if not isinstance(self.compiled_model, LLM):
+                logger.warning("Gradient checkpointing is currently only supported for model_type: llm. Skipping...")
+            elif not hasattr(self.compiled_model, "model") and not hasattr(
+                self.compiled_model.model, "gradient_checkpointing_enable"
+            ):
+                logger.warning("Gradient checkpointing is not supported by this model. Skipping...")
+            elif hasattr(self.compiled_model.model, "gradient_checkpointing_enable"):
+                self.compiled_model.model.gradient_checkpointing_enable()
+                self.compiled_model.model.enable_input_require_grads()
+                logger.info("Gradient checkpointing enabled for training.")
```
justinxzhao, line 232: Nice logs.

ludwig/trainers/trainer.py
```diff
@@ -214,6 +215,24 @@ def prepare(self):
...
+
+        # Enable gradient checkpointing if configured
+        if self.config.enable_gradient_checkpointing:
+            # TODO(Arnav): Add support for gradient checkpointing in the compiled model
+            # when the model is an ECD model using torch.utils.checkpoint (torch.utils.checkpoint.sequential())
+            if not isinstance(self.compiled_model, LLM):
+                logger.warning("Gradient checkpointing is currently only supported for model_type: llm. Skipping...")
+            elif not hasattr(self.compiled_model, "model") and not hasattr(
+                self.compiled_model.model, "gradient_checkpointing_enable"
+            ):
```
justinxzhao, line 227: Is it worth distinguishing two different warnings between these two cases?

justinxzhao (approved):
Nice description! Do you have a sense for which LLMs support gradient checkpointing and which ones don't?

ludwig/trainers/trainer.py
arnavgarg1: I don't think so, since self.compiled_model.model only exists for LLM model types, otherwise its just torch.nn modules with the actual architecture beneath the surface, so I would treat these the same

arnavgarg1:
> Nice description! Do you have a sense for which LLMs support gradient checkpointing and which ones don't?

I think almost all models coming from the `transformer` package support it, but I have the additional check here just in case there is an edge case that hasn't been handled.

PR #3618, opened 2023-09-15
Refactor integration tests into matrix

tgaddair:


.github/workflows/pytest.yml
```diff
@@ -480,9 +262,9 @@ jobs:
           pip list
         shell: bash
 
-      - name: Integration Tests (E)
+      - name: Integration Tests
         run: |
-          RUN_PRIVATE=$IS_NOT_FORK LUDWIG_TEST_SUITE_TIMEOUT_S=7200 pytest -v --timeout 300 --durations 100 -m "not slow and not combinatorial and not horovod and not llm and not integration_tests_a and not integration_tests_b and not integration_tests_c and not integration_tests_d" --junitxml pytest.xml tests/integration_tests
```
tgaddair, line 485: @justinxzhao why is it that only test suite E had all these `not integration_tests_a ...` markers but the others don't?

justinxzhao: Ah, this is because tests for groups a/b/c/d, are explicitly marked with `pytest.mark` while the last bucket (in this case "e"), is a catch-all for all of the rest of unmarked tests. The thinking here was that folks adding a new file in `integration_tests/` shouldn't have to think about assigning a marker.

.github/workflows/pytest.yml
```diff
@@ -199,244 +199,26 @@ jobs:
...
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        test-markers:
+          - "integration_tests_a"
+          - "integration_tests_b"
+          - "integration_tests_c"
+          - "integration_tests_d"
+          - "integration_tests_e"
```
justinxzhao, line 213: This should work if we want to keep the markers unchanged:

```
# catch-all
- "not integration_tests_a and not integration_tests_b and not integration_tests_c and not integration_tests_d"
```

justinxzhao (commented):
Thanks for this refactor!

PR #3623, opened 2023-09-16
Fixed range validation for text generation penalty parameters

tgaddair:
These param ranges were overly restrictive. See https://huggingface.co/docs/transformers/internal/generation_utils for more details.


arnavgarg1 (approved):
Interesting, good catch!

jeffkinnison (approved):
+1 for a more general float field type

PR #3625, opened 2023-09-17
QoL: Default to using fast tokenizer for Llama models

arnavgarg1:
Quality of life improvement - we used to set use_fast to False since the tokenizer was not natively written so it would take 2-4 minutes to load, however in newer `transformers` versions there is first class support so we no longer have to do this.


justinxzhao (approved):
Thanks for following up on this! I was wondering about this and why setting `use_fast=True` didn't seem faster.

PR #3628, opened 2023-09-17
fixed typo in EfficientNet's model variant from v2_ to v2_s

saad-palapa:
# Code Pull Requests

Fixed a typo in `EfficientNet`'s model variant that broke training

# Documentation Pull Requests


justinxzhao (approved):

justinxzhao:
Hi @saad-palapa, thanks for the change.

Looks like this is indeed an error, with v2_s as the [correct variant name](https://pytorch.org/vision/main/models/generated/torchvision.models.efficientnet_v2_s.html).

PR #3632, opened 2023-09-18
[WIP] Add support for LLM models with custom code like `microsoft/phi-1_5` and `tiiuae/falcon-7b` 

arnavgarg1:
Very messy code, WIP, just trying to understand how the code flows and some functionality related things. Will clean this up significantly before actually creating a PR.

### DIFF
--- a/ludwig/schema/llms/base_model.py
+++ b/ludwig/schema/llms/base_model.py
@@ -42,7 +42,7 @@ def BaseModelDataclassField():
         "pretrained model."
     )
 
-    def validate(model_name: str):
+    def validate(model_name: str, trust_remote_code: bool = False):
         """Validates and upgrades the given model name to its full path, if applicable.
 
         If the name exists in `MODEL_PRESETS`, returns the corresponding value from the dict; otherwise checks if the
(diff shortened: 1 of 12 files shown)


ludwig/schema/llms/base_model.py
```diff
@@ -54,7 +54,9 @@ def validate(model_name: str):
             if os.path.isdir(model_name):
                 return model_name
             try:
-                AutoConfig.from_pretrained(model_name)
+                # TODO(Arnav): Why is this called twice, and why is trust_remote_code false the second time?
+                # breakpoint()
+                AutoConfig.from_pretrained(pretrained_model_name_or_path=model_name, trust_remote_code=True)
```
arnavgarg1, line 59: See if there's a potentially sensible/clear CTA to raise to users when trying to use models that require `trust_remote_code` to be set.

TashaSkyUp: Its hard to tell exactly when it happens but i think it may be something to do with the interaction between retrying and the fact that kwargs are popped out rather often:

image

PR #3634, opened 2023-09-18
Remove unnecessary peft config updating 

Infernaught:
Get rid of parameter loading loop for Peft config because it was unnecessary


Infernaught:
Interesting. Give me a moment to figure out why it's still listing so many commits.

PR #3636, opened 2023-09-18
Remove unnecessary peft config updating

Infernaught:
Get rid of parameter loading loop for Peft config because it was unnecessary

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -199,9 +199,18 @@ jobs:
           name: Unit Test Results (Python ${{ matrix.python-version }} ${{ matrix.test-markers }})
           path: pytest.xml
 
-  integration-tests-a:
-    name: Integration Tests (A)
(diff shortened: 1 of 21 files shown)


Infernaught:
Ok. I've been trying to rebase for a while now, and I don't know why it's still showing old commits. Thoughts?

arnavgarg1:
Hey @Infernaught, are you able to explain a bit better what's change and why we're removing this? I'm having a bit of a hard time following this PR

Infernaught:
Yes. I don't know why the commit history is showing so many previous commits, but I'm just removing an unnecessary piece of logic from my previous PR that allowed for loading pretrained adapter weights. I had previously loaded a bunch of parameters into the peft config from our defaults, but after reviewing my code, I realized that I had already eliminated the logic that required this loading in the first place. All I need is the name of the HF repo, so I just eliminated the loading loop.

PR #3646, opened 2023-09-20
Add check to ensure selected backend works with quantization for LLMs 

arnavgarg1:
Quantization can only be used with a local backend. This PR modifies our existing checks since they were not robust enough to catch the 4 total cases we should be accounting for.

### DIFF
--- a/tests/ludwig/schema/test_model_config.py
+++ b/tests/ludwig/schema/test_model_config.py
@@ -912,3 +912,38 @@ def test_llm_finetuning_output_feature_config():
 
     config[OUTPUT_FEATURES] = [{NAME: "text_output", TYPE: "text"}]
     ModelConfig.from_dict(config)
+
+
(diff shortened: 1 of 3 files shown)


justinxzhao (approved):

tests/ludwig/schema/test_model_config.py
```diff
@@ -912,3 +912,38 @@ def test_llm_finetuning_output_feature_config():
...
+
+    # Backend explicitly set to Ray backend
+    config["backend"] = {"type": "ray"}
+    with pytest.raises(ConfigValidationError):
+        ModelConfig.from_dict(config)
+
+    # Start local ray process
+    import ray
+
+    ray.init()
```
tgaddair, line 942: This is risky, as if an error happens this ray cluster will never be terminated. You should use one of the test context managers we have for managing ray clusters.

PR #3649, opened 2023-09-21
[FEATURE] Support Merging LoRA Weights Into Base Model (Issue-3603)

alexsherstinsky:
### Description
This contribution provides an implementation of `LLM.merge_and_unload()`, which is triggered from `LudwigModel.train()`, based on configuration.  The "merge and unload" behavior merges the fine-tuned LoRA weights into the base model so that the users can load one complete model (e.g., from HuggingFace) in a single `AutoModelForCausalLM.from_pretrained()`  call, rather than using two calls (first loading the base model with `AutoModelForCausalLM.from_pretrained()` and then loading the fine-tuned weights with `PeftModelForCausalLM.from_pretrained()`).  This capability facilitates portability of the inference function between Ludwig (`LudwigModel.load()` followed by `model.predict()` for inference) and others (e.g., HuggingFace using `AutoModelForCausalLM.from_pretrained()` followed by `transformers.pipeline()` for inference).

The configuration consists of extending the `adapter` section with the optional `postprocessor` section as follows:
```
adapter:
  type: lora
  postprocessor:
    merge_adapter_into_base_model: true
    progressbar: true
```
(If the `merge_adapter_into_base_model` is kept and set to `false`, then the `progressbar` directive can be omitted.)

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests


arnavgarg1:
Hi @alexsherstinsky! Great work on this PR and thanks for your contribution - I will try and review by EOW!

alexsherstinsky:
> Hi @alexsherstinsky! Great work on this PR and thanks for your contribution - I will try and review by EOW!

Hi @arnavgarg1!  Thank you so much for your help and support.  Could you please wait a little bit -- I want to try to add another test, specific to this particular new feature -- I will let you know once it is ready (hopefully later today).  Thanks a lot!

tgaddair (approved):
Nice! This is an awesome addition, and I think the way it's integrated into the config is quite clean.

ludwig/models/llm.py
```diff
@@ -451,6 +451,21 @@ def generate(
...
+        model as a standalone model.  The implementation calls merge_and_unload() of the underlying LoraModel class
+        (in peft).
+
+        Args:
+            progressbar (bool): whether to show a progressbar indicating the unload and merge process
+        """
+        from peft import LoraModel
+
+        if isinstance(self.model.base_model, LoraModel):
+            self.model.base_model.merge_and_unload(progressbar=progressbar)
```
arnavgarg1, line 465: Nice! In a follow-up PR, I think it could be also good to add support for AdaLoRAModel since I believe it also allows `merge_and_unload`

ludwig/models/llm.py
```diff
@@ -451,6 +451,21 @@ def generate(
...
+        else:
+            raise ValueError("This operation requires an LLM model trained using LoRA as the base model.")
```
arnavgarg1, line 467: nit: Perhaps the right thing to say here is `This operation requires a base LLM model trained with a LoRA adapter` since we always require a base LLM model, but also require it to be adapter based fine-tuning specifically with LoRA? What do you think?

tests/integration_tests/test_llm.py
```diff
@@ -628,23 +780,35 @@ def test_load_pretrained_adapter_weights(adapter):
...
-        dict is loaded into a model. These keys were causing `torch.equal` to raise an exception, so we skip them in the
-        evaluation.
-        """
-        return "weight_format" not in i[0]
+    # Source: https://discuss.pytorch.org/t/check-if-models-have-same-weights/4351/6
 
-    model_1_filtered_state_dict = filter(filter_for_weight_format, model_1.state_dict().items())
-    model_2_filtered_state_dict = filter(filter_for_weight_format, model_2.state_dict().items())
+    if model_1.__class__.__name__ != model_2.__class__.__name__:
+        return False
```
arnavgarg1, line 800: I'm okay with commenting out some of these lines of code for now since we still don't have GPU tests setup for Ludwig (currently a work in progress), but this check for filtering the state dict for weight formats with 8 bit quantization is necessary to make sure the tests for comparing models works correctly when we test fine-tuning with 8-bit quantization!

arnavgarg1: I'd maybe create a comment with a TODO here to re-enable those lines of code when GPU tests are enabled? That should be fine for now

arnavgarg1 (approved):
Really nice work on this! I especially love the way you've written out your tests for making sure it is working as expected - thanks for being so thorough and for this awesome contribution! 

I left some minor comments, but none of them are blocking merging this PR!

arnavgarg1 (approved):
Really nice work on this! I especially love the way you've written out your tests for making sure it is working as expected - thanks for being so thorough and for this awesome contribution! 

I left some minor comments, but none of them are blocking merging this PR!

ludwig/models/llm.py
arnavgarg1: Oh good catch, you're right :) Didn't realize it was a subclass so makes sense. Thank you!

tests/integration_tests/test_llm.py
alexsherstinsky: @arnavgarg1 The issue is that the `filter_for_weight_format()` method (which focuses on 8-bit quantization) is not currently used, and it was causing linter errors.  Could you please suggest what we should do?  Should we keep it commented out for now (I added a TODO in the updated PR), or try to re-enable it and use it?  Thank you.

ludwig/models/llm.py
alexsherstinsky: 🙇

tests/integration_tests/test_llm.py
arnavgarg1: I think it should be okay to keep it commented out for now since we have the TODO - maybe I can take a look at that in a follow-up PR, but for now I don't want to block merging this awesome change into Ludwig because of it since it's causing no harm!

arnavgarg1: cc: @jeffkinnison just a quick FYI so you're not surprised by this change in the test. I will work on adding it back when GPU tests get enabled

PR #3650, opened 2023-09-21
Add sample_size as a global preprocessing parameter

Infernaught:
Adds sample_size as a global preprocessing parameter, allowing users to specify exactly how many samples they want to train on instead of having to calculate the sample_ratio. Adds two integration tests to verify sample_size works as intended.

### DIFF
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -691,3 +691,11 @@ def check_prompt_requirements(config: "ModelConfig") -> None:  # noqa: F821
                     "A template must contain at least one reference to a column or the sample keyword {__sample__} for "
                     "a JSON-serialized representation of non-output feature columns."
                 )
+
+
+@register_config_check
+def check_sample_ratio_and_cap_compatible(config: "ModelConfig") -> None:
+    sample_ratio = config.preprocessing.sample_ratio
+    sample_cap = config.preprocessing.sample_cap
+    if sample_cap and sample_ratio < 1.0:
+        raise ConfigValidationError("sample_cap cannot be used when sample_ratio < 1.0")
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -696,6 +696,6 @@ def check_prompt_requirements(config: "ModelConfig") -> None:  # noqa: F821
 @register_config_check
 def check_sample_ratio_and_cap_compatible(config: "ModelConfig") -> None:
     sample_ratio = config.preprocessing.sample_ratio
-    sample_cap = config.preprocessing.sample_cap
-    if sample_cap and sample_ratio < 1.0:
-        raise ConfigValidationError("sample_cap cannot be used when sample_ratio < 1.0")
+    sample_size = config.preprocessing.sample_size
+    if sample_size and sample_ratio < 1.0:
+        raise ConfigValidationError("sample_size cannot be used when sample_ratio < 1.0")
--- a/ludwig/config_validation/checks.py
+++ b/ludwig/config_validation/checks.py
@@ -694,8 +694,8 @@ def check_prompt_requirements(config: "ModelConfig") -> None:  # noqa: F821
 
 
 @register_config_check
-def check_sample_ratio_and_cap_compatible(config: "ModelConfig") -> None:
+def check_sample_ratio_and_size_compatible(config: "ModelConfig") -> None:
     sample_ratio = config.preprocessing.sample_ratio
     sample_size = config.preprocessing.sample_size
-    if sample_size and sample_ratio < 1.0:
+    if sample_size is not None and sample_ratio < 1.0:
         raise ConfigValidationError("sample_size cannot be used when sample_ratio < 1.0")
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1402,7 +1402,9 @@ def _get_sampled_dataset_df(dataset_df, df_engine, sample_ratio, sample_cap, ran
 
     if sample_cap:
         if sample_cap < len(dataset_df):
-            dataset_df = dataset_df.sample(n=sample_cap, random_state=random_seed)
+            # Cannot use 'n' parameter when using dask DataFrames -- only 'frac' is supported
+            sample_ratio = sample_cap / len(dataset_df)
+            dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
         else:
             logger.warning("sample_cap is larger than dataset size, ignoring sample_cap")
 
--- a/ludwig/data/preprocessing.py
+++ b/ludwig/data/preprocessing.py
@@ -1201,8 +1201,8 @@ def build_dataset(
 
     if mode == "training":
         sample_ratio = global_preprocessing_parameters["sample_ratio"]
-        sample_cap = global_preprocessing_parameters["sample_cap"]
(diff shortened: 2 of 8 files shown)


ludwig/data/preprocessing.py
```diff
@@ -1211,6 +1211,15 @@ def build_dataset(
...
             dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
 
+        sample_cap = global_preprocessing_parameters["sample_cap"]
+        if sample_cap:
+            if sample_ratio < 1.0:
+                raise ValueError("sample_cap cannot be used when sample_ratio < 1.0")
+            if sample_cap < len(dataset_df):
+                dataset_df = dataset_df.sample(n=sample_cap, random_state=random_seed)
+            else:
+                logger.info("sample_cap is larger than dataset size, ignoring sample_cap")
```
arnavgarg1, line 1221: Perhaps logger.warning?

ludwig/data/preprocessing.py
```diff
@@ -1211,6 +1211,15 @@ def build_dataset(
             logger.debug(f"sample {sample_ratio} of data")
             dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
 
+        sample_cap = global_preprocessing_parameters["sample_cap"]
+        if sample_cap:
+            if sample_ratio < 1.0:
+                raise ValueError("sample_cap cannot be used when sample_ratio < 1.0")
```
arnavgarg1, line 1217: Wondering if we can push this up into a schema validation check, i.e., if preprocessing sample_ratio is specified and it is < 1 and sample_cap is also specified, then raise a ConfigValidationError?

arnavgarg1 (approved):
Minor comments, but generally LGTM

ludwig/data/preprocessing.py
justinxzhao: +1. If we can implement this as an auxiliary validation, that would allow the config to fail as early as possible.

ludwig/data/preprocessing.py
```diff
@@ -1211,6 +1211,15 @@ def build_dataset(
             logger.debug(f"sample {sample_ratio} of data")
             dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
 
+        sample_cap = global_preprocessing_parameters["sample_cap"]
```
justinxzhao, line 1214: Could you refactor this section out into a separate function?

```python
dataset_df = get_sampled_dataset_df(dataset_df, sample_ratio, sample_cap)
```

justinxzhao (commented):
Thanks! I like the change overall.

ludwig/data/preprocessing.py
Infernaught: Yep. Done!

ludwig/config_validation/checks.py
Infernaught, line 1: @arnavgarg1 @justinxzhao Is this what you guys were looking for? I've testing this locally and it seems to have the right functionality.

Infernaught, line 1: @arnavgarg1 @justinxzhao is that what you guys were looking for? From what I can tell, this works as expected when tested locally

justinxzhao: Nice! Last request from me is to add a simple test to https://github.com/ludwig-ai/ludwig/blob/master/tests/ludwig/config_validation/test_checks.py since we're adding code to checks.py

ludwig/data/preprocessing.py
```diff
@@ -1396,6 +1389,28 @@ def embed_fixed_features(
...
+            raise ValueError(
+                f"sample_ratio {sample_ratio} is too small for dataset of length {len(dataset_df)}. "
+                f"Please increase sample_ratio or use a larger dataset."
+            )
+
+        logger.debug(f"sample {sample_ratio} of data")
+        dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
+
+    if sample_cap:
+        if sample_cap < len(dataset_df):
```
tgaddair, line 1404: `len(dataset_df)` is a very expensive op for Dask dataframes (this is why we have an explicit check above to skip calling this when `df_engine.partitioned`), so calling it twice is quick succession is definitely not ideal. Let's do this instead:

```
df_len = len(dataset_df)
if sample_cap < df_len:
  # Cannot use 'n' parameter when using dask DataFrames -- only 'frac' is supported
  sample_ratio = sample_cap / df_len
  dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
```

ludwig/data/preprocessing.py
```diff
@@ -1396,6 +1389,28 @@ def embed_fixed_features(
...
+            # Cannot use 'n' parameter when using dask DataFrames -- only 'frac' is supported
+            sample_ratio = sample_cap / len(dataset_df)
+            dataset_df = dataset_df.sample(frac=sample_ratio, random_state=random_seed)
```
tgaddair, line 1407: Note that for Dask this will not be exact, but that's probably okay.

tgaddair (commented):
Not a huge fan of the name `sample_cap` personally. Can we call it something like `sample_size` instead?

ludwig/config_validation/checks.py
```diff
@@ -691,3 +691,11 @@ def check_prompt_requirements(config: "ModelConfig") -> None:  # noqa: F821
                     "A template must contain at least one reference to a column or the sample keyword {__sample__} for "
                     "a JSON-serialized representation of non-output feature columns."
                 )
+
+
+@register_config_check
+def check_sample_ratio_and_cap_compatible(config: "ModelConfig") -> None:
+    sample_ratio = config.preprocessing.sample_ratio
+    sample_cap = config.preprocessing.sample_cap
+    if sample_cap and sample_ratio < 1.0:
```
tgaddair, line 700: Edge case, but this would allow something like:

```
sample_cap: 0
sample_ratio: 0.5
```

So would be more correct to ay:

```
if sample_cap is not None and sample_ratio < 1.0:
```

ludwig/schema/metadata/configs/preprocessing.yaml
```diff
@@ -44,6 +44,22 @@ sample_ratio:
...
+    description_implications:
+        Decreases the amount of data you are inputting into
+        the model. Could be useful if you have more data than you need and you are
+        concerned with computational costs. More useful than sample_ratio if you
+        know the exact number of samples you want to train on instead of knowing the proportion.
+    example_value:
+        - 1000
+    expected_impact: 2
+    suggested_values: Depends on data size
+    ui_display_name: Sample Cap
```
justinxzhao, line 62: nit: Sample Size.

tests/integration_tests/test_preprocessing.py
```diff
@@ -162,6 +162,108 @@ def test_sample_ratio_deterministic(backend, tmpdir, ray_cluster_2cpu):
...
+    model = LudwigModel(config, backend=backend)
+    train_set, val_set, test_set, training_set_metadata = model.preprocess(
+        data_csv,
+        skip_save_processed_input=True,
+    )
+
+    count = len(train_set) + len(val_set) + len(test_set)
+    assert sample_size == count
+
+    # Check that sample cap is disabled when doing preprocessing for prediction
```
justinxzhao, line 199: nit: sample size

justinxzhao (approved):

PR #3653, opened 2023-09-22
fix: The final batch of an epoch is skipped when batch size is 1

jeffkinnison:
When training with batch size 1 using the `RandomAcessBatcher`, the final batch of an epoch is always dropped because of [this condition](https://github.com/ludwig-ai/ludwig/blob/ee92f7d7e29b1ece6ec1178e6bb60d9426d9e4ef/ludwig/data/batcher/random_access.py#L70-L77). Skipping batches leads to downstream inconsistencies with evaluation and metrics logging, and the trainer also runs an additional partial epoch to account for the missing training steps. Notably, this issue heavily impacts LLM fine-tuning, which is almost always run with batch size 1.

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -31,7 +31,7 @@ import psutil
 import torch
 from torch.utils.tensorboard import SummaryWriter
 
-from ludwig.constants import AUTO, LOSS, MAX_CPU_BATCH_SIZE, MINIMIZE, MODEL_ECD, TEST, TRAINING, VALIDATION
+from ludwig.constants import AUTO, LOSS, MAX_CPU_BATCH_SIZE, MINIMIZE, MODEL_ECD, MODEL_LLM, TEST, TRAINING, VALIDATION
 from ludwig.data.dataset.base import Dataset
 from ludwig.distributed.base import DistributedStrategy, LocalStrategy
 from ludwig.globals import (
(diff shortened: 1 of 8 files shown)


ludwig/trainers/trainer.py
```diff
@@ -848,7 +848,9 @@ def train(
                 should_shuffle=self.should_shuffle,
                 random_seed=self.random_seed,
                 distributed=self.distributed,
-                ignore_last=True,
+                ignore_last=(
+                    self.model.type() != MODEL_LLM
```
tgaddair, line 852: This seems like an incidental thing (this would apply equally to ECD with batch size 1, and conversely, not apply if the LLM has batch size > 1). I could see you setting this based on whether `self.batch_size > 1`, but isn't this obviated by the above change to ignore the skipping behavior when the batch size is > 1?

justinxzhao: Instead of checking for batch size in the batcher, I'm thinking that it might be clearer to do that check in the trainer and set `ignore_last` explicitly:

```python
# `ignore_last` skips the last batch of an epoch if the last batch only 
# has one example in it. If the batch size is exactly 1, then we set ignore_last=False.
ignore_last = self.batch_size > 1

...

with training_set.initialize_batcher(
                batch_size=self.batch_size,
                should_shuffle=self.should_shuffle,
                random_seed=self.random_seed,
                distributed=self.distributed,
                ignore_last=ignore_last,
...
```

This way, `ignore_last` continues to straightforwardly ignore the last batch if it only contains one example, and the exceptional case when `batch_size == 1`, which applies to both ECD and LLM model types, is managed outside of the batcher.

I can see an argument that the responsibility of handling the exceptional `batch_size==1` case should belong to the batcher, in which case I think we can keep the trainer code the same as before, with `ignore_last=True` always for both model types. WDYT?

justinxzhao: I went ahead and made the latter change to always use `ignore_last=True` while keeping the batch_size check in the batcher. @tgaddair does this LGTY?

tgaddair (approved):
Can we add a test?

justinxzhao:
> Can we add a test?

Done.

PR #3655, opened 2023-09-22
Add TART decoder config object

jeffkinnison:
Adds config for a binary TART decoder to the existing decoder schemas.

### DIFF
--- a/ludwig/schema/decoders/base.py
+++ b/ludwig/schema/decoders/base.py
@@ -266,3 +266,51 @@ class ClassifierConfig(BaseDecoderConfig):
         description="Initializer for the bias vector.",
         parameter_metadata=DECODER_METADATA["Classifier"]["bias_initializer"],
     )
+
+
+@DeveloperAPI
+@register_decoder_config("tart", [BINARY, CATEGORY], model_types=[MODEL_ECD, MODEL_GBM, MODEL_LLM])
+@ludwig_dataclass
+class TARTDecoderConfig(BaseDecoderConfig):
+    reasoning_head: str = schema_utils.StringOptions(
+        options=[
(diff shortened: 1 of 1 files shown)


ludwig/schema/decoders/base.py
```diff
@@ -266,3 +266,51 @@ def module_name(cls):
         description="Initializer for the bias vector.",
         parameter_metadata=DECODER_METADATA["Classifier"]["bias_initializer"],
     )
+
+
+@DeveloperAPI
+@register_decoder_config("tart", [BINARY, CATEGORY], model_types=[MODEL_ECD, MODEL_GBM, MODEL_LLM])
```
w4nderlust, line 272: not 100% sure of the implications of making a tart decoder available for gbm and llm. as long as we get outputs from them in the same shape as the ones expected from ecd, that's probably ok, but I will make sure of that first.

ludwig/schema/decoders/base.py
```diff
@@ -266,3 +266,51 @@ def module_name(cls):
...
+class TARTDecoderConfig(BaseDecoderConfig):
+    reasoning_head: str = schema_utils.StringOptions(
+        options=[
+            "EleutherAI/gpt-neo-1.3B",
+            "EleutherAI/gpt-neo-125m",
+            "EleutherAI/pythia-1.4b-deduped",
+            "EleutherAI/pythia-2.8b-deduped",
+            "facebook/opt-iml-max-1.3b",
+        ],
+        default=None,
```
jeffkinnison, line 283: set to GPT-2

jeffkinnison:
I'm going to go ahead and merge this so that I can access it for decoder testing.

PR #3664, opened 2023-09-25
Add functionality to allow direct importing of datasets from Hugging Face

Infernaught:
Allow users to specify the path to a Hugging Face Text2Text Generation dataset to directly import that dataset for training/inference rather than requiring them to download and process the data themselves.


Infernaught:
This PR is still incomplete. Wanted to open this up for visibility.

Infernaught:
As of [265c2e3](https://github.com/ludwig-ai/ludwig/pull/3664/commits/265c2e33951b539111475c2973eff9e156d2d317), model.train can now take in a HF dataset and fine-tune on it. Have kicked off a test job to evaluate performance.

ludwig/datasets/loaders/hugging_face.py
```diff
@@ -0,0 +1,66 @@
...
+import logging
+
+import datasets
+import pandas as pd
+
+from ludwig.datasets.loaders.dataset_loader import DatasetLoader
+
+TRAIN = "train"
+VALIDATION = "validation"
+TEST = "test"
```
justinxzhao, line 24: These are already ludwig constants. Can we re-use them?

ludwig/datasets/loaders/hugging_face.py
```diff
@@ -0,0 +1,66 @@
...
+splits = [TRAIN, VALIDATION, TEST]
```
justinxzhao, line 25: Let's capitalize global vars.

ludwig/datasets/loaders/hugging_face.py
```diff
@@ -0,0 +1,66 @@
...
+            if VALIDATION in dataset_dict:
+                validation_df = dataset_dict[VALIDATION]
+            else:
+                logger.warning("No validation set found in provided Hugging Face dataset. Skipping validation set.")
+                validation_df = pd.DataFrame()
+            if TEST in dataset_dict:
+                test_df = dataset_dict[TEST]
+            else:
+                logger.warning("No test set found in provided Hugging Face dataset. Skipping test set.")
+                test_df = pd.DataFrame()
```
justinxzhao, line 58: Simplification:

```python
for split in DEFAULT_SPLITS:
    if split not in dataset_dict:
        logger.warning(f"No {split} set found in provided Hugging Face dataset. Skipping {split} set.")

train_df = dataset_dict[TRAIN] if TRAIN in dataset_dict else pd.DataFrame()
validation_df = dataset_dict[VALIDATION] if VALIDATION in dataset_dict else pd.DataFrame()
test_df = dataset_dict[TEST] if TEST in dataset_dict else pd.DataFrame()
```

ludwig/datasets/loaders/hugging_face.py
```diff
@@ -0,0 +1,66 @@
...
+class HFLoader(DatasetLoader):
+    def load_hf_to_dict(self, hf_id: str, hf_subsample: str) -> dict:
+        # Convert from HF DatasetDict type to dict
+        dataset_dict = datasets.load_dataset(path=hf_id, name=hf_subsample)
+        new_dict = {}
+        for split in dataset_dict:
+            new_dict[split] = dataset_dict[split].to_pandas()
+        return new_dict
+
+    def load(self, split=False, kaggle_username=None, kaggle_key=None) -> pd.DataFrame:
```
justinxzhao, line 37: Since we're already overriding this method, let's remove the kaggle credentials, and pass in the `huggingface_dataset_id` and the `huggingface_subsample` as explicit arguments instead of requiring the the caller set stateful variables implicitly with `setattr`.

ludwig/datasets/__init__.py
```diff
@@ -121,12 +123,81 @@ def load_dataset_uris(
...
+    elif training_set is not None:
+        train_df = test_df = val_df = None
+        training_set_checksum = None
+        if isinstance(training_set, str) and training_set.startswith(HF_PREFIX):
+            # For the training set, we only want to use the TRAINING split of the dataset
+            loader = get_dataset("hugging_face")
+            dataset_name, dataset_subsample = _get_hf_dataset_and_subsample(training_set)
+            setattr(loader.config, "huggingface_dataset_id", dataset_name)
+            setattr(loader.config, "huggingface_subsample", dataset_subsample)
+            train_df, val_df, test_df = loader.load(split=True)
```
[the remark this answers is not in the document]
Infernaught: In the blocks below, we have to use different loaders for the validation and test sets in case the user wants to use validation and test sets from different Hugging Face datasets. We unpack val_df/test_df directly if they come from the same dataset as train_df. Otherwise, we have to load them in using `_load_cacheable_dataset`.

I'm not quite sure what you mean by, "How does the loader know what the training set is?"

ludwig/datasets/__init__.py
```diff
@@ -180,6 +272,9 @@ def get_datasets_output_features(
         for competition in competition_datasets:
             del ordered_configs[competition]
 
+    for hf_dataset in hugging_face_datasets:
```
justinxzhao, line 275: It looks like there's only one config whose name is "hugging_face". Could we instead do:

```python
del ordered_configs["hugging_face"]
```

ludwig/datasets/__init__.py
```diff
@@ -150,12 +236,18 @@ def get_datasets_output_features(
     """
```
justinxzhao, line 236: Let's extend this docstring to include that hugging face datasets are loaded dynamically through a shared connector, so since there is no fixed output feature, HF datasets are excluded.

ludwig/datasets/loaders/dataset_loader.py
```diff
@@ -451,6 +451,28 @@ def load_unprocessed_dataframe(self, file_paths: List[str]) -> pd.DataFrame:
...
+        else:
+            dataframes = self._get_dataframe_with_fixed_splits(
+                train_paths, validation_paths, test_paths, dataset_paths, file_paths
+            )
+        return pd.concat(dataframes, ignore_index=True)
+
+    def _get_dataframe_with_fixed_splits_from_hf(self):
+        dataframes = []
+        splits = ["train", "validation", "test"]
+        data_dict = self.load_hf_to_dataframe(self.config.huggingface_dataset_id, self.config.huggingface_subset)
```
justinxzhao, line 466: Is this new? Why do we need to modify `load_unprocessed_dataframe`? Where is `load_hf_to_dataframe` defined?

justinxzhao:
@alexsherstinsky brought up a great suggestion, which is to make sure that we add a unit test demonstrating the new functionality!

ludwig/datasets/__init__.py
```diff
@@ -118,9 +126,87 @@ def load_dataset_uris(
...
+        if isinstance(dataset, str) and dataset.startswith(HF_PREFIX):
+            dataset_out = _load_cacheable_hf_dataset(dataset, backend)
+        return dataset_out, training_set_out, validation_set_out, test_set_out
+
+    # Because of the conditional logic (_is_hf) in load_dataset_uris, if the above block is not triggered, then
+    # training_set must be a string that starts with HF_PREFIX
+    train_df = test_df = val_df = None
+    loader = get_dataset("hugging_face")
+    hf_id, hf_subsample = _get_hf_dataset_and_subsample(training_set)
+    train_df, val_df, test_df = loader.load(hf_id, hf_subsample, split=True)  # Call hugging_face loader
```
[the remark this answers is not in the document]
Infernaught: I feel like it would be strange to do so here because `_load_cacheable_hf_dataset` would perform extra steps that are unnecessary here. That's why the logic has been split up across the following conditional blocks.

ludwig/datasets/loaders/dataset_loader.py
Infernaught: load_hf_to_dataframe is defined in `ludwig/datasets/loaders/hugging_face.py`

Infernaught: We modify `load_unprocessed_dataframe` in order to handle how HF datasets are loaded in

ludwig/datasets/__init__.py
```diff
@@ -118,9 +126,87 @@ def load_dataset_uris(
...
+            # Reuse the loaded DF from the training split
+            test_df = backend.df_engine.from_pandas(test_df)
+            test_set_out = CacheableDataframe(df=test_df, name=test_set, checksum=None)
+        else:
+            test_set_out = _load_cacheable_hf_dataset(test_set, backend, split_set=TEST)
+
     return dataset_out, training_set_out, validation_set_out, test_set_out
 
 
+def _load_cacheable_hf_dataset(dataset: str, backend: Backend, split_set=None) -> CacheableDataframe:
```
justinxzhao, line 196: nit: `split_set: Optional[Literal[TRAIN, VALIDATION, TEST]] = None`

ludwig/datasets/__init__.py
```diff
@@ -118,9 +126,87 @@ def load_dataset_uris(
...
+    train_df = backend.df_engine.from_pandas(train_df)
+    training_set_out = CacheableDataframe(df=train_df, name=training_set, checksum=None)
+
+    if isinstance(validation_set, str) and validation_set.startswith(HF_PREFIX):
+        if validation_set == training_set:
+            # Reuse the loaded DF from the training split
+            val_df = backend.df_engine.from_pandas(val_df)
+            validation_set_out = CacheableDataframe(df=val_df, name=validation_set, checksum=None)
+        else:
+            validation_set_out = _load_cacheable_hf_dataset(validation_set, backend, split_set=VALIDATION)
```
justinxzhao, line 183: Add a commenthere that in general this handles an edge case where the user has specified different HF datasets for their dataset arguments. Not sure if this is a real use case, but potentially also worth adding a warning.

ludwig/datasets/configs/hugging_face.yaml
```diff
@@ -0,0 +1,7 @@
+version: 1.0
+name: hugging_face
+huggingface_dataset_id: placeholder
```
justinxzhao, line 3: nit: This can be removed?

ludwig/datasets/__init__.py
```diff
@@ -81,16 +83,22 @@ def load_dataset_uris(
 
     Returns the input unmodified for any non-Ludwig datasets.
     """
-    dataset_out = dataset
-    training_set_out = training_set
-    validation_set_out = validation_set
-    test_set_out = test_set
+
+    if dataset is None and training_set is None:
+        raise ValueError("Neither dataset nor training_set has been provided. Please provide one of the two.")
```
justinxzhao, line 88: Are we checking anywhere when `dataset is None and training_set is None`?

justinxzhao (commented):
One more comment! Also, are you planning on adding a test?

ludwig/datasets/__init__.py
Infernaught: I this situation shouldn't come up because I think we raise an error elsewhere if this is true

justinxzhao: I would expect that the place where the error is raised elsewhere also raise this ValueError. Can you check?

justinxzhao (approved):

ludwig/datasets/__init__.py
Infernaught: https://github.com/ludwig-ai/ludwig/blob/2772e9a36fffc1bc9f7d8403a0fbca1ca5b8258e/ludwig/data/preprocessing.py#L1082

Infernaught: Sorry maybe I misunderstood what you were saying. Did you want me to do something with this ValueError in preprocessing.py?

ludwig/datasets/__init__.py
```diff
@@ -81,16 +83,19 @@ def load_dataset_uris(
 
     Returns the input unmodified for any non-Ludwig datasets.
     """
-    dataset_out = dataset
-    training_set_out = training_set
-    validation_set_out = validation_set
-    test_set_out = test_set
+
+    dataset_out, training_set_out, validation_set_out, test_set_out = dataset, training_set, validation_set, test_set
```
w4nderlust, line 87: this is less readable than it was before

ludwig/datasets/__init__.py
```diff
@@ -81,16 +83,19 @@ def load_dataset_uris(
...
+    # Hugging Face datasets should follow the naming convention `hf://<hf_id>--<hf_subsample>`
+    if _is_hf(dataset, training_set):
+        return _load_hf_datasets(dataset, training_set, validation_set, test_set, backend)
 
     # Check that any of the datasets begin with the `ludwig://` prefix denoting a Ludwig dataset URI
     if dataset is not None:
         if isinstance(dataset, str) and dataset.startswith(URI_PREFIX):
             dataset_out = _load_cacheable_dataset(dataset, backend)
-    elif training_set is not None:
+        return dataset_out, training_set_out, validation_set_out, test_set_out
```
w4nderlust, line 97: the logic before allowed for having a ringle return in the function, which is auspicable for readability

ludwig/datasets/__init__.py
```diff
@@ -118,9 +123,97 @@ def load_dataset_uris(
...
+    # Hugging Face datasets should follow the naming convention `hf://<hf_id>--<hf_subsample>`
+    if dataset is not None:
+        if isinstance(dataset, str) and dataset.startswith(HF_PREFIX):
+            dataset_out = _load_cacheable_hf_dataset(dataset, backend)
+        return dataset_out, training_set_out, validation_set_out, test_set_out
+
+    # Because of the conditional logic (_is_hf) in load_dataset_uris, if the above block is not triggered, then
+    # training_set must be a string that starts with HF_PREFIX
+    train_df = test_df = val_df = None
+    loader = get_dataset("hugging_face")
```
w4nderlust, line 168: get_dataset is not a good name for this as it is not returning a dataset, it should be called get_dataset_loader

PR #3666, opened 2023-09-26
Fix dynamic generation config load during `model.predict`

geoffreyangus:
This PR fixes dynamic generation config loading during `model.predict`.

Their were two issues:
1. When passing in a generation config through model.predict, a fresh `transformers.GenerationConfig` was created, meaning that any existing parameters in `self.generation` were not respected by `llm.set_generation_config`. This is fixed by ensuring that we pass in `**{**self.generation.as_dict(), **new_generation_config_dict` as kwargs into the new GenerationConfig.
2. When setting a new generation config, we were not properly communicating with the `TextExtractorDecoder`, which is in charge of extracting the generated outputs from the model outputs. This is fixed by passing in `max_new_tokens` as a kwarg in the decoder forward pass.

### DIFF
--- a/ludwig/decoders/llm_decoders.py
+++ b/ludwig/decoders/llm_decoders.py
@@ -110,21 +110,29 @@ class TextExtractorDecoder(Decoder):
 
     def forward(self, inputs: List[torch.Tensor], **kwargs):
         # Extract the sequences tensor from the LLMs forward pass
+        max_new_tokens = kwargs.get("max_new_tokens", self.max_new_tokens)
         generated_outputs = extract_generated_tokens(
             raw_generated_output_sequences=inputs,
             llm_model_input_lengths=kwargs.get("llm_model_input_lengths", []),
-            max_new_tokens=self.max_new_tokens,
+            max_new_tokens=max_new_tokens,
             pad_sequence=True,
         )
+        # Stack the predictions for each example in the batch. The padding should ensure they are all the same shape.
+        for output in generated_outputs:
+            if output.shape[0] > max_new_tokens:
+                raise ValueError(f"Output {output} is longer than the max_new_tokens {max_new_tokens} during decoding. "
+                                 f"This should never happen– please file an issue on GitHub.")
+
+        generated_outputs = torch.stack(generated_outputs, dim=0)
         outputs_device = generated_outputs.device
 
         return {
             PREDICTIONS: generated_outputs,
             # TODO(Arnav): Add support for probabilities and logits
-            PROBABILITIES: torch.zeros((len(generated_outputs), self.max_new_tokens, self.tokenizer_vocab_size)).to(
+            PROBABILITIES: torch.zeros((len(generated_outputs), max_new_tokens, self.tokenizer_vocab_size)).to(
                 outputs_device
             ),
-            LOGITS: torch.zeros((len(generated_outputs), self.max_new_tokens, self.tokenizer_vocab_size)).to(
+            LOGITS: torch.zeros((len(generated_outputs), max_new_tokens, self.tokenizer_vocab_size)).to(
                 outputs_device
             ),
         }
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -3,7 +3,7 @@ import copy
 import logging
 import os
 import tempfile
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
@@ -135,9 +135,10 @@ class LLM(BaseModel):
         else:
             self.context_len = 2048
 
+        self.generation = GenerationConfig(**self.config_obj.generation.to_dict())
         # max input length value copied from FastChat
         # https://github.com/lm-sys/FastChat/blob/0e958b852a14f4bef5f0e9d7a5e7373477329cf2/fastchat/serve/inference.py#L183  # noqa
-        self.max_new_tokens = self.config_obj.generation.max_new_tokens
+        self.max_new_tokens = self.generation.max_new_tokens
         self.max_input_length = self.context_len - self.max_new_tokens - 8
 
         # TODO(Arnav): This needs be more flexible to account for RoPE Scaling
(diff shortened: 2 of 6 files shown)


ludwig/models/llm.py
```diff
@@ -135,9 +135,10 @@ def __init__(
         else:
             self.context_len = 2048
 
+        self.generation = GenerationConfig(**self.config_obj.generation.to_dict())
         # max input length value copied from FastChat
         # https://github.com/lm-sys/FastChat/blob/0e958b852a14f4bef5f0e9d7a5e7373477329cf2/fastchat/serve/inference.py#L183  # noqa
-        self.max_new_tokens = self.config_obj.generation.max_new_tokens
+        self.max_new_tokens = self.generation.max_new_tokens
```
justinxzhao, line 141: Good catch finding these other variables that also need to be updated when GenerationConfig is updated! OOC, where does 8 come from?

ludwig/decoders/llm_decoders.py
```diff
@@ -110,23 +110,31 @@ def get_prediction_set(self):
 
     def forward(self, inputs: List[torch.Tensor], **kwargs):
```
justinxzhao, line 111: nit: Since we're passing in max_new_tokens explicitly, let's add `max_new_tokens` as a parameter here.

ludwig/decoders/llm_decoders.py
```diff
@@ -110,23 +110,31 @@ def get_prediction_set(self):
...
         # Extract the sequences tensor from the LLMs forward pass
+        max_new_tokens = kwargs.get("max_new_tokens", self.max_new_tokens)
```
justinxzhao, line 113: Conditionally grabbing `max_new_tokens` from either state or kwargs feels brittle. Can we make `max_new_tokens` a parameter that's always passed in and always use that?

ludwig/decoders/llm_decoders.py
```diff
@@ -110,23 +110,31 @@ def get_prediction_set(self):
...
         generated_outputs = extract_generated_tokens(
             raw_generated_output_sequences=inputs,
             llm_model_input_lengths=kwargs.get("llm_model_input_lengths", []),
-            max_new_tokens=self.max_new_tokens,
+            max_new_tokens=max_new_tokens,
             pad_sequence=True,
         )
+        # Stack the predictions for each example in the batch. The padding should ensure they are all the same shape.
+        for output in generated_outputs:
```
justinxzhao, line 121: A sensible check!

justinxzhao (commented):
Thanks for the change @geoffreyangus. Mostly LGTM -- just a few nits.

ludwig/models/llm.py
```diff
@@ -197,13 +192,27 @@ def __init__(
...
+        original_max_new_tokens = self.max_new_tokens
+        original_max_input_length = self.max_input_length
+        try:
+            # no-op if generation_config is None
+            if generation_config_dict is not None:
+                # unwrap the original generation config, update it with the new generation config
+                new_generation_config_dict = {**self.generation.to_dict(), **generation_config_dict}
+                self.generation = GenerationConfig(**new_generation_config_dict)
+                self.max_new_tokens = self.generation.max_new_tokens
+                self.max_input_length = self.context_len - self.max_new_tokens - 8
```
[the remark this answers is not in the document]
geoffreyangus: done

ludwig/models/llm.py
geoffreyangus: from the comment above (FastChat implementation we grandfathered in): https://github.com/lm-sys/FastChat/blob/0e958b852a14f4bef5f0e9d7a5e7373477329cf2/fastchat/serve/inference.py#L183

ludwig/decoders/llm_decoders.py
geoffreyangus: done

ludwig/decoders/llm_decoders.py
geoffreyangus: done

ludwig/models/llm.py
```diff
@@ -197,13 +188,32 @@ def __init__(
...
+        original_max_input_length = self.max_input_length
+        try:
+            # no-op if generation_config is None
+            if generation_config_dict is not None:
+                # unwrap the original generation config, update it with the new generation config
+                new_generation_config_dict = {**self.generation.to_dict(), **generation_config_dict}
+                self._set_generation_config(new_generation_config_dict)
+            yield
+        finally:
+            self.generation = original_generation
```
justinxzhao, line 207: Can we use `_set_generation_config(original_generation)` here?

ludwig/models/llm.py
```diff
@@ -197,13 +188,32 @@ def __init__(
...
         """Sets the generation config for the model."""
-        self.generation = GenerationConfig(**generation_config_dict)
-
-    def reset_generation_config(self):
-        """Sets the generation config for th."""
-        self.generation = self.original_generation_config
+        # Save the original generation config so that we can reset it if/when we change it when self.generation gets is
+        # dynamically mutated during 1-off predict calls after fine-tuning.
+        original_generation = copy.deepcopy(self.generation)
+        original_max_new_tokens = self.max_new_tokens
```
justinxzhao, line 197: If we rely on `_set_generation_config(original_generation)`, then we can remove these two LOC.

justinxzhao (approved):
Thanks!

ludwig/models/llm.py
```diff
@@ -199,13 +189,28 @@ def __init__(
...
+            yield
+        finally:
+            self._set_generation_config(original_generation_config_dict)
+
+    def _set_generation_config(self, new_generation_config_dict: Dict[str, Any]):
+        self.generation = GenerationConfig(**new_generation_config_dict)
+        self.max_new_tokens = self.generation.max_new_tokens
+        # max input length value copied from FastChat
+        # https://github.com/lm-sys/FastChat/blob/0e958b852a14f4bef5f0e9d7a5e7373477329cf2/fastchat/serve/inference.py#L183  # noqa E501
+        self.max_input_length = self.context_len - self.max_new_tokens - 8
```
justinxzhao, line 213: Hi @geoffreyangus, it looks like https://github.com/ludwig-ai/ludwig/pull/3667 was merged. We might need to add this new LOC here as well:

`self.generation.pad_token_id = self.tokenizer.pad_token_id`

CC: @Infernaught @arnavgarg1

geoffreyangus: got it, just added it in!

PR #3667, opened 2023-09-26
Fix eos_token and pad_token issue

Infernaught:
There was a Ludwig user that was running into the following error:`If eos_token_id is defined, make sure that pad_token_id is defined.` There was also this Ludwig issue: https://github.com/ludwig-ai/ludwig/issues/3661

This PR introduces a workaround that fixes this problem. This PR has been successfully tested with the following models:
gpt2
codellama/CodeLlama-7b-hf
meta-llama/Llama-2-7b-h

### DIFF
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -157,6 +157,10 @@ class LLM(BaseModel):
         set_pad_token(self.tokenizer)
 
         self.generation = GenerationConfig(**self.config_obj.generation.to_dict())
+        self.generation.pad_token_id = self.tokenizer.pad_token_id
+        # We need to manually set the pad_token_id to the tokenizer's pad_token_id for certain models like GPT and
+        # CodeLlama to avoid getting an error. This workaround can be found here:
+        # (https://github.com/huggingface/transformers/issues/25353#issuecomment-1669339754)
 
         # Save the original generation config so that we can reset it if/when we change it when self.generation gets is
         # dynamically mutated during 1-off predict calls after fine-tuning.


justinxzhao (commented):
@geoffreyangus is making a change to the generation config setting code in https://github.com/ludwig-ai/ludwig/pull/3666. Let's wait for this change to land first, and then re-integrate.

ludwig/models/llm.py
```diff
@@ -157,6 +157,10 @@ def __init__(
         set_pad_token(self.tokenizer)
 
         self.generation = GenerationConfig(**self.config_obj.generation.to_dict())
+        self.generation.pad_token_id = self.tokenizer.pad_token_id
```
alexsherstinsky, line 160: I believe that I have seen the situation -- with https://huggingface.co/mistralai/Mistral-7B-v0.1 if I am not mistaken -- that even the tokenizer of the base model has the `pad_token_id` as NULL; hence, it would be highly desirable if the proposed solution were verified against this recently published LLM (available on HuggingFace at the above URL).  Thank you.

arnavgarg1: @alexsherstinsky So this will do the right thing actually!

The reason is that in the function called set_pad_token() above, we actually fallback to setting the pad token to the EOS token when the pad token is not set. We do this for Llama styled models as well. This is common practice since many models were pretrained without padding. 

So I do believe this should work? Were you running into issues when you tried it?

alexsherstinsky: @arnavgarg1 Thank you very much for this clarification.  I will try using this branch within 60 -- 120 minutes.  If by that time it gets merged, then I will try on the main branch.  I will keep you posted as to the outcome.  Thank you!

PR #3673, opened 2023-09-27
[MAINTENANCE] Partially reconcile type hints, fix some warnings, and fix comments in parts of the codebase.

alexsherstinsky:
This set of changes aides towards improving code elegance.

It consists of the following improvements in parts of the codebase:
* Solving IDE warnings;
* Adding type hints;
* Fixing broken docstrings;
* Fixing PEP8 warnings and setting up `noqa` and `type ignore` escapes.

Much work towards these improvements can be done; this serves only as a small set surrounding LLM related modules.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -269,11 +269,11 @@ class LudwigModel:
         self,
         config: Union[str, dict],
         logging_level: int = logging.ERROR,
-        backend: Union[Backend, str, None] = None,
-        gpus: Union[str, int, List[int], None] = None,
+        backend: Optional[Union[Backend, str]] = None,
+        gpus: Optional[Union[str, int, List[int]]] = None,
         gpu_memory_limit: Optional[float] = None,
         allow_parallel_threads: bool = True,
-        callbacks: Union[List[Callback], None] = None,
+        callbacks: Optional[List[Callback]] = None,
     ) -> None:
         """Constructor for the Ludwig Model class.
 
@@ -326,29 +326,29 @@ class LudwigModel:
 
         # setup model
         self.model = None
-        self.training_set_metadata: Union[str, dict, None] = None
+        self.training_set_metadata: Optional[str, dict] = None
 
         # online training state
         self._online_trainer = None
 
     def train(
         self,
-        dataset: Union[str, dict, pd.DataFrame, None] = None,
-        training_set: Union[str, dict, pd.DataFrame, Dataset, None] = None,
-        validation_set: Union[str, dict, pd.DataFrame, Dataset, None] = None,
-        test_set: Union[str, dict, pd.DataFrame, Dataset, None] = None,
-        training_set_metadata: Union[str, dict, None] = None,
-        data_format: Union[str, None] = None,
+        dataset: Optional[Union[str, dict, pd.DataFrame]] = None,
+        training_set: Optional[Union[str, dict, pd.DataFrame, Dataset]] = None,
+        validation_set: Optional[Union[str, dict, pd.DataFrame, Dataset]] = None,
+        test_set: Optional[Union[str, dict, pd.DataFrame, Dataset]] = None,
+        training_set_metadata: Optional[Union[str, dict]] = None,
+        data_format: Optional[str] = None,
         experiment_name: str = "api_experiment",
         model_name: str = "run",
-        model_resume_path: Union[str, None] = None,
+        model_resume_path: Optional[str] = None,
         skip_save_training_description: bool = False,
         skip_save_training_statistics: bool = False,
         skip_save_model: bool = False,
         skip_save_progress: bool = False,
         skip_save_log: bool = False,
         skip_save_processed_input: bool = False,
-        output_directory: Union[str, None] = "results",
+        output_directory: Optional[str] = "results",
         random_seed: int = default_random_seed,
         **kwargs,
     ) -> TrainingResults:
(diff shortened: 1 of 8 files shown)


ludwig/api.py
```diff
@@ -111,7 +111,7 @@
 
 @PublicAPI
 @dataclass
-class EvaluationFrequency:
+class EvaluationFrequency:  # noqa F821
```
tgaddair, line 114: Hmm, do you know why the linter complains about this? I don't see anything immediately wrong with the way this class is defined.

ludwig/api.py
```diff
@@ -269,11 +269,11 @@ def __init__(
...
         config: Union[str, dict],
         logging_level: int = logging.ERROR,
-        backend: Union[Backend, str] = None,
-        gpus: Union[str, int, List[int]] = None,
+        backend: Union[Backend, str, None] = None,
+        gpus: Union[str, int, List[int], None] = None,
         gpu_memory_limit: Optional[float] = None,
         allow_parallel_threads: bool = True,
-        callbacks: List[Callback] = None,
+        callbacks: Union[List[Callback], None] = None,
```
tgaddair, line 276: Nit: we tend to use `Optional[List[Callback]]` for params that accept `None` as a valid value. I think it makes sense to use `Union` when there are already a `Union` over multiple types, but `Optional` would be more consistent with the rest of the codebase for situations like this where there is only one valid type that can optionally be `None`. What do you think?

tgaddair (commented):
Nice! Really appreciate PRs like this. Just had a couple nits.

Out of curiosity, with these changes does `mypy` now pass static analysis checks? If so, we may want to add `mypy` to our CI process in a follow-up, which is something I've been wanting to do for some time.

ludwig/api.py
alexsherstinsky: @tgaddair I see the complaint in PyCharm -- they only reason I could think of is that the dataclass we use is not from Python but from Marshmallow, and perhaps PyCharm is unable to introspect it in sufficient depth.

ludwig/api.py
alexsherstinsky: @tgaddair I will change it to `Optional` -- totally understood in terms of supporting the presently established coding pattern.  Having said that, I would like to suggest something entirely different for the future.  If we place `from __future__ import annotations` at the top of the module, we will be able to use the `|` operator, which is a shorthand for `Union`.  That way, we will be able to differentiate the true optionality (i.e., argument can be legitimately omitted, without a default needed to be supplied) from the actual possibility for a parameter to be `None` or other values.  What are your thoughts about this?  Thank you very much.  P.S.: I will notify you once I have made the change to `Optional` -- thanks.

alexsherstinsky:
> Nice! Really appreciate PRs like this. Just had a couple nits.
> 
> Out of curiosity, with these changes does `mypy` now pass static analysis checks? If so, we may want to add `mypy` to our CI process in a follow-up, which is something I've been wanting to do for some time.

Hi @tgaddair -- I did not do all updates to typehints, so `mypy` still fails a lot, just a bit less than before, due to these changes.  In some ways, these changes propose a pattern for how to keep doing this throughout the codebase.  Due to the large number of changes that would be needed, I feel that adding `mypy` is a bit too soon.  However, maybe not -- if we go through the modules meticulously and add the ones we have not done to `pyproject.toml` as exceptions.  Then we can be having small sprints aimed at reducing these exceptions.  This is just one idea for consideration.  Thank you.

ludwig/api.py
alexsherstinsky: @tgaddair I made the changes -- please re-review at your convenience.  Thank you.

ludwig/api.py
tgaddair: Ah, I see. Thanks for the clarification. Sounds like this might require more digging at some point.

cc @ksbrar

ludwig/api.py
tgaddair: Thanks! Totally open to changing the convention if using `|` is the preferred notation going forward. It makes sense to me that it's worth distinguishing between an optional kwarg and something that can be explicitly set to `None`. Do you think it would be better to do this now, or wait until 3.10 (which I believe is the first version to support this officially) is the min supported version? I'm pretty open either way.

tgaddair:
> > Nice! Really appreciate PRs like this. Just had a couple nits.
> > Out of curiosity, with these changes does `mypy` now pass static analysis checks? If so, we may want to add `mypy` to our CI process in a follow-up, which is something I've been wanting to do for some time.
> 
> Hi @tgaddair -- I did not do all updates to typehints, so `mypy` still fails a lot, just a bit less than before, due to these changes. In some ways, these changes propose a pattern for how to keep doing this throughout the codebase. Due to the large number of changes that would be needed, I feel that adding `mypy` is a bit too soon. However, maybe not -- if we go through the modules meticulously and add the ones we have not done to `pyproject.toml` as exceptions. Then we can be having small sprints aimed at reducing these exceptions. This is just one idea for consideration. Thank you.

Thanks for the clarification @alexsherstinsky, that makes sense. Definitely makes sense to do this incrementally in my mind.

cc @justinxzhao

tgaddair (approved):
LGTM!

ludwig/api.py
ksbrar: Thanks for flagging! I have made a note to check this out later.

PR #3677, opened 2023-09-29
Add per-device logging to tensorboard

Infernaught:
This PR adds nvidia-smi logging to tensorboard (specifically, it logs how much memory is being used over time). The numbers are not exactly the same as those outputted by nvidia-smi in the command line (the tensorboard logs show larger numbers than those shown by nvidia-smi in the command line), but are quite close.

[EDIT] After further investigation, nvidia-smi logging is unnecessary. This PR now offers per-device tensorboard logging and adds a new metric -- total_memory_used -- which should be more or less equivalent to what nvidia-smi shows.

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -474,6 +474,13 @@ class Trainer(BaseTrainer):
                 train_summary_writer.add_scalar(
                     f"cuda/device{i}/total_memory_occupied", torch.cuda.mem_get_info(device=device)[1], global_step=step
                 )
+
+                # Total memory used.
+                train_summary_writer.add_scalar(
+                    f"cuda/device{i}/total_memory_used",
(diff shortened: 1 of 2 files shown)


ludwig/trainers/trainer.py
```diff
@@ -453,6 +454,16 @@ def write_step_summary(cls, train_summary_writer, combined_loss, all_losses, ste
                 "cuda/total_memory_occupied", torch.cuda.mem_get_info()[1], global_step=step
             )
 
+        # Log nvidia-smi stats
```
justinxzhao, line 457: Since nvidia-smi is only relevant when using GPUs, let's fold this logging into the block above so that users only get these logs `if torch.cuda.is_available():`.

justinxzhao (approved):

PR #3679, opened 2023-09-29
[BUGFIX] Ensure that full base models and not only adapter weights get saved when merge_and_unload is set

alexsherstinsky:
### Scope
* This change fixes an issue whereby after successful `merge_and_unload()` in terms of model layer structure, only the LoRA adapter weights are persisted.  The fix consists of recognizing that `merge_and_unload` was set and save the base model weights (in addition to fine-tuning weights) at the end of the training phase.  We only save adapter weights during the training; otherwise, the model might be fine-tuning the entire set of weights, instead of only the adapter weights.  Thanks to @arnavgarg1 for guidance.
* New tests check that the expected weights files are saved depending on whether or not `merge_and_unload` is set.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -728,12 +728,18 @@ class LudwigModel:
 
                 self.training_set_metadata = training_set_metadata
 
-                # For an LLM model trained with a LoRA adapter, handle merge and unload postprocessing directives.
-                self._merge_and_unload()
-
                 # Ensure model weights are saved to the driver if training was done remotely
                 if self.backend.is_coordinator() and not skip_save_model:
                     self.model.save(model_dir)
+
+                # For an LLM model trained with a LoRA adapter, handle merge and unload postprocessing directives.
+                do_merge_and_unload: bool = self._is_merge_and_unload_set()
+                if do_merge_and_unload:
+                    self._merge_and_unload()
+                    # Also: Ensure that the full model weights are saved to the driver if training was done remotely.
+
+                    if self.backend.is_coordinator() and not skip_save_model:
+                        self.model.save_base_model(model_dir)
 
                 # Synchronize model weights between workers
                 self.backend.sync_model(self.model)
@@ -813,20 +819,26 @@ class LudwigModel:
 
         self.model = self._online_trainer.train_online(training_dataset)
 
-    def _merge_and_unload(self) -> None:
-        """For an LLM model trained with a LoRA adapter, handle merge and unload postprocessing directives.
+    def _is_merge_and_unload_set(self) -> bool:
+        """First, check that the model is of the "llm" type.  Then check if the "adapter" configuration section
+        contains the "postprocessor" subsection and apply the "merge_adapter_into_base_model" and "progressbar"
+        directives.
 
-        First, check that the model is of the "llm" type.  Then check if the "adapter" configuration section contains
-        the "postprocessor" subsection and apply the "merge_adapter_into_base_model" and "progressbar" directives.
+        # Return
+
+            :return (bool): whether merge_and_unload should be done.
         """
-        if (
+        return (
             self.config_obj.model_type == "llm"
             and self.config_obj.adapter is not None
             and self.config_obj.adapter.postprocessor is not None
             and self.config_obj.adapter.postprocessor.merge_adapter_into_base_model
             and hasattr(self.model, "merge_and_unload")
-        ):
-            self.model.merge_and_unload(progressbar=self.config_obj.adapter.postprocessor.progressbar)
+        )
+
+    def _merge_and_unload(self) -> None:
+        """For an LLM model trained with a LoRA adapter, handle merge and unload postprocessing directives."""
+        self.model.merge_and_unload(progressbar=self.config_obj.adapter.postprocessor.progressbar)
 
     def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_seed):
         """Sets AUTO batch-size-related parameters based on the trainer, backend type, and number of workers.
@@ -1661,7 +1673,8 @@ class LudwigModel:
         ludwig_model.load_weights(model_dir)
 
         # The LoRA layers appear to be loaded again (perhaps due to a potential bug); hence, we merge and unload again.
-        ludwig_model._merge_and_unload()
+        if ludwig_model._is_merge_and_unload_set():  # noqa W0212
+            ludwig_model._merge_and_unload()  # noqa W0212
 
         # load train set metadata
         ludwig_model.training_set_metadata = backend.broadcast_return(
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -648,9 +648,6 @@ class LudwigModel:
                     )
                     (self.model, train_trainset_stats, train_valiset_stats, train_testset_stats) = train_stats
 
-                    # For an LLM model trained with a LoRA adapter, handle merge and unload postprocessing directives.
-                    self._merge_and_unload()
-
                     # Calibrates output feature probabilities on validation set if calibration is enabled.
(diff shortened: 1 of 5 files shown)


arnavgarg1 (approved):
LGTM

tgaddair (approved):

ludwig/config_validation/checks.py
```diff
@@ -639,6 +640,22 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
 
 
+@register_config_check
+def check_qlora_merge_and_unload_compatibility(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks that model.merge_and_unload() is supported by underlying model.save_pretrained() when merging QLoRA
+    layers."""
+    if config.adapter.postprocessor is None or not (
+        config.adapter.postprocessor.merge_adapter_into_base_model and config.quantization
+    ):
+        return
```
arnavgarg1, line 650: Hey @alexsherstinsky , you may need to actually add a couple of additional checks:

1. Check if the config.model_type is LLM since `config.adapter` is only defined for LLM model types
2. Check if `config.adapter` is actually specified in the config, otherwise I believe all of the underlying attributes will cause errors (postprocessor and merge_adapter_into_base_model) since it'll complain that NoneType object has no attribute X

ludwig/api.py
```diff
@@ -1653,7 +1644,9 @@ def load(
         ludwig_model.load_weights(model_dir)
 
         # The LoRA layers appear to be loaded again (perhaps due to a potential bug); hence, we merge and unload again.
-        ludwig_model._merge_and_unload()
+        if config_obj.model_type == MODEL_LLM and ludwig_model.model.is_merge_and_unload_set():
+            # For an LLM model trained with a LoRA adapter, handle merge and unload postprocessing directives.
+            ludwig_model.model.merge_and_unload(progressbar=config_obj.adapter.postprocessor.progressbar)
```
arnavgarg1, line 1649: Perhaps we can add a method called `is_merge_and_unload_set` to the Base Model class (also used by ECD and GBM) where the return value is set to False, and then we can get rid of `config_obj.model_type == MODEL_LLM`. What do you think?

ludwig/models/llm.py
```diff
@@ -658,6 +672,16 @@ def save(self, save_path):
         else:
             logger.info("Skipped saving LLM without weight adjustments.")
 
+    def save_base_model(self, save_path):
+        """Saves the base LLM model to the given path."""
+        # TODO: see the "TODO" statement from "LLM.save()" in this module.
+        if self.config_obj.trainer.type != "none":
+            weights_save_path = os.path.join(save_path, MODEL_WEIGHTS_FILE_NAME)
+            self.model.base_model.save_pretrained(weights_save_path)
+            self.tokenizer.save_pretrained(weights_save_path)
```
arnavgarg1, line 681: Would be nice if you could add a comment here on why the tokenizer is being saved (as we discussed)!

tests/integration_tests/test_llm.py
```diff
@@ -609,6 +619,108 @@ def test_llm_finetuning_strategies_quantized(tmpdir, csv_filename, finetune_stra
...
+            id="qlora-8bit-not-merged",
+        ),
+    ],
+)
+def test_llm_lora_finetuning_merge_and_unload_quantized_accelerate_required(
+    csv_filename, finetune_strategy, adapter_args, quantization, error_raised
+):
+    input_features: List[dict] = [text_feature(name="input", encoder={"type": "passthrough"})]
+    output_features: List[dict] = [text_feature(name="output")]
+
```
arnavgarg1, line 663: Hey Alex! What is the purpose of these tests? Quantization based tests will only work once we have GPU tests enabled, so these will be skipped. I also think that once we have GPU tests enabled, one likely won't run into the `using load_in_8bit=True requires Accelerate` error. Okay to leave it in here for now, but was curious what thi test is intended to do

ludwig/config_validation/checks.py
alexsherstinsky: @arnavgarg1 I will do that -- I thought that the other tests already take care of it.  Please confirm that each validation test has to be self-contained.  Thank you.

arnavgarg1: @alexsherstinsky Yes, each validation test has to be self-contained as it currently stands. This is probably something we can improve in the future

ludwig/api.py
alexsherstinsky: @arnavgarg1 I can certainly do that; however, it feels "not clean", because those other models "do not know what merge_and_unload_set is".  I thought of creating an abstract method, but that just pushes the concern down to the specific models you indicated.  Should we instead try to merge the "merge_and_unload_set = False" setting into the configuration?  I feel that is cumbersome too.  Would then implementing your suggestion with a "TODO" be a good way to go now?  Thanks.

ludwig/models/llm.py
alexsherstinsky: @arnavgarg1 I will certainly do that -- in the next commit.  Thanks.

tests/integration_tests/test_llm.py
alexsherstinsky: @arnavgarg1 This test was simply to avoid commenting out those use cases but just run them and confirm that currently the "NotImportedError" is raised.  Please let me know if it is OK to leave as is for now, or do something else.  Thank you.

ludwig/api.py
alexsherstinsky: @arnavgarg1 I refactored the implementation and added a "TODO" comment.  Thank you.

ludwig/config_validation/checks.py
```diff
@@ -639,6 +640,29 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+        and config.adapter.postprocessor.merge_adapter_into_base_model
+        and config.quantization
+    ):
+        return
+
+    if config.quantization.bits < MIN_QUANTIZATION_BITS_FOR_MERGE_AND_UNLOAD:
+        raise ConfigValidationError(
+            f"""This operation will entail merging LoRA layers on a {config.quantization.bits}-bit \
+quantized model.  Calling "save_pretrained()" on that model is currently unsupported."""
+        )
```
arnavgarg1, line 663: Maybe we can use this as an opportunity to let users know how to rectify the issue - perhaps adding something like "If you want to merge the LoRA adapter weights into the base model, you need to do 8-bit quantization or do non-quantized based training by removing the quantization section from your Ludwig config".

arnavgarg1 (approved):
Only a minor comment, but this PR looks good to me!

ludwig/config_validation/checks.py
alexsherstinsky: @arnavgarg1 This has been incorporated.  Thanks!

ludwig/config_validation/checks.py
```diff
@@ -639,6 +640,31 @@ def check_qlora_requirements(config: "ModelConfig") -> None:  # noqa: F821
...
+@register_config_check
+def check_qlora_merge_and_unload_compatibility(config: "ModelConfig") -> None:  # noqa: F821
+    """Checks that model.merge_and_unload() is supported by underlying model.save_pretrained() when merging QLoRA
+    layers."""
+    if config.model_type != MODEL_LLM or config.trainer.type == "none":
+        return
+
+    if not (
+        config.adapter
+        and config.adapter.type == "lora"
```
arnavgarg1, line 652: lora or adalora?

alexsherstinsky: @arnavgarg1 Thank you for the great catch -- fixed.

PR #3680, opened 2023-09-29
[BUGFIX] Remove spurious test directory at the end of the test_llm.py::test_local_path_loading test run

alexsherstinsky:
### Scope
This change fixes the issue that after "tests/integration_tests/test_llm.py::test_local_path_loading" would run, a spurious directory would remain in the top directory of the Ludwig repository.  The fix consists of removing this directory after the assertions are done (as the last step of the test).

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/tests/integration_tests/test_llm.py
+++ b/tests/integration_tests/test_llm.py
@@ -1,5 +1,4 @@
 import os
-import shutil
 from typing import Dict, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


arnavgarg1 (approved):

arnavgarg1 (changes requested):
Instead, can we just use the tmpdir fixture for this test? That way we won't have to do any of this - what do you think?

alexsherstinsky:
> Instead, can we just use the tmpdir fixture for this test? That way we won't have to do any of this - what do you think?

@arnavgarg1 Yes, great idea -- updated and it works; thank you very much!

arnavgarg1 (approved):

tgaddair (approved):

PR #3683, opened 2023-10-02
[CI] Ensure that mlflow callback cleans up background-saving threads on trainer teardown.

justinxzhao:
https://github.com/ludwig-ai/ludwig/pull/3670 de-duplicates logged metrics to MLFlow by first checking if the metrics for a given training step have already been logged, and skipping if so.

This stateful mechanic ensures that if training completes at the same step as evaluation (default configuration), the `on_eval_end` callback logs metrics and the `on_trainer_teardown` callback skips `self.save_fn()`.

However, the `save_fn` block in `on_trainer_teardown` also takes care of two cleanup steps:
1. `self.save_thread.join()` to join the background-saving thread.
2. `self.save_fn((..., False))` to break the infinite background loop, `_log_mlflow_loop`, which is started in `on_trainer_train_setup` if background saving is enabled.

These were improperly skipped by the new stateful mechanic, causing `pytest` to hang and reach timeout.

In `on_trainer_teardown`, the fix is to 1) always join the `save_thread` and 2) always call `self.save_fn((None, None, None, False))` to ensure that the background saving loop breaks.

This PR also fixes the recent `integration_test_e` timeouts.

Credit to @jeffkinnison who pinpointed the culprit and issue, and worked with me on this solution.

CC: @Infernaught @geoffreyangus

### DIFF
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -161,15 +161,24 @@ class MlflowCallback(Callback):
     def on_eval_end(self, trainer, progress_tracker, save_path):
         if progress_tracker.steps not in self.logged_steps:
             self.logged_steps.add(progress_tracker.steps)
-            self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, True))  # Why True?
+            self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, True))
 
     def on_trainer_train_teardown(self, trainer, progress_tracker, save_path, is_coordinator):
+        print("Hello! Reached on trainer train teardown.")
         if is_coordinator:
             if progress_tracker.steps not in self.logged_steps:
                 self.logged_steps.add(progress_tracker.steps)
-                self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, False))  # Why False?
-                if self.save_thread is not None:
-                    self.save_thread.join()
+                self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, False))
+            else:
+                # TODO(Justin): This should probably live in on_ludwig_end, once that's implemented.
+                # Ensure that we break the MLFlow loop if save_in_background is True.
+                self.save_fn((None, None, None, False))
+
+            # Close the save_thread.
+            if self.save_thread is not None:
+                self.save_thread.join(timeout=5)
+                if self.save_thread.is_alive():
+                    logger.warning("MLFlow save thread timed out and did not close properly.")
 
     def on_visualize_figure(self, fig):
         # TODO: need to also include a filename for this figure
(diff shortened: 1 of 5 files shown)


ludwig/contribs/mlflow/__init__.py
```diff
@@ -161,15 +161,26 @@ def on_trainer_train_setup(self, trainer, save_path, is_coordinator):
...
+                self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, False))
+            # False ensures that the background saving loop breaks.
+            # TODO(Justin): This should probably live in on_ludwig_end, once that's implemented.
+            self.save_fn((None, None, None, False))
+
+            # Close the save_thread.
+            if self.save_thread is not None:
+                self.save_thread.join(timeout=5)
+                if self.save_thread.is_alive():
+                    logger.warning("MLFlow save thread timed out and did not close properly.")
```
jeffkinnison, line 183: This still leaves us with a potential hang. Should we modify this to be a [daemon thread](https://docs.python.org/3/library/threading.html#threading.Thread.daemon)? The downside there is that we wouldn't have detected this issue as quickly if the thread was just shut down on main thread exit.

ludwig/contribs/mlflow/__init__.py
```diff
@@ -161,15 +161,26 @@ def on_trainer_train_setup(self, trainer, save_path, is_coordinator):
...
                 self.logged_steps.add(progress_tracker.steps)
-                self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, False))  # Why False?
-                if self.save_thread is not None:
-                    self.save_thread.join()
+                # Adds a tuple to the logging queue.
+                # False is passed to indicate that the background saving loop should break.
+                self.save_fn((progress_tracker.log_metrics(), progress_tracker.steps, save_path, False))
+            # False ensures that the background saving loop breaks.
+            # TODO(Justin): This should probably live in on_ludwig_end, once that's implemented.
+            self.save_fn((None, None, None, False))
```
jeffkinnison, line 177: Maybe we could add shutdown [event](https://docs.python.org/3/library/threading.html#event-objects) in a followup PR. Then we could potentially get rid of the flag and this dummy queue entry altogether.

jeffkinnison (approved):
LGTM for a quick fix, just a couple of suggestions for how we could potentially iterate on this fix.

ludwig/contribs/mlflow/__init__.py
justinxzhao: Since this seems to have been largely working with regular thread joining, I think we can defer the daemon thread change. If we start seeing additional warnings in our logs, happy to revisit though.

ludwig/contribs/mlflow/__init__.py
justinxzhao: Acknowledged.

justinxzhao:
test_mlflow.py is working with the timeout removal.

jeffkinnison:
> test_mlflow.py is working with the timeout removal.

Oh, interesting. Do we know why the timeout was causing problems?

PR #3684, opened 2023-10-02
Set the metadata only during first training run

Infernaught:
This PR allows users to call model.train multiple times (such that the first training run is on a dataset that contains all possible outputs and all subsequent training runs are on datasets whose outputs are subsets of the first's) by setting the metadata only during the first training run.


ludwig/api.py
```diff
@@ -443,6 +443,14 @@ def train(
             `(training_set, validation_set, test_set)`.
             `output_directory` filepath to where training results are stored.
         """
+        # Only reset the metadata if the model has not been trained before
+        if self.training_set_metadata:
+            logging.warn(
+                "Previous metadata has been detected. Overriding `training_set_metadata` with metadata from previous "
+                "training run."
```
w4nderlust, line 450: Not a huge fan of this warning. The concept of a training set metadata is an internal one to LudwigModel, the user doesn't need to know about it and does not know about it, so the warning is not super useful for that user.
Better warning would sound like "This model has been trained before and its architecture has been defined based on the original training set properties (i.e. the number of output classes for a category output). The new data provided will be mapped into the previous architecture and it is not possible to modify the architecture based on the new training data provided, if you want to achieve that you should concatenate the new data with the previous data and train a new model from scratch." Or soemthing along those lines

justinxzhao (approved):

justinxzhao (commented):
Let's see if there's a test that we can write for this in test_api.py!

justinxzhao (commented):
Hi @Infernaught, the changes LGTM. However, it looks like there was a bad rebase since these changes from previous commits appear in the PR. Could you reconcile these diffs?

justinxzhao (approved):

PR #3687, opened 2023-10-04
Add ability to upload Ludwig models to Predibase.

martindavis:
Adds the ability for Ludwig users to upload their models to Predibase.

Todo:

- [ ] Test code locally
- [ ] Update any relevant documentation

### DIFF
--- a/ludwig/utils/upload_utils.py
+++ b/ludwig/utils/upload_utils.py
@@ -1,5 +1,8 @@
 import logging
 import os
+import tempfile
+import zipfile
+from pathlib import Path
 from abc import ABC, abstractmethod
 from typing import Optional
 
@@ -37,6 +40,8 @@ class BaseModelUpload(ABC):
         private: Optional[bool] = False,
         commit_message: Optional[str] = None,
         commit_description: Optional[str] = None,
+        dataset_file: Optional[str] = None,
+        dataset_name: Optional[str] = None,
     ) -> bool:
         """Abstract method to upload trained model artifacts to the target repository.
 
@@ -61,6 +66,8 @@ class BaseModelUpload(ABC):
         private: Optional[bool] = False,
         commit_message: Optional[str] = None,
         commit_description: Optional[str] = None,
+        dataset_file: Optional[str] = None,
+        dataset_name: Optional[str] = None,
     ):
         """Validate parameters before uploading trained model artifacts.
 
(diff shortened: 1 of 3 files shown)


ludwig/utils/upload_utils.py
```diff
@@ -205,3 +212,88 @@ def upload(
...
+        try:
+            dataset = self.pc.upload_dataset(
+                file_path=dataset_file,
+                name=dataset_name,
+            )
+        except Exception as e:
+            raise Exception(f"Failed to upload dataset to Predibase: {e}")
+            return True
+
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tgaddair, line 279: Is this zipping something we can add to the Predibase SDK, like:

```
pc.upload_model_from_dir(...)
```

Or just check in `upload_model` if the path provided is a directory instead of an archive?

ludwig/utils/upload_utils.py
```diff
@@ -205,3 +212,88 @@ def upload(
...
+                description=commit_description,
+                exist_ok=True,
+            )
+        except Exception as e:
+            raise Exception(f"Failed to create repo in Predibase: {e}")
+            return True
+
+        # Upload the dataset to Predibase
+        try:
+            dataset = self.pc.upload_dataset(
```
tgaddair, line 271: Is it necessary to associate a dataset with the model when uploading?

ludwig/utils/upload_utils.py
```diff
@@ -205,3 +212,88 @@ def upload(
...
+
+            # Upload the zip file to Predibase
+            try:
+                self.pc.upload_model(
+                    repo_id=repo.id,
+                    zip_fp=model_data_zip_path,
+                    dataset=dataset,
+                )
+            except Exception as e:
+                raise Exception(f"Failed to upload model to Predibase: {e}")
```
tgaddair, line 296: For all of these exceptions, I might use the `from e` pattern instead to retain the original stack trace:

```
raise RuntimeException("Failed to upload model to Predibase") from e
```

ludwig/utils/upload_utils.py
martindavis: I think the answer to that question depends on what we think the primary use case will be. If a user specifies a Model Repo with at least one trained Model Version in it, they don't need to specify a dataset. However, if they are trying to create a new Model Repo, they will need a dataset.

ludwig/utils/upload_utils.py
martindavis: Good callout! I thought it was funky as well and will update Predibase!

ludwig/utils/upload_utils.py
tgaddair: Oof, we should definitely remove that restriction.

requirements.txt
```diff
@@ -56,3 +56,6 @@ pyxlsb>=1.0.8       # excel
 pyarrow             # parquet
 lxml                # html
 html5lib            # html
+
+# Allows users to upload
+predibase>=2023.10.2
```
tgaddair, line 61: Please move this to `requirements_extra.txt` instead.

martindavis: Done! @justinxzhao and I talked about this on Friday. Since HF is part of the core requirements, it felt like it could go either way but I agree this keeps Ludwig cleaner.

ludwig/utils/upload_utils.py
martindavis: Agreed offline to explore this at a later date.

ludwig/utils/upload_utils.py
```diff
@@ -205,3 +248,143 @@ def upload(
             return True
 
         return False
+
+
+class Predibase(BaseModelUpload):
+    def __init__(self):
+        self.pc = None
```
justinxzhao, line 255: Could we login at object construction time? That way, instead of needing to do:

```
predibase = Predibase()
predibase.login()
predibase.upload()
```

The caller can just do:

```
predibase = Predibase()
predibase.upload()
```

justinxzhao (commented):
Is there a reasonable test that we can write for this?

martindavis:
> Is there a reasonable test that we can write for this?

Great question @justinxzhao ! @arnavgarg1 and I talked about adding an integration test in the CI to make sure that this works for both HF and Predibase. Our plan is to tackle that late next week!

ludwig/utils/upload_utils.py
```diff
@@ -205,3 +249,144 @@ def upload(
...
+                repo=repo,
+                model_path=model_path,
+                dataset=dataset,
+            )
+        except Exception as e:
+            raise RuntimeError("Failed to upload model to Predibase") from e
+            return True
+
+        logger.info(f"Model uploaded to Predibase with repository name `{repo_id}`")
+        return False
```
tgaddair, line 392: QQ: why does this function return False on success and True on error? Can you add a comment about the return type to the docstring? My intuition here would be that it should return True on success, for what it's worth.

ludwig/utils/upload_utils.py
```diff
@@ -205,3 +249,144 @@ def upload(
...
+                systems. Not used in the base class, but subclasses may use it for specific repository
+                implementations. Defaults to None.
+            commit_description (str, optional): A description of the commit when uploading to version control
+                systems. Not used in the base class, but subclasses may use it for specific repository
+                implementations. Defaults to None.
+
+        Raises:
+            AssertionError: If the repo_id has non-url safe characters.
+        """
+        assert len(repo_id) <= 255, "`repo_id` must be 255 characters or less."
```
tgaddair, line 315: Only use asserts for checking internal (programmer sanity check) invariants. Raise a `ValueError` for user input errors like this.

ludwig/utils/upload_utils.py
```diff
@@ -205,3 +249,144 @@ def upload(
...
+                but subclasses may use it for specific repository implementations. Defaults to False.
+            commit_message (str, optional): A message to attach to the commit when uploading to version control
+                systems. Not used in the base class, but subclasses may use it for specific repository
+                implementations. Defaults to None.
+            commit_description (str, optional): A description of the commit when uploading to version control
+                systems. Not used in the base class, but subclasses may use it for specific repository
+                implementations. Defaults to None.
+
+        Raises:
+            AssertionError: If the repo_id has non-url safe characters.
```
tgaddair, line 313: ValueError

ludwig/utils/upload_utils.py
arnavgarg1: @justinxzhao I agree, that's a cleaner/experience. I do think that most users won't use these classes directly, but instead use it via the LudwigModel API wrapper (https://github.com/ludwig-ai/ludwig/blob/master/ludwig/api.py#L1710), so it may be ok since it feels like an implementation detail. In either case, we can make this change!

ludwig/utils/upload_utils.py
arnavgarg1: Agreed, I am also confused by this

martindavis: Good callout! I mixed up my `True` and `False` because Predibase works a little different than HF. Fixing now!

martindavis: In addition, I've removed the other return statements as they're unreachable.

ludwig/utils/upload_utils.py
martindavis: Oh nice! Ok, done!

ludwig/utils/upload_utils.py
```diff
@@ -142,6 +124,68 @@ def login(self):
...
+                implementations. Defaults to None.
+            commit_description (str, optional): A description of the commit when uploading to version control
+                systems. Not used in the base class, but subclasses may use it for specific repository
+                implementations. Defaults to None.
+
+        Raises:
+            ValueError: If the repo_id does not have both a namespace and a repo name separated by a '/'.
+        """
+        # Validate repo_id has both a namespace and a repo name
+        if "/" not in repo_id:
```
martindavis, line 162: @arnavgarg1 , I've updated the HF code as well. Can you double check this?

arnavgarg1: LGTM!

ludwig/utils/upload_utils.py
martindavis: Addressed this by adding `self.login()` to `__init__` methods

arnavgarg1 (approved):

tgaddair (approved):

PR #3690, opened 2023-10-05
Move on_epoch_end and epoch increment to after run_evaluation loop.

justinxzhao:
https://github.com/ludwig-ai/ludwig/pull/3668 inadvertently moved the `on_epoch_end` callback to be invoked after the last batch has been processed, but before evaluation.

While technically still valid, this is a change from the previous behavior of `on_epoch_end` being called after evaluation, which may come as a surprise to users assuming evaluation results being present in `on_epoch_end`.

Note: `on_eval_end` is the recommended callback for handling fresh eval metrics.

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -1098,6 +1098,7 @@ class Trainer(BaseTrainer):
         """Completes up to one epoch through the data."""
         self.distributed.zero_grad(self.optimizer)
         batch_idx = 0
+        has_called_on_epoch_end = False
         while not batcher.last_batch() and progress_tracker.steps < self.total_steps:
             progress_tracker.learning_rate = self.optimizer.param_groups[0]["lr"]
             self.callback(lambda c: c.on_batch_start(self, progress_tracker, save_path))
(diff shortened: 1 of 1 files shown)


ludwig/trainers/trainer.py
```diff
@@ -1192,6 +1185,13 @@ def _train_loop(
                 if should_break:
                     return should_break
```
jeffkinnison, line 1186: Will we be one epoch off if we return here?

arnavgarg1 (approved):

ludwig/trainers/trainer.py
jeffkinnison: I think when the end of an epoch lines up with eval and early stopping, the final epoch increment won't happen.

ludwig/trainers/trainer.py
```diff
@@ -1189,8 +1183,26 @@ def _train_loop(
                     if self.is_coordinator():
                         progress_tracker.save(os.path.join(save_path, TRAINING_PROGRESS_TRACKER_FILE_NAME))
 
+                # If this was the last batch, then increment the epoch counter and invoke the `on_epoch_end` callback.
+                if not has_called_on_epoch_end and batcher.last_batch():
```
jeffkinnison, line 1187: I'm not a huge fan of the duplicate code blocks. If we remove the early return below, this block should be unnecessary, right?

ludwig/trainers/trainer.py
```diff
@@ -1189,8 +1183,26 @@ def _train_loop(
...
+                    progress_tracker.epoch += 1
+                    self.callback(lambda c: c.on_epoch_end(self, progress_tracker, save_path))
+
+                    # Ensures that we only call `on_epoch_end` once, for example, if evaluation is configured to run
+                    # several times per epoch.
+                    has_called_on_epoch_end = True
+
                 if should_break:
-                    return should_break
+                    return True
```
jeffkinnison, line 1196: The early return here seems odd given that the only following piece of code is an epoch increment and callback that won't happen before the end of the epoch. What if we set `should_break = False` at the top of this method, then returned `should_break` at the end? That would let us remove the duplicate block above, remove this block, and make it so we don't need the `has_called_on_epoch_end` flag.

ludwig/trainers/trainer.py
justinxzhao: +1

ludwig/trainers/trainer.py
justinxzhao: Ah, that's a brain fart on my end -- thanks for the suggestion. Returning `should_break` is definitely simpler!

jeffkinnison (approved):
Nice! This looks great, and it simplifies the second half of `_train_loop` quite a bit

PR #3698, opened 2023-10-06
[MAINTENANCE] Fix the linting warnings in two backend component classes.

alexsherstinsky:
### Scope
* Resolving linting warnings -- making some methods static and including the proper type hints.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/backend/base.py
+++ b/ludwig/backend/base.py
@@ -150,9 +150,6 @@ class Backend(ABC):
 
 
 class LocalPreprocessingMixin:
-    def __init__(self) -> None:
-        pass
-
     @property
     def df_engine(self):
         return PANDAS
@@ -196,9 +193,6 @@ class LocalPreprocessingMixin:
 
 
 class LocalTrainingMixin:
-    def __init__(self) -> None:
-        pass
-
     @staticmethod
     def initialize():
         init_dist_strategy("local")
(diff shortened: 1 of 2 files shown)


ludwig/backend/base.py
```diff
@@ -150,6 +150,9 @@ def supports_batch_size_tuning(self) -> bool:
 
 
 class LocalPreprocessingMixin:
+    def __init__(self) -> None:
```
tgaddair, line 153: Hmm, does the linter complain about the mixins not having initializers? My concern with adding them is it could mess things up if the mixing sits in the middle of the Method Resolution Order. For example:

```
class Derived(Mixin, Base):
   def __init__(self, *args, **kwargs):
       super().__init__(*args, **kwargs)

class Mixin:
    def __init__(self):
        pass

class Base:
    def __init__(self, *args, **kwargs):
        ...
```

One way you could keep the initializer if it is in fact required would be to use [cooperative multiple inheritance](https://stackoverflow.com/questions/16305177/cooperative-multiple-inheritance-issue), where the Mixin initializer is essentially a passthrough to the next class in the MRO:

```
class Mixin
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```

What do you think?

alexsherstinsky: @tgaddair Upon further reflection, I have to agree that for the Mixin use case here it is best to leave no Initializer (because the default initializer in `object` will do nothing, as is best in this case).  Thank you for raising the concern.

tgaddair (approved):

PR #3711, opened 2023-10-10
Define initialize_llm() and generate() methods. Remove extra logging in llm.py

justinxzhao:
Sample code:

```python
model = LudwigModel(config)
model.generate("Hello")  # -> str
model.generate(["Hello", "Goodbye"])  # List[str]
```

Some benchmarking of inference times:

image

Observations:
- Inference time scales linearly with `max_new_tokens` across the board.
- `ludwig.predict()` (single inference) is slightly faster than single inference `ludwig.generate(single)`.
- `ludwig.generate(batch)` batch inference is change is significantly faster than `ludwig.generate(single)`.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -25,6 +25,7 @@ import logging
 import os
 import sys
 import tempfile
+import time
 import traceback
 from collections import OrderedDict
 from pprint import pformat
@@ -103,6 +104,7 @@ from ludwig.utils.misc_utils import (
     set_saved_weights_in_checkpoint_flag,
 )
 from ludwig.utils.print_utils import print_boxed
+from ludwig.utils.tokenizers import HFTokenizer
 from ludwig.utils.torch_utils import DEVICE
 from ludwig.utils.trainer_utils import get_training_report
 from ludwig.utils.types import DataFrame, TorchDevice
@@ -460,6 +462,16 @@ class LudwigModel:
             `(training_set, validation_set, test_set)`.
             `output_directory` filepath to where training results are stored.
         """
+        # Only reset the metadata if the model has not been trained before
+        if self.training_set_metadata:
+            logger.warning(
+                "This model has been trained before. Its architecture has been defined by the original training set "
+                "(for example, the number of possible categorical outputs). The current training data will be mapped "
+                "to this architecture. If you want to change the architecture of the model, please concatenate your "
+                "new training data with the original and train a new model from scratch."
+            )
+            training_set_metadata = self.training_set_metadata
+
         if self._user_config.get(HYPEROPT):
             print_boxed("WARNING")
             logger.warning(HYPEROPT_WARNING)
@@ -903,29 +915,22 @@ class LudwigModel:
         generation_config: Optional[dict] = None,
     ) -> Union[str, List[str]]:
         """A simple generate() method that directly uses the underlying transformers library to generate text."""
-        import time
-
         if self.config_obj.model_type != MODEL_LLM:
             raise ValueError(
                 f"Model type {self.config_obj.model_type} is not supported by this method. Only `llm` model type is "
                 "supported."
             )
-
         if not torch.cuda.is_available():
             raise ValueError("GPU is not available.")
-
-        from ludwig.utils.tokenizers import HFTokenizer
 
         if not self.model.model.config.is_encoder_decoder:
             padding_side = "left"
         else:
             padding_side = "right"
-
         tokenizer = HFTokenizer(self.config_obj.base_model, padding_side=padding_side)
 
-        start_time = time.time()
-
         with self.model.use_generation_config(generation_config):
+            start_time = time.time()
             inputs = tokenizer.tokenizer(input_strings, return_tensors="pt", padding=True)
             input_ids = inputs["input_ids"].to("cuda")
             attention_mask = inputs["attention_mask"].to("cuda")
@@ -939,7 +944,7 @@ class LudwigModel:
                     generation_config=self.model.generation,
                 )
                 decoded_outputs = tokenizer.tokenizer.batch_decode(outputs, skip_special_tokens=True)
-                logging.info(f"Finished generating in: {(time.time() - start_time):.2f}s.")
+                logger.info(f"Finished generating in: {(time.time() - start_time):.2f}s.")
                 if len(decoded_outputs) == 1:
                     return decoded_outputs[0]
                 return decoded_outputs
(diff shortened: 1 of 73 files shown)


arnavgarg1:
Can we add tests?

ludwig/api.py
```diff
@@ -891,6 +910,46 @@ def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_s
...
+        input_strings: Union[str, List[str]],
+        generation_config: Optional[dict] = None,
+    ) -> Union[str, List[str]]:
+        """A simple generate() method that directly uses the underlying transformers library to generate text."""
+        if self.config_obj.model_type != MODEL_LLM:
+            raise ValueError(
+                f"Model type {self.config_obj.model_type} is not supported by this method. Only `llm` model type is "
+                "supported."
+            )
+        if not torch.cuda.is_available():
```
arnavgarg1, line 924: nit: Can also add `or torch.cuda.device_count() == 0`

ludwig/api.py
```diff
@@ -891,6 +910,46 @@ def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_s
...
+            raise ValueError("GPU is not available.")
+
+        if not self.model.model.config.is_encoder_decoder:
+            padding_side = "left"
+        else:
+            padding_side = "right"
```
arnavgarg1, line 930: nit: can we add a comment for why we need to do this?

ludwig/api.py
```diff
@@ -332,6 +334,23 @@ def __init__(
...
+            and self.config_obj.output_features[0].type == "text"
+        ):
+            self._initialize_llm()
+
+    def _initialize_llm(self, random_seed: int = default_random_seed):
+        """Initialize the LLM model.
+
+        Should only be used in a zero-shot (NoneTrainer) setting.
+        """
+        self.model = LudwigModel.create_model(self.config_obj, random_seed=random_seed)
```
arnavgarg1, line 352: It's possible that at the end of `LudwigModel.create()` in `_initialize_llm`, the actual `self.model.model` is on CPU.  Specifically, this is the case when quantization isn't being used to load the model. (https://github.com/ludwig-ai/ludwig/blob/master/ludwig/models/llm.py#L83C1-L87C43), specifically, the `device_map` keyword decides whether it is on CPU or GPU. If not set to `auto` (in our case, when there's no quantization), it ends up being loaded on CPU instead of GPU. 

It's not wrong necessarily to leave the model on CPU and call generate, but I'd double check that `self.model.model.device` is 'cuda', and if not, I'd move it to CUDA because generation will be much faster that way.

arnavgarg1 (commented):
Generally LGTM, but some callouts based on backend config and model's device on load.

ludwig/api.py
```diff
@@ -891,6 +910,46 @@ def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_s
...
+                "supported."
+            )
+        if not torch.cuda.is_available():
+            raise ValueError("GPU is not available.")
+
+        if not self.model.model.config.is_encoder_decoder:
+            padding_side = "left"
+        else:
+            padding_side = "right"
+        tokenizer = HFTokenizer(self.config_obj.base_model, padding_side=padding_side)
```
arnavgarg1, line 931: Do we want to use the tokenizer initialized in `self.model.tokenizer` once https://github.com/ludwig-ai/ludwig/pull/3745 lands?

ludwig/api.py
```diff
@@ -992,6 +1052,7 @@ def predict(
 
                     logger.info(f"Saved to: {output_directory}")
 
+            logger.info(f"Finished decoding in: {(time.time() - start_time):.2f}s.")
```
arnavgarg1, line 1055: Not totally sure we want to keep this one around since we're logging it in the generate call anyway? Unless you were using this to add visibility for `model.predict()` calls in which case I wouldn't set it to `Finished decoding in: ` since this is actually prediction and `decoding` is an implementation detail

justinxzhao: Makes sense. I changed this log to `Finished predicting in...`.

`generate()` and `predict()` are two separate functions, so that's why I added separate logging statements for each.

ludwig/api.py
```diff
@@ -891,6 +910,46 @@ def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_s
...
+        generation_config: Optional[dict] = None,
+    ) -> Union[str, List[str]]:
+        """A simple generate() method that directly uses the underlying transformers library to generate text."""
+        if self.config_obj.model_type != MODEL_LLM:
+            raise ValueError(
+                f"Model type {self.config_obj.model_type} is not supported by this method. Only `llm` model type is "
+                "supported."
+            )
+        if not torch.cuda.is_available():
+            raise ValueError("GPU is not available.")
```
geoffreyangus, line 925: is this required?

ludwig/api.py
justinxzhao: There's extra handling for the padding side that I had to add in this function to bypass warnings from HF.

I've added a TODO to figure out if this should also be folded into `llm.py`.

Regardless, we should still consolidate the tokenizer initialization in `llm.py`.

ludwig/api.py
justinxzhao: Done.

ludwig/api.py
justinxzhao: Done.

ludwig/api.py
justinxzhao: It seems to be for loading quantized models: https://github.com/ludwig-ai/ludwig/issues/3695. This error string might be preferable to the other one. WDYT?

ludwig/api.py
justinxzhao: Ooh, good call out and good to know.

I added a check to move the model to cuda if the model device is CPU and tested this out with a couple of different non-quantized LLMs.

```
Loading large language model...
Done.
Loaded HuggingFace implementation of facebook/opt-350m tokenizer
LLM was initialized on cpu. Moving to GPU for inference.
Finished generating in: 0.57s.
Finished generating in: 0.16s.
Finished generating in: 0.31s.
Finished generating in: 0.61s.
Finished generating in: 1.24s.
Finished generating in: 2.47s.
Finished generating in: 5.04s.
```

ludwig/api.py
geoffreyangus: Nice! Yes I agree, thanks for the clarification. Can you put a comment there linking to issue 3695?

geoffreyangus (approved):

ludwig/api.py
justinxzhao: I didn't realize that `torch.cuda.is_available()` and `torch.cuda.device_count() == 0` could be simultaneously true, though it looks like this [may have been fixed](https://github.com/pytorch/pytorch/issues/90543).

justinxzhao:
@arnavgarg1 Merging for now -- lmk if you have any additional feedback and I'd be happy to follow up.

PR #3716, opened 2023-10-11
Limit model_presets to llama and mistral models

Infernaught:
Limit model presets to Llama and Mistral models to avoid issues with other models


arnavgarg1 (commented):
LGTM. 

One thought - do you think we need a backwards compatibility warning here? I believe this list of model presets was launched as a part of Ludwig 0.8 so it might flag config errors if we just remove them, but the better think to do here is to probably remove them and raise a Runtime error indicating that these preset models are no longer supported and that the full paths must be passed in.

cc: @justinxzhao

martindavis (approved):

PR #3717, opened 2023-10-11
QOL: Update recommended vscode settings.

justinxzhao:
Context: https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions

image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,10 +3,15 @@
         120
     ],
     "editor.formatOnSave": true,
-    "python.formatting.provider": "black",
-    "python.linting.enabled": true,
-    "python.linting.flake8Enabled": true,
-    "python.linting.flake8Args": [
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.black-formatter",
+        "editor.formatOnSave": true
+    },
+    "black-formatter.args": [
+        "--line-length",
+        "120"
+    ],
+    "flake8.args": [
         "--config=setup.cfg"
     ],
     "python.testing.unittestEnabled": false,


.vscode/settings.json
```diff
@@ -3,10 +3,15 @@
         120
     ],
     "editor.formatOnSave": true,
```
arnavgarg1, line 5: should we delete this one as well since its wrapped within [python] > editor.formatOnSave?

.vscode/settings.json
```diff
@@ -3,10 +3,15 @@
...
-    "python.linting.flake8Args": [
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.black-formatter",
+        "editor.formatOnSave": true
+    },
+    "black-formatter.args": [
+        "--line-length",
+        "120"
+    ],
+    "flake8.args": [
```
arnavgarg1, line 14: Does flake8 need to enabled independent of flake8 args?

.vscode/settings.json
justinxzhao: There are other non-python files that this is useful for like JSON.

.vscode/settings.json
justinxzhao: That's right -- flake8 needs to be installed as an extension rather than enabled via `settings.json` (same as black). From [this documentation](https://marketplace.visualstudio.com/items?itemName=ms-python.flake8):

```
Once installed in Visual Studio Code, flake8 will be automatically executed when you open a Python file.

If you want to disable flake8, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension) per workspace in Visual Studio Code.
```

arnavgarg1: Gotcha!

arnavgarg1 (approved):

PR #3719, opened 2023-10-11
QOL: Fail during preprocessing if max sequence lengths are shorter than the prompt template.

justinxzhao:
During preprocessing, raise an error if the max sequence length model can support would truncate all unique info in the training example.

Cases considered:
1. input max sequence length < |prompt template|
2. global max sequence length < |prompt template|

## Case 1: input max sequence length < |prompt template|

- (refactored) During preprocessing, `TextFeature.get_feature_meta()` performs two main steps.
	- a) vocabulary creation: we tokenize the column of data to determine the vocabulary, take note of the true max sequence length of the column, and set vocabulary-index maps.
	- b) max sequence length reconciliation: we apply a simple policy to reconcile the final max sequence length that should be used.
		- If the max sequence length is explicitly specified, we use the minimum of the true maximum sequence length and the explicitly specified value. If the explicitly specified value is less than the true maximum sequence length, we log a warning. If the max sequence length is not specified, we use the true maximum sequence length as the feature's max sequence length.
- (new) Compute `prompt_template_num_tokens` as part of vocabulary creation (step a).
- (new) Add the model configuration, which contains the prompt template, as a new argument to `TextFeature.get_feature_meta()`.
	- This requires making a change to `BaseFeature.get_feature_meta()` and all subclasses to accept the new arugment.
- (new) During max sequence length reconciliation (step b), if the final max sequence length for the input feature is smaller than the number of tokens in the prompt template, a ValueError is raised.

## Case 2: global max sequence length < |prompt template|

- (new) We return `prompt_template_num_tokens` in the return of `Text.Feature.get_feature_meta()` so that it can be referenced in `build_dataset()`.
- A `ValueError` is raised if the number of tokens in the prompt template is larger than the `global_max_sequence_length`.


ludwig/utils/strings_utils.py
```diff
@@ -295,65 +376,49 @@ def create_vocabulary(
...
-            padding_symbol = pad_token
-    elif hasattr(tokenizer, "get_vocab"):
-        vocab = tokenizer.get_vocab()
-        vocab = list(vocab.keys())
-    elif vocab_file is not None:
-        vocab = load_vocabulary(vocab_file)
+    # Number of tokens in template.
+    prompt_template_num_tokens = -1
+    if prompt_template:
+        prompt_template_num_tokens = len(tokenizer(prompt_template))
```
arnavgarg1, line 388: Just to make sure I understand, this would also count the placeholder tokens like `{object}` or `{pronoun}` in the template as tokens and tokenized them right, for e.g.:

```python
from transformers import AutoTokenizer

template = "The {pronoun} sits on the {object}"

tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
tokens = tokenizer.tokenize(template, add_special_tokens=True)
total_tokens = len(tokens)
```

would give you something like:
`['[CLS]', 'the', '{', 'pro', '##no', '##un', '}', 'sits', 'on', 'the', '{', 'object', '}', '[SEP]']` for the tokens and len as 14 (same as just calling tokenizer(data), but it shows you the actual tokens that will get mapped to token IDs).

I'm wondering whether it makes sense to replace the placeholder tokens with blank spaces before calling `tokenizer(prompt_template)` since we ultimately want the "raw" length of the prompt template since these placeholders will actually get replaced by real data. 

```python
import re
from transformers import AutoTokenizer

template = "The {pronoun} sits on the {object}"

tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")


placeholder_pattern = r'\{[^\}]+\}'
placeholder_tokens = re.findall(placeholder_pattern, template)
for token in placeholder_tokens:
    template = template.replace(token, "")

tokens = tokenizer.tokenize(template, add_special_tokens=True)
total_tokens = len(tokens)
```

which would result in `['[CLS]', 'the', 'sits', 'on', 'the', '[SEP]']` as the tokens and a length of 6. 

I think this is probably a more accurate reflection of the tokens in the prompt template but curious what you think!

ludwig/utils/strings_utils.py
```diff
@@ -243,6 +243,84 @@ class Vocabulary:
...
+
+        if pad_token is None and add_special_symbols:
+            logger.warning(
+                "No padding token found in HuggingFace tokenizer. Adding one. "
+                + "NOTE: This will change the vocabulary size and may affect model "
+                + "performance, particularly if the model weights are frozen."
+            )
+            vocab = [padding_symbol] + vocab
+        else:
+            padding_symbol = pad_token
```
arnavgarg1, line 305: Just a general thought - adding these tokens to the vocab is likely going to cause some problems at training time because the embedding layer in pretrained text encoder assumes a fixed vocab size (assuming training from pretrained instead of pertaining from scratch) as one of the arguments in the Embedding layer initialization, so hopefully in our HFTokenizer class implementation, we always set the pad and unk token so that 

```python
pad_token = tokenizer.get_pad_token()
unk_token = tokenizer.get_unk_token()
```

are never none.

arnavgarg1: I just took a look at our implementation. I want to say that unk_token will likely always be set, and our HFTokenizer class does a best effort setting of the pad token with a resolve to set the `pad_token` to token ID 0. I'd say that I've also missed this and am just realizing it, but we probably also want to set tokenizer.pad_token to the corresponding vocab token mapping to token_id 0 after [this](https://github.com/ludwig-ai/ludwig/blob/305fdd2d921c39baba104bdc4e769730ee4b6fad/ludwig/utils/tokenizers.py#L882). 

Otherwise since `tokenizer.get_pad_token()` just returns `self.tokenizer.pad_token`, it may actually return an unknown here and try and insert a different pad token into the vocab.

arnavgarg1: I know this isn't net new code and that you just refactored the create_vocabulary() function into this new private helper method, but just wanted to flag it anyway! We can come back to it in a future PR, of it's easy to do, could be worth just adding a line to also set the pad_token to the corresponding pad_token_id mapping from the vocab.

ludwig/utils/strings_utils.py
```diff
@@ -243,6 +243,84 @@ class Vocabulary:
...
+    add_special_symbols: bool,
+    padding_symbol: str,
+    unit_counts: Counter,
+    num_most_frequent: int,
+) -> Optional[List[str]]:
+    """Returns the vocabulary from the tokenizer_type, tokenizer, or vocab_file.
+
+    If there's no vocab_file or if the tokenizer has no set vocabulary (e.g. space_punct), then the vocabulary is
+    determined from the tokenized data.
+    """
```
arnavgarg1, line 267: nit: Can we copy the docstring from the original `create_vocabulary` function into this helper function

ludwig/utils/strings_utils.py
```diff
@@ -243,6 +243,84 @@ class Vocabulary:
...
+    # The tokenizer has a preset vocabulary.
+    if hasattr(tokenizer, "get_vocab"):
+        vocab = tokenizer.get_vocab()
+        return list(vocab.keys())
+
+    # Load the vocabulary from the vocab file.
+    if vocab_file is not None:
+        return load_vocabulary(vocab_file)
+
+    print(f"padding_symbol: {padding_symbol}")
```
arnavgarg1, line 318: nit: might need to delete this print

ludwig/features/text_feature.py
```diff
@@ -76,6 +77,42 @@ def get_decoded_targets_and_predictions(
     return decoded_targets, decoded_predictions
 
 
+def _get_metadata_reconciled_max_sequence_length(
+    preprocessing_parameters: Dict, vocabulary: List[str]
```
arnavgarg1, line 81: nit: I think `vocabulary` is of type `Vocabulary`?

ludwig/features/text_feature.py
```diff
@@ -99,43 +148,26 @@ def feature_meta(column, preprocessing_parameters: PreprocessingConfigDict, back
...
-                f"preprocessing parameters"
+        max_sequence_length, max_sequence_length_99ptile = _get_metadata_reconciled_max_sequence_length(
+            preprocessing_parameters, vocabulary
+        )
+
+        if is_input_feature and max_sequence_length < vocabulary.prompt_template_num_tokens:
+            raise ValueError(
+                f"The input feature's max sequence length ({max_sequence_length}) is shorter than the prompt template "
+                f"length ({vocabulary.prompt_template_num_tokens}). This will truncate all unique information. "
+                "Consider making the template shorter or increasing the input feature's max sequence length."
```
arnavgarg1, line 167: nit: I think same as before, would be good to be explicit in what exact change to make, something like set your input feature's max_sequence_length to a value >> {vocabulary.prompt_template_num_tokens}. 

Not entirely sure what the best wording would be or if this is too redundant in terms of info in the ValueError message, but wanted to call it out anyway

arnavgarg1 (commented):
Generally LGTM, left a few comments that may warrant a discussion and a couple of minor nits - happy to chat about any of them. 

The main ones are:

1.  How we're calculating the number of tokens in the prompt, which I feel should be the number of tokens in the prompt template without the placeholder tokens in the template as opposed to the current implementation that incldes the placeholder tokens as part of the token count of the template.
2. Accounting for the fact that prompt templates also live in ECD models under text feature preprocessing

ludwig/utils/strings_utils.py
justinxzhao: Thanks for writing out this example.

For longer prompts with single input columns, the token difference feels negligible, however you are totally right that excluding the bracketed elements is more accurate and I've incorporated your suggestion.

ludwig/utils/strings_utils.py
justinxzhao: Discussed offline. Pushed up this commit to ensure that pad token and pad token id is always available for HF tokenizers, and simplify the special symbol special handling in vocab generation: https://github.com/ludwig-ai/ludwig/pull/3719/commits/e9b4666d701d6835c0e100888287e22b515a0540

ludwig/utils/strings_utils.py
```diff
@@ -259,22 +337,25 @@ def create_vocabulary(
...
     then a pre-trained huggingface tokenizer is loaded from `pretrained_model_name_or_path` and that vocabulary is
     used directly.
 
     The UNKNOWN special symbol is always included in the final vocabulary. Additional special symbols (PADDING, START,
-    STOP) are added if add_special_symbols=True.
+    STOP) are added if add_special_symbols=True. If the tokenizer is a pre-trained huggingface tokenizer, then the
+    special symbols are taken from the tokenizer's vocabulary.
 
     Args:
+        prompt_template: The prompt template for the model. Applicable only to LLMs.
```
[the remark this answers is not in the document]
arnavgarg1: Sure! Let's just maybe throw in a to-do for it somewhere in the comments

arnavgarg1 (approved):
LGTM! 🚢

PR #3723, opened 2023-10-12
Fix calculation of dataset size

Infernaught:
We previously calculated a dataset's size from the length of whatever the "dataset" parameter was that was passed in. However, this parameter could have been a filepath, meaning that the dataset.size parameter would be set incorrectly, potentially causing index out of bound errors if the actual length of the dataset was shorter than the length of the filepath string.

This PR fixes this problem by calculating the size after the filepath has been processed into a dataset.

### DIFF
--- a/ludwig/data/batcher/test_batcher.py
+++ b/ludwig/data/batcher/test_batcher.py
@@ -1,0 +1,113 @@
+import logging
+
+import pandas as pd
+import yaml
+
+from ludwig.api import LudwigModel
+from ludwig.data.dataset.pandas import PandasDataset
+
+
+def test_pandas_size():
+    df = pd.DataFrame(
+        {"name": ["joe", "janice", "sara"], "mask": ["green", "black", "pink"], "weapon": ["stick", "gun", "gun"]}
+    )
+    config = yaml.safe_load(
+        """
+    model_type: llm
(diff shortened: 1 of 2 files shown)


arnavgarg1:
Really nice work figuring this one out @Infernaught! If it doesn't balloon the scope, can we add some tests to this PR (or even as a fast follow) to make sure we catch these kinds of errors in the future incase we make changes to the batcher? Some ideas: Check that the number of rows being returned by the batcher is the same as what's in the dataset when using a Pandas DataFrame, Dask DataFrame, and a file path as the dataset.

arnavgarg1 (approved):
You may need to get rid of the print statement!

justinxzhao:
> Really nice work figuring this one out @Infernaught! If it doesn't balloon the scope, can we add some tests to this PR (or even as a fast follow) to make sure we catch these kinds of errors in the future incase we make changes to the batcher? Some ideas: Check that the number of rows being returned by the batcher is the same as what's in the dataset when using a Pandas DataFrame, Dask DataFrame, and a file path as the dataset.

+1. Great find. Another test we should add is one that verifies that one pass through the batcher up to `last_batch()` actually uses each individual example exactly once.

justinxzhao (approved):
Thanks for adding the test!

Issue #3726, opened 2023-10-13
Docker image crashes when checking utilization


khellan:
**Describe the bug**
Training Llama in Docker image crashes because the recently added utilization logging is not available.

**To Reproduce**
 Run docker image **ludwigai/ludwig-gpu:master** with this
```
        ludwig train \
        --config /data/training.yaml \
        --data_format csv \
        --experiment_name llama-20231013-1 \
        --model_name english_classes \
        --test_set english_significant_test.csv \
        --training_set english_significant_train.csv \
        --validation_set english_significant_val.csv
````
My training.yaml:
input_features:
```
 - name: english
   type: text
   encoder:
     type: auto_transformer
     pretrained_model_name_or_path: meta-llama/Llama-2-7b-hf
     trainable: true
     adapter: 
      type: lora


output_features:
 - name: label
   type: category
  

trainer:
  learning_rate: 1.0e-05
  epochs: 3
````

Unfortunately, I can't provide training data.

**Expected behavior**
Training runs as it has before.

**Log**
````
2023-10-13T09:40:15Z [job] ╒══════════╕
2023-10-13T09:40:15Z [job] │ TRAINING │
2023-10-13T09:40:15Z [job] ╘══════════╛
2023-10-13T09:40:15Z [job] Creating fresh model training run.
2023-10-13T09:40:15Z [job] Training for 12579 step(s), approximately 3 epoch(s).
2023-10-13T09:40:15Z [job] Early stopping policy: 5 round(s) of evaluation, or 20965 step(s), approximately 5 epoch(s).
2023-10-13T09:40:15Z [job] Starting with step 0, epoch: 0
2023-10-13T09:40:16Z [job] Traceback (most recent call last):
2023-10-13T09:40:16Z [job] File "/opt/conda/bin/ludwig", line 8, in <module>
2023-10-13T09:40:16Z [job] sys.exit(main())
2023-10-13T09:40:16Z [job] File "/opt/conda/lib/python3.10/site-packages/ludwig/cli.py", line 191, in main
2023-10-13T09:40:16Z [job] CLI()
2023-10-13T09:40:16Z [job] File "/opt/conda/lib/python3.10/site-packages/ludwig/cli.py", line 71, in __init__
2023-10-13T09:40:16Z [job] getattr(self, args.command)()
2023-10-13T09:40:16Z [job] File "/opt/conda/lib/python3.10/site-packages/ludwig/cli.py", line 76, in train
[...]

PR #3727, opened 2023-10-13
Fix issue while logging cuda device utilization to tensorboard

arnavgarg1:
Fix for #3726 

```
2023-10-13T09:40:16Z [job] torch.cuda.device(i).utilization(),
2023-10-13T09:40:16Z [job] AttributeError: 'device' object has no attribute 'utilization'
```

API utilization from docs: https://pytorch.org/docs/stable/generated/torch.cuda.utilization.html#torch.cuda.utilization 

Tested on a GPU and this works correctly


justinxzhao (approved):
TY.

PR #3728, opened 2023-10-13
Add support for data parallel QLoRA training via DeepSpeed Zero stages 0, 1 and 2.

arnavgarg1:
This PR adds support for data parallel QLoRA training using DeepSpeed Stages 0, 1, and 2. 

As a refresher, here is what each DeepSpeed Zero stage corresponds to:
- **Stage 0**: Disabled, i.e., no partitioning of optimizer state, gradients or model parameters. You can still perform optimizer and parameter offloading, as well training using bf16 or fp16 etc.
- **Stage 1**: The optimizer states (e.g., for [Adam optimizer](https://arxiv.org/abs/1412.6980), 32-bit weights, and the first, and second moment estimates) are partitioned across the processes, so that each process updates only its partition.
- **Stage 2**: The reduced 32-bit gradients for updating the model weights are also partitioned such that each process retains only the gradients corresponding to its portion of the optimizer states.
- **Stage 3**: The 16-bit model parameters are partitioned across the processes. ZeRO-3 will automatically collect and partition them during the forward and backward passes.

With this PR, you can now perform QLoRA based training for larger-than-memory datasets. You can find a full attached example in **`ludwig/examples/llm_qlora_data_parallel`**.

For e.g, you can now use a config like the following:

```yaml
model_type: llm

input_features:
  - name: instruction
    type: text

output_features:
  - name: output
    type: text

adapter:
  type: lora

quantization:
  bits: 4

backend:
  type: ray
  trainer:
    use_gpu: true
    strategy:
      type: deepspeed
      zero_optimization:
        stage: 2

trainer:
  type: finetune
  batch_size: 1
  gradient_accumulation_steps: 4
```

to:

1. Create `n` single GPU workers on a single multi-GPU node
2. Load a 4-bit version of Llama-2-7b on each of the GPU training workers
3. Use Ray Datasets to split the total dataset for training across the `n` workers
4. Train using gradient partitioning and optimizer state partitioning across workers (not possible with traditional DDP)

In particular, since this uses DeepSpeed Stage 2 with Ray, it lets you stream datasets into memory at training time per-worker. 

This is a snapshot of GPU utilization a single-node 4 GPU pod using 4xA5000s:
```
Mon Oct 16 07:53:11 2023
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 510.73.05    Driver Version: 510.73.05    CUDA Version: 11.8     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA RTX A5000    On   | 00000000:01:00.0 Off |                  Off |
| 31%   61C    P2   141W / 230W |  10089MiB / 24564MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
|   1  NVIDIA RTX A5000    On   | 00000000:25:00.0 Off |                  Off |
| 30%   59C    P2   131W / 230W |   8509MiB / 24564MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
|   2  NVIDIA RTX A5000    On   | 00000000:41:00.0 Off |                  Off |
| 31%   60C    P2   133W / 230W |   8905MiB / 24564MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
|   3  NVIDIA RTX A5000    On   | 00000000:61:00.0 Off |                  Off |
| 30%   55C    P2   153W / 230W |   8625MiB / 24564MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
```

### DIFF
--- a/examples/llm_qlora_data_parallel/README.md
+++ b/examples/llm_qlora_data_parallel/README.md
@@ -2,6 +2,8 @@
 
 If you have a single-node multi-GPU setup with a large dataset that you would like to train using QLoRA, you can use DeepSpeed Stage 0, 1, or 2.
 
+## DeepSpeed Background
+
 As a refresher, here is what each DeepSpeed Zero stage corresponds to:
 
 - **Stage 0**: Disabled, i.e., no partitioning of optimizer state, gradients or model parameters. You can still perform optimizer and parameter offloading, as well training using bf16 or fp16 etc.
@@ -10,7 +12,16 @@ As a refresher, here is what each DeepSpeed Zero stage corresponds to:
 - **Stage 3**: The 16-bit model parameters are partitioned across the processes. ZeRO-3 will automatically collect and partition them during the forward and backward passes.
 
 _NOTE: Data Parallel QLoRA based training only works with DeepSpeed stages \<= 2. This is because DeepSpeed isn't
-compatible with partitioning/sharding of quantized weights as of DeepSpeed 0.10.3_.
+compatible with partitioning/sharding of quantized weights as of DeepSpeed 0.10.3 when weights are a mixture of dtypes_. See:
+
+- https://github.com/microsoft/DeepSpeed/issues/4295
+- https://github.com/microsoft/DeepSpeed/issues/3620
+
+In particular, this comment here summarizes it well:
+
+> Some code for ZeRO3 assumes that all parameters in a model has the same dtype. This model has uint8 and float32 parameters and it throws the error.
+
+## Example Config
 
 The example `train.py` uses DeepSpeed Stage 2 with the Ray backend as follows to fine-tune a model for natural language to code generation task via instruction fine-tuning.
 
@@ -25,4 +36,24 @@ backend:
         stage: 2
 ```
 
-If you want to use Stage 0 or 1, you can just replace `stage: 2` to the desired zero optimization stage.
+In most cases, stage 2 lets you train large models in distributed fashion across multiple GPUs. However, if you want to use Stage 0 or 1, you can just replace `stage: 2` to the desired zero optimization stage.
+
+## DeepSpeed Zero Stage Benefits
+
+### Benefits of DeepSpeed Stage 0
+
+- **Ease of Use**: Stage 0 is relatively easy to set up and use, making it a good starting point for users looking for memory-efficient training without the complexity of more advanced optimization techniques.
+- **Gradient Accumulation**: Stage 0 enables gradient accumulation, which is beneficial for simulating larger batch sizes even on hardware with memory constraints. This can lead to more stable model training and potentially faster convergence.
+- **Mixed Precision Training**: It supports mixed-precision training, which utilizes lower-precision data types (e.g., float16) to reduce memory usage while maintaining training stability.
+
+### Benefits of DeepSpeed Stage 1
+
+- **Optimizer State Partitioning**: Stage 1 is primarily focused on partitioning the optimizer state, allowing you to train very large models that wouldn't fit within a single GPU's memory.
+- **Memory Efficiency**: It efficiently manages memory by dividing the optimizer state into segments distributed across multiple GPUs. This makes training larger models feasible.
+- **Single-GPU Training**: Stage 1 is especially valuable when you need to train large models on a single GPU, making it an essential step before scaling up to more advanced stages for distributed training.
+- **Limited Configuration Complexity**: It introduces memory efficiency while maintaining a relatively simple configuration setup compared to the more advanced stages like Stage 2 and Stage 3.
+
+### Benefits of DeepSpeed Stage 2
+
+- **Training Extremely Large Models**: ZeRO Stage 2 partitions both the gradients and the optimizer state to reduce memory requirements significantly. By contrast, Stage 0 and Stage 1 do not have the same level of memory optimization to handle models of such magnitude.
+- **Advanced Distributed Training**: ZeRO Stage 2 is designed to handle distributed training at an unprecedented scale. It optimizes communication, gradient aggregation, and synchronization between GPUs and nodes, making it ideal for training large models efficiently in a distributed environment. This advanced distributed training capability is not present in Stage 0 and is more sophisticated than that of Stage 1, which helps in achieving faster training times and handling larger workloads.
--- a/examples/llm_qlora_data_parallel/README.md
+++ b/examples/llm_qlora_data_parallel/README.md
@@ -1,0 +1,28 @@
+# Data-Parallel QLoRA Fine-Tuning
+
+If you have a single-node multi-GPU setup with a large dataset that you would like to train using QLoRA, you can use DeepSpeed Stage 0, 1, or 2.
+
+As a refresher, here is what each DeepSpeed Zero stage corresponds to:
+
+- **Stage 0**: Disabled, i.e., no partitioning of optimizer state, gradients or model parameters. You can still perform optimizer and parameter offloading, as well training using bf16 or fp16 etc.
+- **Stage 1**: The optimizer states (e.g., for Adam optimizer, 32-bit weights, and the first, and second moment estimates) are partitioned across the processes, so that each process updates only its partition.
+- **Stage 2**: The reduced 32-bit gradients for updating the model weights are also partitioned such that each process retains only the gradients corresponding to its portion of the optimizer states.
+- **Stage 3**: The 16-bit model parameters are partitioned across the processes. ZeRO-3 will automatically collect and partition them during the forward and backward passes.
+
+_NOTE: Data Parallel QLoRA based training only works with DeepSpeed stages \<= 2. This is because DeepSpeed isn't
+compatible with partitioning/sharding of quantized weights as of DeepSpeed 0.10.3_.
+
+The example `train.py` uses DeepSpeed Stage 2 with the Ray backend as follows to fine-tune a model for natural language to code generation task via instruction fine-tuning.
+
+```yaml
+backend:
+  type: ray
+  trainer:
+    use_gpu: true
+    strategy:
+      type: deepspeed
+      zero_optimization:
+        stage: 2
+```
+
+If you want to use Stage 0 or 1, you can just replace `stage: 2` to the desired zero optimization stage.
--- a/examples/llm_qlora_data_parallel/train.py
+++ b/examples/llm_qlora_data_parallel/train.py
@@ -108,7 +108,6 @@ trainer:
 )
 
 model = LudwigModel(config=fine_tuning_config, logging_level=logging.INFO)
-results = model.train(dataset=df)
 
 (
     train_stats,  # dictionary containing training statistics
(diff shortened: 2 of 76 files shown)


ludwig/backend/ray.py
```diff
@@ -777,16 +821,29 @@ def batch_evaluation(
...
+            predictor_kwargs = {**self.predictor_kwargs, "collect_predictions": False}
+
+            # If the model is a fine-tuned LLM model that uses an adapter, we can just place the adapter
+            # weights in the Ray object store and load them back on each worker since the underlying base
+            # model's weights aren't modified.
+            adapter_ref = None
+            if self.model.trained_using_adapter:
+                adapter_ref = ray.put(dist_strategy.extract_adapter_weights_for_serialization(self.model))
+
+            optimization_stage = _get_optimization_stage_from_trainer_config(self.trainer_kwargs)
```
arnavgarg1, line 833: hmm I wonder if this could be moved in the dist_strategy base class

jeffkinnison (commented):
Overall looks good, just a few nits. One question that didn't quite fit anywhere: are there any additional validation checks we should add for DS config? Bounding DS stages, DS/qlora, etc.

ludwig/distributed/deepspeed.py
```diff
@@ -122,11 +128,24 @@ def prepare(
         return model_engine, optimizer
 
     def prepare_for_inference(self, model: nn.Module) -> nn.Module:
+        """Prepares the model for inference/evaluation.
+
+        For DeepSpeed Stage 3, we need to wrap the model in a DeepSpeed engine for inference. For all other DeepSpeed
+        stages, we load the base model back. For LLMs, this recreates either the base model or the PEFT model, depending
+        on whether a PEFT adapter was specified.
+        """
+        if self.zero_optimization_stage != 3:
```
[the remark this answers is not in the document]
arnavgarg1: @jeffkinnison I think this is a good call out. One more option is to refactor to have DeepSpeedStrategy be a base class and DeepSpeedStage3 be its own other class that overrides some of the methods or something to that effect. Will take a look!

ludwig/models/llm.py
```diff
@@ -72,6 +78,55 @@ def update(self, modules: Dict[str, torch.nn.Module]) -> None:
...
+        bool: True if the model should be loaded on the driver, False otherwise.
+    """
+    backend_type = _get_backend_type_from_config(config_obj)
+    deepspeed_optimization_strategy = _get_deepspeed_optimization_stage_from_config(config_obj)
+    if backend_type == "ray" and deepspeed_optimization_strategy is not None and deepspeed_optimization_strategy <= 2:
+        # If using deepspeed stage 0, 1 or 2, we only load the model into memory once we're actually inside
+        # of the training workers.
+        return False
+    # If using local backend or deepspeed stage 3, we load the model into memory upon class initialization.
+    return True
```
arnavgarg1, line 102: Going to push this into the backend class and maybe consider passing in the initialized `self.backend` object in the LudwigModel class into the the create_model function so it gets propagated to here, it shouldn't live here like this.

examples/llm_qlora_data_parallel/README.md
```diff
@@ -0,0 +1,28 @@
...
+If you have a single-node multi-GPU setup with a large dataset that you would like to train using QLoRA, you can use DeepSpeed Stage 0, 1, or 2.
+
+As a refresher, here is what each DeepSpeed Zero stage corresponds to:
+
+- **Stage 0**: Disabled, i.e., no partitioning of optimizer state, gradients or model parameters. You can still perform optimizer and parameter offloading, as well training using bf16 or fp16 etc.
+- **Stage 1**: The optimizer states (e.g., for Adam optimizer, 32-bit weights, and the first, and second moment estimates) are partitioned across the processes, so that each process updates only its partition.
+- **Stage 2**: The reduced 32-bit gradients for updating the model weights are also partitioned such that each process retains only the gradients corresponding to its portion of the optimizer states.
+- **Stage 3**: The 16-bit model parameters are partitioned across the processes. ZeRO-3 will automatically collect and partition them during the forward and backward passes.
+
+_NOTE: Data Parallel QLoRA based training only works with DeepSpeed stages \<= 2. This is because DeepSpeed isn't
```
justinxzhao, line 12: What is `\<=`?

examples/llm_qlora_data_parallel/README.md
```diff
@@ -0,0 +1,28 @@
...
+compatible with partitioning/sharding of quantized weights as of DeepSpeed 0.10.3_.
```
justinxzhao, line 13: Is there a link you can provide?

examples/llm_qlora_data_parallel/README.md
```diff
@@ -0,0 +1,28 @@
...
+  type: ray
+  trainer:
+    use_gpu: true
+    strategy:
+      type: deepspeed
+      zero_optimization:
+        stage: 2
+```
+
+If you want to use Stage 0 or 1, you can just replace `stage: 2` to the desired zero optimization stage.
```
justinxzhao, line 28: What are the benefits of using a lower stage of deepspeed?

examples/llm_qlora_data_parallel/train.py
```diff
@@ -0,0 +1,164 @@
...
+
+# Shuffle the array to ensure randomness
+np.random.shuffle(split_values)
+
+# Add the 'split' column to the DataFrame
+df["split"] = split_values
+df["split"] = df["split"].astype(int)
+
+# For now, just use 1000 rows as a demonstration
+df = df.head(n=1000)
```
justinxzhao, line 44: Instead of doing `.head()`, we can use @Infernaught's `preprocessing.sample_size` parameter.

ludwig/distributed/deepspeed.py
```diff
@@ -57,6 +59,7 @@ def __init__(
 
         super().__init__(**kwargs)
         self.zero_optimization = zero_optimization or DEFAULT_ZERO_OPTIMIZATION
+        self.zero_optimization_stage = self.zero_optimization.get("stage", 3)
```
justinxzhao, line 62: Why default to 3?

ludwig/models/llm.py
```diff
@@ -97,9 +152,16 @@ def load_pretrained_from_config(
             else:
                 logger.warning(f"Parameter {param_name} is not supported by {config_obj.base_model}. Skipping.")
 
-    logger.info("Loading large language model...")
     pretrained_model_name_or_path = weights_save_path or config_obj.base_model
-    model: PreTrainedModel = AutoModelForCausalLM.from_pretrained(pretrained_model_name_or_path, **load_kwargs)
+
+    # When performing data parallel training with DeepSpeed, we need to load the model weights on each worker
```
justinxzhao, line 157: Cool!

ludwig/models/llm.py
```diff
@@ -252,8 +317,25 @@ def initialize_adapter(self):
             self.model.print_trainable_parameters()
             logger.info("==================================================")
 
+    @property
+    def trained_using_adapter(self) -> bool:
+        return bool(self.config_obj.adapter)
```
justinxzhao, line 322: Is it possible for the config_obj.adapter to be non-None/non-empty when it's not set by the user explicitly?

ludwig/utils/backend_utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+    This function retrieves the backend type specified in a language model configuration.
+
+    Args:
+        config_obj (ModelConfig): The configuration object containing backend information.
+
+    Returns:
+        str: The backend type, defaulting to "local" if not explicitly specified.
+    """
+    # config_obj.backend maybe None if the backend was not explicitly set in the config
```
justinxzhao, line 33: nit: "may be"

ludwig/distributed/deepspeed.py
arnavgarg1: It's because in the worst case, it makes sense to assume that the model does not fit into memory and we need to do model parallel + data parallel training since the model will not fit into a single GPU. It also assumes no quantization based fine-tuning, which I feel like is a far assumption to make. All of this is to say that we want a config as simple as this "to just work"

```yaml
input_features: ...
output_features: ...
trainer:
    type: finetune
backend:
    type: ray
    trainer:
        strategy:
            type: deepspeed
```

to have the highest chance of succeeding irrespective of LLM model size. This is the worst case scenario that does full fine-tuning with an adapter or without quantization. All of those only reduce the model size, but this is the most useful configuration to default to assuming none of those are set in the config for LLM fine-tuning.

Let me know if this makes sense. I'll also add a comment in the DeepSpeed class explaining why we default to stage 3.

examples/llm_qlora_data_parallel/train.py
arnavgarg1: @Infernaught Can we document this in Ludwig docs as well?

examples/llm_qlora_data_parallel/README.md
arnavgarg1: I believe it's automatically added to escape the italics formatted, here's what it looks like when rendered:

Screenshot 2023-10-19 at 1 41 34 PM

ludwig/models/llm.py
arnavgarg1: I think by default it is set to `None` if not specified. In other parts of the code, we check for if self.config_obj.adapter, and then load the adapter weights, so this is effectively equivalent to that.

ludwig/backend/ray.py
```diff
@@ -663,14 +689,28 @@ def eval_fn(
...
+        #
+        # For LLMs:
+        # If we are using DeepSpeed Stage 3, this wraps the model in DeepSpeed Zero Inference.
+        # For DeepSpeed Stage 2 and below, this loads the base model weights into the model.
+        # If the model was not trained using an adapter, the fine-tuned weights are loaded into the model.
+        # If the model was trained using an adapter, the base model weights are loaded into the model along with the
+        # adapter weights used at initialization. We will updated the state dict to use the weights from the
+        # fine-tuned adapter.
         dist_model = distributed.prepare_for_inference(model)
+        if adapter_ref and (distributed_optimization_stage and distributed_optimization_stage <= 2):
```
tgaddair, line 709: This can be pushed into the DistributedStrategy, again, so we don't need to couple everything to DeepSpeed.

tgaddair (changes requested):
Not a full review yet, but something I see coming up repeatedly is references to the optimiation stage outside of the DeepSpeed strategy. This is a red flag that we should be moving this code into the DistributedStrategy interface rather than coupling these two different abstractions (backend and strategy) together.

PR #3730, opened 2023-10-13
[WIP] Create a binary decoder with a TART reasoning module

jeffkinnison:
This adds a TART reasoning module as a binary decoder.

TODOs:

[] Add the vanilla and LOO embedding protocols
[] Add an embeddings cache
[] Tests


ANarayan:
@jeffkinnison can we pass the tokenizer in to the embedding protocols?

jeffkinnison:
@ANarayan Let's assume that we can pass in the tokenizer and model into the protocol functions. That should simplify initial development, and we can revisit that design if needed. Also, feel free to update the function signatures or convert them into classes, or to make any other changes!

PR #3731, opened 2023-10-13
Limit ray versions to keep 2.4 out of the picture.

trevenrawr:
# Code Pull Requests

- Set a stricter limit on the `ray` version, so `2.4.0` doesn't cause anyone else any issues.


arnavgarg1 (approved):
@trevenrawr Thanks for flagging this!

jeffkinnison (approved):
Thank you!

PR #3732, opened 2023-10-15
[MAINTENACE] Improve typehints and use constants instead of common hardcoded string values.

alexsherstinsky:
### Scope
This small change provides two quality-of-life improvements:
* Enable more of the type hints to be used without the quotation marks.
* For common configuration strings, use defined constants (instead of hardcoded values).

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/distributed/base.py
+++ b/ludwig/distributed/base.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import contextlib
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, TYPE_CHECKING, Union
+from typing import Any, Callable, TYPE_CHECKING
 
 import torch
 from torch import nn
@@ -31,9 +33,9 @@ class DistributedStrategy(ABC):
     def prepare(
         self,
         model: nn.Module,
-        trainer_config: "ECDTrainerConfig",
+        trainer_config: ECDTrainerConfig,
         base_learning_rate: float,
-    ) -> Tuple[nn.Module, Optimizer]:
+    ) -> tuple[nn.Module, Optimizer]:
         """Modifies the model to support distributed training and creates the optimizer.
 
         Args:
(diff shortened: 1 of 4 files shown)


ludwig/distributed/base.py
```diff
@@ -49,7 +51,7 @@ def prepare(
     def prepare_for_inference(self, model: nn.Module) -> nn.Module:
         return model
 
-    def to_device(self, model: "BaseModel", device: Optional[torch.device] = None) -> nn.Module:
+    def to_device(self, model: BaseModel, device: torch.device | None = None) -> nn.Module:
```
justinxzhao, line 54: Out of curiosity, is there a style guide that recommends `... | None` rather than `Optional[...]`?

alexsherstinsky: @justinxzhao I cannot immediately point to a specific style guide (please let me know, and I will look for it).  Having said that, this notation is compatible with Python versions 3.10 and above.  So when we do `from __future__ import annotations`, the `isort` program (one of the pre-commit hooks) will automatically apply the "new Union" syntax (as per `https://www.python.org/dev/peps/pep-0604/`).

To me, the key benefit is that the `typing.Optional` type -- with its multiple interpretations -- is replaced with an explicit type declaration of a `Union` of several types, including `None` (without needing the `typing.Union` itself by providing the `|` shorthand).  This approach also supports arguments that have a default value that is not `None`.  Importantly, the presence of absence of a default value is handled consistently -- any argument may have a default.

If the default value is specified, then in both cases (with a `None` as part of the union, or without), specifying the argument is optional, but the use of the actual `typing.Optional` type -- with its multiple interpretations -- is avoided.

Please let me know if this justifies the change.  Thank you.

justinxzhao: Thank you for the extended explanation @alexsherstinsky! Your reasoning SGTM -- the `Optional` type hint has always been somewhat unsettling to me (multiple interpretations, as you call out).

justinxzhao (approved):

ludwig/distributed/base.py
alexsherstinsky: @justinxzhao My sincere pleasure!  Thank you for your comment.  By the way, after submitting my reasoning, I kept looking for other resources and found `https://adamj.eu/tech/2022/10/17/python-type-hints-old-and-new-syntaxes/`, which IMHO makes the case even better (and it is a pleasant fun read).  Thanks!

PR #3735, opened 2023-10-17
Simplify how we set pad token and pad token ID for huggingface tokenizers.

justinxzhao:
The missing pad token and missing pad token ID causes issues during fine-tuning. To mitigate, Ludwig currently sets the pad token to the eos token, as discussed in these threads (CC: @geoffreyangus)
- https://github.com/huggingface/transformers/issues/2630#issuecomment-1290809338
- https://github.com/huggingface/transformers/issues/2648#issuecomment-616177044

However, vanilla HF `generate()` using the EOS token in place of the pad token raises `RuntimeError: probability tensor contains either `inf`, `nan` or element < 0` when there are multiple examples in the batch. This PR implements a [separate recommendation](https://github.com/facebookresearch/llama/issues/380#issuecomment-1716832417) to set a separate '[PAD]' token.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -30,7 +30,7 @@ jobs:
             torchscript-version: 1.10.2
             ray-version: 2.2.0
           - python-version: "3.9"
-            pytorch-version: 2.0.0
+            pytorch-version: 2.1.0
             torchscript-version: 1.10.2
             ray-version: 2.3.0
           - python-version: "3.10"
@@ -208,6 +208,7 @@ jobs:
           - "integration_tests_c"
           - "integration_tests_d"
           - "integration_tests_e"
+          - "integration_tests_f"
 
     env:
       AWS_ACCESS_KEY_ID: ${{ secrets.LUDWIG_TESTS_AWS_ACCESS_KEY_ID }}
--- a/ludwig/data/batcher/test_batcher.py
+++ b/ludwig/data/batcher/test_batcher.py
@@ -1,0 +1,113 @@
+import logging
+
+import pandas as pd
+import yaml
+
+from ludwig.api import LudwigModel
+from ludwig.data.dataset.pandas import PandasDataset
+
+
+def test_pandas_size():
+    df = pd.DataFrame(
+        {"name": ["joe", "janice", "sara"], "mask": ["green", "black", "pink"], "weapon": ["stick", "gun", "gun"]}
+    )
+    config = yaml.safe_load(
+        """
+    model_type: llm
+    base_model: HuggingFaceH4/tiny-random-LlamaForCausalLM
+    input_features:
+    - name: name
+      type: text
+      preprocessing:
+        max_sequence_length: 256
+      column: name
+    output_features:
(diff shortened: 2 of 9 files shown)


arnavgarg1 (approved):

arnavgarg1:
Hey @justinxzhao, I did some digging over the weekend and I think there's actually an even better and probably correct way to do this that we haven't been doing so far. Here's the code I used to see if this works:

```
from transformers import AutoModelForCausalLM, AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("yujiepan/llama-2-tiny-random")
model = AutoModelForCausalLM.from_pretrained("yujiepan/llama-2-tiny-random")

assert tokenizer.pad_token is None
assert tokenizer.vocab_size  == 32000

tokenizer.add_special_tokens({"pad_token": "[PAD]"})
current_vocab_size = len(tokenizer.vocab.keys()) 
assert tokenizer.pad_token
assert tokenizer.pad_token_id == 32000
assert current_vocab_size == 32001

current_vocab_size_based_on_weights = model.get_output_embeddings().weight.shape[0]
if current_vocab_size_based_on_weights < current_vocab_size:
    model.resize_token_embeddings(current_vocab_size)

assert model.get_output_embeddings().weight.shape[0] == current_vocab_size
```

This will probably work and do the right thing, which is set a new pad_token_id based on the new special token for padding that gets created. But it might require a bit of a refactor on our end to coordinator tokenizer and model initialization.

justinxzhao:
> Hey @justinxzhao, I did some digging over the weekend and I think there's actually an even better and probably correct way to do this that we haven't been doing so far. Here's the code I used to see if this works:
> 
> ```python
> from transformers import AutoModelForCausalLM, AutoTokenizer
> 
> tokenizer = AutoTokenizer.from_pretrained("yujiepan/llama-2-tiny-random")
> model = AutoModelForCausalLM.from_pretrained("yujiepan/llama-2-tiny-random")
> 
> assert tokenizer.pad_token is None
> assert tokenizer.vocab_size  == 32000
> 
> tokenizer.add_special_tokens({"pad_token": "[PAD]"})
> current_vocab_size = len(tokenizer.vocab.keys()) 
> assert tokenizer.pad_token
> assert tokenizer.pad_token_id == 32000
> assert current_vocab_size == 32001
> 
> current_vocab_size_based_on_weights = model.get_output_embeddings().weight.shape[0]
> if current_vocab_size_based_on_weights < current_vocab_size:
>     model.resize_token_embeddings(current_vocab_size)
> 
> assert model.get_output_embeddings().weight.shape[0] == current_vocab_size
> ```
> 
> This will probably work and do the right thing, which is set a new pad_token_id based on the new special token for padding that gets created. But it might require a bit of a refactor on our end to coordinator tokenizer and model initialization.

Ack. I'm less sure this is necessary/desirable since the PAD token is used for mechanically evening out batches. It also seems like we don't need to register it formally in the vocabulary either since the model neither attends to nor generates these tokens.

justinxzhao:
Offline experiments show that this change seems to have no impact on model training/eval.

PR #3741, opened 2023-10-19
Uncap the torchmetrics version, so we can get fixes in 1.0+.

trevenrawr:
While trying to figure out why my air-gapped installation and deployment of Ludwig was failing to detect my cached copy of the `punkt` tokenizer for `nltk`, I came across https://github.com/Lightning-AI/torchmetrics/issues/1779 which got merged into `torchmetrics==1.0.0`... but Ludwig was avoiding that version.  So I forcibly installed `torchmetrics==1.2.0` (latest) in my image (after installing Ludwig), ignored `pip`'s words of caution, and managed to run some LoRA-based fine-tuning without a hitch!  Since that code path worked okay with the updated library, I'm submitting this PR to remove the pin and leaning on CI to help identify if there are issues on other code paths or if it's truly safe to release.

### DIFF
--- a/ludwig/modules/metric_modules.py
+++ b/ludwig/modules/metric_modules.py
@@ -136,7 +136,7 @@ class RMSEMetric(MeanSquaredError, LudwigMetric):
     """Root mean squared error metric."""
 
     def __init__(self, **kwargs):
-        super().__init__(squared=False, **kwargs)
+        super().__init__(squared=False)
 
 
 @register_metric(PRECISION, [BINARY], MAXIMIZE, PROBABILITIES)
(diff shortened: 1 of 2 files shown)


justinxzhao:
Thanks @trevenrawr! Let's see what the CI says -- I recall we had to pin originally to mitigate some test failures, though perhaps the test failures are reasonable to fix now.

arnavgarg1:
Good news is that there's only 1 test failure! Specifically, in the combinatorial tests: `test_number_encoder_loss[config3-dataset3]`. 

I ran this again since I thought it might be transient but seems like it is a persistent error that may require a tiny bit of inspection

justinxzhao:
Looks like my most recent commit fixed the broken test! I haven't seen anyone make use of RMSE kwargs, so I'm inclined to check this in as is.

Metrics unit tests look good.

PR #3742, opened 2023-10-19
Retry LLM model downloads from HF Hub with exponential backoff when there is a Read timeout

arnavgarg1:
Retries when this transient error occurs while downloading an LLM from HF Hub

```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='cdn-lfs.huggingface.co', port=443): Read timed out. 
```

Also adds a suite of new tests to ensure that the default_retry decorator that exists in Ludwig works as intended.


Infernaught (approved):
LGTM

justinxzhao (approved):

PR #3744, opened 2023-10-20
Add NEFTune implementation for Noised Embedding Instruction Fine-Tuning support

arnavgarg1:
This PR adds an implementation of the original [NEFTune paper](https://arxiv.org/pdf/2310.05914.pdf) to add noise to the embedding outputs at train time. Empirical evidence suggests that this improves model performance by up to 35% and reduces overfitting.

To use this in Ludwig with LLMs, just pass in `neftune_noise_alpha` to a non-zero value in the `model_parameters` section of your LLM model config.

```yaml
model_type: llm
model_parameters:
    neftune_noise_alpha: 5
```

---- 

## Implentation 

I implemented this is as a forward hook that gets added to the embedding layer once we're inside the trainer (well after the model is already instantiated) and removed at the end of training. The base class abstraction allows us to modify the behavior in custom ways for this hook as well as future hooks, including using a backforward or forward hook, or even not removing the hook at the end of training (if desired).


ludwig/modules/training_hooks.py
```diff
@@ -0,0 +1,103 @@
...
+    This class provides a template for implementing custom training hooks
+    that can be activated, deactivated, and maintain a handle to the hook.
+
+    Attributes:
+        _hook_handle (Optional[torch.utils.hooks.RemovableHandle]): A handle to the
+            registered forward hook, initially set to None.
+    """
+
+    def __init__(self, **kwargs) -> None:
+        self._hook_handle = None
```
tgaddair, line 21: Should use cooperative multiple inheritance here:

```
def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self._hook_handle = None
```

ludwig/modules/training_hooks.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @abstractmethod
+    def hook_fn(self, module: torch.nn.Module, inputs: torch.tensor, outputs: torch.tensor) -> torch.tensor:
```
tgaddair, line 24: Is it `torch.tensor` or `torch.Tensor`? We tend to use `torch.Tensor` throughout the code, maybe the other is an alias?

ludwig/modules/training_hooks.py
```diff
@@ -0,0 +1,103 @@
...
+    def deactivate_hook(self):
+        """Deactivates and removes the training hook."""
+        if self._hook_handle is not None:
+            self._hook_handle.remove()
+            self._hook_handle = None
+
+
+class NEFTuneHook(TrainingHook):
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
```
tgaddair, line 62: Same thing here with forwarding args in addition to kwargs

tgaddair (approved):
Looks great! Just a couple nits.

ludwig/modules/training_hooks.py
arnavgarg1: They're aliases of each other to my knowledge, but will update to torch.Tensor for consistency

PR #3750, opened 2023-10-21
Update base.py

jetaudio:
Missing config and model variable in line 321

# Code Pull Requests

Please provide the following:

- fix error missing config and model positional variable in line 321 file base.py


tgaddair:
Thanks for the PR @jetaudio, looks like this was an regression introduced in #3709, which removed the config and model from kwargs.

tgaddair (approved):

geoffreyangus (approved):

PR #3752, opened 2023-10-23
[MAINTENANCE] Restrict "fsspec" to not include the latest version "2023.10.0"

alexsherstinsky:
### Scope
The version `2023.10.0` or the `fsspec` library was released on 10/21/2023.  Unfortunately, it breaks the `builder.py::as_dataset()` method of the HuggingFace "datasets" library version in use with the error `NotImplementedError: Loading a dataset cached in a LocalFileSystem is not supported.` being raised.  Hence, we restrict the `fsspec` version for the time being.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ PyYAML>=3.12,<6.0.1,!=5.4.* #Exlude PyYAML 5.4.* due to incompatibility with aws
 absl-py
 kaggle
 requests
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):

justinxzhao:
Thanks @alexsherstinsky!

PR #3757, opened 2023-10-26
Pin getdaft<0.2 in requirements

jaychia:
# Code Pull Requests

Adds an upper-bound to the version of `getdaft`, since Daft introduces a backward-incompatible change in v0.2.0 that deprecates passing in fsspec filesystems into its APIs.

To pass credentials, Daft now includes a `IOConfig` API that can be used.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -42,7 +42,8 @@ sacremoses
 sentencepiece
 
 # requirements for daft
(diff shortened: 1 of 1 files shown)


arnavgarg1:
Thanks @jaychia!

PR #3763, opened 2023-11-03
enh: Add on_checkpoint callback that triggers after every checkpoint even if the model was not evaluated

jeffreyftang:


ludwig/trainers/trainer.py
```diff
@@ -1200,6 +1202,9 @@ def _train_loop(
                     if self.is_coordinator():
                         progress_tracker.save(os.path.join(save_path, TRAINING_PROGRESS_TRACKER_FILE_NAME))
 
+                # Callback that the checkpoint was reached, regardless of whether the model was evaluated or saved.
+                self.callback(lambda c: c.on_checkpoint(self, progress_tracker))
```
tgaddair, line 1206: This should be in the scope of the `if not self.skip_save_progress`, otherwise it seems misleading (since there was no checkpoint written out).

ludwig/trainers/trainer.py
```diff
@@ -1200,6 +1202,9 @@ def _train_loop(
                     if self.is_coordinator():
                         progress_tracker.save(os.path.join(save_path, TRAINING_PROGRESS_TRACKER_FILE_NAME))
```
justinxzhao, line 1203: There's another place where the checkpoint manager saves the model, outside/after 

```
should_break = self._train_loop
```

Should we add this callback there as well?

Also, I'm noticing that this code may be duplicated in both places -- would you be able to refactor into a shared library?

ludwig/trainers/trainer.py
justinxzhao: +1.

justinxzhao (approved):

PR #3764, opened 2023-11-03
[MAINTENANCE] Update contributor instructions with additional steps that may be needed for installation to succeed.

alexsherstinsky:
### Scope
In the present author's experience, Ludwig installation has not gone smoothly (under `Anaconda` with Python 3.10.13) without additional steps and omissions.  This change recommends these steps and work-around procedures.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,7 +79,9 @@ pip install -e .
 1. Finally, if during the installation of `horovod`, the exception `ModuleNotFoundError: No module named 'packaging'`
    occurs, you may need to disable `horovod` (that means you will need train with another backend, just not `horovod`).
    To do that, edit the file `requirements_extra.txt` and comment out the line that begins with `horovod`.  After that,
-   execute the `pip install` command given in the previous step; you should be able to complete the installation fully.
+   please execute the long `pip install` command given in the previous step.  With these work-around provisions, your
+   installation should run to completion successfully.  If you are still having difficulty, please reach out with the
+   specifics of your environment in [Slack](https://join.slack.com/t/ludwig-ai/shared_invite/zt-mrxo87w6-DlX5~73T2B4v_g6jj0pJcQ).
 
 Make sure to not forget to install and run `pre-commit`:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,8 @@ pip install -e .
    To do that, edit the file `requirements_extra.txt` and comment out the line that begins with `horovod`.  After that,
    please execute the long `pip install` command given in the previous step.  With these work-around provisions, your
    installation should run to completion successfully.  If you are still having difficulty, please reach out with the
-   specifics of your environment in [Slack](https://join.slack.com/t/ludwig-ai/shared_invite/zt-mrxo87w6-DlX5~73T2B4v_g6jj0pJcQ).
+   specifics of your environment in [Slack](https://join.slack.com/t/ludwig-ai/shared_invite/zt-mrxo87w6-DlX5~73T2B4v_g6jj0pJcQ)
+   and someone from the Ludwig Community will assist you.
 
 Make sure to not forget to install and run `pre-commit`:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,7 @@ Work on your self-assigned issue and eventually create a Pull Request.
    brew install cmake libomp
    ```
 
-which require `homebrew` to already be installed on your development machine.  Please install these libraries and then repeat the previous step:
+which require `homebrew` to be installed on your development machine.  Please install these libraries and then repeat the previous step:
 
 ```bash
 pip install -e .
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,14 +51,14 @@ Work on your self-assigned issue and eventually create a Pull Request.
    pre-commit install
    ```
 
-1. If this installation approach runs into errors, you may need to install the following prerequisites:
+1. On MacOS with Apple Silicon, if this installation approach runs into errors, you may need to install the following prerequisites:
 
    ```bash
    brew install cmake
    brew install cmake libomp
    ```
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -51,6 +51,44 @@ Work on your self-assigned issue and eventually create a Pull Request.
    pre-commit install
    ```
 
+1. On MacOS with Apple Silicon, if this installation approach runs into errors, you may need to install the following prerequisites:
```
w4nderlust, line 54: all the lines have this 1. preceding them, not sure what the goal is, in theory in markdown that's an ordered list i beleive?

CONTRIBUTING.md
```diff
@@ -51,6 +51,44 @@ Work on your self-assigned issue and eventually create a Pull Request.
...
+
+   ```bash
+   brew install cmake libomp
+   ```
+
+which require `homebrew` to be installed on your development machine.  Please install these libraries and then repeat the previous step:
+
+```bash
+pip install -e .
```
w4nderlust, line 63: this should probably be .[test]

CONTRIBUTING.md
```diff
@@ -51,6 +51,44 @@ Work on your self-assigned issue and eventually create a Pull Request.
...
+1. If the installation still does not complete successfully, then it is possible that the order of the
+   "requirements" files matters.  To carry out the installation in the proper sequence, please execute:
+
+   ```bash
+   pip install --use-pep517 --requirement requirements.txt --requirement requirements_benchmarking.txt \
+      --requirement requirements_distributed.txt --requirement requirements_explain.txt \
+      --requirement requirements_hyperopt.txt --requirement requirements_llm.txt \
+      --requirement requirements_serve.txt --requirement requirements_test.txt \
+      --requirement requirements_tree.txt --requirement requirements_viz.txt \
+      --requirement requirements_extra.txt --editable .
```
w4nderlust, line 75: not sure, but this should be equivalent to  pip install --use-pep517 -e .[test] (a parte fro mthe extras)
not sure what the pep directive changes, but i would verify before suggesting to manually change the requirement installation.

Also, in general, I suggest to checkout the setup.py script as it makes clear both the installation sub-packages and the order of the installation. if there are discoveries that come from reading it, I'd document them here too

CONTRIBUTING.md
alexsherstinsky: @w4nderlust -- indeed I made a typo -- will fix as part of the next commit.  Thank you.

CONTRIBUTING.md
alexsherstinsky: @w4nderlust Yes -- what was happening is that this command would not work in Z-shell; but now I figured it out -- the "[test]" has to be in quotes (after several examples provided by @tgaddair I finally realized that this is what was happening).  Hence, in the next commit, I will tidy it all up and use these commands:
```
pip install -e ."[full]"
pip install -e ."[test]"
```

Thanks!

CONTRIBUTING.md
alexsherstinsky: @w4nderlust Yes -- per my previous reply, I have finally realized what was going on -- you are correct (and I will fix the relevant parts, if any, in the next commit).  Thanks.

CONTRIBUTING.md
alexsherstinsky: @w4nderlust Update -- I checked; not a typo -- continuing the ordered was exactly my intention.  I will check to make sure that things render correctly.  Thanks.

alexsherstinsky:
@w4nderlust After reviewing the different contributor pages, I realized that what I was going for was the right goal -- making the contributor guide clear (in terms of installation); however, my approach was ineffective and incorrect.  Here is the situation highlighting the problem:
* We have the page "https://ludwig.ai/latest/developer_guide/contributing/", which is incomplete in terms of the installation instructions; one following it would only install packages from "requirements.txt" -- the instructions for "installing [full]" are not included.  Given that this is hosted on "ludwig.ai", how do we change it to make it consistent with "https://github.com/ludwig-ai/ludwig/blob/master/CONTRIBUTING.md" (which is the focus of this PR)?  Or is the former sourcing the content from the latter automatically?
* For its part, currently the content on "https://github.com/ludwig-ai/ludwig/blob/master/CONTRIBUTING.md" has slightly different wording, but suffers from the same deficiencies as the content on "https://ludwig.ai/latest/developer_guide/contributing/" -- the lack of instructions for full, developer-ready installations.
* The content that reveals to the developer/contributor that hints about the different `extras_require` that they would be interested in appears on "https://ludwig.ai/latest/getting_started/installation/" (which I cannot change, since it is hosted on "ludwig.ai").

Hence, for the immediate term:
* I changed my suggested modification to the "https://github.com/ludwig-ai/ludwig/blob/master/CONTRIBUTING.md" page in this PR that correctly reflects the availability of the `extras_require` dependencies in `setup.py`, just like you pointed out.
* I created a task in "Linear" ("https://linear.app/predibase/issue/MLX-1516/make-contributor-guide-content-consistent-across-ludwig-sites") to make "https://ludwig.ai/latest/developer_guide/contributing/" consistent (or, perhaps, identical) with this content.

Please re-review.
Thank you.

justinxzhao:
> * I created a task in "Linear" to make "https://ludwig.ai/latest/developer_guide/contributing/" consistent (or, perhaps, identical) with this content.

Hi @alexsherstinsky,

FYI, the contributing page in this repo and the [corresponding page](https://github.com/ludwig-ai/ludwig-docs/blob/master/docs/developer_guide/contributing.md) in ludwig docs is manually kept in sync. If there's an easy way to automate them being in sync (maybe ludwig docs loads it from [raw](https://raw.githubusercontent.com/ludwig-ai/ludwig-docs/master/docs/developer_guide/contributing.md)?), that would be great.

justinxzhao (approved):

alexsherstinsky:
Thank you, @justinxzhao !  I updated the corresponding task in Linear (to explore the idea of loading from "raw").

PR #3766, opened 2023-11-03
[BUGFIX] Pin Transformers version to prevent tests from failing in PEFT

alexsherstinsky:
### Scope
* The Transformers version `4.35.0` was released on 11/2/2023; causing an "IndexError: tuple index out of range" in "peft/tuners/adaption_prompt/utils.py:45" — which originates in executing `model_outputs = self.model(input_ids=self.model_inputs, attention_mask=self.attention_masks).get(LOGITS)` in "ludwig/models/llm.py". 
 Pinning Transformers to be less than that version fixes the problem.
* In addition, since PEFT `0.6.0` was also released (half a day ago), there is no need to use a specific commit, because this latest version incorporates the capabilities needed for LLM fine-tuning, which were added in that pre-release commit.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,7 @@ torchaudio
 torchtext
 torchvision
 pydantic<2.0
-#transformers>=4.33.2
-transformers>=4.33.2,<4.35.0
+transformers>=4.33.2,<4.35.0 # pinning since version 4.35.0 on 11/2/2023 causes IndexError in PEFT
 tokenizers>=0.13.3
 spacy>=2.3
 PyYAML>=3.12,<6.0.1,!=5.4.* #Exlude PyYAML 5.4.* due to incompatibility with awscli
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -11,7 +11,7 @@ torchaudio
 torchtext
 torchvision
 pydantic<2.0
-transformers>=4.33.2
+transformers>=4.33.2,<4.35.0 # pinning since version 4.35.0 on 11/2/2023 causes IndexError in PEFT
```
tgaddair, line 14: Is there an issue on transformers to track, so we can know when to unpin?

alexsherstinsky: @tgaddair No -- good catch; I created it: `https://github.com/huggingface/transformers/issues/27278`.  Thank you.

tgaddair (approved):

PR #3768, opened 2023-11-06
Add k eval predictions to mlflow for LLM fine-tuning

Infernaught:
Adds LLM predictions to MLFlow during evaluation. Local behavior shows inputs, outputs, and prompts added to a csv file in the MLFlow logs.

### DIFF
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -226,7 +226,7 @@ def _log_mlflow_loop(q: queue.Queue, log_artifacts: bool = True):
             break
 
         if log_metrics is not None:
-            if log_metrics["llm_eval_examples"] is not None:
+            if "llm_eval_examples" in log_metrics and log_metrics["llm_eval_examples"] is not None:
                 mlflow.log_table(log_metrics["llm_eval_examples"], artifact_file="llm_eval_examples.json")
                 # Delete the table from the metrics dict so we don't try to log it with the other metrics
                 del log_metrics["llm_eval_examples"]
@@ -247,7 +247,7 @@ def _log_mlflow(log_metrics, steps, save_path, should_continue, log_artifacts: b
     This is used when save_in_background is False.
     """
     if log_metrics is not None:
-        if log_metrics["llm_eval_examples"] is not None:
+        if "llm_eval_examples" in log_metrics and log_metrics["llm_eval_examples"] is not None:
             mlflow.log_table(log_metrics["llm_eval_examples"], artifact_file="llm_eval_examples.json")
             # Delete the table from the metrics dict so we don't try to log it with the other metrics
             del log_metrics["llm_eval_examples"]
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -228,6 +228,7 @@ def _log_mlflow_loop(q: queue.Queue, log_artifacts: bool = True):
         if log_metrics is not None:
             if log_metrics["llm_eval_examples"] is not None:
                 mlflow.log_table(log_metrics["llm_eval_examples"], artifact_file="llm_eval_examples.json")
+                # Delete the table from the metrics dict so we don't try to log it with the other metrics
                 del log_metrics["llm_eval_examples"]
         mlflow.log_metrics(log_metrics, step=steps)
 
@@ -248,6 +249,7 @@ def _log_mlflow(log_metrics, steps, save_path, should_continue, log_artifacts: b
     if log_metrics is not None:
         if log_metrics["llm_eval_examples"] is not None:
             mlflow.log_table(log_metrics["llm_eval_examples"], artifact_file="llm_eval_examples.json")
+            # Delete the table from the metrics dict so we don't try to log it with the other metrics
             del log_metrics["llm_eval_examples"]
         mlflow.log_metrics(log_metrics, step=steps)
         if log_artifacts:
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -242,6 +242,11 @@ def _log_mlflow(log_metrics, steps, save_path, should_continue, log_artifacts: b
     This is used when save_in_background is False.
     """
     if log_metrics is not None:
+        if log_metrics["llm_outputs"] is not None:
+            mlflow.llm.log_predictions(
+                inputs=log_metrics["llm_outputs"]["inputs"], outputs=log_metrics["llm_outputs"]["outputs"]
+            )
+            del log_metrics["llm_outputs"]
         mlflow.log_metrics(log_metrics, step=steps)
         if log_artifacts:
             _log_model(save_path)
--- a/ludwig/contribs/mlflow/__init__.py
+++ b/ludwig/contribs/mlflow/__init__.py
@@ -225,6 +225,14 @@ def _log_mlflow_loop(q: queue.Queue, log_artifacts: bool = True):
             # Break out of the loop if we're not going to log anything.
             break
 
+        if log_metrics is not None:
+            if log_metrics["llm_outputs"] is not None:
+                mlflow.llm.log_predictions(
+                    inputs=log_metrics["llm_outputs"]["inputs"],
+                    prompts=[0] * len(log_metrics["llm_outputs"]["inputs"]),  # This is a workaround
+                    outputs=log_metrics["llm_outputs"]["outputs"],
+                )
+                del log_metrics["llm_outputs"]
         mlflow.log_metrics(log_metrics, step=steps)
 
         if not q.empty():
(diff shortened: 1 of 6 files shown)


ludwig/trainers/trainer.py
```diff
@@ -659,6 +660,34 @@ def run_evaluation(
...
+                o_feat.feature_name: torch.from_numpy(
+                    np.array(training_set.dataset[o_feat.proc_column][:num_eval_examples], copy=True)
+                ).to(self.device)
+                for o_feat in self.dist_model.output_features.values()
+            }
+            eval_example_outputs = self.dist_model((inputs, targets))["predictions"]
+            llm_outputs = {"inputs": [], "outputs": []}
+            input_list = list(inputs.values())[0]
+            for i in range(len(eval_example_outputs)):
+                clean_input = remove_left_padding(input_list[i], self.dist_model.tokenizer)[0]
```
justinxzhao, line 682: Instead of surfacing lower-level code and the raw forward pass, could we re-use `self.evaluation`, which could be modified to return predictions:

```python
    def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tracker):
        predictor = Predictor(
            self.dist_model,
            batch_size=batch_size,
            distributed=self.distributed,
            report_tqdm_to_ray=self.report_tqdm_to_ray,
            model=self.model,
        )
        ### Setting collect_predictions=True will return predictions.
        metrics, _ = predictor.batch_evaluation(dataset, collect_predictions=False, dataset_name=dataset_name)

        return append_metrics(self.model, dataset_name, metrics, metrics_log, progress_tracker)
```

ludwig/trainers/trainer_llm.py
Infernaught, line 1: This is still not a super lightweight solution, but it turns out that setting `collect_predictions=True` causes the code to break when doing batch evaluation because we cannot concatenate the outputs when they are different shapes.

justinxzhao: Yikes! That's good to know -- could you leave a comment in `self.evaluation()` to mention this?

ludwig/trainers/trainer_llm.py
```diff
@@ -434,8 +435,28 @@ def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tr
...
+        llm_eval_examples = {"inputs": [], "targets": [], "outputs": []}
+        for key in inp_tar_out_dict["inputs"]:
+            for inp in inp_tar_out_dict["inputs"][key]:
+                inp = remove_left_padding(inp, tokenizer)[0]
+                llm_eval_examples["inputs"].append(tokenizer.decode(inp))
+
+        for key in inp_tar_out_dict["targets"]:
+            for tar in inp_tar_out_dict["targets"][key]:
+                tar = remove_left_padding(tar, tokenizer)[0]
+                llm_eval_examples["targets"].append(tokenizer.decode(tar))
```
justinxzhao, line 453: Could we do `tokenizer.decode(tar, skip_special_symbols=True)` to avoid manually removing left padding?

ludwig/contribs/mlflow/__init__.py
```diff
@@ -241,7 +249,15 @@ def _log_mlflow(log_metrics, steps, save_path, should_continue, log_artifacts: b
 
     This is used when save_in_background is False.
     """
+    # breakpoint()
     if log_metrics is not None:
+        if log_metrics["llm_eval_examples"] is not None:
+            mlflow.llm.log_predictions(
+                inputs=log_metrics["llm_eval_examples"]["inputs"],
+                prompts=[0] * len(log_metrics["llm_eval_examples"]["inputs"]),
```
justinxzhao, line 257: Based on the example in https://mlflow.org/docs/latest/python_api/mlflow.llm.html, perhaps we should make `inputs=[0] * len(log_metrics["llm_eval_examples"]["inputs"])`  and `prompts=log_metrics["llm_eval_examples"]["inputs"]`. WDYT?

ludwig/contribs/mlflow/__init__.py
```diff
@@ -241,7 +249,15 @@ def _log_mlflow(log_metrics, steps, save_path, should_continue, log_artifacts: b
...
+                outputs=log_metrics["llm_eval_examples"]["outputs"],
+            )
+            del log_metrics["llm_eval_examples"]
```
justinxzhao, line 260: Why is this necessary?

ludwig/contribs/mlflow/__init__.py
```diff
@@ -225,6 +225,14 @@ def _log_mlflow_loop(q: queue.Queue, log_artifacts: bool = True):
...
             break
 
+        if log_metrics is not None:
+            if log_metrics["llm_eval_examples"] is not None:
+                mlflow.llm.log_predictions(
+                    inputs=log_metrics["llm_eval_examples"]["inputs"],
+                    prompts=log_metrics["llm_eval_examples"]["targets"],
+                    outputs=log_metrics["llm_eval_examples"]["outputs"],
+                )
+                del log_metrics["llm_eval_examples"]
```
justinxzhao, line 235: Why is this necessary?

ludwig/contribs/mlflow/__init__.py
```diff
@@ -241,7 +249,15 @@ def _log_mlflow(log_metrics, steps, save_path, should_continue, log_artifacts: b
 
     This is used when save_in_background is False.
     """
+    # breakpoint()
```
justinxzhao, line 252: nit: Remove breakpoint.

ludwig/models/predictor.py
```diff
@@ -388,6 +390,13 @@ def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=Fa
                     outputs = self._predict_on_inputs((inputs, targets))
                     preds = self.model.outputs_to_predictions(outputs)
 
+                    for key in inputs:
+                        example_inputs[key].extend(inputs[key])
+                    for key in targets:
+                        example_targets[key].extend(targets[key])
+                    for key in preds:
+                        example_outputs[key].extend(preds[key]["predictions"])
```
justinxzhao, line 398: Is there a `response` key from the predictions dictionary? I remember adding it for computing decoded-text metrics.

ludwig/contribs/mlflow/__init__.py
Infernaught: Oh wait this is an error. Prompts should actually contain the targets (as per discussion with @geoffreyangus). Let me fix this.

ludwig/contribs/mlflow/__init__.py
Infernaught: Are you referring to `del log_metrics["llm_eval_examples"]`? This is so that mlflow doesn't try to log the llm_eval_examples in `mlflow.log_metrics`, which would cause an error.

ludwig/models/predictor.py
Infernaught: In `preds[key]`, I only see "predictions", "probabilities", and "logits"

ludwig/models/predictor.py
```diff
@@ -351,7 +351,9 @@ def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=Fa
         """
```
justinxzhao, line 351: Update the docstring for `batch_evaluation()`.

ludwig/contribs/mlflow/__init__.py
justinxzhao: Makes sense. Could you add a comment about this in both places?

ludwig/utils/trainer_utils.py
```diff
@@ -179,6 +180,7 @@ def __init__(
         self.train_metrics = train_metrics
         self.validation_metrics = validation_metrics
         self.test_metrics = test_metrics
+        self.llm_eval_examples = llm_eval_examples
```
justinxzhao, line 183: Can you add a comment here with an example of the expected structure and how this is only set for LLM fine-tuning?

> llm_eval_examples is used to capture samples of prompts/outputs for LLM fine-tuning. There are 3 keys (inputs, targets, outputs), which each map to a list of strings.

ludwig/utils/trainer_utils.py
```diff
@@ -210,6 +212,7 @@ def log_metrics(self):
             "best_valid_metric": self.best_eval_metric_value,
             "num_reductions_lr": self.num_reductions_learning_rate,
             "num_increases_bs": self.num_increases_batch_size,
+            "llm_eval_examples": self.llm_eval_examples,
```
justinxzhao, line 215: Maybe add a comment here too about how the structure of llm_eval_examples is different from all other entries and certain callbacks.

> Note that `llm_eval_examples` is the only entry in the map with string leaves. Loggers and other callbacks should handle this accordingly.

justinxzhao, line 215: nit: How about:

```python
if self.llm_eval_examples:
    log_metrics["llm_eval_examples"] = self.llm_eval_examples
```

justinxzhao (commented):
Thanks for the changes @Infernaught! LGTM after adding a few additional comments.

justinxzhao:
If you rebase, the CI should run clean thanks to @alexsherstinsky's recent changes.

ludwig/trainers/trainer.py
```diff
@@ -638,6 +638,7 @@ def run_evaluation(
         all_losses: Dict[str, torch.Tensor],
         early_stopping_steps: int,
         checkpoint_manager: CheckpointManager,
+        num_eval_examples: int = 5,
```
justinxzhao, line 641: Is this being used anywhere anymore?

justinxzhao (approved):

ludwig/models/predictor.py
```diff
@@ -416,9 +427,10 @@ def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=Fa
             metrics = self.model.get_metrics()
             self.model.reset_metrics()
 
-            self.dist_model.train(prev_model_training_mode)  # Restores previous model training mode.
+            inp_tar_out_dict = {"inputs": example_inputs, "targets": example_targets, "outputs": example_outputs}
```
arnavgarg1, line 430: nit, can we rename this to `input_target_output_dict` for clarity?

PR #3773, opened 2023-11-08
Add eval_steps as a flag to help speed up eval

Infernaught:
`eval_steps` effectively samples from the validation and test sets, taking the first few samples of each, depending on the `eval_steps` set in the config and the ratio of validation set size to test set size.

### DIFF
--- a/ludwig/schema/preprocessing.py
+++ b/ludwig/schema/preprocessing.py
@@ -26,6 +26,13 @@ class PreprocessingConfig(schema_utils.BaseMarshmallowConfig):
         parameter_metadata=PREPROCESSING_METADATA["sample_size"],
     )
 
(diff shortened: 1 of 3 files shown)


ludwig/schema/preprocessing.py
```diff
@@ -26,6 +26,13 @@ class PreprocessingConfig(schema_utils.BaseMarshmallowConfig):
         parameter_metadata=PREPROCESSING_METADATA["sample_size"],
     )
 
+    eval_steps: float = schema_utils.NonNegativeInteger(
```
justinxzhao, line 29: I don't think this preprocessing implementation is the right approach.

`eval_steps` should be a trainer parameter and used within `run_evaluation()/batch_evaluate()`.

PR #3774, opened 2023-11-09
Cleanup: Use existing HFTokenizer to consolidate manual pad token setting.

justinxzhao:


arnavgarg1 (approved):
Thanks for the cleanup PR! One thing we can consider in the future is also make the HFtokenizer class a singleton so we don't keep re-instantiating the same tokenizer for every base model!

justinxzhao:
> Thanks for the cleanup PR! One thing we can consider in the future is also make the HFtokenizer class a singleton so we don't keep re-instantiating the same tokenizer for every base model!

Hmm, I'm not sure if a singleton would bring us a lot: 
1) The tokenizer memory footprint seems quite small, and HF reloads from disk/cache on subsequent reloads.
2) Where are we being wasteful with tokenizer instantiation?

justinxzhao:
Thanks for the review!

arnavgarg1:
> > Thanks for the cleanup PR! One thing we can consider in the future is also make the HFtokenizer class a singleton so we don't keep re-instantiating the same tokenizer for every base model!
> 
> Hmm, I'm not sure if a singleton would bring us a lot:
> 
> 1. The tokenizer memory footprint seems quite small, and HF reloads from disk/cache on subsequent reloads.
> 2. Where are we being wasteful with tokenizer instantiation?

It's mostly from the perspective of really nasty/messy logs every time we re-instantiate the tokenizer since it tries to fork and run them in parallel or something to that effect - you can see what I mean if you kick-off any of the fine-tuning tests that tries to train a model.

PR #3776, opened 2023-11-09
Add eval_steps as a flag

Infernaught:
Add eval_steps under trainer this time instead of preprocessing

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -237,6 +237,7 @@ class Predictor(BasePredictor):
                 eval_steps = (
                     self.dist_model.config_obj.trainer.eval_steps
                     if hasattr(self.dist_model.config_obj.trainer, "eval_steps")
+                    and hasattr(self.dist_model, "config_obj")
                     else None
                 )
                 eval_steps_counter = 0
@@ -378,6 +379,7 @@ class LlmFineTunePredictor(Predictor):
                 eval_steps = (
                     self.dist_model.config_obj.trainer.eval_steps
                     if hasattr(self.dist_model.config_obj.trainer, "eval_steps")
+                    and hasattr(self.dist_model, "config_obj")
                     else None
                 )
                 eval_steps_counter = 0
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -236,8 +236,8 @@ class Predictor(BasePredictor):
                 predictions = defaultdict(list)
                 eval_steps = (
                     self.dist_model.config_obj.trainer.eval_steps
-                    if hasattr(self.dist_model.config_obj.trainer, "eval_steps")
-                    and hasattr(self.dist_model, "config_obj")
+                    if hasattr(self.dist_model, "config_obj")
+                    and hasattr(self.dist_model.config_obj.trainer, "eval_steps")
                     else None
                 )
                 eval_steps_counter = 0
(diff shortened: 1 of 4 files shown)


ludwig/models/predictor.py
```diff
@@ -234,7 +234,12 @@ def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=Fa
                 progress_bar = LudwigProgressBar(self.report_tqdm_to_ray, progress_bar_config, self.is_coordinator())
 
                 predictions = defaultdict(list)
+                eval_steps = self.dist_model.config_obj.trainer.eval_steps
```
justinxzhao, line 237: Maybe to accommodate GBMs, do:

```python
eval_steps = self.dist_model.config_obj.trainer.eval_steps if hasattr(self.dist_model.config_obj.trainer, "eval_steps") else None
```

ludwig/schema/trainer.py
```diff
@@ -207,6 +207,13 @@ def __post_init__(self):
         parameter_metadata=TRAINER_METADATA[MODEL_ECD]["train_steps"],
     )
 
+    eval_steps: float = schema_utils.NonNegativeInteger(
```
justinxzhao, line 210: Another option is to only support `eval_steps` for the LLMTrainer config. We're more concerned about slow synchronous evals for fine-tuning LLMs anyway. WDYT? CC: @geoffreyangus

Infernaught: That's kinda what I was thinking, too

ludwig/schema/metadata/configs/trainer.yaml
```diff
@@ -382,6 +382,22 @@ ecd:
             through the dataset to converge while others converge before a single epoch
             through the data.
         ui_display_name: Train Steps
+    eval_steps:
+        default_value_reasoning:
+            The default value is None because we do not want to lower the number of evaluation steps
+            by default, and we do not know the size of an arbitrary dataset.
+            By setting the default to None, we simply evaluate on the full evaluation set.
+        description_implications:
+            The smaller this value of this parameter, the less time evaluation will take.
```
geoffreyangus, line 391: nit: would also mention that smaller values make the evaluation metrics less precise.

ludwig/models/predictor.py
```diff
@@ -366,7 +376,17 @@ def batch_evaluation(self, dataset, collect_predictions=False, collect_logits=Fa
...
+                eval_steps = (
+                    self.dist_model.config_obj.trainer.eval_steps
+                    if hasattr(self.dist_model, "config_obj")
+                    and hasattr(self.dist_model.config_obj.trainer, "eval_steps")
+                    else None
+                )
+                eval_steps_counter = 0
                 while not batcher.last_batch():
+                    if eval_steps and eval_steps_counter >= eval_steps:
+                        logger.info(f"Reached evaluation step {eval_steps}. Ending evaluation.")
```
geoffreyangus, line 388: @Infernaught can you write a test that confirms the first `n` samples of the evaluation are what are evaluated against per epoch?

geoffreyangus (commented):
Nice job so far! The key thing missing is some guarantee that the first `eval_steps` samples of the evaluation step are what are evaluated against. Its probably the case, but we should be sure.

ludwig/models/predictor.py
Infernaught: Done!

tests/ludwig/evaluation/test_evaluation.py
```diff
@@ -0,0 +1,67 @@
...
+    results1 = model.evaluate(df)
+
+    model.config_obj.trainer.eval_steps = 4
+    results2 = model.evaluate(df)
+    results3 = model.evaluate(df)
+
+    diff_exists = False
+    for k in results1[0]["out"]:
+        # Some metrics will be 0 across all runs
+        if results1[0]["out"][k] != 0 and results1[0]["out"][k] != results2[0]["out"][k]:
```
justinxzhao, line 62: Do we want to check somewhere in this test that there are 4 sample outputs?

PR #3779, opened 2023-11-15
feat: add support for tif files

dennisrall:
I am working with tif files at the moment and faced some problems. I tried to fix them in this PR. I made the following changes:

- added `.tif` to the `IMAGE_EXTENSION` tuple
- added tests for the `is_audio_score` function
- changed the implementation of the `is_image_score` function, so it is in line with the `is_audio_score` function (and is correct for a input like `.png/video.mp4`)
- added `read_image_as_tif` function to if the image can't be read as `png` or `numpy`
- some small type hinting improvements


justinxzhao (approved):
LGTM! Thanks @dennisrall for the addition.

PR #3785, opened 2023-11-20
[CI] Upgrade torch on CIs to 2.0.0, 2.1.1, and nightly

justinxzhao:


arnavgarg1:
LGTM, but we should also update the torch version of the docker images: https://github.com/ludwig-ai/ludwig/tree/master/docker

justinxzhao:
> LGTM, but we should also update the torch version of the docker images: https://github.com/ludwig-ai/ludwig/tree/master/docker

Good call out. Instead of upgrading the CI to torch==2.1.0, pytorch==2.0.0 seems to fix the CI and is already consistent with our docker images.

arnavgarg1:
That makes sense and sounds good!

arnavgarg1 (approved):

PR #3792, opened 2023-11-28
QOL: Fail config validation if a user tries to use ECD with a text output feature and an LLM encoder.

justinxzhao:


ludwig/config_validation/checks.py
```diff
@@ -631,6 +631,25 @@ def check_llm_quantization_backend_incompatibility(config: "ModelConfig") -> Non
             pass
 
 
+@register_config_check
+def check_llm_text_encoder_is_not_used_with_ecd(config: "ModelConfig") -> None:
+    """Checks that a pretrained text encoder is not used for ECD models with a text output feature."""
+    if config.model_type != MODEL_ECD:
+        return
+
+    if config.output_features[0].type != TEXT:
```
alexsherstinsky, line 640: @justinxzhao Very minor -- is there a reason to specify "output" first, instead of "input" (cognitive strain)?  Thanks.

ludwig/config_validation/checks.py
```diff
@@ -631,6 +631,25 @@ def check_llm_quantization_backend_incompatibility(config: "ModelConfig") -> Non
...
+        return
+
+    if config.input_features[0].type != TEXT:
+        return
+
+    if (
```
alexsherstinsky, line 646: ❤️

tests/ludwig/config_validation/test_checks.py
```diff
@@ -529,3 +523,24 @@ def test_check_sample_ratio_and_size_compatible():
...
+                    "pretrained_model_name_or_path": "meta-llama/Llama-2-7b-hf",
+                    "trainable": False,
+                },
+                "preprocessing": {"cache_encoder_embeddings": True},
+            }
+        ],
+        "output_features": [{"name": "Answer", "type": "text"}],
+    }
+
+    with pytest.raises(ConfigValidationError):
```
alexsherstinsky, line 545: @justinxzhao Suggestion, use this pattern near the bottom of `test_llm.py::test_llm_lora_finetuning_merge_and_unload_quantized_accelerate_required` in order to confirm the expected text of the error message (to be adapted for the present test's variables and logic):
```
    error_class: type  # noqa [F842]  # incorrect flagging of "local variable is annotated but never used
    error_message: str  # noqa [F842]  # incorrect flagging of "local variable is annotated but never used
    error_class, error_message = error_raised
    with pytest.raises(error_class) as excinfo:
        train_df = generate_data(input_features, output_features, filename=csv_filename, num_examples=3)
        model.train(dataset=train_df)

    assert str(excinfo.value) == error_message
```

Thank you!

alexsherstinsky (approved):
@justinxzhao If you would like to implement the suggestions, that would be great (approving anyway, if you deem it good enough).  Thank you!

arnavgarg1 (approved):
Looks like a reasonable change!

ludwig/config_validation/checks.py
justinxzhao: Done.

tests/ludwig/config_validation/test_checks.py
justinxzhao: Thanks for the suggestion @alexsherstinsky!

I'll say that in general I haven't been a huge fan of checking for error strings because it means that if we ever tweak the error string, we'll have to update the test, with maintenance overhead outweighing test precision (in my expereince).

That said, this file has been growing for a while and there are now many different raises that are all called `ConfigValidationError`, so more precision in this test could be useful.

I've made the change to check for the specific error string with an open mind to see how this pans out!

alexsherstinsky: @justinxzhao For me, it has been kind of similar: start "tighter enforcement/checking and ease off as needed; when we are easing too much, that means -- time for rethinking the abstraction". :-)

PR #3793, opened 2023-11-28
Log LLM test set evaluation outputs to CSV files under the model log directory, stamped by checkpoint number.

Infernaught:
Bypass MLFlow to log LLM outputs because of complications regarding the different MLFlow methods (log_dict does not interact well with our current MLFlow endpoint).

This is what the directory structure looks like: 
image


justinxzhao (approved):

arnavgarg1:
Should we merge?

PR #3796, opened 2023-11-30
Remove dicts from comparison metrics during visualization

Infernaught:
Currently, Ludwig visualization seems to be slightly broken when passing in test_statistics. We try to compare metrics that contain dict values instead of floats, causing the code to break.

This adds the dict metrics to our list of ignored metrics during visualization.

### DIFF
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -1503,11 +1503,7 @@ def compare_performance(
         metric_names = metric_names_sets[0]
         for metric_names_set in metric_names_sets:
             metric_names = metric_names.intersection(metric_names_set)
-        if LOSS in metric_names:
-            metric_names.remove(LOSS)
-        for name in ignore_names:
-            if name in metric_names:
-                metric_names.remove(name)
+        metric_names = metric_names - ignore_names
         metrics_dict = {name: [] for name in metric_names}
 
         for test_stats_per_model in test_stats_per_model_list:
--- a/ludwig/visualize.py
+++ b/ludwig/visualize.py
@@ -1480,7 +1480,15 @@ def compare_performance(
     compare_performance([a_evaluation_stats, b_evaluation_stats], model_names=["A", "B"])
     ```
     """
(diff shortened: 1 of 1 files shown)


ludwig/visualize.py
```diff
@@ -1494,7 +1502,6 @@ def compare_performance(
         metric_names = metric_names_sets[0]
         for metric_names_set in metric_names_sets:
             metric_names = metric_names.intersection(metric_names_set)
-        metric_names.remove(LOSS)
```
w4nderlust, line 1497: I'm not sure why this was here. The structure of the test statistics file changes from datatype to datatype, so maybe this is needed for a specific datatype. I suggest not to remove this but add an if instead

Infernaught: I think this was a mistake on my part. Let me add this back in.

ludwig/visualize.py
```diff
@@ -1494,7 +1502,8 @@ def compare_performance(
...
+        if LOSS in metric_names:
+            metric_names.remove(LOSS)
         for name in ignore_names:
             if name in metric_names:
                 metric_names.remove(name)
```
DavidFarago, line 1509:
```suggestion
        for name in ignore_names:
            metric_names.remove(name)
```

ludwig/visualize.py
```diff
@@ -1480,7 +1480,15 @@ def compare_performance(
...
-    ignore_names = ["overall_stats", "confusion_matrix", "per_class_stats", "predictions", "probabilities"]
+    ignore_names = [
+        "overall_stats",
+        "confusion_matrix",
+        "per_class_stats",
+        "predictions",
+        "probabilities",
+        "roc_curve",
+        "precision_recall_curve",
+    ]
```
DavidFarago, line 1491:
```suggestion
    ignore_names = [
        "overall_stats",
        "confusion_matrix",
        "per_class_stats",
        "predictions",
        "probabilities",
        "roc_curve",
        "precision_recall_curve",
        LOSS,
    ]
```

DavidFarago (changes requested):

justinxzhao (approved):

PR #3799, opened 2023-12-01
Add support for dequantizing 4-bit bitsandbytes base models into fp16

arnavgarg1:
This PR adds support for converting a quantized 4-bit base LLM model (in either `nf4` or `fp4` data types via bitsandbytes) into an upscaled float16 variant.

To use this function, it's as simple as:
1. Create a Ludwig config where quantization is set and bits = 4
2. Instantiate the LudwigModel class
3. Call `model.save_dequantized_base_model(save_path)`

You can also call this function after model training. The only caveat is that you will have to call `LudwigModel.load("/saved/artifact/path")` to reload the model with the trained adapter.

---

# Usage

```python
import yaml

config = """
    model_type: llm
    quantization:
        bits: 4
    ...
"""

config = yaml.safe_load(config)

model = LudwigModel(config)
... # Optionally call model.train(), but it's not required
model.save_dequantized_base_model("/src/llama-2-upscaled")
```

In the future, we will add support for dequantizing 8 bit models into fp16 as well.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -914,6 +914,46 @@ class LudwigModel:
         trainer.eval_batch_size = self.config_obj.trainer.eval_batch_size
         trainer.gradient_accumulation_steps = self.config_obj.trainer.gradient_accumulation_steps
 
+    def upscale_quantized_weights(self, save_path: str) -> None:
+        """Upscales quantized weights of a model and saves the result in a specified folder."""
+        if self.config_obj.model_type != MODEL_LLM:
+            raise ValueError(
+                f"Model type {self.config_obj.model_type} is not supported by this method. Only `llm` model type is "
+                "supported."
+            )
+
+        if not self.config_obj.quantization or self.config_obj.quantization.bits not in {4, 8}:
+            raise ValueError(
+                "Quantization is not enabled or the number of bits is not 4 or 8. "
+                "This method only works with quantized models with 4 or 8 bits."
+            )
+
+        if not torch.cuda.is_available():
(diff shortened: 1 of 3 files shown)


ludwig/api.py
```diff
@@ -914,6 +914,48 @@ def _tune_batch_size(self, trainer, dataset, random_seed: int = default_random_s
...
+            )
+
+        if self.config_obj.quantization.bits != 4:
+            raise ValueError(
+                "This method only works with quantized models with 4 bits. "
+                "Support for 8-bit quantized models will be added in a future release."
+            )
+
+        if not torch.cuda.is_available():
+            raise RuntimeError("GPU is required for quantized models but no GPU found.")
```
justinxzhao, line 951: If this ever becomes a common set of checks, it could be worth creating a reusable function.

justinxzhao (approved):
Nice!

ludwig/api.py
arnavgarg1: Agreed!

ludwig/utils/llm_quantization_utils.py
```diff
@@ -0,0 +1,62 @@
+import torch
+from bitsandbytes.functional import dequantize_4bit
+from bitsandbytes.nn.modules import Linear4bit
+from torch import nn
+
+from ludwig.api_annotations import DeveloperAPI
+
+
+@DeveloperAPI
+class Linear4BitToLinear(nn.Module):
```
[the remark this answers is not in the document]
arnavgarg1: I was tinkering around with some ideas initially, just cleaned this up!

PR #3801, opened 2023-12-03
Added target_modules to LoRA config

tgaddair:
Also changes the default alpha to `2 * r`.

### DIFF
--- a/ludwig/schema/llms/peft.py
+++ b/ludwig/schema/llms/peft.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Optional, Type, TYPE_CHECKING
+from typing import List, Optional, Type, TYPE_CHECKING
 
 from ludwig.api_annotations import DeveloperAPI
 from ludwig.error import ConfigValidationError
@@ -71,6 +71,10 @@ class BaseAdapterConfig(schema_utils.BaseMarshmallowConfig, ABC):
 @register_adapter(name="lora")
 @ludwig_dataclass
 class LoraConfig(BaseAdapterConfig):
+    def __post_init__(self):
+        if self.alpha is None:
+            self.alpha = self.r * 2
+
     type: str = schema_utils.ProtectedString(
         "lora",
         description=LLM_METADATA["adapter"]["lora"]["type"].long_description,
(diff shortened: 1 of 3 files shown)


arnavgarg1 (approved):

ludwig/schema/llms/peft.py
```diff
@@ -71,6 +71,10 @@ def to_config(self, **kwargs) -> "PeftConfig":
 @register_adapter(name="lora")
 @ludwig_dataclass
 class LoraConfig(BaseAdapterConfig):
+    def __post_init__(self):
```
arnavgarg1, line 74: It might also be useful to log a warning in case alpha and r are both set if alpha isn't 2 * r since the best results tend to occur with 2 * r almost always based on experiments. But maybe that's not in scope for this PR.

arnavgarg1 (approved):

ludwig/schema/llms/peft.py
tgaddair: I personally think it's fine without a warning. If this becomes the default, hopefully people won't be changing it unless they know what they're doing. We may want to move its importance from 2 to 1, though.

arnavgarg1: Makes sense!

ludwig/schema/llms/peft.py
```diff
@@ -101,6 +106,18 @@ class LoraConfig(BaseAdapterConfig):
         description="Bias type for Lora.",
     )
 
+    target_modules: Optional[List[str]] = schema_utils.List(
+        default=None,
+        allow_none=True,
+        description=(
+            "List of module names or regex expression of the module names to replace with LoRA. "
+            "For example, ['q', 'v'] or '.*decoder.*(SelfAttention|EncDecAttention).*(q|v)$'. "
+            "Defaults to targeting the query and value matrices of all self-attention and encoder-decoder attention "
```
alexsherstinsky, line 115: @tgaddair Just for my understanding: where is this default specified?  Is it inherent in the Transformers/PEFT implementation?  Thank you.

tests/integration_tests/test_llm.py
```diff
@@ -474,6 +474,11 @@ def _verify_lm_lora_finetuning_layers(
             {"r": 4, "dropout": 0.1},
             id="lora-modified-defaults",
         ),
+        pytest.param(
```
alexsherstinsky, line 477: @tgaddair Question: This additional test case for the present test function confirms that the model can be trained and then loaded and used.  However, it does not verify that the layers are what we would expect.  Do you think we need to verify this "target_modules" test case against something like `_verify_lm_lora_finetuning_layers()`?  If so, I would be happy to add it (as a learning experience).  Thank you.

alexsherstinsky (approved):
LGTM -- I just made a couple of remarks and included a question.  This was educational!

ludwig/schema/llms/peft.py
tgaddair: Yes, it's deep in the weeds of the PEFT code base:

https://github.com/huggingface/peft/blob/main/src/peft/utils/other.py#L481

PR #3810, opened 2023-12-06
Fix LLM integration tests that fail because of PEFT 0.7 release

arnavgarg1:
Released an hour ago: https://pypi.org/project/peft/0.7.0/

Failing tests: https://github.com/ludwig-ai/ludwig/actions/runs/7117174820/job/19377310798

It seems like the latest version of PEFT writes the adapter weights as safetensors instead of binary files.

### DIFF
--- a/tests/integration_tests/test_llm.py
+++ b/tests/integration_tests/test_llm.py
@@ -765,7 +765,7 @@ quantization section from your Ludwig configuration."""
             [
                 "README.md",
                 "adapter_config.json",
-                "adapter_model.bin",
+                "adapter_model.safetensors",
             ],
             id="lora_default_not_merged",
         ),
@@ -777,7 +777,7 @@ quantization section from your Ludwig configuration."""
             [
                 "README.md",
                 "adapter_config.json",
-                "adapter_model.bin",
+                "adapter_model.safetensors",
                 "config.json",
                 "generation_config.json",
                 "merges.txt",
@@ -797,7 +797,7 @@ quantization section from your Ludwig configuration."""
             [
                 "README.md",
                 "adapter_config.json",
-                "adapter_model.bin",
+                "adapter_model.safetensors",
             ],
             id="lora_custom_not_merged",
         ),
@@ -809,7 +809,7 @@ quantization section from your Ludwig configuration."""
             [
                 "README.md",
                 "adapter_config.json",
-                "adapter_model.bin",
+                "adapter_model.safetensors",
                 "config.json",
                 "generation_config.json",
                 "merges.txt",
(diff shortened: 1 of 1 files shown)


justinxzhao (approved):

tests/integration_tests/test_llm.py
```diff
@@ -401,21 +401,24 @@ def _verify_lm_lora_finetuning_layers(
...
+    from peft.tuners.lora.layer import LoraLayer
 
     expected_lora_num_features_orig: tuple[int] = (expected_lora_in_features, expected_lora_out_features)
 
     file_names: list[str] = list_file_names_in_directory(directory_name=model_weights_directory)
     assert set(file_names) == set(expected_file_names)
 
     target_module_name: str
-    target_module_obj: Linear
+    target_module_obj: LoraLayer | torch.nn.Linear
```
alexsherstinsky, line 412: @arnavgarg1 Is this correct, or do we only get the `LoraLayer` type?

tests/integration_tests/test_llm.py
```diff
@@ -401,21 +401,24 @@ def _verify_lm_lora_finetuning_layers(
...
 
     # Not providing default value to "getattr()" so that error is raised if incorrect projection layer name is supplied.
 
     for target_module_name in target_modules:
         target_module_obj = getattr(attention_layer, target_module_name)
-        assert isinstance(target_module_obj, torch.nn.Linear)
+        if merge_adapter_into_base_model:
```
alexsherstinsky, line 418: 💙

tests/integration_tests/test_llm.py
```diff
@@ -809,7 +812,7 @@ def test_llm_lora_finetuning_merge_and_unload_4_bit_quantization_not_supported(l
             [
                 "README.md",
                 "adapter_config.json",
-                "adapter_model.bin",
+                "adapter_model.safetensors",
```
alexsherstinsky, line 815: @arnavgarg1 I see what happened; these changes back and forth from bin to safetensors was in response to PEFT having bugs due to changes in the point releases.

alexsherstinsky (approved):
LGTM -- thank you very much!

tests/integration_tests/test_llm.py
arnavgarg1: @alexsherstinsky When merged, we get the `torch.nn.Linear` like in the base model, but without merging, it's actually the LoraLayer type!

tests/integration_tests/test_llm.py
arnavgarg1: Ah! I see, good to have that context @alexsherstinsky!

tests/integration_tests/test_llm.py
arnavgarg1: Previously, the target module layers were also torch.nn.Linear layers so we were always guaranteed that with or without merge, we'd end up with torch.nn.Linear layers.

PR #3811, opened 2023-12-06
Add binary averaging strategy for precision recall and f1 score to overall stats

arnavgarg1:
For binary and category output features, we currently only return the micro, macro and weighted average values of precision, recall and f1. However, for binary features in particular, we should also return the precision, recall and f1 scores using the binary averaging strategy. 

This PR ensures that anytime there are only 2 labels in the confusion matrix and they are specifically True and False (https://github.com/ludwig-ai/ludwig/blob/master/ludwig/features/binary_feature.py#L320), we also compute these new metrics.

### DIFF
--- a/ludwig/utils/eval_utils.py
+++ b/ludwig/utils/eval_utils.py
@@ -15,6 +15,7 @@
 # ==============================================================================
 import logging
 from collections import OrderedDict
+from typing import Union
 
 import numpy as np
 from sklearn import metrics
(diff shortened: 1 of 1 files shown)


ludwig/utils/eval_utils.py
```diff
@@ -245,6 +254,11 @@ def stats(self):
             "avg_f1_score_weighted": self.avg_f1_score(average="weighted"),
             "kappa_score": self.kappa_score(),
         }
+        if len(self.idx2label) == 2 and set(self.label2idx.keys()) == {"False", "True"}:
```
w4nderlust, line 257: the labels in idx2label are the labels that are there in the original dataset, there is a mechanism for the user to specify which one is the positive one, so i would not limit this just to {"False", "True"}.
Also the same applies to the calls to the score functions, as in sklearn they don't accept only binary values but also other datatypes.

arnavgarg1: @w4nderlust I agree! The reason I did it this way is because for binary features in particular, we've hardcoded the labels that get passed into the ConfusionMatrix class: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/features/binary_feature.py#L320 

I figured that this more or less guarantees that the predictions will always have these 2 labels - True and False. The binary output feature (in combination with the `fallback_true_label` property in the preprocessing parameters for binary output features) will already preprocess and then post-process it so that we only get these 2 labels back. 

What do you think?

PR #3818, opened 2023-12-09
Add support for IA3 PEFT Strategy

arnavgarg1:
Paper: https://arxiv.org/pdf/2205.05638.pdf

Adds support for a new PEFT strategy called IA3, which adds 2 learned vectors to the V and Q projections in attention heads, as well as a learned vector to the feed-forward network. The idea is that these learned vectors can help rescale the attention values and feed-forward network values. l_k, l_v, and l_ff are all initialized with ones so that the overall function computed by the model does not change when they are added.

IA3 makes mixed-task batches possible because each sequence of activations in the batch can be separately and cheaply multiplied by its associated learned task vector (in some ways, very similar to how you can train different rank decomposed matrices with LoRA for each task). In the event that a model will only be used on a single task, the modifications introduced by IA3 can also be applied to weight matrices permanently so that no elementwise multiplication is required and the model’s architecture remains unchanged. This is possible because element-wise multiplications performed in IA3 always co-occur with matrix multiplication, which means that there is no additional computational cost compared to the original model.

### DIFF
--- a/ludwig/schema/llms/peft.py
+++ b/ludwig/schema/llms/peft.py
@@ -461,6 +461,71 @@ class AdaptionPromptConfig(BaseAdapterConfig):
 
 
 @DeveloperAPI
+@register_adapter("ia3")
+@ludwig_dataclass
+class IA3Config(BaseAdapterConfig):
+    type: str = schema_utils.ProtectedString(
+        "ia3",
+        description=LLM_METADATA["adapter"]["ia3"]["type"].long_description,
+    )
+
+    target_modules: Optional[List[str]] = schema_utils.List(
+        default=None,
+        allow_none=True,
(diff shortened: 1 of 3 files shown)


alexsherstinsky (approved):
LTGM -- @arnavgarg1 -- really nice!  So, if I interpret this correctly, the IA3 algorithm is already supported by PEFT in HuggingFace -- and this change enables Ludwig to accept, validate, and pass the IA3 configuration as part of the `adapter` configuration section in case the `model_type` is `llm`.  Is that correct?  Thank you!  P.S.: Thanks for sharing the reference to the paper, which introduces the IA3 technique.

arnavgarg1:
@alexsherstinsky Yes, that is exactly right :)

PR #3821, opened 2023-12-11
Move reusable LLM model methods to utility functions

jeffkinnison:
This is a small PR to move two methods out of the LLM object and into utils so they can be reused for text encoders. The two methods/functions are:

- `initialize_adapter` - wraps a pretrained model with a PEFT model, with support for new and pretrained adapters
- `to_device` - moves a pretrained model to device with logic to shard the model over multiple GPUs.

### DIFF
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -248,7 +248,7 @@ class LLM(BaseModel):
         self.model = prepare_model_for_kbit_training(self.model, use_gradient_checkpointing=False)
 
     def to_device(self, device):
-        self.model = to_device(self.model, self.config_obj, self.curr_device)
+        self.model = to_device(self.model, device, self.config_obj, self.curr_device)
         self.curr_device = torch.device(device)
         return self
 
--- a/ludwig/models/llm.py
+++ b/ludwig/models/llm.py
@@ -26,6 +26,7 @@ from ludwig.utils.llm_utils import (
     generate_merged_ids,
     get_context_len,
     get_realigned_target_and_prediction_tensors_for_inference,
+    initialize_adapter,
     pad_target_tensor_for_fine_tuning,
     remove_left_padding,
 )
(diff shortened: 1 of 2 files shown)


justinxzhao (approved):

ludwig/models/llm.py
```diff
@@ -267,59 +248,8 @@ def prepare_for_quantized_training(self):
...
-                    self.model = AutoModelForCausalLM.from_pretrained(
-                        tmpdir,
-                        **model_kwargs,
-                    )
-
-        else:
-            self.model = self.model.to(device)
-
-        self.curr_device = device
+        self.model = to_device(self.model, self.config_obj, self.curr_device)
```
arnavgarg1, line 251: Seems like this also requires device to be passed in right? So it should be:

```python
self.model = to_device(self.model, device, self.config_obj. self.curr_device)
```

arnavgarg1: That you can also return `device` which becomes a `torch.device(device)` object within `to_device()`

arnavgarg1 (commented):
I think the function call `to_device()` may need to be updated

ludwig/models/llm.py
jeffkinnison: Ah, yeah, small typo

jeffkinnison: Okay, that should be fixed along with a circular import

arnavgarg1 (approved):

ludwig/utils/llm_utils.py
```diff
@@ -1,14 +1,17 @@
...
 import logging
-from typing import Dict, Tuple
+import tempfile
+from typing import Dict, Tuple, Union
 
 import torch
 import torch.nn.functional as F
 from bitsandbytes.nn.modules import Embedding
-from transformers import AutoConfig, AutoModelForCausalLM, PreTrainedTokenizer
+from peft import get_peft_model, MODEL_TYPE_TO_PEFT_MODEL_MAPPING, PeftConfig, PeftModel, TaskType
```
geoffreyangus, line 9: @jeffkinnison seems like tests are failing due to a peft import failure. Can you wrap this in try/except logic like we do here? https://github.com/ludwig-ai/ludwig/blob/8f10afe609fd10fb08afcb6063a9dfb8ad7f41e9/ludwig/utils/visualization_utils.py#L31-L52

arnavgarg1: @jeffkinnison Another option is to just import these functions inline - that way they only get imported when the method is called, which is fine because it means that the user has the llm dependencies already installed (I believe that is what we were doing before)

jeffkinnison: Done, I reverted it to the inline format.

PR #3823, opened 2023-12-11
[MAINTENANCE] Support Adapter Model ".safetensors" File Format During HF Upload

alexsherstinsky:
### Scope
* This change updates Ludwig to handle changes introduced in HuggingFace PEFT 0.7.0: Adapter weights are now stored in the ".safetensors" format, instead of the PyTorch default Pickle ".bin" format.  This change caused Ludwig to error when uploading the trained model to HuggingFace, because Ludwig verifies acceptable file extensions.  With this change, both are allowed for now; however, there is a "TODO" to formalize the deprecation policy, and subsequently discontinue supporting ".bin" format.
* The new test module for testing upload functionality is created and contains the initial tests for the new model file format.
* Enhancements in type hinting and improvements to exception types in "upload_utils.py" is also included with this change.

### Remarks
* Please note that even though HuggingFace PEFT now uses "safetensors" as the default, it is backward-compatible, meaning that "adapter_mode.bin" files, uploaded via the earlier versions of PEFT, will still be read-in correctly (please see "https://github.com/huggingface/peft/blob/e73967edea0b337fb11fdde3da5b9823512f8f68/src/peft/utils/save_and_load.py#L236" for details):
```
    if use_safetensors:
        adapters_weights = safe_load_file(filename, device=device)
    else:
        adapters_weights = torch.load(filename, map_location=torch.device(device))
```

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/tests/ludwig/utils/test_upload_utils.py
+++ b/tests/ludwig/utils/test_upload_utils.py
@@ -11,7 +11,14 @@ from ludwig.utils.upload_utils import HuggingFaceHub
 logger = logging.getLogger(__name__)
 
 
-def _build_tmp_model_repo(destination_directory: str, file_names: list[str]) -> None:
+def _build_fake_model_repo(
+    destination_directory: str,
+    experiment_name: str,
+    file_names: list[str],
+    *,
+    model_directory_name: str = "model",
+    model_weights_directory_name: str = "model_weights",
+) -> None:
     """This utility function accepts the "destination_directory" and list of file names on input.
 
     It then makes directory hierarchy "my_simple_experiment_run" / "model" / "model_weights" under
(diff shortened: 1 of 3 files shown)


arnavgarg1 (approved):
PR looks great! Thanks for adding this backwards compatibility check and also adding a nice test set for this :)

tests/ludwig/utils/test_upload_utils.py
```diff
@@ -0,0 +1,167 @@
...
+@pytest.mark.unit
+def test_upload_to_hf_hub__validate_upload_parameters(
+    output_directory_manager, file_names: list[str], error_raised: tuple[type, str] | None
+):
+    """Test "HuggingFaceHub._validate_upload_parameters()", which is executed in the path of upload to HuggingFace
+    Hub; for example: `upload hf_hub -repo_id "hf-account/repo-name" --model_path.
+
+    /content/results/api_experiment_run`.
+
+    Eeach test case consists of: 1) Populating the temporary output directory ("training_results_output) with zero or
```
justinxzhao, line 137: nit: Each

justinxzhao (approved):

tests/ludwig/utils/test_upload_utils.py
alexsherstinsky: Nice catch -- fixed -- thank you!

PR #3828, opened 2023-12-13
Add LLM Text Encoder

jeffkinnison:
# Overview

This adds a text encoder for ECD that wraps a pretrained LLM and passes the model's hidden state downstream to the combiner. We reuse large portions of the LLM model type's utitilies, refactored as utility functions rather than `LLM` methods.

## LLM Encoder

The encoder subclasses `SequenceEncoder` and implements custom behavior for working with LLMs, with pieces borrowed from `LLM`. Whereas the `LLM` model type is focused on text generation, the encoder is focused on 1) passing hidden state downstream for predictive tasks and 2) packaging the adapter with the rest of the ECD architecture. Major methods include

- `__init__`: Loads the pretrained model and adapter from config
- `forward`: Takes tokenized text as input and returns the hidden state of the last layer
- `_save_to_state_dict`: Normally, adapter weights are saved with `PEFTModel.save_pretrained`, which writes the weights to file. In ECD, however, additional parameters need to be recorded, particularly the combiner and decoder weights. Since these weights will have been trained alongside a particular adapter, we want to package them together. This method adds custom logic to extract adapter weights as part of the dictionary returned by `model.state_dict()`.
- `load_from_state_dict`: Under the hood, PEFT alters the names of the adapter parameters. This leads to errors loading the state dict back in. This adds additional logic to load the state dict with a PEFT utility then update the `load_state_dict` args to reflect that the adapter state was loaded in.

## Refactored `LLM` Methods

The following methods have been moved from the `LLM` class into `ludwig.utils.llm_utils`:

- `initialize_adapter`: This now takes a `PretrainedModel` as an argumentand returns a `PEFTModel`
- `load_pretrained_from_config`: This was moved from `ludwig.models.llm` with no changes
- `to_device`: This now takes a model and device as arguments and returns the model on device and destination device

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -2406,13 +2406,13 @@ class LLMEncoder(Encoder):
         # When merging input IDs and target IDs for LLM fine-tuning, we want to make sure that the merged tensor is
         # not longer than the global maximum sequence length. This is provided in the preprocessing config. We never
         # want to exceed the maximum possible context length so we also check for that.
-        # if self.encoder_config.max_sequence_length:
-        #     max_sequence_length = self.encoder_config.max_sequence_length
-        #     self.max_sequence_length = (
-        #        max_sequence_length if max_sequence_length <= self.context_len else self.context_len
-        #     )
-        # else:
-        self.max_sequence_length = self.context_len
+        if self.config.max_sequence_length:
+            max_sequence_length = self.config.max_sequence_length
+            self.max_sequence_length = (
+                max_sequence_length if max_sequence_length <= self.context_len else self.context_len
+            )
+        else:
+            self.max_sequence_length = self.context_len
 
         # Initialize tokenizer
         self.tokenizer = HFTokenizer(self.config.base_model).tokenizer
@@ -2482,8 +2482,21 @@ class LLMEncoder(Encoder):
 
             sd = get_peft_model_state_dict(self.model)
             destination.update(sd)
+
         else:
             super()._save_to_state_dict(destination, prefix=prefix, keep_vars=keep_vars)
+
+    def state_dict(self, *args, destination=None, prefix="", keep_vars=False):
+        destination = super().state_dict(destination, prefix=prefix, keep_vars=keep_vars)
+
+        if self.config.adapter:
+            adapter_type_prefix = self.ADAPTER_PARAM_NAME_PREFIX[self.config.adapter.type]
+            exclude_model_keys = [k for k in destination.keys() if adapter_type_prefix not in k]
+
+            for k in exclude_model_keys:
+                del destination[k]
+
+        return destination
 
     def _load_from_state_dict(
         self, state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -2415,6 +2415,8 @@ class LLMEncoder(Encoder):
 
         self.attention_masks = None
 
+        self.prepare_for_training()
+
         clear_data_cache()
 
     @staticmethod
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -56,7 +56,7 @@ from ludwig.schema.encoders.text_encoders import (
 from ludwig.schema.llms.peft import BaseAdapterConfig
 from ludwig.utils.data_utils import clear_data_cache
 from ludwig.utils.hf_utils import load_pretrained_hf_model_with_hub_fallback
-from ludwig.utils.llm_utils import get_context_len, load_pretrained_from_config
+from ludwig.utils.llm_utils import get_context_len, initialize_adapter, load_pretrained_from_config
 from ludwig.utils.tokenizers import HFTokenizer
 from ludwig.utils.torch_utils import FreezeModule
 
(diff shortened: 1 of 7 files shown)


ludwig/encoders/text_encoders.py
```diff
@@ -2374,3 +2380,99 @@ def output_shape(self) -> torch.Size:
...
+    def initialize_adapter(self):
+        """If an adapter config is provided, we want to wrap the model with a PEFT model for fine-tuning."""
+        if self.config.adapter:
+            self.model = initialize_adapter(self.model, self.config)
+
+            logger.info("==================================================")
+            logger.info("Trainable Parameter Summary For LLM Encoder Fine-Tuning")
+            logger.info(f"Fine-tuning with adapter: {self.config.adapter.type}")
+            self.model.print_trainable_parameters()
+            logger.info("==================================================")
```
arnavgarg1, line 2445: nit: Should we potentially consider moving this within initialize_adapter?

jeffkinnison: Sure! I wanted the logs to distinguish between the LLM mode type and the LLM encoder, but we can probably do that more nicely

ludwig/encoders/text_encoders.py
```diff
@@ -2374,3 +2380,99 @@ def output_shape(self) -> torch.Size:
...
+
+        self.model = prepare_model_for_kbit_training(self.model, use_gradient_checkpointing=False)
+
+    def forward(self, inputs: torch.Tensor, mask: Optional[torch.Tensor] = None):
+        # Wrap with flash attention backend for faster generation
+        with torch.backends.cuda.sdp_kernel(enable_flash=True, enable_math=False, enable_mem_efficient=False) if (
+            torch.cuda.is_available() and self.curr_device.type == "cuda"
+        ) else contextlib.nullcontext():
+            # TODO (jeffkinnison): Determine why the 8-bit `SCB` and `CB` matrices are deleted in the forward pass
+            model_outputs = self.model(input_ids=inputs, return_dict=True, output_hidden_states=True).hidden_states[-1]
```
geoffreyangus, line 2464: is `.last_hidden_state` possible here? If so, maybe use that for readability

ludwig/encoders/text_encoders.py
```diff
@@ -2374,3 +2380,99 @@ def output_shape(self) -> torch.Size:
...
+        return {ENCODER_OUTPUT: model_outputs}
+
+    def _save_to_state_dict(self, destination: Dict, prefix: str, keep_vars: bool):
```
geoffreyangus, line 2467: OOC, could you drop a code pointer here for where this is called? Would like to understand where this `destination` dict is coming from!

tests/integration_tests/test_llm.py
```diff
@@ -1190,3 +1192,25 @@ def test_llm_finetuning_with_embedding_noise(
     preds = convert_preds(preds)
 
     assert preds
+
+
+def test_llm_encoding(tmpdir):
```
geoffreyangus, line 1197: lmk when this is fleshed out and passing and I'll give this PR another pass!

ludwig/encoders/text_encoders.py
arnavgarg1: Totally makes sense! Will review the rest soon :)

ludwig/encoders/text_encoders.py
jeffkinnison: It seems like `last_hidden_state` isn't included in `CasualLMOutput` or `CausalLMWithPastOutput`, or at the very least it isn't in my local testing.

ludwig/encoders/text_encoders.py
```diff
@@ -2374,3 +2380,150 @@ def output_shape(self) -> torch.Size:
...
+        # Initialize tokenizer
+        self.tokenizer = HFTokenizer(self.config.base_model).tokenizer
+
+        self.attention_masks = None
+
+        self.prepare_for_training()
+
+        if not self.config.adapter:
+            out_module = list(self.model.modules())[-1]
+            out_module.requires_grad_(requires_grad=False)
```
arnavgarg1, line 2426: Can we add a comment explaining what this code does, i.e., why are we turning off gradients for the last module? I'd also recommend moving this to within the `prepare_for_training()` encapsulation since it seems in spirit of that function!

jeffkinnison: Done!

tests/ludwig/encoders/test_llm_encoders.py
```diff
@@ -0,0 +1,137 @@
...
+
+    #     # Load the weights of encoder1 back into encoder2
+    #     encoder2.load_state_dict(encoder1_sd)
+    #     encoder2_sd = encoder2.state_dict()
+
+    #     # The LoRA weights should now be equal again
+    #     assert all(map(lambda k: torch.equal(encoder1_sd[k], encoder2_sd[k]), adapter_keys))
+
+    #     # The remaining weights should still be unequal
+    #     assert all(map(lambda k: not torch.equal(encoder1_sd[k], encoder2_sd[k]), model_keys))
```
arnavgarg1, line 137: Is this test commented out intentionally?

ludwig/encoders/text_encoders.py
```diff
@@ -2374,3 +2380,153 @@ def output_shape(self) -> torch.Size:
...
+            # When using an adapter, only the adapter weights are saved, and so we only want to load those weights.
+            # Under the hood, PEFT alters the names of the parameters, which leads to an "unexpected keys" error when
+            # using strict mode. This block uses PEFT's version of `load_state_dict` to handle loading in weights.
+            from peft.utils.save_and_load import set_peft_model_state_dict
+
+            adapter_type_prefix = self.ADAPTER_PARAM_NAME_PREFIX[self.config.adapter.type]
+            peft_model_state_dict = {k: v for k, v in state_dict.items() if adapter_type_prefix in k}
+            set_peft_model_state_dict(self.model, peft_model_state_dict)
+
+            if strict:
```
geoffreyangus, line 2528: I see, so this code block just removes the peft weights from the unexpected keys dictionary that is being passed around in Torch?

ludwig/encoders/text_encoders.py
```diff
@@ -2374,3 +2380,153 @@ def output_shape(self) -> torch.Size:
...
+        destination = super().state_dict(destination, prefix=prefix, keep_vars=keep_vars)
+
+        if self.config.adapter:
+            adapter_type_prefix = self.ADAPTER_PARAM_NAME_PREFIX[self.config.adapter.type]
+            exclude_model_keys = [k for k in destination.keys() if adapter_type_prefix not in k]
+
+            for k in exclude_model_keys:
+                del destination[k]
+
+        return destination
```
arnavgarg1, line 2506: Very neat!

arnavgarg1 (approved):
Really nice work! 🚢

arnavgarg1 (approved):
LGTM! Really nice work @jeffkinnison 👏🏽 

In a follow-up PR, we should address moving the `prepare_for_training()` call to happen only once we're inside each trainer so that it makes fewer assumptions abut the model being loaded on a CPU head node / GPU head in a multi-node setup.

tests/ludwig/encoders/test_llm_encoders.py
jeffkinnison: Yes, there's an issue with the custom logic for adapters when you run `load_state_dict` on the model itself. I'm digging into it, but since the integration tests are passing and do a load as part of training, I think we can add this back in as a followup.

ludwig/encoders/text_encoders.py
jeffkinnison: Yes. We have to do this because PEFT renames the parameters under the hood. With the normal `.save_pretrained`/`from_pretrained` model checkpointing logic, the conversion seems to be handled behind the scenes. When we do it like this though, torch thinks we just have extra weights, and in strict mode that raises and exception.

tests/ludwig/encoders/test_llm_encoders.py
arnavgarg1: Sounds good!

PR #3840, opened 2023-12-18
Update LLM trainer defaults

arnavgarg1:
Updates LLM trainer defaults to use parameters that work well out of the box so that they don't need to be set in the trainer config!


alexsherstinsky (approved):
LGTM -- ❤️ -- thank you!  This will enable us to omit major section of configuration for the LLM fine-tuning mode!

PR #3843, opened 2023-12-20
 fix: Handle missing and unexpected keys during LLMEncoder state dict load, part 2

jeffkinnison:
Followup to #3841. When loading the state dict with fully-qualified parameter names, the post-load hook was unable to properly compare missing keys with local model parameter names. For example, when attempting to remove the key `encoder.model.base_model.model.model.embed_tokens.weight` from the missing keys list, the hook would only be have access to the parameter name `model.base_model.model.model.embed_tokens.weight`.

This adds a parameter name prefix computation to the hook and updates the load_state_dict unit tests to include a wrapper for validating that fully-qualified names are loaded correctly.

### DIFF
--- a/ludwig/encoders/text_encoders.py
+++ b/ludwig/encoders/text_encoders.py
@@ -2535,7 +2535,7 @@ class LLMEncoder(Encoder):
         lists in `incompatible_keys` must be made in-place.
 
         Args:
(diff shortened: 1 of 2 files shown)


geoffreyangus (approved):
nice change!

PR #3852, opened 2024-01-03
Add default LoRA target modules for Mixtral and Mixtral instruct

arnavgarg1:
PEFT 0.7.1 doesn't have a default LoRA target module mapping: https://github.com/huggingface/peft/blob/8665e2b5719faa4e4b91749ddec09442927b53e0/src/peft/utils/constants.py#L49

This PR adds support to default to `q_proj` and `v_proj` for mixtral and mixtral instruct in the instance that the `target_modules` aren't explicitly specified for mixtral.


alexsherstinsky (approved):
LGTM! ❤️

PR #3858, opened 2024-01-04
Revert hack that leads to OOM during fine-tuning

arnavgarg1:
A few weeks ago, we merged https://github.com/ludwig-ai/ludwig/pull/3830 to temporarily get around an issue of fine-tuning with gradient checkpointing that was introduced with Transformers 4.36. The original issue can be seen here: https://github.com/huggingface/transformers/issues/28023

Since then, they've released a patch release (Transformers 4.36.2) that fixes the original issue for all model types, including Llama-2, Mixtral, Phi etc. However, it seems like the overall interaction between the new transformers version and our hacky fix leads to memory ballooning because we manually map each module to the correct mode, and in the process, set some modules to train mode when they shouldn't be which causes the memory to balloon. 

This PR is no longer needed if transformers is set to use 4.36.2 since it has the patch release. I've pinned the minimum version of transformers to this version as part of this PR.

### DIFF
--- a/ludwig/distributed/base.py
+++ b/ludwig/distributed/base.py
@@ -48,15 +48,6 @@ class DistributedStrategy(ABC):
         """
         pass
 
-    def eval(self, model: nn.Module):
-        model.eval()
-
-    def train(self, model: nn.Module, prev_model_training_mode: bool = None):
-        if prev_model_training_mode is not None:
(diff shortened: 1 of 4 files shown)


alexsherstinsky (approved):
LGTM

tgaddair (approved):

PR #3863, opened 2024-01-08
fix: correct typo in FeatureCollection

dennisrall:
I think there is a bug/typo in the FeatureCollection class


arnavgarg1 (approved):
Nice catch!

PR #3866, opened 2024-01-08
Scrub credentials from model_hyperparameters.json and description.json

Infernaught:
Scrub credentials from artifacts

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -102,7 +102,6 @@ from ludwig.utils.misc_utils import (
     get_from_registry,
     get_output_directory,
     set_saved_weights_in_checkpoint_flag,
-    scrub_creds
 )
 from ludwig.utils.print_utils import print_boxed
 from ludwig.utils.tokenizers import HFTokenizer
@@ -539,7 +538,7 @@ class LudwigModel:
                 # save description
                 if self.backend.is_coordinator():
                     description = get_experiment_description(
-                        scrub_creds(self.config_obj.to_dict()),
+                        self.config_obj.to_dict(),
                         dataset=dataset,
                         training_set=training_set,
                         validation_set=validation_set,
@@ -1899,8 +1898,7 @@ class LudwigModel:
         """
         os.makedirs(save_path, exist_ok=True)
         model_hyperparameters_path = os.path.join(save_path, MODEL_HYPERPARAMETERS_FILE_NAME)
-        config_dict = self.config_obj.to_dict()
-        save_json(model_hyperparameters_path, scrub_creds(config_dict))
+        save_json(model_hyperparameters_path, self.config_obj.to_dict())
 
     def to_torchscript(
         self,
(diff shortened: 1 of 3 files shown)


ludwig/api.py
```diff
@@ -1898,7 +1898,13 @@ def save_config(self, save_path: str) -> None:
...
         os.makedirs(save_path, exist_ok=True)
         model_hyperparameters_path = os.path.join(save_path, MODEL_HYPERPARAMETERS_FILE_NAME)
-        save_json(model_hyperparameters_path, self.config_obj.to_dict())
+        save_json(model_hyperparameters_path, self._scrub_creds(self.config_obj).to_dict())
+
+    def _scrub_creds(self, config_obj: ModelConfig) -> ModelConfig:
+        """Returns a copy of the config with all sensitive fields scrubbed."""
+        config_dict = config_obj.to_dict()
+        config_dict.get("backend", {})["credentials"] = {}
+        return ModelConfig.from_dict(config_dict)
```
arnavgarg1, line 1907: @Infernaught i think what we want to do is to temporary store credentials in a variable, write the artifact to disk, and then insert the credentials back into the object. 

In most cases, we do need to persist these credentials in memory in the process since we will likely use it downstream, but we just want to make sure it doesn't make it to disk.

arnavgarg1: Ah I see, nvm! Looks good to me

ludwig/api.py
```diff
@@ -1898,7 +1898,13 @@ def save_config(self, save_path: str) -> None:
         """
         os.makedirs(save_path, exist_ok=True)
         model_hyperparameters_path = os.path.join(save_path, MODEL_HYPERPARAMETERS_FILE_NAME)
-        save_json(model_hyperparameters_path, self.config_obj.to_dict())
+        save_json(model_hyperparameters_path, self._scrub_creds(self.config_obj).to_dict())
+
+    def _scrub_creds(self, config_obj: ModelConfig) -> ModelConfig:
+        """Returns a copy of the config with all sensitive fields scrubbed."""
+        config_dict = config_obj.to_dict()
+        config_dict.get("backend", {})["credentials"] = {}
```
arnavgarg1, line 1906: I think we'll have to check if `credentials` is in the backend config. The reason is that the `backend` is the only part of a Ludwig config that hasn't been schemified yet. Otherwise it'll complain about an access/type error if the `credentials` key ins't in the backend config

ludwig/api.py
```diff
@@ -538,7 +539,7 @@ def on_epoch_end(self, trainer, progress_tracker, save_path):
                 # save description
                 if self.backend.is_coordinator():
                     description = get_experiment_description(
-                        self.config_obj.to_dict(),
+                        scrub_creds(self.config_obj.to_dict()),
```
justinxzhao, line 542: To be even stricter, can we make `.to_dict()` call `scrub_creds()`?

I imagine that any callers of `confict_obj.to_dict()` can do without rendered credentials. Consumers of credentials via config_obj should use the backend object directly, or via something like `config_obj.backend.credentials`.

Infernaught: I can def make the change, but I don't know if that'll break something. Let's find out.

arnavgarg1 (approved):

justinxzhao (approved):

PR #3867, opened 2024-01-09
Add per-step token utilization to tensorboard and progress tracker.

justinxzhao:
Add per-step token utilization to tensorboard and progress tracker.

## Progress Tracker

7 new keys:
- `1: cumulative_step_token_usage`: The cumulative number of tokens used, keyed by train step.

    ```
    "0": 364,
    "1": 888,
    "2": 1142,
    "3": 1610,
    "4": 1886,
    "5": 2610,
    ```

    And the checkpoint-based mapping, `2: cumulative_checkpoint_token_usage`.

- `3: incremental_step_token_usage`: The incremental number of tokens used, keyed by train step.

    ```
    "0": 364,
    "1": 524,
    "2": 254,
    "3": 468,
    "4": 276,
    "5": 724,
    ```

    And the checkpoint-based mapping, `4: incremental_checkpoint_token_usage`.

- `5: total_tokens_used`: The total number of tokens used, over the entire training run.

    ```
    "total_tokens_used": 553843
    ```

- `6: checkpoint_to_step`: Map of checkpoint number to step number.
- `7: checkpoint_to_epoch`: Map of checkpoint number to epoch number.

## Tensorboard

Token usage is also logged to tensorboard.

Sample ECD model:

image

Sample BERT text classification model:

Screen Shot 2024-01-08 at 8 00 06 PM

Sample Llama model:

image

### How tokens are counted

| model_type | # of tokens                                |
|------------|--------------------------------------------|
| gbm        | len(input + output)                        |
| llm        | len(input + output) non-padding token IDs. |
| ecd        | len(input + output) tensors.               |


ludwig/accounting/used_tokens.py
```diff
@@ -0,0 +1,55 @@
...
+    binary, category, and number features, and variable number of tokens for text features, for each example in the
+    batch.
+
+    Args:
+        inputs: The input tensors for one forward pass through ecd.
+        targets: The target tensors for one forward pass through ecd.
+    """
+    used_tokens = 0
+    for input_feature_tensor in inputs.values():
+        used_tokens += torch.flatten(input_feature_tensor).shape[0]
```
arnavgarg1, line 37: Is my understanding here correct - this line (like in the gbm function) essentially flattens an Nd array into a single dimensional array and we take the shape which accounts for number of features "per row" * batch size? 

If yes, the net effect of this is that for any N dimensional vector feature, text, sequence and timeseries features, we will treat each value in the N dimensional input as an individual token?

ludwig/utils/trainer_utils.py
```diff
@@ -191,8 +198,14 @@ def __init__(
...
 
+        # Token usage.
+        self.incremental_token_usage = incremental_token_usage
+        self.cumulative_token_usage = cumulative_token_usage
+        self.total_tokens_used = total_tokens_used
+
     def save(self, filepath):
-        save_json(filepath, self.__dict__)
+        # sort_keys=False to ensure that token usage dictionaries (keyed by integers) are encodable.
+        save_json(filepath, self.__dict__, sort_keys=False)
```
arnavgarg1, line 208: Oh super interesting - how has this worked historically with current epochs/steps? Would it perhaps makes sense to change the dictionary mapping to `Dict[str, int]` instead of `Dict[int,int]` so that we don't have to set `sort_keys=False`? Might also help with consistency with metrics stored per epoch which I believe are keyed by dataset (train, test, val) + metric + epoch, all of which are strings?

tests/ludwig/accounting/test_used_tokens.py
```diff
@@ -0,0 +1,26 @@
...
+    inputs = {"input1": torch.tensor([[1, 2], [3, 4]]), "input2": torch.tensor([5, 6])}
+
+    assert get_used_tokens_for_gbm(inputs) == 3
+
+
+def test_get_used_tokens_for_ecd():
+    inputs = {"input1": torch.tensor([[1, 2], [3, 4]]), "input2": torch.tensor([5, 6])}
+    targets = {"output": torch.tensor([7, 8, 9])}
+
+    assert get_used_tokens_for_ecd(inputs, targets) == 9
```
arnavgarg1, line 16: Is it worth parameterizing this to also cover the case where targets can be None?

justinxzhao: Added a test.

ludwig/utils/trainer_utils.py
justinxzhao: It turns out that we do want to index by strings because deserializing from JSON when restarting from an existing checkpoint, any dictionary's keys get automatically serialized as strings anyway. I've added a comment about this in the ProgressTracker docstring as this behavior is not at all obvious.

PR #3868, opened 2024-01-09
qol: Fix some lints.

justinxzhao:


ludwig/models/ecd.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         except KeyError as e:
             raise KeyError(
                 f"An input feature has a name that conflicts with a class attribute of torch's ModuleDict: {e}"
-            )
+            ) from e
```
arnavgarg1, line 45: Should we get rid of the `{e}` in the logging statement since it will get printed out anyway?

ludwig/utils/data_utils.py
```diff
@@ -356,7 +356,7 @@ def read_stata(data_fp, df_lib, **kwargs):
 
 @DeveloperAPI
 @spread
-def read_hdf5(data_fp, **kwargs):
+def read_hdf5(data_fp, **_kwargs):
```
arnavgarg1, line 359: Why does this need to be `**_kwargs`?

arnavgarg1 (approved):

PR #3869, opened 2024-01-09
Add a utility to detect LLM usage in a config

jeffkinnison:
The LLM encoder in ECD needs to share large parts of the code base with the LLM model type, but branching currently inspects configs for model type only. This adds a utility that detects LLM model type or the usage of LLM encoder in text features for ECD models (LLM encoder is not currently valid for GBM). This can be used as a replacement for expressions like `if config.model_type =="llm":`.

### DIFF
--- a/ludwig/utils/config_utils.py
+++ b/ludwig/utils/config_utils.py
@@ -105,7 +105,7 @@ def has_pretrained_encoder(config: ModelConfig) -> bool:
     return False
 
 
-def is_or_uses_llm(config: Union[Dict[str, Any], ModelConfig]) -> bool:
+def config_uses_llm(config: Union[Dict[str, Any], ModelConfig]) -> bool:
     """Determine if a config uses an LLM.
 
     Args:
--- a/ludwig/utils/config_utils.py
+++ b/ludwig/utils/config_utils.py
@@ -1,7 +1,20 @@
-from typing import Set
+from typing import Any, Dict, Set, Union
 
 from ludwig.api_annotations import DeveloperAPI
-from ludwig.constants import DECODER, ENCODER, IMAGE, PREPROCESSING, SEQUENCE, TEXT, TIMESERIES, TYPE
+from ludwig.constants import (
+    DECODER,
+    ENCODER,
+    IMAGE,
+    INPUT_FEATURES,
+    MODEL_ECD,
+    MODEL_LLM,
+    MODEL_TYPE,
+    PREPROCESSING,
+    SEQUENCE,
+    TEXT,
+    TIMESERIES,
+    TYPE,
+)
 from ludwig.features.feature_registries import get_input_type_registry
 from ludwig.schema.model_config import ModelConfig
 from ludwig.types import FeatureConfigDict, FeatureTypeDefaultsDict, PreprocessingConfigDict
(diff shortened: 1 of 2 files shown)


ludwig/utils/config_utils.py
```diff
@@ -90,3 +103,42 @@ def has_pretrained_encoder(config: ModelConfig) -> bool:
...
+            uses_llm = True
+        elif INPUT_FEATURES in config:
+            for feature in config.get(INPUT_FEATURES, []):
+                if feature.get(ENCODER, {}).get(TYPE) == MODEL_LLM:
+                    uses_llm = True
+                    break
+        else:
+            raise ValueError(
+                "Invalid config cannot be checked for LLM usage because it has no input features." f"Config: {config}"
+            )
```
arnavgarg1, line 140: Do we think we need these ValueErrors here? I'd imagine that if the config has made it past config validation, then it is a valid config that should have both `input_features` and `output_features` in the dictionary or config object. What do you think?

tests/ludwig/utils/test_config_utils.py
```diff
@@ -57,3 +72,149 @@ def test_set_fixed_preprocessing_params_cache_embeddings(encoder_params: Dict[st
...
+def test_is_or_uses_llm(config, expectation, request):
+    """Test LLM detection on a variety of configs. Configs that use an LLM anywhere should return True, otherwise
+    False.
+
+    Args:
+        config: The name of the config fixture to test
+        expectation: The expected result
+        request: pytest `request` fixture
+    """
+    config = request.getfixturevalue(config)
```
arnavgarg1, line 209: @jeffkinnison Just curious - is there a reason you decided to create all of the tests as individual (and nested) fixtures as opposed to just creating each of them directly as parameterized tests?

ludwig/utils/config_utils.py
jeffkinnison: This is specifically for the dict path because there isn't any validation on dicts; config objects aren't subject to the same check. We could also just load dicts into `ModelConfig` here and run the validations, but that seems like an expensive operation just to check for the existence of LLM(s) in a config.

tests/ludwig/utils/test_config_utils.py
jeffkinnison: It was mostly for readability on my part, though it does also allow us to reuse the configs for both the dict and object cases.

ludwig/utils/config_utils.py
arnavgarg1: That makes sense! I guess my followup question is, when do we envision this utility function being called in Ludwig code?

tests/ludwig/utils/test_config_utils.py
```diff
@@ -57,3 +72,149 @@ def test_set_fixed_preprocessing_params_cache_embeddings(encoder_params: Dict[st
...
+    return ModelConfig.from_dict(ecd_config_dict_no_text_features)
+
+
+@pytest.fixture()
+def gbm_config_dict_llm_encoder() -> Dict[str, Any]:
+    return {
+        MODEL_TYPE: MODEL_GBM,
+        INPUT_FEATURES: [{TYPE: TEXT, NAME: "in1", ENCODER: {TYPE: "tf_idf"}}],
+        OUTPUT_FEATURES: [{TYPE: BINARY, NAME: "out1"}],
+    }
```
arnavgarg1, line 169: Is this supposed to have `encoder` type `llm`? Or is it just named this was for consistency with the other fixtures?

jeffkinnison: No, `tf_idf` is the only valid text encoder for GBM right now. This is in so that we have at least one GBM check.

jeffkinnison: Fixed the fixture name

tests/ludwig/utils/test_config_utils.py
arnavgarg1: Got it! Thanks for the changes with making them module scoped. I think that because they're relatively light weight fixtures, this is okay to do and shouldn't slow things down or cause memory related issues.

arnavgarg1 (approved):
Thanks for the utility function and comprehensive testing!

tests/ludwig/utils/test_config_utils.py
arnavgarg1: Last callout is that because the fixtures are dependent on each other, I think it may prevent pytest from parallelizing some of the tests. Again, it seems okay in this case given that the test itself is mostly an in place config or dict check and nothing too heavy, so each of the tests should run quickly.

PR #3871, opened 2024-01-09
Add batch size tuning for LLMs

Infernaught:
This PR extends Ludwig's batch size tuning functionality to LLMs.

For each batch size, we generate synthetic data in the following way:
We consider three values:
(1) The sum of the max_sequence_lengths of the input feature and the output feature
(2) The global_max_sequence_length
(3) The model's context length

If (1) is the smallest, then we generate synthetic inputs and outputs with the corresponding max_sequence_lengths.
If (2) is the smallest, then we generate synthetic inputs and outputs with length global_max_sequence_length/2 + 1.
If (3) is the smallest, then we generate synthetic inputs and outputs with length context_len/2 + 1.

### DIFF
--- a/ludwig/trainers/trainer_llm.py
+++ b/ludwig/trainers/trainer_llm.py
@@ -1,9 +1,9 @@
 import logging
 import os
 import time
-import torch
 from typing import Callable, Dict, List, Optional, Union
 
+import torch
 from torch.utils.tensorboard import SummaryWriter
 
 from ludwig.constants import MINIMUM_BATCH_SIZE, TEST, TRAINING, VALIDATION
@@ -472,11 +472,32 @@ class FineTuneTrainer(Trainer):
 
         progress_tracker.llm_eval_examples = llm_eval_examples
         return append_metrics(self.model, dataset_name, metrics, metrics_log, progress_tracker)
-    
-    def tune_batch_size(self, config: ModelConfigDict, training_set: Dataset, random_seed: int = default_random_seed, max_trials: int = 20, halving_limit: int = 3, snapshot_weights: bool = True, on_best_batch_size_updated: Optional[Callable[[int, float, int], None]] = None, tune_for_training: bool = True, global_max_sequence_length: Optional[int] = None) -> int:
+
+    def tune_batch_size(
+        self,
+        config: ModelConfigDict,
+        training_set: Dataset,
+        random_seed: int = default_random_seed,
+        max_trials: int = 20,
+        halving_limit: int = 3,
+        snapshot_weights: bool = True,
+        on_best_batch_size_updated: Optional[Callable[[int, float, int], None]] = None,
+        tune_for_training: bool = True,
+        global_max_sequence_length: Optional[int] = None,
+    ) -> int:
         if global_max_sequence_length is None:
             global_max_sequence_length = self.model.global_max_sequence_length
-        return super().tune_batch_size(config, training_set, random_seed, max_trials, halving_limit, snapshot_weights, on_best_batch_size_updated, tune_for_training, global_max_sequence_length)
+        return super().tune_batch_size(
+            config,
+            training_set,
+            random_seed,
+            max_trials,
+            halving_limit,
+            snapshot_weights,
+            on_best_batch_size_updated,
+            tune_for_training,
+            global_max_sequence_length,
+        )
 
     def _create_batch_size_evaluator(self) -> BatchSizeEvaluator:
         trainer = self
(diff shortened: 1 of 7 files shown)


ludwig/trainers/trainer_llm.py
Infernaught, line 1: This does not take into account the length of the prompt template yet.

ludwig/trainers/trainer_llm.py
```diff
@@ -471,6 +473,65 @@ def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tr
...
+                trainer.model.reset_metrics()
+                trainer.optimizer.zero_grad()
+
+            def step(self, batch_size: int, max_sequence_length: Optional[int] = None):
+                # max_sequence_length here is the smaller value between the global max sequence length of the model and the model's context length
+                trainer.distributed.set_batch_size(trainer.dist_model, batch_size)
+                input_feature_name, input_feature = trainer.model.input_features.items()[0]
+                output_feature_name, output_feature = trainer.model.output_features.items()[0]
+                input_msl = input_feature.input_shape[0]
+                output_msl = output_feature.output_shape[0]
```
justinxzhao, line 516: Is this the most reliable way to get the MSL? Or should we be looking up properties in the feature's preprocessing configuration?

ludwig/trainers/trainer_llm.py
```diff
@@ -471,6 +473,65 @@ def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tr
...
+        self,
+        config: ModelConfigDict,
+        training_set: Dataset,
+        random_seed: int = default_random_seed,
+        max_trials: int = 20,
+        halving_limit: int = 3,
+        snapshot_weights: bool = True,
+        on_best_batch_size_updated: Optional[Callable[[int, float, int], None]] = None,
+        tune_for_training: bool = True,
+        max_sequence_length: Optional[int] = None,
```
justinxzhao, line 486: nit: Rename to `global_max_sequence_length`?

ludwig/trainers/trainer_llm.py
justinxzhao: I believe you're in the clear here since the prompt template would be resolved at preprocessing time so by the time we've reached `trainer.train_step()`, the prompt template is already accounted for (within the input feature).

ludwig/trainers/trainer_llm.py
Infernaught: Based on what I see [here](https://github.com/ludwig-ai/ludwig/blob/56e42c0573612fc1781555043269cda13d6bd835/ludwig/features/text_feature.py#L274), it looks like at least the input_shape should provide a tighter upper bound than looking at the preprocessing configuration, but maybe output_shape needs to be changed.

justinxzhao: Ok! Can you leave a quick comment about this in the code?

ludwig/trainers/trainer_llm.py
```diff
@@ -503,37 +503,53 @@ def _create_batch_size_evaluator(self) -> BatchSizeEvaluator:
...
 
         class _TrainerBatchSizeEvaluator(BatchSizeEvaluator):
+            def __init__(self):
+                self.input_feature_name, self.input_feature = trainer.model.input_features.items()[0]
+                self.output_feature_name, self.output_feature = trainer.model.output_features.items()[0]
+
+                # Get the length of the longest input sequence from the training data
+                self.input_msl = self.input_feature.input_shape[0]
+                # Get the length of the longest output sequence from the training data
+                self.output_msl = self.output_feature.output_shape[0]
```
Infernaught, line 513: I don't know if this actually gets the longest output sequence. I checked the code for output_shape, and I saw that it actually generated outputs from the input sequence and didn't seem to use the targets provided.

Infernaught: nvm scratch this

ludwig/trainers/trainer_llm.py
```diff
@@ -471,6 +473,87 @@ def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tr
...
+
+            def reset(self):
+                trainer.model.reset_metrics()
+                trainer.optimizer.zero_grad()
+
+            def step(self, batch_size: int, global_max_sequence_length: int):
+                trainer.distributed.set_batch_size(trainer.dist_model, batch_size)
+
+                if not self.vocab_size:
+                    self.vocab_size = len(trainer.model.config_obj.input_features[0].encoder.vocab)
```
arnavgarg1, line 527: nit: Since the `trainer` object is available outside of the `_TrainerBatchSizeEvaluator` class, we don't actually need to add this if condition. Instead, in the constructor itself, we can just do

`self.vocab_size = len(trainer.model.config_obj.input_features[0].encoder.vocab)`

ludwig/trainers/trainer_llm.py
```diff
@@ -471,6 +473,87 @@ def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tr
...
+
+                # max_sequence_length here is the smaller value between the global max sequence length of the model
+                # and the model's context length
+                if trainer.model.config_obj.output_features[0].preprocessing.max_sequence_length:
+                    self.output_msl = trainer.model.config_obj.output_features[0].preprocessing.max_sequence_length
```
arnavgarg1, line 532: I think this can also move into the constructor, something like this:

```python
self.output_msl = self.output_feature.output_shape[0]
if trainer.model.config_obj.output_features[0].preprocessing.max_sequence_length:
    self.output_msl = trainer.model.config_obj.output_features[0].preprocessing.max_sequence_length
```

ludwig/trainers/trainer_llm.py
```diff
@@ -471,6 +473,87 @@ def evaluation(self, dataset, dataset_name, metrics_log, batch_size, progress_tr
...
+
+                if self.input_msl + self.output_msl > global_max_sequence_length:
+                    # In this case, we just need to make sure that the length of the synthetic data exceeds
+                    # max_sequence_length by at most a small amount
```
arnavgarg1, line 536: What is the reason for this? Is it so that we leave some headroom just in case?

arnavgarg1: Perhaps another way of writing this could be:

```python
input_msl = self.input_msl
output_msl = self.output_msl
if self.input_msl + self.output_msl > global_max_sequence_length:
    # In this case, we just need to make sure that the length of the synthetic data exceeds
    # max_sequence_length by at most a small amount
    input_msl = global_max_sequence_length // 2 + 1
    output_msl = global_max_sequence_length // 2 + 1
```

ludwig/utils/batch_size_tuner.py
```diff
@@ -51,7 +52,9 @@ def _is_valid_batch_size(batch_size):
             gc.collect()
 
             try:
-                samples_per_sec = self.evaluate(batch_size, total_steps=5)
+                samples_per_sec = self.evaluate(
+                    batch_size, total_steps=5, global_max_sequence_length=global_max_sequence_length
```
justinxzhao, line 56: nit: Make this a constant in the file.

justinxzhao (approved):

ludwig/utils/batch_size_tuner.py
Infernaught: I'm not quite sure I understand. Why should `samples_per_sec` be a constant?

tests/integration_tests/test_llm.py
```diff
@@ -1254,6 +1255,39 @@ def test_llm_encoding(llm_encoder_config, adapter, quantization, tmpdir):
...
+    model_type: llm
+    input_features:
+        - name: instruction
+          type: text
+    output_features:
+        - name: output
+          type: text
+    prompt:
+        template: >-
+            {instruction}
```
arnavgarg1, line 1271: @Infernaught Seeing the same test twice? on line 1258 and line 1348

ludwig/utils/batch_size_tuner.py
Infernaught: Ohh I see

tests/integration_tests/test_llm.py
arnavgarg1: I think pre-commit is also complaining about this

Infernaught: Interesting. Probably an issue with merging?

arnavgarg1 (approved):

PR #3872, opened 2024-01-10
[Maintenance] Fix docker images by pinning ray==2.3.1, daft==0.1.20, unpinning proto, and using torch 2.1.1.

justinxzhao:


arnavgarg1:
Nice @justinxzhao! Looks like most of this is passing - we may need to just be less strict with the minimum Ray version. I'm currently trying to build an image with it - if it works, I can push that change up!

requirements_distributed.txt
```diff
@@ -12,4 +12,4 @@ awscli
 deepspeed!=0.11.0  # https://github.com/microsoft/DeepSpeed/issues/4473
 
 # requirements for daft
-getdaft[ray]
+getdaft[ray]==0.1.20
```
arnavgarg1, line 15: @justinxzhao OOC, did you try pinning this to <2? Or does it specifically need getdaft 0.1.20? For reference, in the main requirements.txt file, I have this pinned to <2

justinxzhao: I did try to leave this unpinned. What I observed was that the 8th step of the docker build slowed down considerably since for some reason pip seems to try versions 0.1.18 down through 0.1.13 (over 50 minutes and counting). It's possible that it would get to 0.1.20 eventually, but I instead opted to restart with a pinned version, which brought the 8th step of the docker build time back down to 20 minutes.

arnavgarg1: Gotcha, good to know! In that case, I think it should be safe to leave this pinned. Perhaps, we can also pin it to 0.1.20 in the main `requirements.txt` file for consistency, but we can do that in a follow-up PR!

arnavgarg1 (approved):
LGTM and thanks for the fix! I think it should be fine to use Ray 2.3.1 for our entire testing suite given that there are many newer Ray releases and we are pretty far behind.

requirements_distributed.txt
justinxzhao: Done.

PR #3875, opened 2024-01-10
[BUGFIX] Guard against UnicodeEncodeError when saving validation results in Google Colab environment

alexsherstinsky:
### Scope
If the textual data in the validation dataset contains non-ASCII characters, saving the results to storage may encounter errors, such as:
```
UnicodeEncodeError: 'ascii' codec can't encode character '\u2260' in position 2699: ordinal not in range(128)
```

The fix consists of defensively converting the contents to the `UTF-8` format.

### Remark

This error was not observed in isolated GPU environments such as while using EC2 outside of Colab, but this change assure stable behavior in all environments.


arnavgarg1 (approved):

justinxzhao:
It's probably worth specifying the encoding everywhere we are writing files to disk!

PR #3878, opened 2024-01-11
Add streaming support for zero shot inference

arnavgarg1:
This PR introduces a new boolean flag called `streaming` to the LudwigModel.generate() API which allows users to see streaming output when performing zero-shot inference on single or multiple samples. 

## Demo

https://github.com/ludwig-ai/ludwig/assets/106701836/50fc01db-837b-4689-8e8a-8da491481da1

## Config to reproduce the demo video

```python
import yaml
import logging

config = yaml.safe_load(
    """
model_type: llm
base_model: meta-llama/Llama-2-7b-chat-hf

quantization:
  bits: 4

input_features:
  - name: instruction
    type: text

output_features:
  - name: output
    type: text

generation:
  max_new_tokens: 64
  temperature: 0.1

trainer:
    type: none

backend:
  type: local
"""
)

model = LudwigModel(config, logging_level=logging.INFO)

# Single sample - Normal generation
output = model.generate("What is the meaning of life?", generation_config={"max_new_tokens": 32})

# Single sample - Streaming generation
output = model.generate("What is the meaning of life?", generation_config={"max_new_tokens": 32}, streaming=True)

# Multi sample - Normal generation
output = model.generate(["What is the meaning of life?", "What is the weather like in Germany around December?"], generation_config={"max_new_tokens": 64})

# Multi sample - Streaming generation
output = model.generate(["What is the meaning of life?", "What is the weather like in Germany around December?"], generation_config={"max_new_tokens": 64}, streaming=True)
```

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -1037,6 +1037,9 @@ class LudwigModel:
         for i in range(len(input_ids)):
             with torch.no_grad():
                 logger.info(f"Input: {input_strings[i]}\n")
+                # NOTE: self.model.model.generation_config is not used here because it is the default
+                # generation config that the CausalLM was initialized with, rather than the one set within the
+                # context manager.
                 generated_output = self.model.model.generate(
                     input_ids=input_ids[i].unsqueeze(0),
                     attention_mask=attention_mask[i].unsqueeze(0),
@@ -1066,6 +1069,9 @@ class LudwigModel:
             torch.Tensor: Tensor of generated outputs.
         """
         with torch.no_grad():
+            # NOTE: self.model.model.generation_config is not used here because it is the default
+            # generation config that the CausalLM was initialized with, rather than the one set within the
+            # context manager.
             return self.model.model.generate(
                 input_ids=input_ids,
                 attention_mask=attention_mask,
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -1002,12 +1002,12 @@ class LudwigModel:
             tokenized_inputs = tokenizer.tokenizer(input_strings, return_tensors="pt", padding=True)
             input_ids = tokenized_inputs["input_ids"].to("cuda")
             attention_mask = tokenized_inputs["attention_mask"].to("cuda")
-            streamer = create_text_streamer(tokenizer.tokenizer) if streaming else None
 
-            if streamer:
+            if streaming:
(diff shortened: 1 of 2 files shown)


justinxzhao (commented):
Very cool!

ludwig/api.py
```diff
@@ -983,31 +994,90 @@ def generate(
...
+            logger.info(f"Finished generating in: {(time.time() - start_time):.2f}s.")
+
+            return decoded_outputs[0] if len(decoded_outputs) == 1 else decoded_outputs
+
+    def _generate_streaming_outputs(
+        self,
+        input_strings: Union[str, List[str]],
+        input_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        streamer: Union[TextStreamer, None],
```
alexsherstinsky, line 1022: Very minor -- if we are already calling this method, then `streamer` should not be `None`...  Is this right?

justinxzhao (approved):
LGTM

ludwig/api.py
```diff
@@ -983,31 +994,90 @@ def generate(
...
+                logger.info("----------------------")
+                outputs.append(generated_output)
+        return torch.cat(outputs, dim=0)
+
+    def _generate_non_streaming_outputs(
+        self,
+        _input_strings: Union[str, List[str]],
+        input_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+        streamer: Union[TextStreamer, None],
```
alexsherstinsky, line 1058: @arnavgarg1 So here, we would not include the `streamer`, because we know that it is `None` -- and thus pass `None` to `self.model.generate()`?  Is this reasonable?

alexsherstinsky (changes requested):
@arnavgarg1 This is great -- I just made some minor comments (please let me know whether or not they will help).  Thank you!

arnavgarg1:
@alexsherstinsky I think I had a poor interface defined! Can you take a look now and see if it makes sense and your comments are addressed?

ludwig/api.py
arnavgarg1: It is reasonable since it defaults to None by default in the generate() call in the model API, but I cleaned this up so that we only pass in the argument into generate when the streamer is actually present. Let me know if that helps!

alexsherstinsky (approved):
LGTM -- so nice!

PR #3882, opened 2024-01-15
Set use_reentrant to False for gradient checkpointing

arnavgarg1:
Fixes the following warnings in newer versions of Torch (>= 2.1.0)

```
/home/ray/anaconda3/lib/python3.8/site-packages/torch/utils/checkpoint.py:429: UserWarning: torch.utils.checkpoint:
 please pass in use_reentrant=True or use_reentrant=False explicitly. The default value of use_reentrant will be 
updated to be False in the future. To maintain current behavior, pass use_reentrant=True. It is recommended that 
you use use_reentrant=False. Refer to docs for more details on the differences between the two variants.
```

and 

```
`use_cache=True` is incompatible with gradient checkpointing. Setting `use_cache=False`...
```

### DIFF
--- a/ludwig/trainers/trainer.py
+++ b/ludwig/trainers/trainer.py
@@ -28,6 +28,7 @@ import time
 from typing import Callable, Dict, List, Optional, Tuple
 
 import numpy as np
+import packaging
 import pandas as pd
 import psutil
 import torch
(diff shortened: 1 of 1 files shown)


alexsherstinsky (approved):
LGTM!  Thank you so much for the solution for this ever-present warning!

PR #3896, opened 2024-01-18
Enable AdaLoRA tests for LLM adapter

jeffkinnison:
This adds adalora to the list of adapters available to `LLMEncoder` and adds adalora to the existing test suite.


arnavgarg1 (approved):
LGTM! Seems like there's just some failing combinatorial tests that may need to be fixed

arnavgarg1:
@jeffkinnison Actually I think it was a transient failure, so should be good to merge, especially given that your tests don't touch either module and you only added new tests.

PR #3904, opened 2024-01-21
Improve ludwig feature dict

dennisrall:
Some minor improvements to the `LudwigFeatureDict`:
- remove strange `__next__` method
- remove `internal_key_to_original_name_map ` dict as the values were never read
- use `MutableMapping` from the `collections.abc` &rarr; get useful and more performant methods
- added tests for the new functionality and split up the old test (I kept the old test to show it is still running, but I think it can be removed?)

### DIFF
--- a/ludwig/features/feature_utils.py
+++ b/ludwig/features/feature_utils.py
@@ -14,7 +14,8 @@
 # limitations under the License.
 # ==============================================================================
 import re
-from typing import Dict, Iterator, List, Optional, Tuple, Union
+from collections.abc import MutableMapping
+from typing import Iterator, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
@@ -157,7 +158,7 @@ def get_name_from_module_dict_key(key: str, feature_name_suffix_length: int = FE
     return name[:-feature_name_suffix_length]
 
 
-class LudwigFeatureDict(torch.nn.Module):
+class LudwigFeatureDict(torch.nn.Module, MutableMapping):
     """Torch ModuleDict wrapper that permits keys with any name.
 
     Torch's ModuleDict implementation doesn't allow certain keys to be used if they conflict with existing class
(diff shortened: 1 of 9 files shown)


ludwig/models/llm.py
```diff
@@ -65,13 +65,13 @@ def __iter__(self) -> None:
         return iter(self.obj.keys())
 
     def keys(self) -> List[str]:
-        return self.obj.keys()
+        return self.obj.key_list()
```
alexsherstinsky, line 68: @dennisrall Do you think it might be simpler to retain the coding pattern of returning `list(my_object.keys())` or `list(my_object.items())`, etc., instead of introducing a special method `i.tem_list()`?  It seems to me that doing so will be consistent with other cases in these Ludwig modules as well as general Python collection patterns.  Thank you.

ludwig/features/feature_utils.py
```diff
@@ -157,7 +158,7 @@ def get_name_from_module_dict_key(key: str, feature_name_suffix_length: int = FE
     return name[:-feature_name_suffix_length]
 
 
-class LudwigFeatureDict(torch.nn.Module):
+class LudwigFeatureDict(torch.nn.Module, MutableMapping):
```
alexsherstinsky, line 161: @dennisrall Thank you for incorporating my previous suggestion -- I think that part looks clean now.  Thank you for the idea and the implementation!

For this one, I am not sure if the benefits due to adding the MutableMapping subclassing justify taking the risk brought about the multiple inheritance.  Do the test cover all the eventualities that might happen with this change?

Thank you. /cc @justinxzhao @arnavgarg1 @Infernaught

tests/ludwig/features/test_feature_utils.py
```diff
@@ -5,6 +5,111 @@
 from ludwig.features import feature_utils
 
 
+@pytest.fixture
+def to_module() -> torch.nn.Module:
+    return torch.nn.Module()
+
+
+@pytest.fixture
+def type_module() -> torch.nn.Module:
```
alexsherstinsky, line 14: @dennisrall This fixture and the `to_module()` one appear the same, and just instantiate the `PyTorch` `Module()` class.  Without a docstring, it is a bit difficult to justify it.  Thank you.

ludwig/features/feature_utils.py
dennisrall: No problem, I was just playing around a bit.

I can't think of any problems about the multiple inheritance, but you know the code better than me😉

It is also possible to remove the `MutableMapping` inheritance and implement the other methods by hand. But I think this way it is a bit cleaner, if it doesn't cause any problems...

tests/ludwig/features/test_feature_utils.py
dennisrall: Thanks for your feedback. I added a docstring for both fixtures and also for the `hash` method of the `LudwigFeatureDict`

justinxzhao (approved):
Overall PR LGTM, thanks for the change @dennisrall.

I pushed up a couple of commits, largely around fixing the CI.

It appears we'll need to bump up integration tests CI to torch==2.1.0, which we're absolutely ok also since that's what our Docker images are using now.

mhabedank:
LGTM can be merged as soon as possible

w4nderlust:
Thank you @dennis-rall for the LudwigFeatureDict improvements! This was approved by multiple maintainers (justinxzhao, mhabedank) and the changes are genuinely valuable — making `LudwigFeatureDict` a proper `MutableMapping`, removing the incorrect `__next__` method, and eliminating the redundant `internal_key_to_original_name_map`.

Unfortunately, the PR no longer applies cleanly due to significant codebase changes (GBM/LightGBM removal, import modernization). We plan to port the core idea manually in a future PR, as the improvements are still relevant. The key changes needed are in `feature_utils.py`, `ecd.py`, `llm.py`, and the explainer files. Thanks for the excellent contribution — credit will go to you when we implement this!

Issue #3910, opened 2024-01-23
Remove target_module LoRA mapping for Phi-2 model


arnavgarg1:
The current PEFT LoRA modules map to the wrong target modules for Phi-2 since the architecture was recently updated.

I created a PR on the PEFT repo to update this to new target modules: https://github.com/huggingface/peft/pull/1375 

For now, we will fall back to target_modules for Phi-2 that lead to good model performance.

PR #3911, opened 2024-01-23
Add default LoRA target modules for Phi-2

arnavgarg1:
These modules maximize performance, also the current LoRA target modules for Phi-2 are wrong on the PEFT repo.

See #3910

### DIFF
--- a/ludwig/schema/model_types/utils.py
+++ b/ludwig/schema/model_types/utils.py
@@ -315,7 +315,13 @@ def set_llm_parameters(config: "ModelConfig") -> None:
 
     # HACK(Arnav): Set Mixtral target modules when using LoRA
     # GitHub issue: https://github.com/ludwig-ai/ludwig/issues/3853
+    # PEFT PR: https://github.com/huggingface/peft/pull/1376
     _set_mixtral_target_modules(config)
+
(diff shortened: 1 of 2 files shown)


ludwig/schema/model_types/utils.py
```diff
@@ -421,9 +427,35 @@ def _set_mixtral_target_modules(config: "ModelConfig") -> None:
...
+def _set_phi2_target_modules(config: "ModelConfig") -> None:
+    """If the base model is Phi-2, LoRA is enabled and the target modules are not set, set the target modules to
+    maximize performance."""
+    if config.base_model not in {
+        "susnato/phi-1_dev",
+        "susnato/phi-1_5_dev",
+        "susnato/phi-2",
+        "microsoft/phi-1",
+        "microsoft/phi-1_5",
+        "microsoft/phi-2",
```
justinxzhao, line 445: I feel like I've seen this list in other places.

Is it worth it to make this in to a shared map and do `.keys()`?

arnavgarg1: Yeah I can do that

justinxzhao (approved):

PR #3924, opened 2024-02-05
Add example for base model dequantization/upscaling

arnavgarg1:


examples/llm_base_model_dequantization/phi_2_dequantization.py
```diff
@@ -0,0 +1,43 @@
...
+
+import yaml
+
+from ludwig.api import LudwigModel
+from ludwig.utils.hf_utils import upload_folder_to_hfhub
+
+base_model_name = "microsoft/phi-2"
+dequantized_path = "microsoft-phi-2-dequantized"
+save_path = "/home/ray/" + dequantized_path
+hfhub_repo_id = "arnavgrg/" + dequantized_path
```
Infernaught, line 11: Nonblocking: 
Maybe we should use a placeholder instead of "arnavgrg/". Maybe something like:
`{HF_USER}/` + dequantized_path`

Infernaught: Otherwise, LGTM.

arnavgarg1: Agreed! Will update

examples/llm_base_model_dequantization/phi_2_dequantization.py
```diff
@@ -0,0 +1,45 @@
+import logging
+
+import yaml
+from huggingface_hub import whoami
+
+from ludwig.api import LudwigModel
+from ludwig.utils.hf_utils import upload_folder_to_hfhub
+
+hf_username = whoami().get("name")
```
Infernaught, line 9: I did not know that this was a thing. If this works like it's supposed to, then this is sick!

Infernaught (approved):

PR #3930, opened 2024-02-12
Rename the argument of dense encoder from layers to fc_layers

metamath1:
# Code Pull Requests

Please provide the following:

- a clear explanation of what your code does: fc_layers setting not working for the dense encoder for the input feature. To fix this, layers in the DenseEncoder class has been renamed to fc_layers.
- if applicable, a reference to an issue: I posted a question on Slack, but haven't received an answer yet. https://ludwig-ai.slack.com/archives/C01PN6M2RSM/p1707631595275339
- a reproducible test for your PR (code, config and data sample): 
https://gist.github.com/metamath1/0dc3893e11a887cdd2544372d9e46852

### DIFF
--- a/ludwig/encoders/generic_encoders.py
+++ b/ludwig/encoders/generic_encoders.py
@@ -65,7 +65,7 @@ class DenseEncoder(Encoder):
     def __init__(
         self,
         input_size,
-        layers=None,
+        fc_layers=None,
         num_layers=1,
         output_size=256,
         use_bias=True,
@@ -87,7 +87,7 @@ class DenseEncoder(Encoder):
         logger.debug("  FCStack")
         self.fc_stack = FCStack(
             first_layer_input_size=input_size,
-            layers=layers,
+            layers=fc_layers,
             num_layers=num_layers,
             default_output_size=output_size,
             default_use_bias=use_bias,


ludwig/encoders/generic_encoders.py
```diff
@@ -65,7 +65,7 @@ class DenseEncoder(Encoder):
     def __init__(
         self,
         input_size,
-        layers=None,
+        fc_layers=None,
```
alexsherstinsky, line 68: @metamath1 Thank you for proposing this change.  Do you think it would potentially cause backward compatibility issues if other users already passed `layers` to the constructor?  Thank you. /cc @arnavgarg1

alexsherstinsky: 🙇

arnavgarg1: @metamath1 Actually, I am a bit confused about this PR - what's the intended scope and reason for the change? 

I misunderstood the change - this may not require any backward incompatibility changes since you're modifying the argument in the encoder class as opposed to the schema directly.

metamath1: > @alexsherstinsky Yes, that is right.
> 
> We have a file to add support for backward compatibility here: https://github.com/ludwig-ai/ludwig/blob/ea890d99e72fbc32fad8aae18070da9969ddb422/ludwig/utils/backward_compatibility.py
> 
> And an associated test file https://github.com/ludwig-ai/ludwig/blob/ea890d99e72fbc32fad8aae18070da9969ddb422/tests/ludwig/utils/test_backward_compatibility.py to add a test to make sure it works.
> 
> @metamath1 Are you able to add backward compatibility support and an associated test for it?

Here's the setup for creating multiple Linear layers in a dense encoder

```
config = {
    'preprocessing': {
        'split': {
            'type': 'fixed',
            'column': 'split'
        }
    },
    'input_features': [
        {
            'name': 'image',
            'type': 'vector',
            'encoder': {
                'type': 'dense',
                # This part doesn't work.
		# This part is ignored and always creates a layer with an output size of 256.
		'fc_layers': [
                    {'output_size': 64},
                    {'output_size': 32}
                ]
            }
        }
    ],
    'output_features': [
        {'name': 'label', 'type': 'category'}
    ],
    'trainer': {'epochs': 1, 'batch_size': 64, 'eval_batch_size': 128}
}
```

However, the `fc_layers` setting does not work.
This is because the argument to the DenseEncoder is named `layers`, 
but the config dictionary object stores the setting as `fc_layers`.
The settings in the user's `fc_layers` are not being passed to the DenseEncoder.

w4nderlust: This really needs to be fixed.

For retrocompatibility we can have a couple lines fo code that internally rename `layers` to `fc_layers` and warn the user of the change and that it is deprecated.

Also, if it was not working anyway, there's not a lot of damage in beckward incompatibility.

Finally, I'm surprised by the fact that the wrong keyword was not captured by either the config validar or the dereferencing of the kwargs, the should have been a check somewhere.

mhabedank:
Seems to be important, need merge

w4nderlust:
Thank you @metamath1! The fc_layers fix has been merged into main as part of PR #4061. This was an important bug fix — the DenseEncoder was silently ignoring user fc_layers configuration. Appreciate the contribution!

PR #3934, opened 2024-02-21
Update ludwig version to 0.9.4

alexsherstinsky:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/globals.py
+++ b/ludwig/globals.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
-LUDWIG_VERSION = "0.9.dev"
+LUDWIG_VERSION = "0.9.4"
 
 MODEL_WEIGHTS_FILE_NAME = "model_weights"
 MODEL_HYPERPARAMETERS_FILE_NAME = "model_hyperparameters.json"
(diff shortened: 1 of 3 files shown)


arnavgarg1 (changes requested):
Can we please decouple these changes into two separate PRs?

ludwig/globals.py
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
-LUDWIG_VERSION = "0.9.dev"
+LUDWIG_VERSION = "0.9.4"
```
arnavgarg1, line 17: We want to make this Ludwig 0.10 right?

tests/integration_tests/test_llm.py
```diff
@@ -531,16 +531,18 @@ def _verify_lm_lora_finetuning_layers(
...
-            "adaption_prompt",
-            {},
-            id="adaption_prompt-defaults",
-        ),
-        pytest.param(
-            "adaption_prompt",
-            {"adapter_len": 6, "adapter_layers": 1},
-            id="adaption_prompt-modified-defaults",
-        ),
+        # TODO: <Alex>02/21/2024: Disabling AdaptionPrompt (waiting for PEFT release to fix
```
arnavgarg1, line 534: I think in addition to commenting out the tests, we should also unregister the `adaption_prompt` adapter.

PR #3947, opened 2024-02-28
[MAINTENANCE] Update Ludwig development version.

alexsherstinsky:
# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/globals.py
+++ b/ludwig/globals.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
-LUDWIG_VERSION = "0.10.dev"
+LUDWIG_VERSION = "0.10.0.dev"
 
 MODEL_WEIGHTS_FILE_NAME = "model_weights"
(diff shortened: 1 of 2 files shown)


ludwig/globals.py
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 
-LUDWIG_VERSION = "0.10.dev"
+LUDWIG_VERSION = "0.10.1.dev"
```
arnavgarg1, line 17: Should this be 0.10.2.dev since we're developing 0.10.2?

alexsherstinsky: @arnavgarg1 This is a very good question.  To me, the future version is not know (i.e., we could be developing 1.0).  Hence, my approach was in the style of "current_version_+" -- meaning that "what we are doing on top of the currently shipping version".  What do you think?  Thank you.

arnavgarg1: Sounds good to me!

arnavgarg1 (approved):

PR #3957, opened 2024-03-07
Add support for eval batch size tuning for LLMs on local backend

arnavgarg1:
One can now set `eval_batch_size` to `auto` for LLMs when using the local backend during fine-tuning. This may help select a larger eval batch size to help reduce time spent during evaluation.


geoffreyangus (approved):

Infernaught:
Very cool!

PR #3959, opened 2024-03-08
[BUGFIX] Fixing integration test failures.

alexsherstinsky:
### Scope
We have had consistent failures from two tests involving image processing.  Through troubleshooting, the cause was determined to be the running out of storage space on the test host Linux machine in GitHub actions.  The fixes are:
1. Improve error handling in Carton utilities module.
2. Add GitHub Action steps to clean out unneeded storage space.

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests


.github/workflows/pytest.yml
```diff
@@ -231,10 +231,10 @@ jobs:
     timeout-minutes: 90
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: "3.10"
```
arnavgarg1, line 237: Is the python version bump strictly necessary?

alexsherstinsky: @arnavgarg1 It is not; I just felt that testing on the latest -- if all tests pass -- would be better.  What do you think?

justinxzhao (commented):
This is a lovely change -- thank you!

justinxzhao (approved):

.github/workflows/pytest.yml
arnavgarg1: Fair enough!

arnavgarg1 (approved):

ludwig/utils/carton_utils.py
```diff
@@ -108,26 +110,40 @@ def export_carton(model: LudwigModel, carton_path: str, carton_model_name="ludwi
...
+                    required_framework_version="=2",
+                    model_name=carton_model_name,
+                    inputs=_get_input_spec(model),
+                    outputs=_get_output_spec(model),
+                )
+            except Exception as e:
+                exception_message: str = 'An Exception inside "pack()" occurred.\n'
+                exception_traceback: str = traceback.format_exc()
+                exception_message += f'{type(e).__name__}: "{str(e)}".  Traceback: "{exception_traceback}".'
+                sys.stderr.write(exception_message)
```
arnavgarg1, line 135: Is it possible to use the logging directory here? Or do we actually need to write to sys stderr when we're raising it through the ValueError anyway?

alexsherstinsky: @arnavgarg1 I was thinking of showing these messages on the screen, so that in GitHub log review, we can see exactly what went wrong.  We could create a directory, but that might hide the failure error messages.  It was very challenging to troubleshoot even this one, because the code is wrapping Rust.  Happy to discuss, of course.  Thanks!

PR #3969, opened 2024-03-18
enh: enable loading model weights from training checkpoint

geoffreyangus:
Title. Should enable users to use models given a pytorch training checkpoint instead of just finalized models. Useful if jobs error midway through training.

### DIFF
--- a/ludwig/api.py
+++ b/ludwig/api.py
@@ -66,6 +66,7 @@ from ludwig.globals import (
     LUDWIG_VERSION,
     MODEL_HYPERPARAMETERS_FILE_NAME,
     set_disable_progressbar,
+    TRAINING_CHECKPOINTS_DIR_PATH,
     TRAIN_SET_METADATA_FILE_NAME,
 )
 from ludwig.models.base import BaseModel
@@ -1282,9 +1283,12 @@ class LudwigModel:
                     self.model.output_features, predictions, dataset, training_set_metadata
                 )
                 eval_stats = {
-                    of_name: {**eval_stats[of_name], **overall_stats[of_name]}
-                    # account for presence of 'combined' key
-                    if of_name in overall_stats else {**eval_stats[of_name]}
+                    of_name: (
+                        {**eval_stats[of_name], **overall_stats[of_name]}
+                        # account for presence of 'combined' key
+                        if of_name in overall_stats
+                        else {**eval_stats[of_name]}
+                    )
                     for of_name in eval_stats
                 }
 
@@ -1765,6 +1769,7 @@ class LudwigModel:
         gpu_memory_limit: Optional[float] = None,
         allow_parallel_threads: bool = True,
         callbacks: List[Callback] = None,
+        from_checkpoint: bool = False,
     ) -> "LudwigModel":  # return is an instance of ludwig.api.LudwigModel class
         """This function allows for loading pretrained models.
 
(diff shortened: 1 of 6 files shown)


Infernaught (approved):
LGTM, but why do we set this to False by default? Is there a reason why this shouldn't be the default behavior?

geoffreyangus:
@Infernaught it's an interesting point– we want it False by default just because that preserves the existing user behavior. We can consider changing that if it is an overall better experience, but for now don't want to introduce confusion.

ludwig/api.py
```diff
@@ -1788,6 +1793,9 @@ def load(
         :param callbacks: (list, default: `None`) a list of
             `ludwig.callbacks.Callback` objects that provide hooks into the
             Ludwig pipeline.
+        :param from_checkpoint: (bool, default: `False`) if `True`, the model
+            will be loaded from the latest checkpoint (training_checkpoints/)
+            instead of the final model weights.
```
arnavgarg1, line 1798: Yeah I guess this is a fair thing to do, wondering though what the reason would be to load from_checkpoint if `model/model_weights` is already present? Perhaps a no-op in that case and we can always make this True? Okay keeping it like this for now as well, just wanted to call it out

geoffreyangus: Yeah– I think it's okay to have it be explicit and not make assumptions on behalf of the user for now. 

We can revisit this if people get confused, but because it's default False, I'm hopeful this won't interrupt anyone's experience (until of course someone really needs it, at which point we can direct them to this flag)

arnavgarg1 (approved):

ludwig/api.py
```diff
@@ -1868,7 +1880,16 @@ def load_weights(
         ```
         """
-            self.model.load(model_dir)
+            if from_checkpoint:
+                with self.backend.create_trainer(
```
alexsherstinsky, line 1884: 💡

ludwig/trainers/trainer.py
```diff
@@ -821,6 +821,11 @@ def save_checkpoint(self, progress_tracker: ProgressTracker, save_path: str, che
         # Callback that the checkpoint was reached, regardless of whether the model was evaluated.
         self.callback(lambda c: c.on_checkpoint(self, progress_tracker))
 
+    def create_checkpoint_handle(self):
```
alexsherstinsky, line 824: 💡

alexsherstinsky (approved):
@geoffreyangus ✅ -- code LGTM (and it was illuminating -- thanks!) -- I was just wondering, how did/do we test something like this, where we called `distributed`?  Thank you!

justinxzhao (approved):

geoffreyangus:
I added a test for this @alexsherstinsky, if you want to take a look!

tests/integration_tests/test_model_save_and_load.py
```diff
@@ -32,6 +32,79 @@
 )
 
 
+def test_model_load_from_checkpoint(tmpdir, csv_filename, tmp_path):
```
alexsherstinsky, line 35: ❤️ @geoffreyangus Thank you -- so cool!  For my edification, which checkpoint are we comparing?  It looks like the loaded model from storage (`ludwig_model_2`) is the result of training `ludwig_model_1` all the way through the `1` epoch -- as opposed to some intermediate checkpoint (e.g., saved after a few steps).  Is this correct?  Thank you!

geoffreyangus: yeah that is correct– it's just the latest checkpoint deposited into training_checkpoints/ during the course of training. In the 1 epoch case, it is equivalent to the model weights at the end of training.

For some reason on my local, the models weren't equivalent after 2 epochs– my hunch is that this is because there was a difference between the "best" checkpoint (loaded at the end of training) and the "latest" checkpoint

alexsherstinsky: 🙇

PR #3977, opened 2024-03-22
Fix for 'upload_to_hf_hub()' path mismatch with 'save()'

sanjaydasgupta:
The code in this PR provides a solution for this issue: [upload_to_hf_hub model path mismatch with model.save](https://github.com/ludwig-ai/ludwig/issues/3925) while also remaining backward compatible with the previous behavior of [upload_to_hf_hub](https://ludwig.ai/latest/user_guide/api/LudwigModel/#upload_to_hf_hub).

With this change, `upload_to_hf_hub` can accept any of the following kinds of path as its `model_path` argument:

1. a path to the `experiment` folder (existing implementation).
2. a path to the `model` folder (added by this PR).

The names `experiment` and `model` refer to places in the folder hierarchy created when a model is trained:

`results / experiment / model / model_weights`

The `model` folder is additionally distinguished by containing the `model_hyperparameters.json` file. The `model_weights` folder contains one or more of these files: `pytorch_model.bin`, `adapter_model.bin`, `adapter_model.safetensors`.

The changes to `upload_to_hf_hub()` cause it to test the `model_path` argument to determine which kind of path was passed. <ins>If the path is a `model` folder, then the parent path (the corresponding `experiment` path) is obtained and used instead. So the logic of `upload_to_hf_hub` and its supporting functions do not have to change at all.</ins> ~It uses that knowledge for its own purposes and also passes it down to all supporting functions (as an additional argument `model_path_is_experiment_path: bool = True`) for their use. The default value of `model_path_is_experiment_path` has been chosen to ensure the backward compatible behavior.~

~A unit test has been added (`test_upload_to_hf_hub__validate_upload_parameters2()`) for the new use-case.~ And the feature has been integration tested with custom code on a local build.


sanjaydasgupta:
Hi @alexsherstinsky, please take a look. Thanks.

ludwig/api.py
```diff
@@ -1938,39 +1939,53 @@ def upload_to_hf_hub(
...
             `f"Upload {path_in_repo} with huggingface_hub"`
-        :param commit_description (`str` *optional*):
+        :param commit_description: (`str` *optional*)
             The description of the generated commit
 
         # Returns
 
         :return: (bool) True for success, False for failure.
         """
+        if os.path.exists(os.path.join(model_path, "model", "model_weights")) and os.path.exists(
```
alexsherstinsky, line 1965: @sanjaydasgupta The idea/direction seems great!  I just have a couple of minor suggestions for you to consider (because if they work, they might improve the code quality slightly):
1) Can we create the constants for the two directory names, `"model"` and `"model_weights"` and replace codebase-wise?
2) Is it possible instead of passing the boolean `model_path_is_experiment_path`, determine the "experiments" directory and standardize all methods on using it -- so the "experiment" directory would be passed to them.  I think that even if a function is passed the "model" directory, and in those cases we have to make the assumption that "experiments" is "model/..", this would still be reasonable and acceptable.  What do you think?

Thank you very much!

ludwig/api.py
```diff
@@ -1938,39 +1939,53 @@ def upload_to_hf_hub(
...
+            os.path.join(model_path, "model", MODEL_HYPERPARAMETERS_FILE_NAME)
+        ):
+            model_path_is_experiment_path = True
```
alexsherstinsky, line 1968: If we cannot implement my suggestion above, then we should declare `model_path_is_experiment_path: bool` at the top.  We can even initialize it to `False` as in `model_path_is_experiment_path: bool = False` (or `True` depending on how you prefer your logic to flow).  Thank you!

ludwig/api.py
sanjaydasgupta: @alexsherstinsky, responding to your (2) above - sounds like a great idea! Let me take a closer look at the code.

sanjaydasgupta: > @sanjaydasgupta The idea/direction seems great! I just have a couple of minor suggestions for you to consider (because if they work, they might improve the code quality slightly):
> 
> 1. Can we create the constants for the two directory names, `"model"` and `"model_weights"` and replace codebase-wise?
> 2. Is it possible instead of passing the boolean `model_path_is_experiment_path`, determine the "experiments" directory and standardize all methods on using it -- so the "experiment" directory would be passed to them.  I think that even if a function is passed the "model" directory, and in those cases we have to make the assumption that "experiments" is "model/..", this would still be reasonable and acceptable.  What do you think?
> 
> Thank you very much!

Hi @alexsherstinsky, both your suggestions have been implemented.

There was already a label `MODEL_WEIGHTS_FILE_NAME` defined (and used in some places) for the literal string "model_weights". So the choice of a label for "model" fell naturally on `MODEL_FILE_NAME` -- not the best, but there it is for the time being.

alexsherstinsky (changes requested):
@sanjaydasgupta I worry that it may not be safe to change a global constant as it may adversely impact backward compatibility (as in -- a user might rely on it).  Is there any way for us in the present PR to constrain the changes in such a way that we do not modify the existing constants but reuse them and define a new one, if needed?  Thank you!

sanjaydasgupta:
> @sanjaydasgupta I worry that it may not be safe to change a global constant as it may adversely impact backward compatibility (as in -- a user might rely on it). Is there any way for us in the present PR to constrain the changes in such a way that we do not modify the existing constants but reuse them and define a new one, if needed? Thank you!

I think there is a misunderstanding, let me clarify what I meant:

The global constant MODEL_WEIGHTS_FILE_NAME (for the literal string “model_weights”) existed previously, and continues to be used to denote “model_weights”. I have not changed that.

I added the (previously unused) global constant MODEL_FILE_NAME for the literal string “model”.

My comment about their being poor choices was due to the use of “FILE” instead of “DIRECTORY” or “FOLDER”. Both constants refer to folders, not file-names.

Please let me know if I misunderstood your comment. Thanks.

alexsherstinsky (changes requested):
@sanjaydasgupta You are correct (I just got concerned about a 38-module change!).  Please sync with the latest `"master"` branch, and I will approve.  Thank you very much!

sanjaydasgupta:
> @sanjaydasgupta You are correct (I just got concerned about a 38-module change!). Please sync with the latest `"master"` branch, and I will approve. Thank you very much!

Hi @alexsherstinsky this PR is now ready for merge.

alexsherstinsky (approved):
LGTM!  Beautiful enhancement!  Thank you very much, @sanjaydasgupta!

PR #3981, opened 2024-04-02
Support for freezing pretrained vision model layers with regex

### LINKED ISSUE #3733: Option to unfreeze encoder layers when training a computer vision model
Our computer vision solution is missing an important feature to improve accuracy. 

**Describe the use case**
In the past, I've followed this [Tensorflow tutorial](https://www.tensorflow.org/tutorials/images/transfer_learning#fine_tunin) to train an image classifier. This is a transfer learning technique that has two training rounds:
1. **Training:** add a classification head to a pretrained base model. The base model is frozen and the classification layers are trainable. This is currently what we have right now.
2. **Fine-tuning:** unfreezing layers from the base model and doing [10 epochs of low learning rate training](https://www.tensorflow.org/tutorials/images/transfer_learning#compile_the_model_2).

There are two ways to implement this:
1. Update the configuration to let users choose to automatically run the 2nd fine-tuning round. Maybe this can be the default.
2. Rely on users to code their own fine-tuning round.

Both require a way to choose how to freeze layers. There are two approaches to freezing layers:
1. By percentage or layer count of the base model starting from the first hidden layer
2. Using a regex to choose which layers to freeze. [From example](https://github.com/google/automl/blob/c7392f2bab3165244d1c565b66409fa11fa82367/efficientdet/tf2/README.md#8-finetune-on-pascal-voc-2012-with-detector-coco-ckpt):
```
var_freeze_expr: '(efficientnet|fpn_cells|resample_p6)'
```

The later layers encode high level features of the image while the early ones encode low level features (corners, edges, colors, gradients, etc). It's not good to fine-tune the lower levels because the model will overfit those features.

When fine-tuning, we unfreeze the later layers of the pretrained base model. The number of layers to unfreeze depends on the task similarity and dataset size relative to the pretrained encoder.

image

The alternative is to keep things as they are and rely on users to implement their own 2-round transfer-learning job.

ethanreidel:
Allows the user to input a regular expression in the yaml config which freezes specific layers of a pretrained model. Adds new CLI option "pretrained_summary" to let users access string representations of model layers for freezing via regex. Currently all pretrained torchvision models are accessible. 

trainer:
    layers_to_freeze_regex: (regex here)

ludwig pretrained_summary -m (model name here)

(I am aware that the collect_summary CLI command is similar, however it only accepts a preexisting directory so I thought creating a separate command to strictly output layer names was appropriate for this feature.)

Closes #3733 

Future plans -> expand this capability to implement gradual unfreezing

Test: pytest tests/ludwig/modules/test_regex_freezing.py


ludwig/cli.py
```diff
@@ -191,6 +192,11 @@ def upload(self):
 
         upload.cli(sys.argv[2:])
 
+    def pretrained_summary(self):
+        from ludwig.utils import pretrained_summary
+
+        pretrained_summary.cli_summarize_pretrained(sys.argv[2:])
+
```
skanjila, line 199: it might be good to add in the docs some example runs and outputs

ludwig/schema/metadata/configs/trainer.yaml
```diff
@@ -70,6 +70,15 @@ ecd:
             In many large-scale training runs, evaluation is often configured to run on
             a sub-epoch time scale, or every few thousand steps.
         ui_display_name: Checkpoints per epoch
+    layers_to_freeze_regex:
+        default_value_reasoning:
+            Default no layers will be frozen for fine-tuning a pretrained model.
```
skanjila, line 75: what is the repercussion of not freezing any layers

ludwig/utils/pretrained_summary.py
```diff
@@ -0,0 +1,152 @@
...
+# limitations under the License.
+# ==============================================================================
+import argparse
+import importlib
+
+from ludwig.api_annotations import DeveloperAPI
+from ludwig.contrib import add_contrib_callback_args
+from ludwig.globals import LUDWIG_VERSION
+from ludwig.utils.print_utils import print_ludwig
+
```
skanjila, line 23: wait are we really supporting all of these models, I thought we were just going to go out the door with a couple of models to start?

ludwig/utils/trainer_utils.py
```diff
@@ -408,3 +410,16 @@ def get_rendered_batch_size_grad_accum(config: "BaseTrainerConfig", num_workers:
                 gradient_accumulation_steps = 1
 
     return batch_size, gradient_accumulation_steps
+
+
+def freeze_layers_regex(config: "BaseTrainerConfig", model: ECD) -> None:
+    """Freezes layers based on provided regular expression."""
```
skanjila, line 416: lets add all of the comments around inputs/outputs as well

ludwig/utils/pretrained_summary.py
ethanreidel: For this specific feature (simple regex freezing), as long as you have access to the string representation of layers + actual model architecture, you can freeze any layers that you'd like. It wasn't any extra work adding support for all torchvision models besides adding to this list. I however don't like the look of this long model array though

ludwig/schema/metadata/configs/trainer.yaml
ethanreidel: No repercussion. If no layers are frozen, normal full fine tuning (adjusting all parameters) will take place.

ethanreidel:
@saad-palapa

ludwig/utils/pretrained_summary.py
```diff
@@ -0,0 +1,152 @@
...
+models = [
```
saad-palapa, line 24: This list will quickly become obsolete. It also requires maintenance as people add/remove models. Is it really needed?

ludwig/utils/trainer_utils.py
```diff
@@ -408,3 +410,16 @@ def get_rendered_batch_size_grad_accum(config: "BaseTrainerConfig", num_workers:
...
+    try:
+        pattern = re.compile(config.layers_to_freeze_regex)
+    except re.error:
+        logger.warning("Invalid regex input.\n")
```
saad-palapa, line 420: This should be logger.error. Also print the regex in the message. No need for \n

ludwig/utils/pretrained_summary.py
```diff
@@ -0,0 +1,152 @@
...
+    "vit_h_14",
+    "vit_l_16",
+    "vit_l_32",
+    "wide_resnet101_2",
+    "wide_resnet50_2",
+]
+
+
+def pretrained_summary(model_name, **kwargs) -> None:
+    if model_name in models:
```
saad-palapa, line 118: Unless I'm missing something, I'm not sure models is needed here.

ethanreidel: Would you say that a better/cleaner alternative would be providing the user (for example in the ludwig docs) a link to this: https://pytorch.org/vision/0.8/models.html#id1 which provides a list of models available for this purpose?

ludwig/cli.py
alexsherstinsky: @ethanreidel To second @skanjila -- would it be possible to show an example of running this command in terms of how it is different from the existing one -- and an example output.  Thank you very much.

ludwig/schema/metadata/configs/trainer.yaml
alexsherstinsky, line 75: Minor: "Default" or "By default"?

ludwig/schema/metadata/configs/trainer.yaml
```diff
@@ -70,6 +70,15 @@ ecd:
...
+        description_implications:
+            Freezing specific layers can improve a pretrained models performance in a number
```
alexsherstinsky, line 77: Minor rypo: "models -> model's" -- thanks!

ludwig/trainers/trainer.py
```diff
@@ -225,6 +227,10 @@ def prepare(self):
             base_learning_rate *= lr_scale_fn(self.distributed.size())
         self.base_learning_rate = base_learning_rate
 
+        # Given that regex is supplied, freeze layers
```
alexsherstinsky, line 230: @ethanreidel Question: By placing this capability at the top-level "trainer.py" module, we imply that all subclasses will inherit this capability.  While for Computer Vision, based on yours and others' work, it will work and we believe will be a useful capability to have, what can we say about such trainers as `FineTuneTrainer` (for LLM fine-tuning) and `NoneTrainer` (for LLM predictions) -- and also the GBM Trainer (albeit not as popular nowadays).  In particular, I am curious whether or not it will work for LLM architectures and, if affirmative, what you think about how we can take advantage of it?  Thanks!

ludwig/utils/pretrained_summary.py
alexsherstinsky: > For this specific feature (simple regex freezing), as long as you have access to the string representation of layers + actual model architecture, you can freeze any layers that you'd like. It wasn't any extra work adding support for all torchvision models besides adding to this list. I however don't like the look of this long model array though

@ethanreidel Sorry, could you please point me to this "long model array"?  Which line in your code has it?  Thanks!

ludwig/utils/trainer_utils.py
alexsherstinsky: @ethanreidel I think that if you put `rom __future__ import annotations` as the very first line in the module, you would not need to quote the types.  Would you like to give it a try and see if it works?  Thanks!

ludwig/utils/trainer_utils.py
```diff
@@ -408,3 +410,20 @@ def get_rendered_batch_size_grad_accum(config: "BaseTrainerConfig", num_workers:
                 gradient_accumulation_steps = 1
 
     return batch_size, gradient_accumulation_steps
+
+
+def freeze_layers_regex(config: "BaseTrainerConfig", model: ECD) -> None:
+    """Freezes layers based on provided regular expression."""
+    try:
+        pattern = re.compile(config.layers_to_freeze_regex)
```
alexsherstinsky, line 418: @ethanreidel Would you be interested if I gave you a reasonably well-featured RegEx utility so that you can just put it into the utils and use it -- it will save a lot of boilerplate like this.  Please let me know any time.  Thanks!

alexsherstinsky (changes requested):
@ethanreidel I love this work!  Of course others need to look at it as well; I made a few largely minor comments and suggestions (and raised a few questions) for your consideration.  More examples (e.g., into Ludwig Docs) would be wonderful!

Thank you very much!

ludwig/cli.py
ethanreidel: Certainly.

ludwig/utils/pretrained_summary.py
ethanreidel: For your first question Alex: as long as access to the model layers + their requires_grad parameter is available, in theory, this feature should work on LLMs/text. I'm not too familiar with LLM architecture and I'll have to do some quick checks, but I'm 99% sure it is an easy addition. Second question: in a previous commit, I had a pretty hacky solution where users had another command line option (under pretrained_summary) which would list all available model names. Those names were stored in a Python list which had a few issues namely having to expand it regularly/many lines of unnecessary code. Saad made a good point and said to fully remove it (it was not needed), so it's no longer there.

ludwig/cli.py
ethanreidel: Qq: when you say you'd like an example, do you mean an example in the Ludwig docs or how would you prefer it?

ludwig/utils/trainer_utils.py
ethanreidel: Fixed. Thanks

ludwig/utils/pretrained_summary.py
ethanreidel: Just checked and sure enough you can apply the same regex freezing technique to an LLM

ludwig/utils/trainer_utils.py
ethanreidel: I tested the annotations import, and it worked, but the git pre-commit was forcing changes (e.g. converting all uppercase Dicts to lowercase dicts) that I didn't like.

ludwig/utils/trainer_utils.py
ethanreidel: Yeah that sounds good. Thanks

ludwig/cli.py
arnavgarg1: @ethanreidel One option is to create an example in the `examples/` top level directory in Ludwig

ludwig/schema/trainer.py
```diff
@@ -86,6 +86,17 @@ class BaseTrainerConfig(schema_utils.BaseMarshmallowConfig, ABC):
...
+        default=None,
+        allow_none=True,
+        description=(
+            "Freeze specific layers based on provided regex. Freezing specific layers can improve a  "
+            "pretrained model's performance in a number of ways. At a basic level, freezing early layers can  "
+            "prevent overfitting by retaining more general features (beneficial for small datasets). Also can  "
+            "reduce computational resource use and lower overall training time due to less gradient calculations. "
+        ),
+    )
+
```
arnavgarg1, line 99: Instead of putting this in the base trainer config, what if we put it in the `ECDTrainerConfig`? For now, that will be good enough to ensure that this only works for ECD supported models and is not a valid argument/parameter for GBMs and LLMs. If it also works for LLMs, then we can also duplicate adding it in `FineTuneTrainerConfig` which is used by LLMs. It also means we don't have to modify any other trainers for now.

ludwig/trainers/trainer.py
arnavgarg1: @alexsherstinsky This wouldn't be particularly useful for the NoneTrainer, but it could be useful for the FineTuneTrainer. I left a comment above that might help simplify this by only enabling this parameter for the ECDTrainerConfig for now, which would then make it a no-op for LLMs and GBMs entirely.

ludwig/utils/pretrained_summary.py
arnavgarg1: > Just checked and sure enough you can apply the same regex freezing technique to an LLM
@ethanreidel That's awesome. Maybe we can then use one of my earlier comments to only add this parameter to the `ECDTrainerConfig` and `FineTuneTrainerConfig` for now. 

As part of the examples you have, it would be good to create 2 example Python files:
1. To show how to use it with a computer vision model
2. To show how to use it with an LLM base model

What do you think?

ludwig/utils/pretrained_summary.py
```diff
@@ -0,0 +1,51 @@
...
+from ludwig.utils.print_utils import print_ludwig
+
+
+def pretrained_summary(model_name, **kwargs) -> None:
+    module = importlib.import_module("torchvision.models")
+    encoder_class = getattr(module, model_name)
+    model = encoder_class()
+
+    for name, _ in model.named_parameters():
+        print(name)
```
arnavgarg1, line 31: We generally don't like to use Prints in Ludwig code - can we use `logger.info()` instead?

```python
import logging

logger.info("message")
```

ludwig/utils/trainer_utils.py
arnavgarg1: @ethanreidel I think that makes sense. 

Are you able to expand on the docstring itself for this function? Also, if it also supports `LLM`, can we make `model` a union of ECD and LLM?

ludwig/utils/trainer_utils.py
```diff
@@ -408,3 +410,20 @@ def get_rendered_batch_size_grad_accum(config: "BaseTrainerConfig", num_workers:
...
+    except re.error:
+        logger.error(f"Invalid regex input: {config.layers_to_freeze_regex}")
+        exit()
```
arnavgarg1, line 421: Instead of `exit()`, let's raise a RuntimeError() with the same message. 

In fact, here's a thought I have: We can move this check to earlier in the code path, that is, at config validation time. Specifically, you can create a `__post_init__()` hook for `ECDTrainerConfig` and `FineTuneTrainerConfig` that tries to do re.compile() and if it fails, throws a ConfigValidationError with the error message. That way, we don't have to wait for all of preprocessing etc to be done before catching this error. 

Here's an example explaining the same idea in a different part of the Ludwig codepath: https://github.com/ludwig-ai/ludwig/blob/master/ludwig/schema/llms/peft.py#L443

ludwig/utils/trainer_utils.py
```diff
@@ -408,3 +410,20 @@ def get_rendered_batch_size_grad_accum(config: "BaseTrainerConfig", num_workers:
...
+
+    matched = False
+    for name, p in model.named_parameters():
+        if re.search(pattern, str(name)):
+            p.requires_grad = False
+            matched = True
+    if not matched:
+        logger.error(f"No regex match for {config.layers_to_freeze_regex}! Check layer names and regex syntax.")
```
arnavgarg1, line 429: Two thoughts here:
1. Instead of `logger.error`, perhaps we can do a `logger.warning()`? I think it's okay if there are no matches, but we just want to claim that as a warning so the users can notice it as opposed to calling it an error (which it sort of is). 
2. One thing that could be very useful here, is to also create a set of all the layers where the regex search actually returns true and requires grad gets set to false, and then log that full list! Observability can be super helpful

arnavgarg1 (commented):
Overall, really nice work and clean implementation @ethanreidel! I left a few suggestions that might help simplify edge cases as well as considerations to add more observability into which layers are frozen.

arnavgarg1:
I would also recommend installing `pre-commit` via `pip install pre-commit`, then running `pre-commit install` within the Ludwig repo. That will help fix some of the pre-commit related styling errors here: https://results.pre-commit.ci/run/github/163346054/1714080170.FEtyFVFyR8m3t6xqARcnDQ

arnavgarg1:
@ethanreidel Let us know when this is ready for re-review!

ethanreidel:
Hi @arnavgarg1! Thanks for checking in. The PR is good for re-review. 👍

alexsherstinsky (approved):
LGTM!  @ethanreidel -- thank you for this amazing and very useful contribution!

arnavgarg1 (approved):

PR #4008, opened 2024-05-23
[MAINTENANCE] Refactor and clean up.

alexsherstinsky:
### Scope
* Make the implementation for the fix of the `ViTEncoder` to ensure that the `transformers.ViTModel` returns the `output_attentions` more elegant (and cuts on the amount of code).

# Code Pull Requests

Please provide the following:

# Documentation Pull Requests

### DIFF
--- a/ludwig/encoders/image/base.py
+++ b/ludwig/encoders/image/base.py
@@ -382,16 +382,17 @@ class ViTEncoder(ImageEncoder):
             raise ValueError("img_height and img_width should be identical.")
         self._input_shape = (in_channels, img_height, img_width)
 
+        config_dict: dict
         if use_pretrained and not saved_weights_in_checkpoint:
+            config_dict = {
+                "pretrained_model_name_or_path": pretrained_model,
+            }
             if output_attentions:
-                transformer = ViTModel.from_pretrained(
-                    pretrained_model_name_or_path=pretrained_model,
-                    attn_implementation="eager",
-                )
-            else:
-                transformer = ViTModel.from_pretrained(pretrained_model_name_or_path=pretrained_model)
+                config_dict["attn_implementation"] = "eager"
+
+            transformer = ViTModel.from_pretrained(**config_dict)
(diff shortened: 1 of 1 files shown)


ludwig/encoders/image/base.py
```diff
@@ -383,31 +383,49 @@ def __init__(
...
-                intermediate_size=intermediate_size,
-                hidden_act=hidden_act,
-                hidden_dropout_prob=hidden_dropout_prob,
-                attention_probs_dropout_prob=attention_probs_dropout_prob,
-                initializer_range=initializer_range,
-                layer_norm_eps=layer_norm_eps,
-                gradient_checkpointing=gradient_checkpointing,
-            )
+            if output_attentions:
+                config = ViTConfig(
```
arnavgarg1, line 395: Can we just create a dictionary mapping, then optionally add "attn_implementation" if output_attention and pass the dictionary as **kwargs into the Config? Will reduce boilerplate :)

alexsherstinsky: @arnavgarg1 Done -- sorry about missing the obvious!  Thanks!

ludwig/encoders/image/base.py
```diff
@@ -383,31 +383,33 @@ def __init__(
...
+                "intermediate_size": intermediate_size,
+                "hidden_act": hidden_act,
+                "hidden_dropout_prob": hidden_dropout_prob,
+                "attention_probs_dropout_prob": attention_probs_dropout_prob,
+                "initializer_range": initializer_range,
+                "layer_norm_eps": layer_norm_eps,
+                "gradient_checkpointing": gradient_checkpointing,
+            }
+            config_dict["attn_implementation"] = "eager"
+            config = ViTConfig(**config_dict)
```
arnavgarg1, line 410: Do we no longer need to do the if else so we can pass these custom args in both cases?

arnavgarg1 (approved):

Issue #4009, opened 2024-05-28
MNIST Dataset can't be downloaded


mhabedank:
**Describe the bug**
It's not possible to download the MNIST dataset from it's source. 
For example the URL http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz is just forbidden even on the browser.

**To Reproduce**

Running this code
```
from ludwig.datasets import mnist
train_df, test_df = mnist.load(split=True)
```

will result in the error

```
Finding fallback mirrors to download the dataset. Downloading from the original source failed with the following error HTTP Error 403: Forbidden.
No fallback mirror found. Failed to download dataset MNIST.
```

**Expected behavior**
The file should get downloaded.

**Screenshots**
<img width="994" alt="image" src="https://github.com/ludwig-ai/ludwig/assets/43589214/e4fa4166-df4a-4f05-8f73-7e1fa1998b3a">
**Environment (please complete the following information):**

- OS: MacOS on Arm64
- Version Sonoma 14.4.1
- Python 3.11.8
- Ludwig 0.10.3

**Additional context**
It's not really a bug from ludwig. The source withholds the content.
Maybe there should be different sources, or it could be possible to just to use torchs or sklearns data zoo.

PR #4011, opened 2024-05-29
Fix mnist source

mhabedank:
# Code Pull Requests

The MNIST dataset could not get acquired from the original source because the data is not available anymore.
This PR will solve the following issue #4009 

Beside new tests the follow two lines will work again:

```
from ludwig.datasets import mnist

a, b = mnist.load(split=True)

```


alexsherstinsky (approved):
LGTM!  @mhabedank -- Thank you very much for making this fix!

PR #4022, opened 2024-08-05
Fixed options dublicate in documentation for EmbeddingInitializerField

mhabedank:
The options for the EmbeddingInitializerField are displayed twice in the documentation. 

This is due to the fact that both the field in the Ludwig repro and the build logic in the documentation append the options list to the description string in this particular case.

The problem has been solved by removing this logic from the field. The documentation takes over the task of appending this information.

See issue in [documentation repo](https://github.com/ludwig-ai/ludwig-docs/issues/366).

There should be a discussion about what the right path is. At the moment, different fields handle the task differently. Sometimes the documentation is supposed to assemble the string, sometimes the field does it itself.

### DIFF
--- a/ludwig/schema/common_fields.py
+++ b/ludwig/schema/common_fields.py
@@ -139,15 +139,13 @@ def WeightsInitializerField(
 def EmbeddingInitializerField(
     default: Optional[str] = None, description: str = None, parameter_metadata: ParameterMetadata = None
 ) -> Field:
-    initializers_str = ", ".join([f"`{i}`" for i in initializer_registry.keys()])
     description = description or "Initializer for the embedding matrix."
-    full_description = f"{description} Options: {initializers_str}."
     parameter_metadata = parameter_metadata or COMMON_METADATA["embedding_initializer"]
     return schema_utils.StringOptions(
         list(initializer_registry.keys()),
         default=default,
         allow_none=True,
-        description=full_description,
+        description=description,
         parameter_metadata=parameter_metadata,
     )
 


alexsherstinsky:
@mhabedank Hi!  Thank you for looking into this issue.  Could you please point out what the duplication issue actually is?  By reading the documentation, I am seeing the `embedding_initializer` included in different encoder documentation pages that use it.  Perhaps you could provide screenshot of the documentation and highlight the duplication you are seeing?  Thank you  for helping to clarify your observation.

mhabedank:
Here in category features encoding section: https://ludwig.ai/latest/configuration/features/category_features/#encoders

Bildschirmfoto 2024-08-05 um 18 26 15

alexsherstinsky:
> Here in category features encoding section: https://ludwig.ai/latest/configuration/features/category_features/#encoders
> 
> Bildschirmfoto 2024-08-05 um 18 26 15

@mhabedank I see -- thank you very much for highlighting.  Ideally, the solution would be the one that also includes the "(default: null)" part (with everything being the same).  Does your fix make it so?  Thank you.

mhabedank:
The default part is not part of the issue. The green marked part should be gone.

alexsherstinsky:
> The default part is not part of the issue. The green marked part should be gone.

@mhabedank Got it.  Does your fix properly delete the green part, without causing problems in other parts of the documentation?  Thank you.

mhabedank:
`EmbeddingInitializerField` is only used in two places within the category feature. To the best of my knowledge, I cannot see a problem.

alexsherstinsky:
> `EmbeddingInitializerField` is only used in two places within the category feature. To the best of my knowledge, I cannot see a problem.

🙇

alexsherstinsky (approved):
LGTM!  Thank you very much, @mhabedank

Issue #4035, opened 2024-10-21
Refactor BERTTokenizer


mhabedank:
The [BERTTokenizer](https://github.com/ludwig-ai/ludwig/blob/00c51e0a286c3fa399a07a550e48d0f3deadc57d/ludwig/utils/tokenizers.py#L1109) is using torchtext. We want to remove torchtext as a dependency so this Tokenizer has to be refactored not using it.

thc1006:
Hi @ManishSharma1609,

  I just saw your message from January—apologies for the late response!

  I noticed you were asking about how to contribute to this issue. I've actually been working on the same problem recently and have completed some testing work. If you're still interested, perhaps we could collaborate?

  My current plan is to:
  1. First reach out to the PR #4041 authors to see if we can help push that PR forward
  2. Or, if needed, prepare a focused, smaller-scope PR

  The technical solution in PR #4041 is already quite complete. The main gap is in testing and validation, which I've completed:
  - ✅ Docker deployment testing
  - ✅ All 5 tokenizers validated
  - ✅ Test results documentation

  If you're interested in working together on this, I'd love to discuss! Or if you have other ideas, I'd be happy to hear them as well.

  The main challenge right now is coordinating with the existing PR #4041 to avoid duplicate work. But with good communication, I believe we can find the best approach.

  Looking forward to hearing from you!

  Best regards,
Hsiu-Chi Tsai

Issue #4032, opened 2024-10-21
Refactor SentencePieceTokenizer


mhabedank:
The [SentencePieceTokenizer](https://github.com/ludwig-ai/ludwig/blob/00c51e0a286c3fa399a07a550e48d0f3deadc57d/ludwig/utils/tokenizers.py#L1014) is using torchtext. We want to remove torchtext as a dependency so this Tokenizer has to be refactored not using it.

w4nderlust:
this is done! we removed the torchtext dependency entirely in Ludwig v0.11.0. all tokenizers have been refactored to use HuggingFace transformers or standalone implementations. you can install with `pip install ludwig==0.11.0`.

Issue #4031, opened 2024-10-21
Replace Ngramm Tokenizer


mhabedank:
https://github.com/ludwig-ai/ludwig/blob/00c51e0a286c3fa399a07a550e48d0f3deadc57d/ludwig/utils/tokenizers.py#L135

nqbao:
if you can provide an example i can help with the rest

mhabedank:
If we decide to replace the dependency, this would be about  5 lines of code: https://pytorch.org/text/stable/_modules/torchtext/data/utils.html#ngrams_iterator

torchtext is used here: https://github.com/ludwig-ai/ludwig/blob/00c51e0a286c3fa399a07a550e48d0f3deadc57d/ludwig/utils/tokenizers.py#L143

nqbao:
can we just copy the code over?

mhabedank:
yeah that would probably be the solution for this tokenizer.

w4nderlust:
this is done! we removed the torchtext dependency entirely in Ludwig v0.11.0. all tokenizers have been refactored to use HuggingFace transformers or standalone implementations. you can install with `pip install ludwig==0.11.0`.

PR #4038, opened 2024-10-25
WIP: Remove torchtext dependency

mhabedank:
torchtext is deprecated but used for some tokenizers in Ludwig.
This PR will remove the dependency.

Related to

#4031 
#4032 
#4033 
#4034 
#4035


mhabedank:
this is merged into #4041

PR #4051, opened 2025-07-02
fix: handle loss config access for class_similarities

MohamedAliBouhaouala:
### Summary
The following PR fixes the logic for accessing `class_similarities` in the loss config.  Previously, the code assumed a fixed structure, which could break when `feature_config.loss` was a dict or an object. The updated logic handles both cases safely. 
While training my model with class_similarities, I encountered this exception:
```bash
 TypeError: argument of type 'SoftmaxCrossEntropyLossConfig' is not iterable
```

### Changes

- Added conditional checks to support both dict and object access styles for `feature_config.loss`.
- Ensured compatibility with different configuration formats (e.g., YAML or Python objects).

### Additional Note

I also modified `.pre-commit-config.yaml` based on a previous PR [4039](https://github.com/ludwig-ai/ludwig/pull/4039) to ensure that pre-commit hooks pass correctly. This change was required to avoid commit failures due to missing staged updates in the pre-commit config.


w4nderlust:
Thank you @MohamedAliBouhaouala for identifying the class_similarities bug! The core issue (using `in` operator on a config object instead of proper attribute access) has been fixed in PR #4061 with a simpler approach:

```python
# Before (raises TypeError):
if "class_similarities" in feature_config.loss:

# After:
if feature_config.loss.class_similarities is not None:
```

The fix was applied to both `category_feature.py` and `sequence_feature.py`. We went with direct attribute access since `class_similarities` is always a defined attribute on the loss config object. Thanks for catching this!

PR #4058, opened 2026-02-23
Add JSON sanitization to handle float inf/-inf/nan values

dhoepp:
## Summary

When training without a validation set, `training_progress.json` contains `Infinity` / `-Infinity` values which are not valid JSON (RFC 8259). This causes failures with standard JSON tools like `jq`.

## Root cause

`get_initial_validation_value()` returns `float("inf")` / `float("-inf")` as sentinel values for `best_eval_metric_value` and `best_increase_batch_size_eval_metric`. Without a validation set, these are never overwritten. `save_json()` passes them to `json.dump()`, which writes non-standard `Infinity` / `-Infinity` tokens.

## Fix

Added `_sanitize_for_json()` in `data_utils.py` that recursively converts `inf`, `-inf`, and `nan` floats to `None` (`null` in JSON) before serialization. Applied inside `save_json()` so all callers benefit.

## Reproducible test

```python
import json, tempfile, os
from ludwig.utils.data_utils import save_json

data = {
    "best_eval_metric_value": float("-inf"),
    "best_increase_batch_size_eval_metric": float("inf"),
}

with tempfile.NamedTemporaryFile(suffix=".json", delete=False) as f:
    tmp = f.name

save_json(tmp, data)

with open(tmp) as f:
    output = f.read()

assert "Infinity" not in output
parsed = json.loads(output)
assert parsed["best_eval_metric_value"] is None
assert parsed["best_increase_batch_size_eval_metric"] is None
os.unlink(tmp)

### DIFF
--- a/ludwig/utils/data_utils.py
+++ b/ludwig/utils/data_utils.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 import base64
+import math
 import collections.abc
 import contextlib
 import csv
@@ -428,8 +429,20 @@ def load_json(data_fp):
     return data
 
 
+def _sanitize_for_json(obj):
+    """Recursively replace float inf/-inf/nan with None for JSON compliance."""
+    if isinstance(obj, float) and (math.isinf(obj) or math.isnan(obj)):
+        return None
+    elif isinstance(obj, dict):
+        return {k: _sanitize_for_json(v) for k, v in obj.items()}
+    elif isinstance(obj, list):
+        return [_sanitize_for_json(v) for v in obj]
+    return obj
+
+
 @DeveloperAPI
 def save_json(data_fp, data, sort_keys=True, indent=4):
+    data = _sanitize_for_json(data)
     with open_file(data_fp, "w") as output_file:
         json.dump(data, output_file, cls=NumpyEncoder, sort_keys=sort_keys, indent=indent)
 


w4nderlust:
Thank you @dhoepp for identifying the JSON serialization issue with inf/nan values! The problem you identified is real — `float('inf')` values in the ProgressTracker are not valid JSON per RFC 8259.

However, we found that converting inf/-inf/nan to `None` would break training resume, because the comparison functions (`improved_fn`) in the trainer would then try to compare `None` with floats, raising a `TypeError`. The sentinel values need to survive the serialization round-trip.

A proper fix would need either a custom JSON encoder/decoder pair that preserves inf semantics, or changing `get_initial_validation_value()` to use very large finite floats instead of inf. We'll address this separately. Thanks for the contribution!

dhoepp:
> Thank you @dhoepp for identifying the JSON serialization issue with inf/nan values! The problem you identified is real — `float('inf')` values in the ProgressTracker are not valid JSON per RFC 8259.
> 
> However, we found that converting inf/-inf/nan to `None` would break training resume, because the comparison functions (`improved_fn`) in the trainer would then try to compare `None` with floats, raising a `TypeError`. The sentinel values need to survive the serialization round-trip.
> 
> A proper fix would need either a custom JSON encoder/decoder pair that preserves inf semantics, or changing `get_initial_validation_value()` to use very large finite floats instead of inf. We'll address this separately. Thanks for the contribution!

Thank you for the feedback! I see now what you mean. My thoughts were to either convert it to none or min/max but I didn't have enough context to the benefit of the function to understand the further scenarios that would break from it. And didn't want to assume min/max would be any better. 

I'll be sure to communicate MO more moving forward. 

I learned a lot from this though! Thanks for the time!

PR #4182, opened 2026-05-17
fix: stream collect_activations to disk to prevent OOM on large models

w4nderlust:
## Summary

`batch_collect_activations` previously accumulated all batch activations as CPU tensors in RAM, then called `torch.cat()` to concatenate them. For large models (e.g. LLaMA 3 8B) with >10k samples, this OOMs even on CPU.

**Fix**: each batch's activations are written to a temporary `.npy` file immediately after inference using `numpy.save`. After all batches complete, files are loaded one at a time and concatenated. The temporary directory is cleaned up automatically.

Peak RAM is now bounded to ~one batch of activations rather than the full dataset.

Addresses C6 from the codebase quality improvement plan: *"TODO -> Fix OOM on large models e.g. llama 3 8B. Loads all activations for all samples into RAM at once."*

## Test plan
- [ ] Existing collect tests pass
- [ ] Manual test: collect activations on a large dataset doesn't OOM

### DIFF
--- a/ludwig/models/predictor.py
+++ b/ludwig/models/predictor.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+import tempfile
 from abc import ABC, abstractmethod
 from collections import defaultdict, OrderedDict
 from pprint import pformat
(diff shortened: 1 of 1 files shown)


w4nderlust:
Superseded by #4186 (all changes merged into feat/full-improvement-plan)

